PR #484, opened 2026-02-18
Add ratchet to require public function arguments to be pos-only or kw-only

qi-imbue:
Because there are so many violations now, I had to use a dummy date for all the violations otherwise the test takes too long running `git blame`


weishi-imbue:
Please continue this PR on imbue-ai/mngr-internal#69 — development has moved to the private repository (this public repository is now a read-only mirror, so PRs cannot merge here). The branch and this PR's description were carried over.

PR #529, opened 2026-02-20
fix issue where VolumeId requires vol- prefix but doesn't get it from host volumes

evgunter:
name the host's volume vol-<id> rather than host-<id> since it's required to start with vol-

---
claude's summary:

## Problem

`mng rm -f` crashes with `InvalidRandomIdError: VolumeId must start with 'vol-'` during post-destroy GC.

The Docker provider stores host volume data in directories named `volumes/host-<uuid>/` on the state volume (created by `_ensure_host_volume_dir`). But `list_volumes()` wraps every directory name in `VolumeId()`, which requires the `vol-` prefix. So any `host-*` directory crashes it.

## Fix

Change `_ensure_host_volume_dir` to create `vol-<same_hex_as_host_id>` directories instead of `host-<host_id>`. Since HostId is `host-<32hex>` and VolumeId is `vol-<32hex>`, we can reuse the hex part. Then list_volumes/delete_volume become trivial -- directory name IS the VolumeId.

### DIFF
--- a/libs/mng/imbue/mng/providers/docker/testing.py
+++ b/libs/mng/imbue/mng/providers/docker/testing.py
@@ -8,6 +8,7 @@ from imbue.mng.errors import MngError
 from imbue.mng.primitives import ProviderInstanceName
 from imbue.mng.providers.docker.config import DockerProviderConfig
 from imbue.mng.providers.docker.instance import DockerProviderInstance
+from imbue.mng.providers.local.volume import LocalVolume
 from imbue.mng.utils.testing import get_short_random_string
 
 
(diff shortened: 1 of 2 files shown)


libs/mng/imbue/mng/providers/docker/testing.py
```diff
@@ -21,6 +22,20 @@ def make_docker_provider(mng_ctx: MngContext, name: str = "test-docker") -> Dock
...
+    volume_root: Path,
+) -> DockerProviderInstance:
+    """Create a Docker provider using a LocalVolume instead of a real Docker volume.
+
+    This avoids needing a running Docker daemon for tests that only exercise
+    state-volume logic (list_volumes, delete_volume, host store, etc.).
+    """
+    provider = make_docker_provider(mng_ctx)
+    provider.__dict__["_state_volume"] = LocalVolume(root_path=volume_root)
+    return provider
```
evgunter: is there a better way to do this?

PR #534, opened 2026-02-20
Add -b/--remove-branch flag to mng destroy

qi-imbue:
Store the created git branch name in the agent's data.json during creation. Add a -b/--remove-branch flag to mng destroy that deletes the branch from the source repository after the agent is destroyed.

The source repo is discovered by reading the worktree's .git file, which contains a gitdir pointer back to the main repository. Branch deletion runs after GC (which removes the worktree), so git allows the branch to be deleted.

Also fixes GC's worktree removal to run git from the main repository (via the same .git file parsing), so git properly unregisters the worktree instead of just removing the directory.

### DIFF
--- a/libs/mng/imbue/mng/hosts/test_host.py
+++ b/libs/mng/imbue/mng/hosts/test_host.py
@@ -2481,7 +2481,7 @@ def test_create_work_dir_cross_host_generates_unique_paths(
         data_options=AgentDataOptions(is_rsync_enabled=True),
     )
 
-    work_dir_1 = target_host.create_agent_work_dir(source_host, source_path, options)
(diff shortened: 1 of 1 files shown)


libs/mng/imbue/mng/agents/base_agent.py
```diff
@@ -128,6 +128,10 @@ def set_labels(self, labels: Mapping[str, str]) -> None:
         data["labels"] = dict(labels)
         self._write_data(data)
 
+    def get_branch_name(self) -> str | None:
+        data = self._read_data()
+        return data.get("branch_name")
```
joshalbrecht: to be clearer, this should be "original_branch_name"

joshalbrecht:
works for me!

PR #538, opened 2026-02-20
faster tabcomplete

evgunter:
built on https://github.com/imbue-ai/mng/pull/572

- tabcomplete reads from pre-written caches of commands and agent names, and has minimal dependencies
  - it does need to actually cache the commands instead of having a hardcoded list of them to work with plugins
- tabcomplete kicks off `mng list` asynchronously whenever it runs (and it's been 30 seconds since the last time it did that)
- `mng list` always writes out the caches of commands and agent names, whether it's called from tabcomplete or not
- tab completion-associated functions silently swallow errors and never fail (pros and cons, there)

note that the cache is in `MNG_COMPLETION_CACHE_DIR`, by default in tmp. we monkeypatch this in the test autouse fixture to be unique for each test

other random stuff in this PR:
- get rid of manual list of click aliases
- add an `atomic_write` util and make various local writes use it
- `disable_modal_for_subprocesses` fixture

---
claude's summary:

### What changed

Every TAB press used to import the full CLI (click, pydantic, pluggy, all plugins). Now TAB presses invoke `python -m imbue.mng.cli.complete`, a stdlib-only module that reads a JSON cache file and prints results. No click, no pydantic, no plugins loaded.

### Architecture

- **`complete.py`** -- stdlib-only entrypoint. Reads `COMP_WORDS`/`COMP_CWORD` from env, resolves context from cache, prints completions (one per line). Also generates shell scripts via `--script zsh|bash`.
- **`completion_writer.py`** -- walks the click command tree and writes the extended cache (commands, subcommands, options, choices, aliases, agent_name_arguments). Called from `mng list`.
- **`click_utils.py`** -- `detect_alias_to_canonical()` and `detect_aliases_by_command()` derive alias mappings from the click group, replacing the old hardcoded `COMMAND_ALIASES` dict.

### Cache format

`.command_completions.json` now includes:
- `commands` -- all registered command names (including aliases)
- `aliases` -- auto-detected alias->canonical mapping
- `subcommand_by_command` -- subcommands for group commands
- `options_by_command` -- `--option` names per command (dot-separated keys for subcommands like `config.get`)
- `option_choices` -- valid values for `click.Choice` options (e.g. `config.get.--scope` -> `["user", "project", "local"]`)
- `agent_name_arguments` -- commands whose positional args complete against agent names

### Removed

- `completion.py` -- deleted entirely. Constants/functions moved to `completion_writer.py`.
- `CachedCompletionGroup` -- config, plugin, snapshot now use plain `click.Group`.
- `COMMAND_ALIASES` / `_ALIAS_TO_CANONICAL` -- replaced by auto-detection from click group.
- Click's `_MNG_COMPLETE` completion mechanism is no longer used.

### Shell script

The `.zshrc` addition is 7 lines:
```zsh
_mng_complete() {
    local -a completions
    (( ! $+commands[mng] )) && return 1
    completions=(${(@f)"$(COMP_WORDS="${words[*]}" COMP_CWORD=$((CURRENT-1)) /path/to/python -m imbue.mng.cli.complete)"})
    compadd -U -V unsorted -a completions
}
compdef _mng_complete mng
```

Generated via `python -m imbue.mng.cli.complete --script zsh` (bakes in the correct python path at install time).

### Manual testing (tmux)

All verified interactively in a tmux session with the generated zsh script:

| Input | TAB result |
|---|---|
| `mng cr<TAB>` | `mng create` |
| `mng <TAB>` | All canonical commands (ask, cleanup, clone, config, connect, create, ...) -- no aliases shown |
| `mng config <TAB>` | `edit  get  list  path  set  unset` |
| `mng config get --<TAB>` | `--context  --disable-plugin  --format  --help  --scope  ...` |
| `mng config get --scope <TAB>` | `user  project  local` |
| `mng connect <TAB>` | Agent names from cache |

### Performance

~70ms per TAB press (without `uv run` overhead), well under the 100ms target.

### Test results

`cd libs/mng && uv run pytest` -- 2455 passed. (1 pre-existing failure in `test_no_type_errors` from `mng_tutor` referencing `PROFILES_DIRNAME` introduced by main merge, unrelated.)

---

## Other changes

**New `utils/click_utils.py` with alias auto-detection**

`detect_alias_to_canonical` and `detect_aliases_by_command` dynamically detect command aliases by comparing registered names to `cmd.name` in a Click group. Replaces the hand-maintained `COMMAND_ALIASES` dict and `_ALIAS_TO_CANONICAL` reverse mapping in `main.py` -- aliases can no longer get out of sync with actual registrations.

**New `utils/file_utils.py` with `atomic_write`**

Reusable `atomic_write(path, content)` that writes to a temp file, fsyncs, and atomically renames. Also preserves existing file permissions. `claude_config.py` now uses this instead of its own 19-line inline temp-file-and-rename logic.

**`api/list.py` decoupled from completion cache format**

`_write_completion_cache` and `COMPLETION_CACHE_FILENAME` moved out of `api/list.py` into `cli/completion_writer.py`. The list API now extracts agent names as a plain `list[str]` and passes them to `write_agent_names_cache()`, decoupling list from the cache format. Tests simplified accordingly (no longer need a full `MngContext`).

**`install.sh` updated**

Generates the completion script via `python3 -m imbue.mng.cli.complete --script <shell_type>` instead of the old `eval "$(_MNG_COMPLETE=...)"` approach.

**Import path updates across CLI commands**

`complete_agent_name` moved from `cli.completion` to `cli.completion_writer` -- pure import change in connect, destroy, exec, logs, message, pull, push, rename, start, stop, and `mng_pair/cli.py`.

**Test/fixture updates**

- `conftest.py`: `MNG_COMPLETION_CACHE_DIR` env var added to test isolation; new `disable_modal_for_subprocesses` fixture for tests that spawn subprocesses without Modal credentials.
- `test_ratchets.py`: Thresholds adjusted (broad_exception_catch 1->4, trailing_comments 0->4, builtin_exception_raises 0->1, direct_subprocess 47->48, num_prefix 2->0).

### DIFF
--- a/libs/mng/imbue/mng/cli/completion_writer.py
+++ b/libs/mng/imbue/mng/cli/completion_writer.py
@@ -32,18 +32,27 @@ def get_completion_cache_dir() -> Path:
     return cache_dir
 
 
-def complete_agent_name(
-    ctx: click.Context,
-    param: click.Parameter,
-    incomplete: str,
-) -> list:
-    """Marker callback for click arguments that accept agent names.
-
-    The cache writer detects this callback (via identity check) to populate
-    the agent_name_arguments field in the completions cache. The lightweight
-    completer (complete.py) handles the actual completion at runtime.
-    """
-    return []
+# Commands whose positional arguments should complete against agent names.
+# This list is used by the cache writer to populate agent_name_arguments
+# in the completions cache. The lightweight completer (complete.py) reads
+# this field to decide when to offer agent name completions.
+_AGENT_NAME_COMMANDS: Final[frozenset[str]] = frozenset(
+    {
+        "connect",
+        "destroy",
+        "exec",
+        "limit",
+        "logs",
+        "message",
+        "pair",
+        "provision",
+        "pull",
+        "push",
+        "rename",
+        "start",
+        "stop",
+    }
+)
 
 
 # =============================================================================
@@ -77,17 +86,6 @@ def _extract_choices_for_command(cmd: click.Command, key_prefix: str) -> dict[st
     return choices
 
 
-def _has_agent_name_argument(cmd: click.Command) -> bool:
-    """Check if a command has an argument using complete_agent_name for shell completion."""
-    for param in cmd.params:
-        if isinstance(param, click.Argument):
-            # click stores the shell_complete callback in _custom_shell_complete
-            custom_complete = vars(param).get("_custom_shell_complete")
-            if custom_complete is complete_agent_name:
-                return True
-    return False
-
-
 def write_cli_completions_cache(cli_group: click.Group) -> None:
     """Write all CLI commands, options, and choices to the completions cache (best-effort).
 
@@ -109,14 +107,15 @@ def write_cli_completions_cache(cli_group: click.Group) -> None:
         subcommand_by_command: dict[str, list[str]] = {}
         options_by_command: dict[str, list[str]] = {}
         option_choices: dict[str, list[str]] = {}
-        agent_name_arguments: list[str] = []
 
+        canonical_names: set[str] = set()
         for name, cmd in cli_group.commands.items():
             # Skip alias entries -- only process canonical command names
             if name in alias_to_canonical:
                 continue
 
             canonical_name = cmd.name or name
+            canonical_names.add(canonical_name)
 
             if isinstance(cmd, click.Group) and cmd.commands:
                 if canonical_name not in subcommand_by_command:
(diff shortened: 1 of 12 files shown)


libs/mng/imbue/mng/cli/completion.py
```diff
@@ -19,6 +21,29 @@ def _get_host_dir() -> Path:
...
+        is_fd_closed = True
+        os.replace(tmp_path, path)
+    except OSError:
+        if not is_fd_closed:
+            os.close(fd)
+        try:
+            os.unlink(tmp_path)
+        except OSError:
+            pass
+        raise
```
evgunter: we do other atomic writes in the codebase; should have a shared helper for them

libs/mng/imbue/mng/cli/completion.py
```diff
@@ -90,3 +96,92 @@ def complete_agent_name(
...
+    try:
+        path = _get_cli_completions_path()
+        if not path.is_file():
+            return None
+        data = json.loads(path.read_text())
+        if not isinstance(data, dict):
+            return None
+        return data
+    except (json.JSONDecodeError, OSError):
+        return None
```
evgunter: no logging to avoid importing logger (which wouldn't be configured), but i could change this

libs/mng/imbue/mng/utils/test_ratchets.py
```diff
@@ -311,7 +314,7 @@ def test_prevent_direct_subprocess_usage() -> None:
     # Docker provider uses subprocess for docker build/run CLI pass-through.
     # connect.py uses os.execvp/os.execvpe for process replacement (not child spawning).
     chunks = check_ratchet_rule(PREVENT_DIRECT_SUBPROCESS, _get_mng_source_dir(), TEST_FILE_PATTERNS)
-    assert len(chunks) <= snapshot(47), PREVENT_DIRECT_SUBPROCESS.format_failure(chunks)
+    assert len(chunks) <= snapshot(48), PREVENT_DIRECT_SUBPROCESS.format_failure(chunks)
```
evgunter: don't want to import concurrencygroup in the cache reader (which uses subprocess to kick of a `mng list` to update its cache)

PR #541, opened 2026-02-21
don't intersperse warnings in streaming `mng list` output

evgunter:
https://github.com/user-attachments/assets/dfcd411e-54d1-4dd6-a985-333db4851376

claude's summary:

## Summary
- During `mng ls`, loguru warnings (e.g. "Skipped container: no host record") were interleaving with the streaming table, corrupting the ANSI status line
- In TTY mode, stderr is now intercepted and routed through the streaming renderer, which keeps warnings pinned at the bottom of the table above the status line
- When new agents arrive, they are inserted above the warnings via ANSI cursor manipulation so warnings always stay at the bottom
- In non-TTY mode, warnings go to stderr normally with clean stream separation (no duplicates)

## Test plan
- [x] 6 new unit tests covering: TTY warning emission, warning ordering below agents, non-TTY behavior, warnings before any agents, multiple warnings
- [x] Full test suite passes (2343 tests, 0 failures, 81.23% coverage)
- [x] Smoke tested non-TTY mode: stdout has clean table, stderr has warnings, no duplicates
- [x] Autofix passed (one fix applied: guard warning re-write with is_tty check)

### DIFF
--- a/libs/mng/imbue/mng/cli/list_test.py
+++ b/libs/mng/imbue/mng/cli/list_test.py
@@ -759,21 +759,6 @@ def test_streaming_renderer_tty_erases_status_on_finish() -> None:
     assert output.endswith("\r\x1b[K")
 
 
-def test_streaming_renderer_emit_warning_tty() -> None:
-    """TTY streaming should show warnings below agent rows, above the status line."""
-    captured = StringIO()
-    renderer = _create_streaming_renderer(fields=["name"], is_tty=True, output=captured)
-    renderer.start()
-    renderer(make_test_agent_info(name="agent-1"))
-    renderer.emit_warning("WARNING: something bad\n")
-    renderer.finish()
-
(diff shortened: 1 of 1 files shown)


libs/mng/imbue/mng/utils/terminal.py
```diff
@@ -0,0 +1,76 @@
...
+ANSI_DIM_GRAY: Final[str] = "\x1b[38;5;245m"
+ANSI_RESET: Final[str] = "\x1b[0m"
+
+
+def ansi_cursor_up(lines: int) -> str:
+    """ANSI escape sequence to move the cursor up by the given number of lines."""
+    return f"\x1b[{lines}A"
+
+
+class StderrInterceptor(MutableModel):
```
evgunter: this seems like a reasonable solution given the constraints but it is (inevitably?) a bit hacky

PR #563, opened 2026-02-23
pollute gitconfig nonlocal

evgunter:
limit https://github.com/imbue-ai/mng/pull/486 to local


libs/mng/imbue/mng/providers/docker/test_docker_create.py
```diff
@@ -205,6 +206,48 @@ def test_mng_create_with_dockerfile_on_docker(
     assert "Done." in result.stdout, f"Expected 'Done.' in output: {result.stdout}"
 
 
+@pytest.mark.timeout(120)
+def test_mng_create_with_git_source_on_docker(
```
evgunter: NOT a regression test - this test does pass on main

evgunter: this test was written to show why this change was necessary, but it did not in fact show that, and i don't know if the change is necessary (i made this pr as a followup in response to your comment on a previous pr)

PR #570, opened 2026-02-24
send messages concurrently so one failure does not block others

evgunter:
- if messaging multiple agents at once, do it concurrently so that they don't wait for each others' timeouts or all die if one encounters an exception it doesn't handle
- also makes the raise builtin exception ratchet ignore test files - this seems reasonable to me because we have legitimate reasons to test non-mng exceptions and it seems like it doesn't have the same potential to be confusing that it does in non-test code

---
claude's summary:

## Summary
- Split `send_message_to_agents` into two phases: sequential host/agent resolution, then concurrent message sends via `ConcurrencyGroupExecutor`
- Each agent's message send runs in its own thread with shared `MessageResult` protected by a lock, so one failure never blocks or kills others
- Added regression test: one stopped agent's failure does not prevent a running agent from receiving its message


joshalbrecht:
thanks! will merge

PR #587, opened 2026-02-24
iTerms2 connection plugin

qi-imbue:
This is a little bit suspicious since I try to make it disabled by default, and Claude seems to be doing it in a somewhat twisted way


weishi-imbue:
Please continue this PR on imbue-ai/mngr-internal#71 — development has moved to the private repository (this public repository is now a read-only mirror, so PRs cannot merge here). The branch and this PR's description were carried over.

PR #596, opened 2026-02-25
draft: use autofix in reviewer

evgunter:
right now it just replaces the reviewer with this but i could make them be separate agent types instead


.claude/skills/autofix/SKILL.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Phase 2: Fix Loop
+
+Repeat up to 10 times:
+
+1. Record the current HEAD as `pre_iteration_head`.
+2. Read the supporting file [verify-and-fix.md](verify-and-fix.md) from this
+   skill's directory. Spawn a single Task subagent
+   (`subagent_type: "general-purpose"`) with its contents as the prompt.
+   Prepend the line `Base branch for this project: {base_branch}` to the prompt.
```
evgunter: kind of silly to have the orchestrator read this instead of telling the reviewer subagent to do it, but this means that it's easy to pass instructions through to the reviewer subagents, since the orchestrator can just edit their prompt

PR #630, opened 2026-02-27
make reliable tmux send separable from mng message

evgunter:
## Summary

- Extract the marker-based tmux send protocol from `BaseAgent` into `utils/tmux.py` as a standalone utility
- Introduce `TmuxCommandRunner` callable abstraction with two implementations: `CgCommandRunner` (local subprocess via ConcurrencyGroup) and `HostCommandRunner` (remote via host.execute_command/SSH)
- Both BaseAgent and standalone callers use the same protocol code path with different runners, eliminating duplication
- BaseAgent's `_send_message_with_marker`, `_capture_pane_content`, and all marker helpers now delegate to the shared utility

## Test plan

- [x] All 2465 tests pass (`cd libs/mng && uv run pytest`)
- [x] New `tmux_test.py` tests cover: capture, backspace, signal send/timeout, bad target, error formatting
- [x] Ruff and all ratchets clean


evgunter:
giving up on having the reviewer in a separate tmux pane

PR #645, opened 2026-02-27
Make config reading permissive

qi-imbue:


joshalbrecht:
LGTM *except* please default it to strict, and make a developer-only env var that we can set to get the non-strict mode when we're using it ourselves (and even when non-strict, make sure that it emits a warning)

PR #653, opened 2026-02-27
mark used resources in pytest

evgunter:
requires https://github.com/imbue-ai/mng/pull/652

## Summary

- Adds a resource guard system that enforces pytest marks for external tool dependencies (tmux, rsync, unison)
- PATH wrapper scripts intercept binary calls during the test call phase and block unmarked tests
- Must-use enforcement catches superfluous marks (tests marked but never invoking the resource)
- Blocked-invocation tracking catches tests that handle the guard's exit code gracefully
- Marks added to ~330 tests across 40+ files
- Fixes a CPython posix_spawn bug where `subprocess.Popen` intermittently uses stale environment (cpython#113119)
- **Found and fixed a real bug in `UnisonSyncer.is_running`** (see below)

## Bug found: UnisonSyncer.is_running was lying

The must-use enforcement caught a real bug: `UnisonSyncer.is_running` returned `True` as soon as `subprocess.Popen` created the OS process, before unison had actually initialized. Tests that called `start()` then immediately checked `is_running` and called `stop()` were killing the process within ~4ms of spawning -- before the unison binary even loaded. These tests appeared to pass but were testing nothing.

**Fix:** `is_running` now requires that unison has produced at least one line of output (tracked via a `threading.Event` set by the `on_output` callback). This ensures unison has actually started before callers can observe it as "running."

## How it works

1. At session start, bash wrapper scripts are created for each guarded binary and prepended to PATH
2. During the test call phase (not fixture setup/teardown), wrappers check env vars to allow/block
3. If a test invokes a resource without `@pytest.mark.<resource>`, the wrapper writes a `blocked_<resource>` tracking file and exits 127
4. After the test, `makereport` checks for blocked tracking files (missing mark) and unused allowed resources (superfluous mark), failing the test in either case

The blocked-invocation tracking is important: without it, a test that expects a failure (e.g., checking `is_running` returns False) could silently pass by handling the guard's exit 127 as a legitimate error. The tracking file ensures `makereport` catches it regardless. This caught additional tests that were missing marks across unit, integration, and acceptance tests.

## Demo

Five test scenarios showing the guard system in action:

```python
def test_no_mark_no_call():
    """No mark, no call -- should pass."""
    assert 1 + 1 == 2

@pytest.mark.tmux
def test_has_mark_and_calls_tmux():
    """Has @pytest.mark.tmux and actually calls tmux -- should pass."""
    result = subprocess.run(["tmux", "-V"], capture_output=True, text=True)
    assert result.returncode == 0

def test_no_mark_but_calls_tmux():
    """Missing @pytest.mark.tmux but calls tmux -- should FAIL (guard blocks it)."""
    result = subprocess.run(["tmux", "-V"], capture_output=True, text=True)
    assert result.returncode == 0

def test_no_mark_handles_error_gracefully():
    """Missing mark, but test handles the 127 -- should still FAIL via makereport."""
    result = subprocess.run(["tmux", "-V"], capture_output=True, text=True)
    assert result.returncode != 0

@pytest.mark.tmux
def test_has_mark_but_never_calls_tmux():
    """Has @pytest.mark.tmux but never calls tmux -- should FAIL (superfluous mark)."""
    assert 1 + 1 == 2
```

Results: 2 passed, 3 failed -- exactly as expected:

```
PASSED  test_no_mark_no_call
PASSED  test_has_mark_and_calls_tmux
FAILED  test_no_mark_but_calls_tmux
FAILED  test_no_mark_handles_error_gracefully
FAILED  test_has_mark_but_never_calls_tmux
```

**Missing mark (assertion failure)** -- the wrapper intercepts the call and exits 127:
```
_________________________ test_no_mark_but_calls_tmux __________________________

    def test_no_mark_but_calls_tmux():
        result = subprocess.run(["tmux", "-V"], capture_output=True, text=True)
>       assert result.returncode == 0
E       assert 127 == 0
E        +  where 127 = CompletedProcess(
E              args=['tmux', '-V'], returncode=127, stdout='',
E              stderr="RESOURCE GUARD: Test invoked 'tmux' without @pytest.mark.tmux mark.\n
E                      Add @pytest.mark.tmux to the test, or remove the tmux usage.\n"
E            ).returncode
```

**Missing mark (error handled gracefully)** -- the test's own assertion passes, but makereport catches the blocked tracking file:
```
____________________ test_no_mark_handles_error_gracefully _____________________
Test invoked 'tmux' without @pytest.mark.tmux.
Add @pytest.mark.tmux to the test, or remove the tmux usage.
```

**Superfluous mark** -- the tracking file was never created, so the test is failed after the call phase:
```
______________________ test_has_mark_but_never_calls_tmux ______________________
Test marked with @pytest.mark.tmux but never invoked tmux.
Remove the mark or ensure the test exercises tmux.
```

## Test plan

- [x] All 3010 tests pass with guards active
- [x] Blocking catches tests that use resources without marks
- [x] Blocked-invocation tracking catches tests that handle the exit code gracefully
- [x] Must-use catches tests with superfluous marks
- [x] Guards only active during call phase (fixtures can use tools freely)
- [x] Works with xdist parallel workers
- [x] UnisonSyncer.is_running fix verified with 10/10 passes on previously-failing tests


libs/imbue_common/imbue/imbue_common/resource_guards.py
```diff
@@ -0,0 +1,253 @@
...
+- Track invocation if the test has the mark (catches superfluous marks)
+
+Docker and Modal use Python SDKs (not CLI binaries), so they are not guarded here.
+
+Usage:
+    Call create_resource_guard_wrappers() during pytest_sessionstart and
+    cleanup_resource_guard_wrappers() during pytest_sessionfinish. Register the
+    three runtest hooks (pytest_runtest_setup, pytest_runtest_teardown,
+    pytest_runtest_makereport) into the conftest namespace.
+"""
```
evgunter: should i remove this

PR #654, opened 2026-02-27
make pinning-warnings-in-ls-to-the-bottom work for multiline warnings

evgunter:
before:
Screenshot 2026-02-27 at 2 50 36 PM
...
Screenshot 2026-02-27 at 2 51 06 PM

after:

https://github.com/user-attachments/assets/c25ae521-89bf-4cc8-b6c6-839af44b818b

(resizing while it's running to show that that mostly-works; it can strand warning lines but it can't eat list lines)

## Summary
- Warnings in `mng ls` that exceed the terminal width were leaving ghost copies between every agent row because the ANSI cursor-up logic only counted explicit `\n` characters, not terminal line wrapping
- Added `count_visual_lines()` to `terminal.py` that computes actual visual line count accounting for wrapping and ANSI escape codes
- The streaming renderer now uses this for correct cursor management

## Test plan
- [x] Regression test verifies `cursor-up(2)` is emitted for a 209-char warning at 120 columns (fails on pre-fix code, passes with fix)
- [x] All 2472 existing tests pass
- [x] 11 new unit tests for `count_visual_lines`

### DIFF
--- a/libs/mng/imbue/mng/utils/terminal.py
+++ b/libs/mng/imbue/mng/utils/terminal.py
@@ -55,6 +55,29 @@ def count_visual_lines(text: str, terminal_width: int) -> int:
 
 
 @pure
+def _tokenize_ansi_text(text: str) -> list[str]:
+    """Split text into single visible characters and multi-character ANSI escape sequences."""
+    tokens: list[str] = []
+    idx = 0
+    while idx < len(text):
+        if text[idx] == "\x1b":
+            match = _ANSI_ESCAPE_PATTERN.match(text, idx)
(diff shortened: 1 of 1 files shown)


joshalbrecht:
the amount of code for this feature sketches me out (it's also hard to read warnings that are being scrolled, and I've never see another program do this)

@qi-imbue what do other programs do in such a case? And ideas if there is a better cleaner way to do this?

I'm thinking even generically about the warning scrolling stuff--this current PR is even more cursed than that, which is why I wanted to take a step back and see if we were Doing It Wrong

PR #660, opened 2026-02-28
misuse the originally-for-tabcomplete cache to skip checking all providers for an agent if we found it in the cache

evgunter:
pretty sketchy and kind of a misuse of the cache

---
claude's summary:

## Summary

- Extends the agent completion cache to store which provider each agent belongs to, so commands like `mng exec <agent> "..."` can skip querying Docker (or other irrelevant providers)
- `find_agents_by_identifiers_or_state` now checks the cache first and only queries the relevant providers, falling back to a full scan on cache miss
- `load_all_agents_grouped_by_host` writes the extended cache on any full unfiltered scan (`provider_names is None`), so the cache is populated by all commands that do full scans -- including the fallback path on first use
- Moved agent cache functions from `cli/completion_writer.py` to `utils/agent_cache.py` to fix the import layer violation (`api` was importing from `cli`)

## Manual verification

Tested on a machine where Docker is installed but the daemon is not running:

1. `mng list` populates the cache with 82 agents (79 local, 3 modal)
2. `mng exec dont-double-bump "echo cache-test-success"` (a local agent):
   - Docker, Modal, and SSH providers all skipped (`Skipped backend docker (not in provider filter)`)
   - No Docker daemon warning
   - Command succeeded
3. `mng exec bench-ep-cache4 "echo hello"` (a stopped Modal agent):
   - First attempt correctly skips Docker, queries only modal+local
   - Agent not found because `find_agents_by_identifiers_or_state` uses `include_destroyed=False` which excludes stopped Modal hosts -- this is a **pre-existing issue** (same failure on `main` branch)
   - Falls back to full scan (triggers Docker warning) -- this fallback is correct behavior for a stale/incomplete cache

## Test plan

- [x] All 2472 tests pass (`cd libs/mng && uv run pytest`)
- [x] Unit tests for cache format, round-trip lookups (by name and ID), cache misses, corrupt files, old format backward compat
- [x] Manual verification: local agent exec skips Docker entirely, no warning
- [ ] Pre-existing: `mng exec` on stopped Modal agents fails on both `main` and this branch (separate issue)


evgunter:
doesn't handle multiple agents with the same name well if one is cached and the other is not

PR #661, opened 2026-02-28
add modal and docker resource guards

evgunter:
built on https://github.com/imbue-ai/mng/pull/653

## Summary

- Adds SDK-level resource guards that monkeypatch chokepoint methods in the Modal gRPC client (`UnaryUnaryWrapper.__call__`, `UnaryStreamWrapper.unary_stream`) and Docker API client (`APIClient.send`)
- Enforces the same two-way invariant as binary guards: tests using these SDKs must have the corresponding `@pytest.mark.modal` / `@pytest.mark.docker` mark, and tests with the mark must actually use the SDK
- All patches call `_enforce_sdk_guard()` which reuses the existing env var + tracking file mechanism -- no changes to the pytest hook implementations

## Test plan

- [x] 4 unit tests for `_enforce_sdk_guard` (block, allow, skip outside call phase, skip when no phase set)
- [x] 5 pytester end-to-end tests mirroring existing binary guard tests (marked+triggered passes, unmarked+triggered fails, catches-exception-still-fails, superfluous mark fails, no-mark-no-trigger passes)
- [x] Full monorepo test suite passes (3019 passed)
- [ ] CI validates existing `@pytest.mark.docker` and `@pytest.mark.modal` tests still work with guards active

### DIFF
--- a/libs/mng/imbue/mng/providers/docker/instance_test.py
+++ b/libs/mng/imbue/mng/providers/docker/instance_test.py
@@ -1,10 +1,14 @@
 import json
+from datetime import datetime
+from datetime import timezone
 from pathlib import Path
 
 import pytest
 
 from imbue.mng.config.data_types import MngContext
 from imbue.mng.errors import MngError
+from imbue.mng.hosts.offline_host import OfflineHost
+from imbue.mng.interfaces.data_types import CertifiedHostData
 from imbue.mng.primitives import HostId
 from imbue.mng.primitives import HostName
 from imbue.mng.primitives import ProviderInstanceName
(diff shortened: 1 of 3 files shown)


uv.lock
```diff
@@ -1836,7 +1836,7 @@ requires-dist = [
     { name = "concurrency-group", editable = "libs/concurrency_group" },
     { name = "coolname", specifier = ">=3.0.0" },
     { name = "cryptography", specifier = ">=42.0" },
-    { name = "docker", specifier = ">=7.0" },
+    { name = "docker", specifier = "==7.1.0" },
```
evgunter: we're doing stuff to docker internals here, so pinning

PR #705, opened 2026-03-03
Add plugin CLI option support: param filtering and flag_value

evgunter:
## Summary

- Fix `setup_command_context` to handle plugin-registered CLI options gracefully
- Before constructing the command options class, filter `ctx.params` to only include fields known to `command_class`, preventing pydantic validation errors from extra plugin keys
- Store plugin params in `ctx.meta["plugin_cli_params"]` so plugins can access their registered option values

## Details

When plugins register custom CLI options via `register_cli_options`, click parses them into `ctx.params`. But `setup_command_context` passed all of `ctx.params` to `command_class(...)` (a pydantic FrozenModel with `extra="forbid"`), causing validation errors for the unknown plugin keys.

The fix adds `_split_known_and_plugin_params()` which uses `command_class.model_fields` to partition params. This is called twice: once before the initial options construction, and again after config defaults and plugin overrides are applied. The separated plugin params are stored in `ctx.meta["plugin_cli_params"]`, consistent with the existing `ctx.meta` pattern used for `logging_config`, `hook_command_name`, etc.

Lifecycle hooks (`on_before_command`, `on_after_command`, `on_error`) continue to receive the full params dict unchanged.

## Test plan

- [x] 3 unit tests for `_split_known_and_plugin_params` (separates known/extra, all known, empty params)
- [x] 3 integration tests exercising the full path: plugin registers CLI option, command invoked, `setup_command_context` succeeds, plugin params accessible in `ctx.meta`
- [x] Full test suite: 2972 passed, 1 skipped, 84.14% coverage

### DIFF
--- a/libs/mng/imbue/mng/plugins/hookspecs.py
+++ b/libs/mng/imbue/mng/plugins/hookspecs.py
@@ -218,6 +218,11 @@ class OptionStackItem(FrozenModel):
         default=None,
         description="Environment variable to read value from",
     )
+    flag_value: Any = Field(
+        default=None,
+        description="Value to use when the option is provided without an argument. "
+        "Enables dual flag/value behavior when set (e.g., --opt uses flag_value, --opt VALUE uses VALUE).",
+    )
 
     def to_click_option(self, group: OptionGroup | None = None) -> click.Option:
         """Convert this spec to a click.Option instance.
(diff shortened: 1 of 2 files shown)


libs/mng/imbue/mng/plugins/hookspecs.py
```diff
@@ -218,6 +218,11 @@ class OptionStackItem(FrozenModel):
         default=None,
         description="Environment variable to read value from",
     )
+    flag_value: Any = Field(
+        default=None,
+        description="Value to use when the option is provided without an argument. "
+        "Enables dual flag/value behavior when set (e.g., --opt uses flag_value, --opt VALUE uses VALUE).",
+    )
```
evgunter: claude said it needed this for the plugin i was working on

PR #706, opened 2026-03-03
pull claude utility out of ask and make it slightly less bad

evgunter:
## Summary

`ask.py` previously managed the Claude subprocess directly via `subprocess.Popen`, with its own `SubprocessClaudeBackend` class, `ClaudeBackendInterface` ABC, and `_extract_text_delta` helper. This PR extracts that logic into a shared utility at `libs/mng/imbue/mng/utils/claude.py` so other consumers can reuse it.

The new utility provides two functions:
- `query_claude()`: non-streaming, returns complete response text or None on failure
- `query_claude_streaming()`: streaming, yields text chunks as they arrive

The key behavioral change is that both functions use `ConcurrencyGroup` for subprocess lifecycle management instead of raw `subprocess.Popen`. This means the spawned Claude process is tracked by the concurrency group and cleaned up properly on shutdown, rather than relying on manual `terminate()`/`kill()` logic.

`ask.py` now calls `query_claude_streaming()` directly, removing ~120 lines of subprocess management code.

## Test plan
- [x] All 2968 existing tests pass (1 skipped), 84.23% coverage
- [x] New `claude_test.py` covers `_extract_text_delta` edge cases and error paths for both `query_claude` and `query_claude_streaming`
- [x] Updated `ask_test.py` with new `_FakeQueryStreaming` test double
- [x] Type checker (`ty check`) passes with no errors
- [x] Manually verified `mng ask "how do I list my agents?"` in tmux -- streams response and exits cleanly

### DIFF
--- a/libs/mng/imbue/mng/utils/claude.py
+++ b/libs/mng/imbue/mng/utils/claude.py
@@ -1,5 +1,4 @@
 import json
-import shutil
 import tempfile
 from collections.abc import Iterator
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


evgunter:
doing https://github.com/imbue-ai/mng/pull/785 instead

PR #723, opened 2026-03-03
require non-ancient bash for stream_transcript.sh

evgunter:
## Summary

- Fix `stream_transcript.sh` failing on macOS when launched via shebang (`#\!/bin/bash` resolves to bash 3.2 which lacks `declare -A`). Changed `claude_background_tasks.sh` to invoke with `bash "$STREAM_SCRIPT"` so PATH-resolved bash (Homebrew 5+) is used.
- Add bash 4+ version check to `install.sh` integrated into the existing dependency installation flow. If PATH-resolved bash is too old, it shows up in the missing deps prompt and gets installed via `brew install bash` / `apt-get install bash`.

## Test plan

- [x] Existing 22 stream_transcript tests pass (they already invoke via PATH-resolved bash)
- [x] Full libs/mng suite passes (3029 passed, 1 skipped)
- [x] Manual: confirmed `/bin/bash` (3.2.57) fails on `stream_transcript.sh` line 45 (`declare: -A: invalid option`, exit code 2), while PATH-resolved `bash` (5.3.3) succeeds
- [ ] Manual: verify install.sh prompts for bash on a system with only bash 3.2 on PATH

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -216,7 +216,8 @@ if [ "$_NEED_MODERN_BASH" = true ]; then
             warn "PATH-resolved bash is still version $_POST_BASH_VER after install."
             warn "Ensure /opt/homebrew/bin (Apple Silicon) or /usr/local/bin (Intel) is before /bin in your PATH."
         else
-            warn "PATH-resolved bash is still version $_POST_BASH_VER. Please install bash 4+ and ensure it is on your PATH."
+            warn "PATH-resolved bash is still version $_POST_BASH_VER after install."
+            warn "Ensure the newly installed bash is before the old one in your PATH."
         fi
     fi
 fi


scripts/install.sh
```diff
@@ -186,6 +206,22 @@ else
...
+
+if [ "$_NEED_MODERN_BASH" = true ]; then
+    _POST_BASH_VER="$(bash -c 'echo ${BASH_VERSINFO[0]}' 2>/dev/null || echo 0)"
+    if [ "$_POST_BASH_VER" -lt 4 ] 2>/dev/null; then
+        if [ "$OS" = "macos" ]; then
+            warn "PATH-resolved bash is still version $_POST_BASH_VER after install."
+            warn "Ensure /opt/homebrew/bin (Apple Silicon) or /usr/local/bin (Intel) is before /bin in your PATH."
+        else
+            warn "PATH-resolved bash is still version $_POST_BASH_VER after install."
+            warn "Ensure the newly installed bash is before the old one in your PATH."
```
evgunter: i dont know if trying to apt-get install bash would even fix it if you're on linux and your bash is ancient, so perhaps this message is misleading, but also i hope no one will ever see it

PR #736, opened 2026-03-04
write agent command to run.sh so it looks nicer when we run it with send-keys

evgunter:
just look at it the summary is longer than the diff

---

## Summary

- Write the full agent startup command to `$MNG_AGENT_STATE_DIR/run.sh` instead of sending it as literal keys to the tmux pane
- The tmux command is now just `bash "$MNG_AGENT_STATE_DIR/run.sh"` instead of the full multi-part command
- Avoids `#\!/bin/bash` shebang to prevent macOS bash 3.2 issues (script is always invoked explicitly with `bash`)
- Hitting up+enter in the tmux pane still re-runs the same command

## Test plan

- [x] All 2892 unit/integration tests pass
- [x] Manually verify a `mng create` session starts correctly and shows the short command in the tmux pane
- [x] Verify up+enter re-runs the command successfully


weishi-imbue:
Please continue this PR on imbue-ai/mngr-internal#73 — development has moved to the private repository (this public repository is now a read-only mirror, so PRs cannot merge here). The branch and this PR's description were carried over.

PR #740, opened 2026-03-04
Launch a TUI to install plugins during installation

qi-imbue:
This adds a new `mng plugin install-wizard` command, which just shows a hardcoded list of recommended plugins for now.

In future it may make sense to have a central registry of plugins (a JSON file hosted in the Git repo should work well enough) and have a full-fledged TUI for managing plugins, but this feels like a reasonable start.


joshalbrecht:
this is super cool, thanks! :)

PR #771, opened 2026-03-05
Switch default command to an empty command

qi-imbue:
Most of the changes here are actually doc changes. I'm a bit surprised that the default subcommand of `mng` is actually coupled with the default sub-subcommand of `mng snapshot` (because they are both `create`) but the code delta looks reasonable.

Tested manually, `mng` now prints help text and `mng snapshot` is still equivalent to `mng snapshot create`.


joshalbrecht:
Let's make them consistent at least?

```By default, `mng` with no subcommand shows help, while `mng snapshot` defaults to `create`.```

seems like it should default to the same behavior?  other than that, lgtm, feel free to merge

qi-imbue:
Changed `mng snapshot` too! Merging.

PR #774, opened 2026-03-05
Trivial flag reorganizations for mng create

qi-imbue:
This PR batches all the relatively straightforward flag changes discussed in https://www.dropbox.com/scl/fi/k0k3cgmw8apspk5p94vt7/Simplifying-mng-create.paper?rlkey=cx9kndyfsjbe7ythrur6ej6gd&dl=0:

- Renames
- Removing aliases
- Moving flags to a different section

Since we decided to remove the `--json` and `--jsonl` aliases, for consistency we're also removing those aliases for all the other subcommands.

I've skimmed at the entire diff and it doesn't seem to be doing anything beyond this scope.


joshalbrecht:
LGTM after at least noticing my one comment and adding it to the backlog of things to fix!

PR #790, opened 2026-03-06
Basic E2E tests

qi-imbue:


libs/skitwright/README.md
```diff
@@ -0,0 +1,83 @@
...
+```
+
+## Transcript format
+
+Each test session produces a transcript with annotated lines:
+
+- `$ ` prefixes the shell command that was run
+- `  ` (two spaces) prefixes each line of stdout
+- `! ` prefixes each line of stderr
+- `? ` prefixes the exit code
```
joshalbrecht: just add "#: prefixes a comment", and then this should be possible to easily transform from mega_tutorial!  

It's important to express "intent" so that the LLM kinda knows what was supposed to happen

PR #820, opened 2026-03-10
tab completion for "everything"

evgunter:
https://github.com/user-attachments/assets/bf2d21bb-f587-4ad1-b503-a6f73f029a1d

## Summary

Extends the tab completion system with comprehensive completion support for CLI options, positional arguments, and config values.

### New completion sources
- **Host names** for create --host, create --target-host (from discovery stream), and events positional
- **Plugin names** for --plugin/--enable-plugin/--disable-plugin options and plugin enable/disable positional
- **Config keys** for config get/set/unset positional (flattened from MngConfig)
- **Config values** for config set KEY VALUE -- offers valid values based on field type:
  - Bool fields: true/false
  - Enum fields: enum values (e.g. TRACE/DEBUG/BUILD/INFO/WARN/ERROR/NONE for log levels)
  - AgentTypeName fields: registered agent type names
  - ProviderBackendName fields: registered backend names
  - Dynamic dict-keyed fields (plugins.modal.enabled, providers.modal.is_enabled, etc.)
- **Dynamic option choices** for create --type, --template, --in, --new-host and list --provider
- Derived config keys (disabled_plugins, enabled_backends) excluded from completion

### Per-position positional completion
- Different positional argument positions can have different completion sources
- `config set <TAB>` offers config keys, but `config set prefix <TAB>` correctly offers nothing (freeform VALUE)
- Positional nargs limits extracted from click.Argument params, enforced at completion time
- Handles interleaved options: `config set --scope user KEY <TAB>` correctly counts positionals

### Architecture improvements
- `CompletionCacheData` NamedTuple as shared cache schema (completion_cache.py)
- `COMPLETION_CACHE_FILENAME` and `get_completion_cache_dir()` deduplicated into shared module
- `_CompletionContext` NamedTuple separates parsing from dispatch in the completer
- `_get_option_value_candidates` and `_get_positional_candidates` extracted as clean helpers
- Cache write runs in background via concurrency_group (does not block list output)
- `_build_dynamic_completions` moved into completion_writer.py (next to its consumer)
- `_FIELD_TYPE_COMPLETION_SOURCES` registry for extensible type-to-source mapping

### Integration tests
- Reverse test: every --long option on the real CLI appears in the cache
- Forward tests: every hand-maintained constant (git_branch_options, host_name_options, plugin_name_options, option_choices) references options that actually exist

## Test plan
- [x] ~80 new tests across complete_test.py, completion_writer_test.py, and test_completion.py
- [x] Full test suite passes (3223 passed)
- [x] NamedTuple ratchet bumped from 0 to 2 (CompletionCacheData + _CompletionContext, both avoiding pydantic in the lightweight completion hot path)

### DIFF
--- a/libs/mng/imbue/mng/config/completion_writer.py
+++ b/libs/mng/imbue/mng/config/completion_writer.py
@@ -67,8 +67,7 @@ _GIT_BRANCH_OPTIONS: Final[frozenset[str]] = frozenset(
 # event stream. Uses the same "command.--option" notation.
 _HOST_NAME_OPTIONS: Final[frozenset[str]] = frozenset(
     {
-        "create.--host",
-        "create.--target-host",
+        "create.--target",
     }
 )
 
@@ -95,7 +94,7 @@ _EXCLUDED_CONFIG_KEY_PREFIXES: Final[frozenset[str]] = frozenset(
 _DYNAMIC_CHOICE_OPTIONS: Final[dict[str, str]] = {
     "create.--type": "agent_type_names",
     "create.--template": "template_names",
-    "create.--in": "provider_names",
+    "create.--provider": "provider_names",
     "create.--new-host": "provider_names",
     "list.--provider": "provider_names",
 }
--- a/libs/mng/imbue/mng/config/completion_writer_test.py
+++ b/libs/mng/imbue/mng/config/completion_writer_test.py
@@ -107,15 +107,14 @@ def _read_cache(cache_dir: Path) -> dict:
 
 
 def test_write_cli_completions_cache_includes_host_name_options(completion_cache_dir: Path) -> None:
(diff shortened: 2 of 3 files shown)


libs/mng/imbue/mng/cli/complete.py
```diff
@@ -14,13 +14,32 @@
 import subprocess
 import sys
 from pathlib import Path
+from typing import Any
+from typing import NamedTuple
```
evgunter: didn't want pydantic in something that gets loaded all the time

joshalbrecht:
LGTM after you and Qi resolve the above comment  (all I'll say is that it is ok for us to have config vars that do not end up being parsed on the command line of a command, eg, some esoteric option for `mng create`, but I see no reason why valid config vars would not be accepted by `mng config set` )

PR #840, opened 2026-03-12
don't print scary errors when interrupting concurrencygroup in mng

evgunter:
## Summary
This PR adds signal handling to gracefully manage ConcurrencyGroup cleanup when a user interrupts a CLI command with Ctrl+C (SIGINT). It distinguishes between intentional shutdowns (where exceptions during cleanup are suppressed) and real failures (where exceptions are propagated).

## Key Changes
- **Signal handler installation**: Added `_install_sigint_shutdown_handler()` to install a custom SIGINT handler that sets the ConcurrencyGroup's shutdown event before delegating to the original handler. This only works from the main thread.
- **Graceful cleanup**: Added `_close_concurrency_group()` to handle ConcurrencyGroup exit with special logic:
  - Suppresses `ConcurrencyExceptionGroup` exceptions when the group is shutting down (Ctrl+C case)
  - Re-raises exceptions when not shutting down (real failures)
  - Always restores the original SIGINT handler
- **Signal handler restoration**: Added `_restore_sigint_handler()` to safely restore the original handler, with defensive error handling for non-main thread scenarios
- **Module-level state**: Added `_sigint_handler_state` dict to track the active shutdown event and original handler (necessary because signal handlers are process-global)
- **Integration**: Updated `setup_command_context()` to install the SIGINT handler and use the new cleanup function

## Implementation Details
- The SIGINT handler is only installed from the main thread (signal.signal() requirement)
- Module-level state is used for the handler because signal handlers are inherently global per-process
- Click exceptions are still passed as None to `__exit__()` to prevent wrapping in ConcurrencyExceptionGroup
- Comprehensive test coverage includes: suppression behavior, exception propagation, handler restoration, and non-main-thread scenarios

https://claude.ai/code/session_01KMErLZdtV8Kkvr7rfLBXQX


weishi-imbue:
Please continue this PR on imbue-ai/mngr-internal#75 — development has moved to the private repository (this public repository is now a read-only mirror, so PRs cannot merge here). The branch and this PR's description were carried over.

PR #905, opened 2026-03-18
Fix auto-generated host name collisions with existing hosts

qi-imbue:
When creating a host with an auto-generated name (e.g. `mng c @.modal`), the name was generated randomly without checking against existing hosts, which could produce collisions. Now discovers existing hosts on the provider first and retries name generation until a unique name is found.

Fixed-name providers (like local, which always returns "localhost") are detected by observing that get_host_name returns the same value on retry, and the name is accepted as-is since such providers reuse hosts by design.


weishi-imbue:
Please continue this PR on imbue-ai/mngr-internal#77 — development has moved to the private repository (this public repository is now a read-only mirror, so PRs cannot merge here). The branch and this PR's description were carried over.

PR #949, opened 2026-03-23
let agents directly use existing branches in worktree mode

evgunter:
## Summary

- Previously, `--worktree` mode unconditionally required a new branch (`--branch BASE:NEW`). Now `--branch EXISTING_BRANCH` works: if the branch exists and isn't checked out in another worktree, it's checked out directly.
- When a branch *is* already checked out, the error message now includes git's own diagnostic and suggests alternatives: `--branch BASE:` to create a new branch, or `--in-place` from the existing worktree directory.
- `created_branch_name` is correctly set to `None` for existing branch checkouts, so `mng destroy --remove-created-branch` won't delete a branch the user created independently.
- Removes unnecessary `git worktree remove` cleanup from tests (pytest tmpdir handles it).

## Test plan

- [x] Unit tests pass (3035 passed)
- [x] `test_worktree_with_existing_branch` -- checks out existing branch, verifies `created_branch_name` is None
- [x] `test_worktree_already_checked_out_gives_helpful_error` -- verifies improved error message
- [ ] CI passes

### DIFF
--- a/libs/mng_recursive/imbue/mng_recursive/watcher_common.py
+++ b/libs/mng_recursive/imbue/mng_recursive/watcher_common.py
@@ -302,9 +302,7 @@ def mtime_poll_directories(
     return is_changed
 
 
-_NON_CHANGE_EVENT_TYPES: Final[frozenset[str]] = frozenset(
-    {"opened", "closed", "closed_no_write"}
-)
+_NON_CHANGE_EVENT_TYPES: Final[frozenset[str]] = frozenset({"opened", "closed", "closed_no_write"})
 
 
 class ChangeHandler(FileSystemEventHandler):
--- a/libs/mng_recursive/imbue/mng_recursive/watcher_common_test.py
+++ b/libs/mng_recursive/imbue/mng_recursive/watcher_common_test.py
@@ -4,6 +4,7 @@ import json
 import os
 import threading
 from pathlib import Path
+
 import pytest
 from loguru import logger
 from watchdog.events import FileClosedEvent


libs/mng/imbue/mng/hosts/host.py
```diff
@@ -1592,25 +1592,43 @@ def _create_work_dir_as_git_worktree(
...
+            if base_branch:
+                worktree_args.append(shlex.quote(base_branch))
+            cmd = " ".join(worktree_args)
+            created_branch = new_branch_name
 
             result = self.execute_command(cmd)
             if not result.success:
-                raise MngError(f"Failed to create git worktree: {result.stderr}")
+                stderr = result.stderr or ""
+                if "already checked out" in stderr or "already used by worktree" in stderr:
```
evgunter: possibly fragile

libs/mng/imbue/mng/hosts/host.py
```diff
@@ -1592,25 +1592,42 @@ def _create_work_dir_as_git_worktree(
...
+                    raise UserInputError(
+                        f"{stderr.strip()}\n"
+                        f"To create a new branch instead, use --branch BASE: or --branch BASE:new-name"
```
evgunter: will be somewhat confusing in the case that the error is caused by something like `--branch BASE:another-branch-that-already-exists`

libs/mng/imbue/mng/api/test_create.py
```diff
@@ -282,55 +283,46 @@ def test_create_agent_with_worktree(
...
-            branch_name = result.stdout.strip()
-            assert branch_name.startswith("mng/")
-            assert str(agent_name) in branch_name
-        finally:
-            if worktree_path is not None:
-                subprocess.run(
-                    ["git", "worktree", "remove", "--force", str(worktree_path)],
-                    cwd=temp_git_repo,
-                    capture_output=True,
-                )
```
evgunter: pretty sure we don't need to do this since the whole temp dir is cleaned up anyway

PR #955, opened 2026-03-25
let users specify gitignored files to symlink/copy in the project repo

evgunter:
## Summary

- Adds `work_dir_extra_paths` config field: a dict mapping relative paths to transfer mode (`"SHARE"` or `"COPY"`)
- `"SHARE"` creates symlinks on same host, falls back to rsync on different hosts
- `"COPY"` always uses rsync for independent per-work-dir copies
- Integrated into both worktree and copy creation paths
- Config merging follows per-key override (same as `pre_command_scripts`)
- All remote operations are batched to minimize SSH round trips (at most 3 commands regardless of path count)
- Implementation uses `execute_command` uniformly -- no `is_local` branching

Example config:
```toml
[work_dir_extra_paths]
".mng/settings.local.toml" = "SHARE"
".venv" = "COPY"
".test_output" = "COPY"
```

### Remote operation batching

- Source-exists check: single shell command tests all paths, outputs those that exist
- Symlink creation: single shell script handles conflict detection, parent dir creation, and symlink creation for all SHARE paths
- File copying: single rsync call with `--files-from` for all COPY paths

## Test plan

- [x] Unit tests for config defaults, merging, parsing, and TOML round-trip (loader regression tests)
- [x] Unit tests for symlink creation, idempotency, missing source warning, non-symlink conflict, copy mode, absolute path rejection, path traversal rejection, nested path parent creation
- [x] Full test suite passes (3294 passed, 82.72% coverage)
- [x] Manual verification: created a worktree agent with `work_dir_extra_paths` config and verified:

| Path | Mode | Result |
|------|------|--------|
| `.mng/settings.local.toml` | `SHARE` | Symlink to source (same host) |
| `.venv` | `COPY` | Independent copy via rsync |
| `.test_output` | `COPY` | Independent copy via rsync |


libs/mng/imbue/mng/hosts/host.py
```diff
@@ -1432,16 +1437,108 @@ def _transfer_extra_files(
...
+            normalized = os.path.normpath(rel_path_str)
+            if os.path.isabs(normalized):
+                raise UserInputError(f"work_dir_extra_paths: absolute paths are not allowed: {rel_path_str}")
+            if normalized.startswith(".."):
+                raise UserInputError(f"work_dir_extra_paths: path escapes project root: {rel_path_str}")
+
+            source_abs = source_path / rel_path_str
+            result = source_host.execute_command(
+                f"test -e {shlex.quote(str(source_abs))} || test -L {shlex.quote(str(source_abs))}"
+            )
```
joshalbrecht: we dont want to be doing this--I just fixed a bunch of this kind of stuff (for loops that execute remote commands). They make the deploys really slow

Could the remote case get a more automatic lumping of this data into the rsync commands? We're already calling rsync anyway...

PR #995, opened 2026-03-26
skip unnecessary providers if we already know the agents we want to discover

evgunter:
## Summary

- When targeting a specific agent (e.g. `mng exec my-agent "echo hi"`), the system previously queried ALL providers (local, docker, modal, ssh) to find it. This is slow when Docker/Modal are configured but the agent is local.
- Uses the discovery event stream (`~/.mng/events/mng/discovery/events.jsonl`) to resolve agent identifiers to their provider names before querying. Only the relevant providers are then queried.
- Falls back to a full scan if the event stream is stale or missing (no regression).
- Renames `discover_all_hosts_and_agents` -> `discover_hosts_and_agents` with explicit parameters (no defaults).

## Timing results

Measured on a machine with local + modal providers configured (Docker daemon down):

| Scenario | Time (3-run avg) |
|---|---|
| Full scan (all providers) | ~3.9s |
| Optimized (local only) | ~1.9s |
| **Speedup** | **~2x** |

Command: `mng exec ui "echo test"` targeting a local agent.

The optimization resolves "ui" to provider "local" via the event stream and skips querying Modal entirely. With Docker daemon running, the speedup would be even larger (Docker discovery adds more latency than Modal).

## Test plan

- [x] All existing tests pass in libs/mng (3294 passed, 82.73% coverage)
- [x] All existing tests pass in libs/mng_wait (124 passed, 80.47% coverage)
- [x] All existing tests pass in libs/mng_kanpan (269 passed, 85.97% coverage)
- [x] All existing tests pass in libs/mng_file (122 passed, 78.76% coverage)
- [x] All existing tests pass in libs/mng_schedule (234 passed, 77.85% coverage)
- [x] New unit tests for `resolve_provider_names_for_identifiers()` (10 tests)
- [x] Manual timing test with optimization enabled vs disabled


joshalbrecht:
(waiting on flag that we discusssed IRL, otherwise LGTM)

evgunter:
man i don't know how they keep committing those. fixing

PR #1055, opened 2026-03-30
Followup for - support

qi-imbue:
- Remove builtin support for filtering from subcommands that support the `-` placeholder.
- Mention the `-` placeholder in error messages.

Claude also identified `archive` as a candidate for this change, but since its `--all` actually only looks for stopped agents, I've not included it in this PR. We can remove its `--all` too and the user needs to either:

- `mngr list --stopped --ids | mngr archieve -`
- Or `mng list --ids | mngr archieve -` and live with the warnings about running agents


qi-imbue:
Mentioning `-` support in the help texts explicitly now.

PR #1100, opened 2026-03-31
Enable offload acceptance tests and print failure logs in CI

jsk11235:
## Summary
- Uncomments the previously disabled offload acceptance test run step
- Adds a step to print failure logs via `offload -c offload-modal-acceptance.toml logs --failures` when the run fails, so test output is visible directly in the CI job

## Test plan
- [ ] Verify the offload acceptance workflow triggers on a PR
- [ ] Confirm failure logs are printed when tests fail

### DIFF
--- a/.github/workflows/test-offload-acceptance.yml
+++ b/.github/workflows/test-offload-acceptance.yml
@@ -54,6 +54,9 @@ jobs:
             cargo install offload@0.4.0
           fi
 
-#      - name: Run acceptance tests via offload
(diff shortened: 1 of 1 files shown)


weishi-imbue:
Please continue this PR on imbue-ai/mngr-internal#78 — development has moved to the private repository (this public repository is now a read-only mirror, so PRs cannot merge here). The branch and this PR's description were carried over.

PR #1118, opened 2026-04-01
Skip code review stop hook when there are no changes to review

qi-imbue:
Add two fast paths to stop_hook_gates.sh:

1. "Last passed" check: when all gates pass (or diff is empty), record the HEAD hash. On subsequent invocations, if HEAD matches, skip immediately. This handles "no changes since last review."

2. Empty diff check: if the branch has no changes relative to the base branch, skip all gates. There is nothing to autofix, no architecture to verify, and a no-op session is not worth conversation review.

Previously, even when an agent made no changes, missing marker files would cause the stop hook to block and force the agent to spawn review skills that each independently discovered there was nothing to do.


weishi-imbue:
Please continue this PR on imbue-ai/mngr-internal#79 — development has moved to the private repository (this public repository is now a read-only mirror, so PRs cannot merge here). The branch and this PR's description were carried over.

PR #1175, opened 2026-04-05
implement mngr schedule run and mngr schedule remove

evgunter:
based on https://github.com/imbue-ai/mngr/pull/1203 (for the e2e test)

## Summary

- Implements `mngr schedule run` -- executes a scheduled trigger immediately on its deployed provider
  - Local triggers: executes `run.sh` (same code path as cron)
  - Modal triggers: invokes the deployed function via `modal.Function.from_name().remote()` (same code path as Modal cron)
- Implements `mngr schedule remove` -- removes triggers and all associated artifacts
  - Local: removes crontab entry, trigger directory, and creation record
  - Modal: stops the Modal app and deletes the creation record from the state volume
- Adds `run` to `ScheduledMngrCommand` so triggers can use `mngr run` (headless agents) in addition to create/start/message/exec
- Extracts shared `load_schedule_provider()` utility to reduce duplication across schedule subcommands
- Adds lookup-by-name functions for both local and modal providers

## Bug fixes (pre-existing)

- Fixed `list_schedule_creation_records` doubled volume path prefix -- `schedule list --provider modal` was always returning empty results
- Fixed `stage_deploy_files` PermissionError on stale read-only git objects (wipe staging dir before re-staging)
- Removed invalid `--await-ready` flag from release test args (flag was removed from CLI in March)
- Improved `cron_runner.py` error diagnostics: `RuntimeError` now includes the last 50 lines of command output

## Behavioral change: `run_scheduled_trigger()` return value

`run_scheduled_trigger()` (the Modal cron function) now returns the full captured stdout+stderr of the mngr command it runs, instead of returning `None`. This allows `mngr schedule run` to display the command output to the user. The return value is propagated through `fn.remote()` -> `invoke_modal_trigger_function()` -> CLI stdout.

## Test plan

- [x] Unit tests for local trigger execution (run_test.py)
- [x] Unit tests for local trigger removal (remove_test.py)
- [x] Unit tests for lookup-by-name functions (deploy_test.py)
- [x] Unit tests for shared provider utility (provider_utils_test.py)
- [x] Release test: full lifecycle -- add trigger, list it, run it (verifies "hello-from-schedule-run" in output via `mngr run headless_command`), remove it, verify it's gone from list. Passes locally.
- [ ] CI passes all tests including ratchets


libs/mngr_schedule/imbue/mngr_schedule/testing.py
```diff
@@ -0,0 +1,190 @@
...
+def resolve_modal_environment(deploy_output: str) -> str:
+    """Extract the Modal environment name from schedule add output.
+
+    Parses the 'env: <name>' from the deploy log line. Falls back to
+    'main' if not found.
+    """
+    match = re.search(r"env:\s*(\S+)\)", deploy_output)
+    if match:
+        return match.group(1)
+    return "main"
```
evgunter: fragile; this is part of something intended patch the preexisting bug where some test envs were getting stranded

joshalbrecht:
I'm very tempted to get rid of mngr schedule, actually... I'm not using it for anything actively, and I think we're going  to have better, simpler ways to do this

evgunter:
hmm how would you suggest i set up the changelog agent then? i have it using mngr schedule

joshalbrecht:
lgtm after our discussion!

PR #1187, opened 2026-04-06
say to record whole output of commands before filtering in claude.md

evgunter:
## Summary
- Adds a bullet to "Important commands and conventions" advising to never pipe non-instant commands directly through filters (grep, jq, awk) without preserving the full output
- Recommends writing to a file first or using `tee`, so that if the filter misses something you don't have to rerun the entire command

## Test plan
- [ ] Read the new bullet and confirm it's clear and well-placed

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -28,6 +28,7 @@ Only after doing all of the above should you begin writing code.
 # Important commands and conventions:
 
 - Never run `uv sync`, always run `uv sync --all-packages` instead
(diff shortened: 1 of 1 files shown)


joshalbrecht:
add a bit more there with head / tail, etc, then LGTM

maybe we could even give it a nice convenient form that it should generally use for long-running commands... (eg, that also captures exit code, makes files, etc)

evgunter:
added head/tail mention

PR #1193, opened 2026-04-07
skip disabled providers with unknown backends

evgunter:
## Summary
- `is_enabled = false` on a provider block in settings config now correctly prevents backend resolution in `_parse_providers()`, avoiding `UnknownBackendError` when the plugin isn't installed
- Uses `list_registered_provider_backend_names()` to check upfront whether the backend is known, and skips disabled providers with unknown backends before attempting resolution
- When the backend IS installed, `is_enabled=false` is preserved in the parsed config so config layer merging works correctly (a higher-priority layer's `is_enabled=false` properly overrides a lower-priority layer's `is_enabled=true`)
- Adds 4 unit tests covering the fix and edge cases

## Test plan
- [x] `test_parse_providers_skips_disabled_provider_with_unknown_backend` -- disabled provider with unknown backend is silently skipped
- [x] `test_parse_providers_preserves_disabled_provider_with_known_backend` -- disabled provider with known backend preserved for merge
- [x] `test_parse_providers_still_raises_on_unknown_backend_when_enabled` -- error still raised when is_enabled=true
- [x] `test_parse_providers_still_raises_on_unknown_backend_when_is_enabled_unset` -- error still raised when is_enabled not set
- [x] Full test suite: 3468 passed, 1 skipped, 82.85% coverage


joshalbrecht:
lgtm

PR #1196, opened 2026-04-07
fix mngr ask startup race 

evgunter:
thanks @hynek-urban for reporting this bug!

---

## Summary
- Fix startup race condition where `mngr ask` would fail immediately on slower machines because the lifecycle check saw bash (not claude) and concluded the agent was done before claude started
- Add 10-second startup grace period before trusting lifecycle state
- Regression tests for the grace period logic

## Startup race fix

The lifecycle detection checks the tmux pane's `current_command`. On slower machines (or first run via nvm), claude takes seconds to start. During that time, bash is the current command, which the lifecycle check interprets as DONE. This causes `mngr ask` to immediately report "claude exited without producing output" before claude even begins.

The fix adds a two-phase poll in `_wait_for_stdout_file`:
1. **Phase 1 (grace period, 10s):** Only check if the stdout file exists, ignoring lifecycle state entirely
2. **Phase 2 (remaining timeout):** Also check lifecycle state, so we don't wait forever if claude genuinely failed

This was diagnosed from a real user report where `claude --print ... "just say hi"` worked fine directly but `mngr ask "just say hi"` failed immediately.

## Test plan
- [x] All headless unit tests pass (including 3 new grace period regression tests)
- [x] All ratchets pass
- [x] Manual verification: `mngr ask "just say hi"` returns "Hi!"
- [ ] CI


joshalbrecht:
gross, but lgtm

PR #1200, opened 2026-04-07
fix bug where normal terminal was detected as non-interactive

evgunter:
## Summary

- Add `has_interactive_terminal()` to `urwid_utils` that checks `sys.stdin.isatty()` then probes `/dev/tty`, fixing the false "No interactive terminal detected" message when running via `uv run` or other piped-stdin contexts
- Extend `create_urwid_screen_preserving_terminal()` to open `/dev/tty` as urwid input when stdin is piped, so all TUI callers benefit

## Test plan

- [ ] CI passes
- [ ] Manual: `uv run mngr plugin install-wizard` shows the TUI instead of "No interactive terminal detected"


joshalbrecht:
lgtm

PR #1205, opened 2026-04-08
don't clobber plugins in install script

evgunter:
## Summary
- Use `uv tool upgrade` instead of `uv tool install` when mngr is already installed
- `uv tool install imbue-mngr` wipes all `--with` plugin dependencies, breaking idempotency
- `uv tool upgrade` preserves the existing receipt (including plugin extras)

## Test plan
- [x] Verified `uv tool upgrade imbue-mngr` preserves all plugin extras in receipt
- [x] Ran full `bash scripts/install.sh` end-to-end -- correctly detects existing install, uses upgrade, all steps pass
- [x] Verified fresh install path still works (uses `uv tool install` when tool not present)


joshalbrecht:
lgtm

PR #1206, opened 2026-04-08
fix plugin paths pointing to agent worktrees

evgunter:
### Root cause fix: skip installed_plugins.json rewrite for local hosts
When provisioning a local agent, `_setup_per_agent_config_dir` would rewrite `installed_plugins.json` paths from `~/.claude/...` to agent-scoped paths. Because `_sync_user_resources` creates child-level symlinks for the `plugins/` directory, the rewritten file was written through the symlink back into `~/.claude/plugins/installed_plugins.json`, corrupting the user's global config. Now we only rewrite for remote hosts where `~/.claude/` doesn't exist.

### Best-effort rewrite for non-standard paths (remote hosts)
Instead of crashing when `installPath` doesn't match the expected `~/.claude/` prefix, the rewrite now attempts a best-effort fix using the last `/plugins/` segment as a marker. This handles:
- **Stale mngr agent paths** (e.g. `~/.mngr/agents/<id>/plugin/claude/anthropic/plugins/cache/...`): rewrites successfully, warns with the expected persistent path and suggests uninstalling/reinstalling
- **Local-path plugins** (e.g. `~/code-guardian/plugins/...`): rewrites successfully, warns with a symlink workaround suggestion
- **No `/plugins/` marker at all**: keeps path as-is with a warning that it may not work remotely

This merges Josh's permissive best-effort approach (171f9433b) with actionable warning messages.

### Known remaining issue
Installing plugins via `/plugin` while inside an mngr agent still writes agent-scoped paths through the symlinked `installed_plugins.json` back to `~/.claude/`. This is Claude Code's own plugin installer behavior (not mngr). The best-effort rewrite handles this gracefully now, but the underlying path corruption persists until the user reinstalls the plugin.

## Test plan
- [x] `test_write_generated_files_writes_through_symlink_safely` -- verifies symlink is preserved when `installed_plugins.json` is not in generated_files
- [x] `test_rewrite_installed_plugins_paths_best_effort_for_mngr_agent_path` -- stale agent path is rewritten via /plugins/ marker
- [x] `test_rewrite_installed_plugins_paths_best_effort_for_non_matching_prefix` -- local-path plugin is rewritten via /plugins/ marker
- [x] `test_rewrite_installed_plugins_paths_best_effort_for_similar_prefix` -- .claude2/ path is rewritten via /plugins/ marker
- [x] All existing rewrite tests pass (normal prefix matching, empty plugins, multiple plugins, field preservation)
- [x] Type check (`ty check`) passes
- [x] End-to-end verified on Modal: local-path plugin (`~/code-guardian`) fails with expected error, succeeds after symlink workaround, and now also succeeds without the workaround via best-effort rewrite
- [ ] CI passes


libs/mngr/docs/commands/primary/create.md
```diff
@@ -262,16 +262,6 @@ Provider: ssh
...
-Provider: vultr
-  VPS-specific args (--vps- prefix, consumed by provider):
-    --vps-region=REGION  Vultr region (default: ewr)
-    --vps-plan=PLAN      Vultr plan (default: vc2-1c-1gb)
-    --vps-os=OS_ID       Vultr OS ID (default: 2136 = Debian 12 x64)
-
-  All other build args are passed to 'docker build' on the VPS.
-  Example: -b --vps-plan=vc2-2c-4gb -b --file=Dockerfile -b .
-  Start args are passed directly to 'docker run'. Run 'docker run --help' for details.
-
```
joshalbrecht, line 274: Erm, I think there's something wrong here, might need to regenerate after merging qi's fix to make sure everything is always enabled for doc generation?

joshalbrecht:
lgtm (except above comment, noting in slack)

merging because I want to see if this helps fix my autofix settings

libs/mngr/docs/commands/primary/create.md
evgunter: ah oops, i kept getting that because agents didn't run uv sync

PR #1207, opened 2026-04-08
add e2e tests for docker hosts

evgunter:
i've only glanced at these but they seem like an improvement

---

## Summary

Addresses Josh's review comment on #1142 requesting extensive testing of host and agent state definitions and transitions.

- New `test_docker_state_transitions.py`: 8 tests verifying `discover_hosts()` returns correct `host_state` at each Docker lifecycle stage (RUNNING, STOPPED, CRASHED, FAILED) using real containers
- New `test_gc_docker.py`: 4 tests verifying GC behavior when Docker daemon is offline and with real Docker hosts
- Moved duplicated `docker_provider` fixture into shared `conftest.py`

## Test plan

- [x] All 12 new tests pass with Docker running (110s total)
- [x] Full test suite passes: 3470 passed, 0 failed, 82.91% coverage
- [x] Ratchet tests pass
- [x] Autofix clean


joshalbrecht:
lgtm, though I worry about duplication between what is being tested here and any other existing docker tests (esp given our current test load is getting towards the limit...)

so, I'm not merging this right now, but you can merge as soon as we've gotten ourselves a bit more testing overhead + checked to make sure we're not duplicating (acceptance) tests--release tests are more ok to have duplication

evgunter:
the agent that wrote these affirms that there still isn't duplication (though i haven't checked its work)

PR #1208, opened 2026-04-08
fix gitignore check when .claude is a symlink

evgunter:
## Summary
- `_check_settings_local_gitignored` fails with "fatal: pathspec is beyond a symbolic link" when `.claude` is a symlink (e.g. `.claude -> .agents`), then raises a misleading error telling users to add `.claude/settings.local.json` to `.gitignore`
- Fix: resolve symlinks via `realpath` before passing to `git check-ignore`, so the check and error message use the resolved path (e.g. `.agents/settings.local.json`)
- Handle edge case where symlink target is outside the repo (silently skip -- git won't track it)

## Test plan
- [x] 3 new unit tests: symlink with correct gitignore, symlink with wrong gitignore, symlink outside repo
- [x] 4 existing gitignore tests still pass
- [x] Full suite: 370 passed, 81.81% coverage
- [x] Manual verification in repo with `.claude -> .agents` symlink


joshalbrecht:
lgtm

PR #1209, opened 2026-04-08
fix observe crash on stale agent references

evgunter:
## Summary
- Pass `--on-error continue` to the discovery subprocess spawned by `AgentObserver`, so stale agent references (`AgentNotFoundOnHostError`) don't crash the entire observe loop
- Widen exception catches from `MngrError` to `BaseMngrError` in `discovery_events.py` and `observe.py` error handlers, since transient errors like `AgentNotFoundOnHostError` extend `BaseMngrError` (not `MngrError`) and were slipping through

## Context
The observer's own direct `list_agents` calls already used `ErrorBehavior.CONTINUE`. The inconsistency was that the child discovery subprocess inherited the CLI default of `ABORT`. Additionally, several `except MngrError` clauses missed `BaseMngrError` subclasses entirely.

## Test plan
- [x] 3292 unit/integration tests pass (106s)
- [x] 3475 tests pass with coverage (80.42%, autofix run)
- [ ] CI passes

### DIFF
--- a/libs/mngr/imbue/mngr/api/discovery_events.py
+++ b/libs/mngr/imbue/mngr/api/discovery_events.py
@@ -24,6 +24,7 @@ from imbue.imbue_common.logging import generate_log_event_id
 from imbue.imbue_common.pure import pure
 from imbue.mngr.config.data_types import MngrConfig
 from imbue.mngr.config.data_types import MngrContext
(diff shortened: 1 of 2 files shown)


joshalbrecht:
lgtm

PR #1217, opened 2026-04-08
share claude credentials across sessions on macos

evgunter:
kind of cursed but what can you do

---

## Summary

- On macOS, Claude Code hashes `CLAUDE_CONFIG_DIR` into keychain labels, giving each mngr agent isolated keychain entries. Credentials were only copied at provisioning time, so re-authenticating in one session (`/login`) didn't propagate to others.
- Adds a `Notification:auth_success` hook that, after successful login, syncs the new credential to all per-agent keychain entries and the default (unsuffixed) entry. Implemented as a standalone Python script (stdlib only, no mngr imports) provisioned alongside existing hook scripts.
- On Linux, adds the option to copy `.credentials.json` instead of symlinking, for credential isolation.
- New config field: `sync_credentials_on_login` (default `True`). On macOS, controls hook installation. On Linux, controls symlink vs copy.

## Test plan

- [x] `just test-offload`: 6981 passed, 0 failed
- [x] Local `libs/mngr_claude` suite: 368 passed, 81.84% coverage
- [x] Architecture verification passed
- [x] Autofix clean (4 fixes applied)
- [ ] Manual verification: provision agent on macOS, confirm `auth_success` hook in `.claude/settings.local.json`, run `/login` in one session and verify other sessions pick up the change


joshalbrecht:
lgtm. Agreed that it is disgusting

PR #1218, opened 2026-04-08
fix inconsistent aliased update error on git mirror push

evgunter:
i'm not sure if this is the right fix, especially since i'm not sure why others aren't getting the same error (i get this whenever i try to make a modal agent)

---

## Summary

- Replace `git push --mirror` with explicit refspecs (`+refs/heads/*:refs/heads/*` and `+refs/tags/*:refs/tags/*`) plus `--force --prune` in all mirror push call sites
- This avoids pushing remote-tracking refs (`refs/remotes/*`), which include symbolic refs like `refs/remotes/origin/HEAD` that cause "inconsistent aliased update" errors on git 2.45+
- Add regression test verifying remote-tracking refs are not pushed to the target
- Update docs/help strings to not reference the removed `--mirror` flag

## Test plan

- [x] Regression test: `test_mirror_push_refspecs_do_not_push_remote_tracking_refs` verifies branches and tags are pushed while `refs/remotes/*` are excluded
- [x] Validated regression: with `--mirror` (old behavior), `refs/remotes/origin/HEAD` and `refs/remotes/origin/main` are pushed to the target, failing the test assertion
- [x] Full test suite: 3466 passed, 82.88% coverage
- [ ] Manual verification: start a modal agent from a repo with `refs/remotes/origin/HEAD`


joshalbrecht:
lgtm -- will merge and let you know if I see any issues when I try to run them in the future

PR #1222, opened 2026-04-08
Consolidate --from/--source/--source-agent/--source-host/--source-path into a single --from flag

qi-imbue:
The new --from DSL encodes four variants:
1. AGENT[@HOST[.PROVIDER]]     - agent's host + agent's work_dir
2. AGENT[@HOST[.PROVIDER]]:PATH - agent's host + explicit path
3. @HOST[.PROVIDER]:PATH       - host + path (@ signals host reference)
4. :PATH                       - local path

A bare name now refers to an agent, not a directory (use :dirname for relative paths). Paths starting with /, ./, ~/, ../ are still recognized directly as a convenience.

The source string is now parsed exactly once by parse_source_string() in api/find.py, and the resulting ParsedSourceLocation is passed to resolve_source_location() for resolution - no re-parsing.

Removes SourceLocation (dead code), _find_source_location, _parse_source_string, ParsedSourceString, and _host_str_from_address_components from cli/create.py. Updates clone.py to use --from instead of --from-agent.


qi-imbue (commented):
Overall I'm happy with this, especially the fact that it ends up simplifying a lot of the code and results in fewer lines of code!

PR #1223, opened 2026-04-08
detect global-only gitignore rules in preflight check

evgunter:
## Summary
- The preflight check for `.claude/settings.local.json` being gitignored now detects when the rule only exists in the user's global gitignore (e.g., `~/.config/git/ignore`) and not in the repository itself
- This prevents expensive remote host creation (e.g., on Modal) only to fail during provisioning because the remote has no global gitignore
- Uses `git -c core.excludesFile= check-ignore` to verify the rule comes from the repo, not just global config

## Test plan
- [x] New test: `test_preflight_check_raises_when_only_global_gitignore` verifies the error is raised when the rule is only in the global gitignore
- [x] Existing tests pass (349 passed, 82.34% coverage)
- [ ] CI passes


joshalbrecht:
lgtm, will let you merge and resolve conflicts

PR #1224, opened 2026-04-08
fix flaky test_on_after_provisioning_adopts_session_by_id

evgunter:
## Summary
- Remove unnecessary `provision()` calls from three `on_after_provisioning` tests that were causing intermittent timeouts against the 10-second pytest timeout
- `provision()` does expensive work (shell commands, git checks, background thread management, script provisioning) that `on_after_provisioning()` does not depend on
- Matches the pattern already used by the sibling test `test_on_after_provisioning_raises_when_session_not_found`, which calls `on_after_provisioning()` directly

## Test plan
- [x] All 370 tests in `libs/mngr_claude` pass (81.68% coverage)
- [x] Target test runs consistently in ~2s (down from occasional 7-8s)
- [x] Ran target test 10x in isolation and 10x with full suite -- no failures


joshalbrecht:
lgtm

PR #1229, opened 2026-04-08
fix bug in finding .claude.json when ORIGINAL_CLAUDE_CONFIG_DIR is set

evgunter:
## Summary
- Renames `get_user_claude_config_path()` to `find_user_claude_config()` -- the function now searches disk for the config file rather than constructing a single path
- When `ORIGINAL_CLAUDE_CONFIG_DIR` is set (e.g. `~/.claude`), searches candidate paths in priority order: inside-dir (`~/.claude/.claude.json`), beside-dir (`~/.claude.json`), then `~/.claude.json` as default. Returns the first that exists, or the first candidate if none exist.
- This fixes trust checks failing when a mngr agent launches a nested agent, because the old code always returned `~/.claude/.claude.json` which doesn't exist in the default Claude Code layout (trust entries live in `~/.claude.json`)
- Removes unused `get_claude_config_path()` and `get_claude_config_backup_path()` (were only used in tests, never in production)

## Test plan
- [x] Unit tests for candidate-list search (inside-dir, beside-dir, default, preference order)
- [x] Existing trust/dialog plugin tests pass (197 passed)
- [x] Ratchet tests pass
- [ ] CI (integration shard 3 timed out at 152s/150s limit -- flake, not related)


joshalbrecht:
lgtm

PR #1238, opened 2026-04-08
fix atomic_write breaking symlinks

evgunter:
## Summary
- `atomic_write` used `os.replace(tmp_path, path)` which replaces a symlink with a regular file, breaking symlinked `settings.toml` files (e.g. symlinked from dotfiles)
- Now resolves the path via `path.resolve()` before the replace, so the write goes through to the symlink target and the link is preserved
- Temp file is also created in the resolved parent directory to ensure the rename is atomic (same filesystem)

## Test plan
- [x] Added `test_atomic_write_preserves_symlink` -- verifies symlink is still a symlink after write, content is correct at both link and target
- [x] Added `test_atomic_write_preserves_symlink_permissions` -- verifies target file permissions are preserved when writing through a symlink
- [x] All 3518 existing tests pass, 82.88% coverage


joshalbrecht:
lgtm

PR #1239, opened 2026-04-09
add skill for messaging another mngr agent

evgunter:
## Summary

- Adds a new `/message-agent` Claude skill that teaches agents how to send messages to other mngr agents
- Messages are prefixed with `[from: <agent name>]` (read from `$MNGR_AGENT_NAME`) and end with a reply instruction pointing recipients to the same skill
- Advises using `--message-file` for multiline messages to avoid shell quoting issues
- Includes agent name resolution (with fuzzy matching against `mngr list`) following the same pattern as `wait-for-agent`

## Test plan

- [ ] Verify the skill appears in the Claude skill list (confirmed locally)
- [ ] Test invoking `/message-agent some-agent tell them to rebase` and verify the composed message format
- [ ] Verify `$MNGR_AGENT_NAME` is read correctly in an agent session

### DIFF
--- a/.claude/skills/find-agent/SKILL.md
+++ b/.claude/skills/find-agent/SKILL.md
@@ -1,0 +1,24 @@
+---
+name: find-agent
+argument-hint: <agent_name_or_description>
+description: Resolve an agent name or description to an exact mngr agent name. Used by other skills that target agents.
+allowed-tools: Bash(uv run mngr list *)
+---
+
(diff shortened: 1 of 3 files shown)


joshalbrecht:
lgtm, I've been wanting us to add more skills like this for using mngr anyway--want to go make a bunch more and then we can publish these to various skills marketplaces, etc? :)

joshalbrecht:
(leaving open just so you see the above)

PR #1241, opened 2026-04-09
misc resource cleanup

evgunter:
## Summary

Complementary FD leak fixes on top of main's provider instance cache (which landed independently via commit `e6a4a977`).

- **Per-executor `on_thread_exit` callback** on `ConcurrencyGroupExecutor` — lets callers register thread-local cleanup to run after each task, without global state.
- **Gevent Hub cleanup** — `cleanup_thread_local_resources()` destroys the thread-local gevent Hub (and its OS pipe) when pyinfra worker threads exit. Wired up via `mngr_executor()` factory.
- **Host disconnect in discovery path** — `connected_host()` context manager ensures SSH connections are closed after `discover_agents()`. `get_host_and_agent_details()` also disconnects the initial host in a `finally` block.
- **`disconnect()` method on `HostInterface`** — default no-op; online host implementations close their connection.
- **Repro/isolation scripts** under `scripts/qi/fd_leak/` for diagnosing future FD leaks.

The provider cache from an earlier iteration of this branch was reverted in favor of main's equivalent cache.

## Test plan
- [x] All existing tests pass (3773 passed locally)
- [x] Unit tests for host disconnect in `discover_hosts_and_agents`
- [ ] CI validates no regressions


libs/concurrency_group/imbue/concurrency_group/executor.py
```diff
@@ -49,6 +57,9 @@ def _run() -> None:
                     future.set_exception(e)
                 else:
                     future.set_result(result)
+                finally:
+                    if self._on_thread_exit is not None:
+                        self._on_thread_exit()
```
evgunter: alarmingly general change--might be best to skip this one and only include the other kinds of leak cleanup

libs/mngr/imbue/mngr/api/providers.py
```diff
@@ -11,10 +11,12 @@
...
-# is reused across calls within the same context. This prevents accumulating
-# duplicate instances (and their SSH connections) when discovery runs repeatedly.
-_instance_cache: dict[tuple[ProviderInstanceName, int], BaseProviderInstance] = {}
+# Cache of provider instances by name. Reusing instances avoids leaking
+# resources (gRPC connections, SSH connections) that accumulate when new
+# instances are created on every call. The cache is invalidated when the
+# MngrContext changes (tracked by _cache_ctx).
+_instance_cache: dict[ProviderInstanceName, BaseProviderInstance] = {}
+_cache_ctx: MngrContext | None = None
 _atexit_registered: dict[str, bool] = {"registered": False}
```
evgunter, line 18: i really don't think that a global `_cache_ctx` is the right solution for this, but i'm not sure what the right solution is for caching provider instances, or if we even need to cache provider instances

joshalbrecht:
I believe I fixed the underlying issue by caching the provider instances already?

And I did some investigation of related leaks (of connections) and fixed them (maybe was the same issue?)

Basically--is this still an issue? I don't think it should be (so I'd rather close this, bc as you said, it is alarmingly general)

evgunter:
you fixed one of the issues that was originally addressed here, but there were like 3 different kinds of resource leaks addressed. in particular i think the gevent hub one is still important. i can skip the alarmingly general change and wire up the gevent hub in some way that doesn't involve passing new info into concurrency group

joshalbrecht:
yeah, please do that--I'm uncomfortable with the change it made to concurrency group

(but I see what you're saying about the other fixes still being good)

evgunter:
cg change is reverted

joshalbrecht:
lgtm, thanks both for tracking down and fixing!

evgunter:
(@qi-imbue did all the tracking down, not me!)

PR #1242, opened 2026-04-09
fix install script again

evgunter:
## Summary

- macOS kqueue does not support `EVFILT_READ` on `/dev/tty` (the virtual controlling-terminal device), returning `EINVAL`. When `curl -fsSL ... | bash` piped stdin, `create_urwid_screen_preserving_terminal()` opened `/dev/tty` for urwid's TUI input. urwid's event loop then tried to register this fd with kqueue, which rejected it.
- Fix: resolve the actual pty device path (e.g. `/dev/ttys003`) via `os.ttyname()` on stdout/stderr, instead of using `/dev/tty` directly. The real pty device works with kqueue on all macOS versions.
- Add acceptance tests that exercise the actual kqueue + tty interaction in tmux, covering both piped-stdin and direct-stdin scenarios. These tests catch regressions that unit-level mocks cannot detect.

## Test plan

- [x] Unit tests for `resolve_real_tty_path()` (mock-based, fast)
- [x] Acceptance test: kqueue registration with piped stdin in tmux (simulates `curl | bash`)
- [x] Acceptance test: kqueue registration with direct stdin in tmux (regression check)
- [x] Full test suite passes (3520 passed, 1 skipped, 0 failures)
- [x] Verified fix via tmux: both `cat script.sh | bash` and `bash script.sh` resolve to real pty and register successfully with kqueue

### DIFF
--- a/libs/imbue_common/imbue/imbue_common/ratchet_testing/standard_ratchet_checks.py
+++ b/libs/imbue_common/imbue/imbue_common/ratchet_testing/standard_ratchet_checks.py
@@ -98,8 +98,13 @@ def check_global_keyword(source_dir: Path, max_count: int) -> None:
     assert_ratchet(PREVENT_GLOBAL_KEYWORD, source_dir, max_count)
 
 
-def check_bare_print(source_dir: Path, max_count: int) -> None:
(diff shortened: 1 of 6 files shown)


joshalbrecht:
lgtm

PR #1244, opened 2026-04-09
let agents specify arbitrary provisioning commands

evgunter:
something like this is needed to support the config options affecting the reviewer after we move it fully out of mngr into code-guardian, but i'm not certain that this is the right way to do it

---

## Summary

- Adds 7 new `tuple[str, ...]` fields to `AgentTypeConfig`: `extra_provision_command`, `upload_file`, `append_to_file`, `prepend_to_file`, `create_directory`, `env`, `env_file`
- Field names match the existing `CreateCliOptions` and `AgentProvisioningOptions` fields, bringing agent type config to parity with the provisioning options already available via CLI and create templates
- These fields use concatenation merge semantics (parent + child = combined), matching existing `cli_args` behavior
- At provision time, agent type provisioning fields are parsed into typed specs and prepended before CLI entries, so agent type provisioning runs first and CLI options can override
- `PROVISIONING_FIELD_MAP` in `interfaces/host.py` is the shared mapping between raw string field names and their typed parsers, used by both `create.py` (CLI path) and `host.py` (agent type merge path) to prevent drift
- `merge_cli_args` renamed to `merge_tuples` to reflect generalized usage across all tuple fields
- `AGENT_TYPE_CONCAT_TUPLE_FIELDS` is the single source of truth for which fields use concatenation merge

Example TOML:
```toml
[agent_types.my_claude]
parent_type = "claude"
extra_provision_command = [
    "printf '{\"base_branch\": \"%s\"}' \"$GIT_BASE_BRANCH\" > .reviewer/settings.local.json"
]
env = ["REVIEWER_ENABLED=true"]
```

## Test plan

- [x] Unit tests for `AgentTypeConfig.merge_with()` concatenation on new fields
- [x] Unit tests for `_apply_custom_overrides_to_parent_config()` concatenation
- [x] Unit tests for `_normalize_tuple_fields_for_construct` with new fields
- [x] Unit tests for `_merge_agent_type_provisioning` parsing and prepend logic
- [x] Integration test: agent type with provisioning fields creates agent and verifies commands ran, env vars set
- [x] Integration test: CLI provisioning stacks on top of agent type provisioning
- [x] Integration test: child type inherits parent type provisioning fields
- [x] Full test suite passes (3539 passed, 84% coverage)


libs/mngr/imbue/mngr/cli/create.py
```diff
@@ -1249,11 +1249,15 @@ def _parse_agent_opts(
...
     label_options = resolve_labels(opts.label)
 
-    # Parse provisioning options
-    provisioning = AgentProvisioningOptions(
-        extra_provision_commands=opts.extra_provision_command,
-        upload_files=tuple(UploadFileSpec.from_string(f) for f in opts.upload_file),
-    )
+    # Parse provisioning options using the shared field map.
+    # getattr with default () because not all map entries have CLI flags
+    # (e.g. create_directory is agent-type-only).
```
evgunter, line 1304: apparently these were just removed from the cli flags, but `AgentProvisioningOptions` still supports them; not sure whether they belong here

libs/mngr/imbue/mngr/hosts/test_host.py
```diff
@@ -2876,3 +2878,165 @@ def test_create_work_dir_cross_host_generates_unique_paths(
...
+    updated_ctx = host.mngr_ctx.model_copy_update(
+        to_update(host.mngr_ctx.field_ref().config, updated_config),
+    )
+
+    agent = _create_minimal_agent(host, temp_dir)
+
+    options = CreateAgentOptions(
+        name=AgentName("prov-agent-type"),
+        agent_type=agent_type_name,
+        command=CommandString("sleep 1"),
```
evgunter, line 2819: allegedly a placeholder command that is never run which matches other tests already here

joshalbrecht:
A few things:
1. qi changed the interfaces, there are no more append and prepend
2. this seems broadly really nice--it'd be much better if our claude integrations could be simplified to be more about generic commands, etc, and didn't have to get into so much custom rewriting, etc...

evgunter:
huh apparently the cli flags were removed but the actual append and prepend fields on the types remain? is this as intended? (i removed append_to_file and prepend_to_file before realizing this, but this means it doesn't match CreateCliOptions and AgentProvisioningOptions)

joshalbrecht:
They should probably also be removed from the types--I think the previous claude just failed to be exhaustive in its removal (and I dont think they're used anywhere).

You can always prepend/append with a command anyway, so it doesnt matter, and then it's your fault for not making it idempotent

joshalbrecht:
Also, on this PR--I think I'm ok (LGTM) on this one, but something just feels off about provisioning still at a conceptual level... it's a messy mix of host and agent dependence...

PR #1248, opened 2026-04-09
Remove dead --interactive flag from create command

qi-imbue:
## Summary
- The `--interactive/--no-interactive` flag on `mngr create` was stored in `ConnectionOptions.is_interactive` but never read by any consumer
- Interactive mode is actually controlled by `--headless`, `config.headless`, and TTY auto-detection in `setup_command_context`
- Removed the flag, the `ConnectionOptions.is_interactive` field, the `CreateCliOptions.interactive` field, and the docs reference

## Test plan
- [x] All 3518 tests pass with coverage (83.99%)
- [x] Type checker passes (was catching the dead field reference in conftest)


qi-imbue (commented):
Simple change, and I'm convinced that `--interactive` has been dead.

PR #1250, opened 2026-04-09
Add :PATH syntax to agent address for target path

qi-imbue:
## Summary
- Added `:PATH` suffix support in the agent address positional argument: `mngr create addr:path` or `mngr create :path`
- `--target-path` is kept as an alternative flag (like `--provider` is an alternative to `.PROVIDER`)
- Uses the same first-colon splitting convention as `parse_source_string`
- Conflict detection: specifying different paths via both `:PATH` and `--target-path` is an error

## Test plan
- [x] All 3530 libs/mngr tests pass (82.91% coverage)
- [x] Ratchets pass
- [ ] CI passes
- [ ] Release tests pass (test_create_modal_target_path updated to new syntax)


qi-imbue (commented):
Makes sense - accepts `:PATH` suffix as an alternative to `--target-path`.

We wanted to remove the flag, but we have some templates that use this, so I've kept it as an alternative way to specify the target path - like how `--provider modal` is an alternative to `@.modal`.

joshalbrecht:
lgtm

PR #1251, opened 2026-04-09
Replace --yes flag with [local_system_mutations] config section

qi-imbue:
## Summary

- Replace the interactive prompting approach for local system mutations with a declarative `[local_system_mutations]` config section with three policies: `accept_permission_dialogs` (YES/ERROR/IGNORE), `install_agents` (YES/ERROR), `install_skills` (YES/ERROR)
- Remove `--yes` flag from `create` command and `is_auto_approve` from `MngrContext`
- Remove `auto_dismiss_dialogs` from `ClaudeAgentConfig` (behavior now controlled at config root level)
- Add `mngr claude dismiss` CLI command group for manual dialog dismissal
- Consolidate the double dialog check: ERROR/IGNORE validates in `on_before_provisioning` (read-only), YES auto-dismisses in `provision()`

## Test plan

- [x] All tests pass in libs/mngr (3518 passed, 82.90% coverage)
- [x] All tests pass in libs/mngr_claude (368 passed, 82.20% coverage)
- [x] All tests pass in libs/mngr_pi_coding (76 passed, 88.36% coverage)
- [x] All tests pass in libs/mngr_claude_mind (242 passed, 86.02% coverage)
- [ ] CI acceptance tests pass
- [ ] Manual test: `mngr claude dismiss all /some/path` writes to `~/.claude.json`
- [ ] Manual test: `mngr create` works without `--yes` (default YES policy auto-approves)


libs/mngr_claude/imbue/mngr_claude/plugin.py
```diff
@@ -394,8 +384,8 @@ def _build_claude_json(
                 worktree_config["_mngrSourcePath"] = str(source_path)
                 projects[worktree_path_str] = worktree_config
 
-    # Trust work_dir if unattended or auto_dismiss_dialogs
-    if should_trust_work_dir(config, ctx):
+    # Trust work_dir if unattended (remote/deploy agents)
+    if should_trust_work_dir(ctx):
```
qi-imbue, line 386: I find this a bit suspicious, and in fact I'm a bit confused of the original behavior: we're building the per-agent .claude.json file, it seems that we should always accept the trust dialog, because we're not mutating anything global. I'm not sure why the condition was here in the first place, and I don't actually remember seeing trust dialogs for my local agents. Not sure what's going on here, but the new code doesn't seem to have made things worse?

qi-imbue (commented):
Big PR, but overall makes sense to me:

- Adds config section
- Adds `claude dismiss` subcommand
- Removes all the prompting code
- New code to check the settings

libs/mngr_claude/imbue/mngr_claude/plugin.py
joshalbrecht: I added auto_dismiss_dialogs recently, and the logic was: it's a dialog, thus it should be dimissed if that is set (regardless of where you are)

Claude is pretty annoying about throwing up all sorts of dialogs. The option wasn't intended for this particular location (it was for other things), but it at least seemed logically consistent to apply it here too

evgunter:
mostly looks good, but i'm not sure why we would need a whole new command for this rather than using `mngr config set`. edit: stacked pr to remove the new command and just recommend `config set` https://github.com/imbue-ai/mngr/pull/1267

qi-imbue:
The reason for the extra commands is that it has a different semantics to the config section:
- The `claude dismiss` subcommands each mutates one bit of the local Claude config file
- The `config set` subcommand tells mngr to apply those mutations whenever necessary, and without prompting

The original discussion thread is https://imbue-ai.slack.com/archives/C0AA6LFV5S5/p1773498191015639 and the `claude dismiss` subcommand corresponds to Josh's

> one for "accept just this one setting" (will need the mng_claude plugin to effectively expose a CLI so that it can mutate such settings)

I don't feel too strongly for keeping this functionality - @joshalbrecht?

qi-imbue:
Talked on Slack - we decided to hold off this change for now, and can revisit when we have more experience with other agent harnesses

PR #1261, opened 2026-04-10
Skip libs without pyproject.toml during editable mngr installation

gnguralnick:
## Summary
- The editable install logic in `mngr_recursive` discovers plugin libs by iterating `libs/mngr_*` directories, but did not verify each directory is a valid Python package
- Stray or incomplete directories (e.g. `mngr_*` with no `pyproject.toml`, leftover from previous versions / bad git state) caused `uv tool install --with-editable` to fail, preventing per-agent mngr installation during mind provisioning
- Filter out directories missing `pyproject.toml` in both local and remote editable install paths

## Test plan
- [x] All 131 mngr_recursive tests pass (85.47% coverage)
- [x] Updated `test_agent_editable_local_mode_builds_correct_command` to create `pyproject.toml` in mock lib dirs

### DIFF
--- a/libs/mngr_recursive/imbue/mngr_recursive/provisioning.py
+++ b/libs/mngr_recursive/imbue/mngr_recursive/provisioning.py
@@ -317,7 +317,7 @@ def _install_mngr_editable_remote(
             lib_names = ls_result.stdout.strip().split()
             install_parts = [f"{_UV_PATH_PREFIX}{uv_env}cd {remote_repo_dir} && uv tool install -e libs/mngr"]
             for lib_name in lib_names:
-                if lib_name != "imbue-mngr":
+                if lib_name != "imbue-mngr" and lib_name.startswith("mngr_"):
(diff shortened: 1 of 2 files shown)


evgunter:
when it gets into this state, I think it also causes failures on a meta-ratchet which checks if every libs/mngr* dir has a pyproject.toml; so, since the directories will need to be cleaned up anyway, it might be better to raise a better error about it or something like that? (i assume you're encountering this on your local machine rather than in some programmatic context)

gnguralnick:
yeah i just had an error related to this - happened when creating a mind. not tied to this solution.

the error i got was
```
WARNING: Failed to install mngr for agent 'selene': Failed to install mngr in editable mode: error: /Users/gabeguralnick/utilities/mngr/libs/mngr_lima does not appear to be a Python project, as neither `pyproject.toml` nor `setup.py` are present in the directory
```

this is informative enough for me to find the stray folder and delete it so i suppose it's not necessary to add any logic here if there are issues

PR #1263, opened 2026-04-10
add --event flag to mngr wait

evgunter:
## Summary
- Adds `--event` flag to `mngr wait` that watches an agent's lifecycle event stream for specific events (e.g., `AGENT_READY`)
- Introduces `LifecycleEventType` enum and `emit_agent_lifecycle_event()` helper in new `lifecycle_events.py` module
- Emits paired `AGENT_STARTING`/`AGENT_READY` lifecycle events from `mngr create`. `AGENT_READY` is emitted after verified readiness (via `wait_for_ready_signal`) when there's an initial message, or immediately after start otherwise
- `--event` is mutually exclusive with state arguments and requires an agent target
- Fixes `HeadlessClaude.wait_for_ready_signal` to be a no-op instead of raising `NotImplementedError` (needed because MRO would otherwise resolve to `ClaudeAgent`'s readiness polling, which hangs for headless agents)

## Test plan
- [x] Unit tests for `emit_agent_lifecycle_event`, path helpers
- [x] Unit tests for `_check_latest_lifecycle_event` with real local host
- [x] Unit tests for `wait_for_event` (match, timeout, host rejection, stale event)
- [x] Unit tests for `_validate_event_type` (valid types, case insensitivity, invalid)
- [x] CLI integration tests via CliRunner (event match, timeout, mutual exclusivity, state match)
- [x] All mngr tests pass (3592 passed, 83.9% coverage)
- [x] All mngr_wait tests pass (146 passed, 88.6% coverage)


libs/mngr/imbue/mngr/api/lifecycle_events.py
```diff
@@ -0,0 +1,73 @@
...
+
+def get_lifecycle_events_path(host_dir: Path, agent_id: AgentId) -> Path:
+    """Return the path to an agent's lifecycle events JSONL file."""
+    return get_lifecycle_events_dir(host_dir, agent_id) / "events.jsonl"
+
+
+def emit_agent_lifecycle_event(
+    host: OnlineHostInterface,
+    agent_id: AgentId,
+    event_type: LifecycleEventType,
```
evgunter, line 41: probably confusing to have 'lifecycle events' as a separate entity from agent lifecycle; not sure what a better name would be, though

libs/mngr_wait/imbue/mngr_wait/api.py
```diff
@@ -345,3 +349,99 @@ def _detect_state_changes(
...
+                    elapsed_seconds=time.monotonic() - start_time,
+                    state_changes=(),
+                )
+        except HostConnectionError as exc:
+            logger.debug("Host unreachable while checking lifecycle events: {}", exc)
+
+        if timeout_seconds is not None and elapsed >= timeout_seconds:
+            is_waiting = False
+        else:
+            time.sleep(interval_seconds)
```
evgunter, line 436: wait_for_state

joshalbrecht:
Hmm, I'm not entirely sold on this one--I think the point of "mngr wait" is that it waits for an agent to enter a state.

We could either:
1. make these things into states
2. make a different command that waits for a certain type of event (actually that's possible today I think via `... | jq '.something.state' | some_cli_command_that_exits_once_it_sees_some_line` )

evgunter:
hm, if we want everything to use the event stream, it seems like we would want the ability to wait on other events in the event stream, so i think 2 makes sense; and i'm not sure what the benefit is of having this be separate from `wait`. (the jq command you have is also probably more fragile than what i have, since it lacks the ID i use to avoid having to worry about races between the watcher and the event writer)

joshalbrecht:
one of the nice things about "mngr wait" is that it does NOT use the event stream (it polls). This is good because it does not require "mngr observe", which kinda wants to be a singleton...

there's a logical issue in that "agent enters a state" is NOT really an event (because you can miss state transitions). I believe I made a fix for this somewhere as I was making minds, but it ended up being quite tricky and brittle (you have to sort of know the full state diagram to be able to impute certain transitions that you may have missed, etc). It was really gross.

evgunter:
not needed for the case i wanted it for (`mngr files` suffices)

PR #1264, opened 2026-04-10
check that modal test environment names include timestamp

evgunter:
## Summary
- Adds `_validate_test_environment_name()` in `backend.py` that checks Modal environment names during DIRECT-mode provider creation in tests
- Uses the existing resource guard's `_PYTEST_GUARD_PHASE` env var to detect test context (no-op in production)
- Rejects environment names that don't start with `mngr_test-YYYY-MM-DD-HH-MM-SS`, catching production prefixes (`mngr-{uuid}`) and UUID-based test prefixes (`mngr_{uuid}-{user_id}`)
- Only enforced for `ModalMode.DIRECT` (real Modal); `ModalMode.TESTING` (fake Modal) is excluded since it never creates real environments

## Test plan
- [x] Added 5 unit tests for `_validate_test_environment_name()`: accepts valid timestamped names, rejects production prefixes, UUID prefixes, non-timestamped mngr_test- names, and is a no-op outside test phase
- [x] All 379 mngr_modal tests pass with coverage at 77.29%
- [x] Existing `test_build_provider_instance_testing_mode` continues to pass (TESTING mode is excluded from validation)


joshalbrecht:
lgtm

we *may* need to be more aggressive with this if we continue to see leftover modal envs (eg, we could put this in the production code to check if we're testing, and if so, ensure this name pattern)

PR #1277, opened 2026-04-12
Merge template vars with config defaults instead of overwriting

joshalbrecht:
## Summary
- Fix `apply_create_template` to concatenate list/tuple params (like `env`, `pass_env`, etc.) with existing values from `[commands.create]` defaults and earlier templates, instead of overwriting them
- Add empty-list-resets semantics to `merge_list_fields` and `merge_dict_fields`: an explicit `[]` or `{}` means "clear everything from earlier in the precedence chain," while `None` means "not set"
- CLI-specified values are always preserved (concatenated with template values for lists, unchanged for scalars)

## Test plan
- [x] Unit tests for template list concatenation with config defaults
- [x] Unit tests for multiple templates concatenating list values
- [x] Unit tests for empty list reset behavior
- [x] Unit tests for reset-then-add-back across templates
- [x] Unit tests for CLI values concatenating with template values
- [x] Unit tests for empty list not resetting CLI values
- [x] Updated merge_list_fields and merge_dict_fields tests
- [x] All 3377 mngr tests passing with 80.26% coverage

### DIFF
--- a/.claude/skills/minds-dev-iterate/SKILL.md
+++ b/.claude/skills/minds-dev-iterate/SKILL.md
@@ -1,0 +1,157 @@
+---
+name: minds-dev-iterate
+description: Set up and iterate on the minds app stack (desktop client, workspace server, mngr, forever-claude-template) with a running Docker agent
+---
+
+# Minds Dev Iteration Loop
+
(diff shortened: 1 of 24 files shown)


joshalbrecht:

PR #1281, opened 2026-04-13
Updated prototyping code for minds

joshalbrecht:
Lumping together a bunch of changes and fixes for minds (not mentioned)

Changes for mngr:

added modules:
    mngr_lima: a lima-based provider. Mostly works, still early and experimental

removed modules:
    apps/claude_web_chat (became minds_workspace_server and changed its purpose)
    libs/mngr_claude_mind
    libs/mngr_llm
    libs/mngr_mind
    libs/mngr_mind_chat

security:
    fix LLM trying to work around known hosts file

bug fixes:
    disabled plugins should prevent them from mutating command line arguments
    options with list values from templates and cli config should be merged (ex: --env and env in a toml should 
concatenate)
    cache provider instances to fix some bug I forgot the details of
    makes things work when sudo is required (ex: removing directories)

performance:
    remove pointless discovery delay for server events by pre-creating the events file
    blow up earlier if a docker container already has the same name (before building it)

features:
    add auto_allow_permissions field for claude plugin (to skip ALL permissions prompts)

refactor:
    abstract how we collect host details so it can be properly shared across provider (see listing_utils.py)

testing:
    fix flaky tests (see https://github.com/imbue-ai/mngr/pull/1283 )


joshalbrecht:
Merging because the tests passed here: https://github.com/imbue-ai/mngr/pull/1283

PR #1287, opened 2026-04-13
fix bug for `mngr ask` error reporting which meant the final pane capture fallback was always skipped

evgunter:
## Summary

- Fixed a logic bug in `_raise_no_output_error` where tmux pane capture was only attempted when both stdout and stderr redirect files were missing. Shell redirects create empty files even on immediate failure, so the pane capture (which would show the actual error) was being skipped, resulting in the unhelpful "no details available" message.
- Pane capture is now always attempted as a last resort when no other error details are found, regardless of whether redirect files exist.
- Added a test for the specific bug scenario (redirect files exist but empty, pane content should still be surfaced).

## Test plan

- [x] Unit tests pass for `base_headless_agent_test.py` (16 tests)
- [x] Unit tests pass for `headless_claude_agent_test.py` (30 tests)
- [x] Unit tests pass for `headless_command_agent_test.py` (9 tests)
- [ ] Full test suite via `just test-offload` (running)
- [ ] `test_ask_simple_query` should now produce a more informative error message when it fails in CI, enabling root-cause diagnosis of the sandbox failure


joshalbrecht:
lgtm, but this is marked as a draft, so I'll let you merge when ready

PR #1289, opened 2026-04-13
use mngr create to run headless agents with --foreground flag

evgunter:
includes the changes from https://github.com/imbue-ai/mngr/pull/1175 but you can skip most of those by not looking at mngr_schedule, and the changes from https://github.com/imbue-ai/mngr/pull/1338 but they're short

---

## Summary

- When `create` detects a `StreamingHeadlessAgentMixin` agent type (e.g. `headless_command`), it automatically uses the headless flow: temp directory, stream output, auto-destroy
- Adds `--foreground` flag, required for headless agent types and rejected for non-headless types (makes the behavior explicit)
- Removes the separate `mngr run` command entirely
- Removes the `--command` CLI flag (merged from #1325); headless agents get their command from agent type config or templates via `make_test_sleep_agent_type`
- Rejects flags that are silently ignored on the headless path (`--env`, `--branch`, `--reuse`, `--message`, etc.) with a clear error
- Headless path supports local and remote hosts via create's existing address parsing

## Test plan

- [x] Unit tests for agent type resolution, headless detection, \`--foreground\` validation, incompatible flag rejection
- [x] Integration test for \`_create_headless\` streaming output (uses \`write_agent_type_to_settings_toml\` fixture)
- [x] Manual verification: \`mngr create --type headless_command --foreground\` with command configured via template streams output
- [x] Manual verification: \`mngr run\` shows "No such command"
- [x] Full test suite: 3787 passed, coverage 86.00% (above 80% threshold)


libs/mngr/docs/commands/primary/create.md
```diff
@@ -64,8 +70,8 @@ mngr create [OPTIONS] [POSITIONAL_NAME] [POSITIONAL_AGENT_TYPE] [AGENT_ARGS]...
 | `-n`, `--name` | text | Agent address (alternative to positional argument, mutually exclusive) [default: auto-generated] | None |
 | `--id` | text | Explicit agent ID [default: auto-generated] | None |
 | `--name-style` | choice (`coolname` &#x7C; `english` &#x7C; `fantasy` &#x7C; `scifi` &#x7C; `painters` &#x7C; `authors` &#x7C; `artists` &#x7C; `musicians` &#x7C; `animals` &#x7C; `scientists` &#x7C; `demons`) | Auto-generated name style | `coolname` |
-| `--type` | text | Which type of agent to run [default: claude] | None |
-| `--command` | text | Run a literal command using the generic agent type (mutually exclusive with --type) | None |
+| `--type` | text | Which type of agent to run [default: claude, or generic when -c is used] | None |
+| `--command`, `-c` | text | Shell command for the agent to run [default --type: generic] | None |
```
evgunter: not a big fan of the dependence between these two but i'm not sure how to do better without making the interface more verbose

joshalbrecht: I was planning to remove --command entirely?

It doesn't seem like it's actually needed...

joshalbrecht:
I don't love this--I want to reduce the amount of magical defaults in our parameters, not add more

Why do we want this vs just setting those flags?

libs/mngr/docs/commands/primary/create.md
evgunter: it could become the exclusive domain of headless agents, which would remove the dependence

evgunter: it also could just be removed, i suppose; it was just to enable headless agents running a user-specified command, but it's not clear that that's something there's meaningful demand for. `mngr ask` doesn't use it

evgunter:
you mean the `--connect-command` misuse? it really doesn't work well for the motivating case, which was `mngr ask`; it would require calling separate python scripts from the connect command (or porting them to bash, yikes). i think it's actually pretty clean to have the agent type determine whether we connect or stream the output

libs/mngr/docs/commands/primary/create.md
evgunter: removed (in https://github.com/imbue-ai/mngr/pull/1325)

libs/mngr/docs/commands/secondary/schedule.md
```diff
@@ -173,6 +179,12 @@ mngr schedule run [OPTIONS] NAME
 ```
 **Options:**
 
+## Execution
+
+| Name | Type | Description | Default |
+| ---- | ---- | ----------- | ------- |
+| `--provider` | text | Provider on which the trigger is deployed (e.g. 'local', 'modal'). | None |
+
```
evgunter, line 187: from https://github.com/imbue-ai/mngr/pull/1175

libs/mngr/imbue/mngr/cli/create.py
```diff
@@ -147,6 +153,180 @@ def _split_address_and_target_path(raw: str) -> tuple[str, Path | None]:
...
+    # --host-label is intentionally *not* listed: when the headless address
+    # creates a new host (NAME@.PROVIDER or --new-host), _parse_target_host
+    # reads opts.host_label to populate host tags, so the flag is honored.
+    # This matches the treatment of other host-level flags (--host-env,
+    # --snapshot, --idle-timeout, ...) which are also only meaningful on
+    # new-host creation.
+    ("id", "--id"),
+    ("label", "--label"),
+    ("project", "--project"),
+)
```
evgunter, line 227: a lot of these are specific to how we use it in `mngr ask`, but i figure there's not a lot of reason to disentangle "runs independently of a project", "streams output", "automatically destroys when done", etc until something other than `mngr ask` uses this

joshalbrecht: I don't think that most of those are actually incompatible...  is there a reason why so many are banned?

joshalbrecht:
> When create detects a StreamingHeadlessAgentMixin agent type (e.g. headless_command), it automatically uses the headless flow: temp directory, stream output, auto-destroy

I thought we were going to get rid of that magic and just error? Is that just an outdated comment in the PR description?

joshalbrecht:
(also, since this is stacked, I'll need to review once the other one is merged so that I can more easily see the diff, but I suspect it will be fine)

evgunter:
it is outdated-ish in that it also requires the --foreground flag to be passed explicitly

libs/mngr/imbue/mngr/cli/create.py
evgunter: they are all actually not wired up here. i'm doing the wiring up in https://github.com/imbue-ai/mngr/pull/1339 (mostly as-needed to make it work for the changelog agent, but iirc it does end up supporting almost all of the flags listed here)

evgunter: yeah, all but 7 are removed there

joshalbrecht: ahh, that makes me feel a lot better about this one then!

merging (and assuming that the other PR will fix this)

joshalbrecht:
err, actually, not merging--will let you merge after fixing whatever is wrong in CI

PR #1290, opened 2026-04-13
fix test_help_nonexistent_topic + move log capture utility to testing.py

evgunter:
## Summary
- `test_help_nonexistent_topic` fails because loguru's default handler stores a reference to the original `sys.stderr` at import time, so when Click's `CliRunner` replaces `sys.stderr` with its own capture buffer, loguru continues writing to the original stream
- Fix the test by replacing loguru's default handler with `_dynamic_stderr_sink` (reused from production `setup_logging`), which resolves `sys.stderr` at write time and follows CliRunner's replacement
- No production code changes -- the `logger.error()` calls in the help command remain as-is

## Test plan
- [x] `test_help_nonexistent_topic` passes (was failing)
- [x] Full mngr test suite passes (3587 passed, 1 skipped)
- [x] All ratchet tests pass
- [x] Coverage: 82.82%

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -227,6 +227,9 @@
 # Local Claude settings backup files
 **/.claude/*.bak
 
+# Claude worktress
+**/.claude/worktrees
+
 # Local mngr settings (not committed)
 **/.mngr/settings.local.toml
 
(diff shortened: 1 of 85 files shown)


joshalbrecht:
lgtm

PR #1291, opened 2026-04-13
improve mngr test coverage

evgunter:
lots of error path testing with fake objects + some silly tests for transcript rendering and ssh utils. pretty low value tests but i went through and tried to catch ones which were vacuous 

---

## Summary

Adds and improves unit/integration tests across 5 files that had the lowest coverage in the mngr library:

- **`utils/detail_renderer.py`**: 10% -> 98% (58 tests). Pure function tests for ANSI-to-HTML conversion, transcript rendering, and test detail HTML generation. No mocks.
- **`providers/ssh_utils.py`**: 19% -> 87% (32 tests). Tests for SSH keypair generation/save/load, known_hosts management, and pyinfra host creation. No mocks.
- **`api/gc.py`**: 57% -> 88% (~40 tests). Tests for provider discovery errors, work dir cleanup on local hosts, host auth/offline errors, snapshot/volume/log GC error paths, and git worktree removal.
- **`api/cleanup.py`**: 61% -> 93% (9 tests). Tests for agent-not-found graceful degradation, ABORT error behavior for destroy/stop, offline host handling, and post-cleanup GC errors.
- **`api/list.py`**: 70% -> 99% (~15 tests). Tests for ErrorBehavior.CONTINUE/ABORT, provider discovery errors, CEL filter application, host-level error handling, and SSH info emission.

Also extracts `make_ctx_with_plugins()` shared helper into `testing.py`, replacing duplicated plugin manager construction across 4 test files.

Overall coverage: **82.82% -> 84.96%** (3763 passed, 1 skipped, 0 failed)

## Test plan
- [x] All new tests pass locally
- [x] Coverage threshold met (84.96% > 80%)
- [x] No ratchet violations
- [x] CI passes (35/35 checks green)

### DIFF
--- a/libs/mngr/imbue/mngr/api/cleanup_test.py
+++ b/libs/mngr/imbue/mngr/api/cleanup_test.py
@@ -42,7 +42,6 @@ from imbue.mngr.primitives import LOCAL_PROVIDER_NAME
 from imbue.mngr.primitives import ProviderBackendName
 from imbue.mngr.primitives import ProviderInstanceName
 from imbue.mngr.providers.local.instance import LocalProviderInstance
-from imbue.mngr.providers.registry import load_local_backend_only
 from imbue.mngr.utils.polling import wait_for
 from imbue.mngr.utils.testing import make_test_agent_details
 
@@ -59,7 +58,6 @@ def _make_ctx_with_plugin(temp_mngr_ctx: MngrContext, plugin: object) -> MngrCon
     """Create a MngrContext that has the given plugin registered in its plugin manager."""
     pm = pluggy.PluginManager("mngr")
     pm.add_hookspecs(hookspecs)
-    load_local_backend_only(pm)
     pm.register(plugin)
     return temp_mngr_ctx.model_copy_update(to_update(temp_mngr_ctx.field_ref().pm, pm))
 
@@ -77,9 +75,9 @@ class _OfflineHostProvider(LocalProviderInstance):
     type checker (ty) skip the incompatibility check.
     """
 
-    def get_host(self, host):
+    def get_host(self, host: HostId | HostName):
+        host_id = host if isinstance(host, HostId) else HostId.generate()
         now = datetime.now(timezone.utc)
-        host_id = HostId.generate()
         certified_data = CertifiedHostData(
             created_at=now,
             updated_at=now,
--- a/libs/mngr/imbue/mngr/api/cleanup_test.py
+++ b/libs/mngr/imbue/mngr/api/cleanup_test.py
@@ -1,6 +1,8 @@
 """Unit tests for cleanup API functions."""
 
+from collections.abc import Generator
 from collections.abc import Sequence
+from contextlib import contextmanager
 from datetime import datetime
 from datetime import timezone
 from pathlib import Path
@@ -15,6 +17,7 @@ from imbue.mngr.api.cleanup import execute_cleanup
 from imbue.mngr.api.cleanup import find_agents_for_cleanup
 from imbue.mngr.api.create import CreateAgentOptions
 from imbue.mngr.api.data_types import CleanupResult
+from imbue.mngr.api.providers import _instance_cache
 from imbue.mngr.config.data_types import MngrContext
 from imbue.mngr.config.data_types import ProviderInstanceConfig
 from imbue.mngr.errors import MngrError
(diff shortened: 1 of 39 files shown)


libs/mngr/imbue/mngr/api/cleanup_test.py
```diff
@@ -1,25 +1,134 @@
...
+class _OfflineHostProvider(LocalProviderInstance):
+    """Provider that returns an offline HostInterface from get_host().
+
+    destroy_host() raises LocalHostNotDestroyableError (a MngrError), which
+    triggers the error path in _execute_destroy when the match arm falls through
+    to HostInterface (the offline branch).
+
+    get_host() has no return type annotation because it returns OfflineHost, which
+    satisfies HostInterface but is not a subclass of Host (the parent's declared
+    return type). Adding a return annotation would produce a type error.
```
evgunter, line 79: hm

libs/mngr/imbue/mngr/utils/testing.py
```diff
@@ -688,6 +692,50 @@ def make_mngr_ctx(
...
+    for a given (name, context) pair. The entry is removed on exit.
+    """
+    from imbue.mngr.api.providers import _instance_cache
+
+    cache_key = (name, id(mngr_ctx))
+    _instance_cache[cache_key] = instance
+    try:
+        yield
+    finally:
+        _instance_cache.pop(cache_key, None)
```
evgunter: pretty cursed. the combo inline-and-private import is a nice touch........

libs/mngr/imbue/mngr/providers/ssh_utils_test.py
```diff
@@ -0,0 +1,566 @@
...
+    _, public_openssh = generate_ssh_keypair()
+    assert public_openssh.startswith("ssh-rsa ")
+
+
+def test_generate_ssh_keypair_rsa_4096_bits() -> None:
+    """The RSA key should be 4096 bits."""
+    private_pem, _ = generate_ssh_keypair()
+    private_key = load_pem_private_key(private_pem.encode("utf-8"), password=None)
+    assert isinstance(private_key, rsa.RSAPrivateKey)
+    assert private_key.key_size == 4096
```
evgunter: these tests are kind of dumb, but seem fast enough to be harmless?

libs/mngr/imbue/mngr/utils/detail_renderer_test.py
```diff
@@ -0,0 +1,467 @@
...
+def test_render_transcript_stderr_line_gets_stderr_prefix_class() -> None:
+    """Lines starting with '! ' render the prefix in stderr-prefix span."""
+    result = render_transcript("$ cmd\n! error message\n? 1")
+    assert '<span class="stderr-prefix">! </span>' in result
+    assert "error message" in result
+
+
+def test_render_transcript_output_is_wrapped_in_pre_transcript() -> None:
+    """The rendered output is wrapped in a <pre class="transcript"> element."""
+    result = render_transcript("$ ls\n? 0")
```
evgunter, line 268: i think we don't have any snapshot tests yet, but all these little tests really seem like they wants to be one big snapshot test

libs/mngr/imbue/mngr/api/cleanup_test.py
```diff
@@ -1,25 +1,139 @@
...
+    mngr_ctx: MngrContext,
+    instance: LocalProviderInstance,
+) -> Generator[None, None, None]:
+    """Temporarily inject a provider instance into the provider cache."""
+    cache_key = (name, id(mngr_ctx))
+    _instance_cache[cache_key] = instance
+    try:
+        yield
+    finally:
+        _instance_cache.pop(cache_key, None)
```
evgunter, line 59: not sure if this is the best way to do this

libs/mngr/imbue/mngr/utils/testing.py
joshalbrecht: that is quite disgusting...

not sure if there's an amazing way around that one though :(

libs/mngr/imbue/mngr/providers/ssh_utils_test.py
joshalbrecht: for tests like this--there should really only be *one* test of the function (it's just there for coverage reasons)

maybe not as part of this PR, but it WOULD be nice to clean up our tests a bit

having tests like this is not quite harmless--it increases the test count, which slows down discovery. Currently discovery actually takes a really long time...

libs/mngr/imbue/mngr/utils/detail_renderer_test.py
joshalbrecht: correct--would love a ratchet against this kind of crap (and the above), but not quite sure how to construct it...

libs/mngr/imbue/mngr/api/cleanup_test.py
joshalbrecht: wait--isn't this almost the same as the above?

hmmm....  is there a better way?

joshalbrecht:
lgtm, all of my comments are advisory and/or future work

libs/mngr/imbue/mngr/api/cleanup_test.py
evgunter: above was moved here

PR #1304, opened 2026-04-15
ruff meta ratchets

evgunter:
not sure why it split them into two tests; i can combine them into one if that would be better

---

## Summary
- Adds `test_no_ruff_pre_commit_errors` and `test_no_ruff_format_errors` to `test_meta_ratchets.py`
- Mirrors the pre-commit hook's ruff rules (UP006, UP007, I, F401) and format check as CI-enforced meta ratchets
- Catches violations even when the pre-commit hook is bypassed

## Test plan
- [x] Both new tests pass locally (12/12 meta ratchets pass)
- [x] imbue_common tests pass (214/214) -- no subprocess ratchet regression
- [ ] CI passes


joshalbrecht:
probably better to combine, otherwise lgtm

PR #1305, opened 2026-04-15
modal usage guide

evgunter:
- add usage guide for modal, including example template
- add GH_TOKEN to our settings.toml (verified that this works)

---

## Summary
- Add `libs/mngr/docs/modal_usage.md` -- step-by-step guide for running coding agents on Modal, covering creation, templates, getting changes back (via `mngr pull` or git credentials), snapshots, and a pointer to the full reference

## Test plan
- [x] CI passes

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.mngr/settings.toml
+++ b/.mngr/settings.toml
@@ -24,6 +24,7 @@ provider = "modal"
 build_arg = ["file=libs/mngr/imbue/mngr/resources/Dockerfile", "context-dir=.mngr/dev/build/", "volume=code-review-json:/code_reviews"]
 extra_window = ["github_setup='ssh-keyscan github.com >> ~/.ssh/known_hosts && git remote set-url origin https://github.com/imbue-ai/mngr.git && gh auth setup-git'"]
 agent_args = ["--dangerously-skip-permissions"]
+pass_env = ["GH_TOKEN"]
 target_path = "/code/mngr"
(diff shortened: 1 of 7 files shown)


.mngr/settings.toml
```diff
@@ -24,6 +24,7 @@ provider = "modal"
 build_arg = ["file=libs/mngr/imbue/mngr/resources/Dockerfile", "context-dir=.mngr/dev/build/", "volume=code-review-json:/code_reviews"]
 extra_window = ["github_setup='ssh-keyscan github.com >> ~/.ssh/known_hosts && git remote set-url origin https://github.com/imbue-ai/mngr.git && gh auth setup-git'"]
 agent_args = ["--dangerously-skip-permissions"]
+pass_env = ["GH_TOKEN"]
```
evgunter, line 27: is there a reason not to have this here?

joshalbrecht: it's ~fine

I wanted a better way to do this--putting that token in their gives the claudes more github access than I would really want, and hynek and I are working on better ways to do such auth. So for now, yes, we can use this, but we should revisit in order to restrict their access a bit more

joshalbrecht:
lgtm

PR #1308, opened 2026-04-15
Fix docker destroy cleanup errors

joshalbrecht:
## Summary
- Investigating and fixing errors during `mngr destroy` for docker-hosted agents
- `gc_snapshots` fails with "Host not found" because `gc_machines` already destroyed the host in the same GC run
- Observer shows stale host briefly because it only processes full discovery snapshots, not incremental destroy events

## Test plan
- [ ] Verify `mngr destroy -f <agent>` on a docker host produces no errors
- [ ] Verify host is cleaned up promptly after destroy
- [ ] Run existing test suite


joshalbrecht:

PR #1309, opened 2026-04-15
error if missing worktree on agent start

evgunter:
## Summary

- When an agent's `work_dir` points to a deleted git worktree, `tmux new-session -c` silently falls back to `$HOME`. The agent launches in `~/`, shows a trust dialog [edit: or not, and confuses the agent because it's in the wrong place], and times out.
- Adds `_ensure_work_dir_exists` check in `start_agents` before building the tmux command. If the worktree is gone, raises `AgentStartError` with the exact `git worktree add` command the user needs to run to recreate it.
- If no branch is recorded on the agent, errors with a plain "does not exist" message.

## Test plan

- [x] Unit test: no-op when work_dir exists
- [x] Unit test: raises AgentStartError when dir missing and no branch recorded
- [x] Unit test: error includes `git worktree add` recovery command with correct branch
- [x] CI passes (7307 tests)


joshalbrecht:
lgtm

Issue #1256, opened 2026-04-10
Support alternative SSH authentication methods


rebornix:
Following up on the plugin system discussion on https://github.com/imbue-ai/mngr/issues/1158#issuecomment-4190472042, I found that the SSH transport layer currently assumes key-based authentication everywhere. Daytona uses a token for the username when it connects to SSH (no static keys), and other providers like Proxmox VE use a password for SSH. The current code won't support these integrations.

This is also a good example for the plugin extensibility story from #1158. Even if I make a provider plugin as a separate package, I'd still need to make some changes to the main program to make the SSH transport work with non-key authentication. Ideally, a community provider plugin should be able to use their preferred authentication method without changing the core code.

I put together a poc [branch](https://github.com/rebornix/mngr/commit/64390b7b94ce78b95eaa9b99f80dbcc202c93b1b) to explore what changes would be needed. It shows where the key-based assumptions live. Not a complete solution though (since this fix doesn't really scale‚ but hope it's enough to see the scope of the problem.

joshalbrecht:
Thanks for pointing this out!

We'll take a look at it and fix it--you're right that we cannot assume the use of keys everywhere.

For the hosts that *we* create in provider (modal, docker, etc), we do set up sshd, and do need to know how to connect to them (and I'm reluctant to add a ton of configuration complexity there right now). But for the SSH provider, we should definitely allow that to accept whatever types of auth are possible.

Does that work for you? Or did you want something changed about SSH across all of the different providers? And if so, do you have any ideas about how to make that configuration convenient / simple?

PR #1310, opened 2026-04-15
working towards ssh password auth

evgunter:
i'm not confident yet that the security is handled properly here! in particular, there are multiple different methods used for transporting the password between hosts right now, and i'm not sure if any of them are how we actually want to do it

towards fixing https://github.com/imbue-ai/mngr/issues/1256, but doesn't actually wire up a daytona provider or anything

---

## Summary

Infrastructure for extensible SSH authentication. **This PR does not enable password auth for any user-facing workflow** -- no shipped provider uses `SSHPasswordAuth`. It adds the abstraction layer (`SSHAuthMethod` ABC) and a standalone `imbue-mngr-ssh-password-auth` package so future providers (e.g. Proxmox, Daytona) can bring their own auth without patching core.

### What this PR delivers
- `SSHAuthMethod` ABC + `SSHKeyAuth` core implementation
- Polymorphic `build_transport_command()` / `connect_paramiko()` / `configure_pyinfra_host_data()` -- zero branches on auth type in core
- All providers (docker, modal, lima, vps_docker, ssh) construct `SSHKeyAuth` and pass it via `Host(ssh_auth=...)` at construction time
- `SSHInfo` carries `auth: SSHAuthMethod` instead of `key_path` + `command`
- Desktop client uses polymorphic `connect_paramiko()` -- no auth-type branches
- `HOST_AUTH_INFO` stdout-only discovery event carries unmasked credentials to the desktop client without writing to `events.jsonl`
- Runtime registry (`SSHAuthMethod._registry`) populated via `__init_subclass__`; deserialization dispatches through a single `_deserialize_ssh_auth` helper (pydantic `BeforeValidator` + `PlainSerializer` on `SSHAuthField`)

### What this PR does NOT deliver
- No user-facing way to create a host with password auth
- No provider wired up to use `SSHPasswordAuth`
- `imbue-mngr-ssh-password-auth` is a standalone package that ships `SSHPasswordAuth`, but no provider depends on it yet; users who want password auth will need to write (or wait for) a provider that does

When a real password-auth provider is added, its own e2e tests will exercise `SSHPasswordAuth` against real infrastructure (sshd in docker, etc.).

### The abstraction

`SSHAuthMethod` is a Pydantic ABC with an `auth_type: Literal[...]` discriminator and four polymorphic methods:
- `build_transport_command(port, known_hosts_file) -> SSHTransportCommand` -- for rsync/git subprocess transport
- `connect_paramiko(client, hostname, port, username)` -- for direct paramiko SSH sessions
- `configure_pyinfra_host_data(host_data)` -- populates pyinfra connector data
- `get_display_command(user, hostname, port) -> str` -- human-readable SSH command, never includes secrets

`SSHTransportCommand` is the key data shape for CLI transport:
```python
class SSHTransportCommand(FrozenModel):
    command: str                      # e.g. "sshpass -e ssh -p 22 -o ..."
    env: dict[str, SecretStr] = {}    # e.g. {"SSHPASS": SecretStr("...")}
```

`SSHKeyAuth` produces `env={}` (no secret env needed -- key path goes in `command`). `SSHPasswordAuth` produces `env={"SSHPASS": SecretStr(...)}` and a command that begins with `sshpass -e` (reads SSHPASS from env).

### Secret masking at rest

`SSHPasswordAuth.password: SecretStr` -- Pydantic's `SecretStr` wraps values so:
- `repr()` / `str()` / default `model_dump()` all produce `"**********"`
- `.get_secret_value()` is the only way to extract the plaintext
- Every exposure callsite is greppable via `get_secret_value` or the helper `expose_secrets_for_subprocess`

### Credential transport: local -> remote host

Flow: user's local mngr process spawns rsync/git, talking over SSH to a remote host.

```python
transport = conn.auth.build_transport_command(port, known_hosts)
env = {
    **os.environ,
    **expose_secrets_for_subprocess(transport.env),   # unwraps SecretStr -> str
    "GIT_SSH_COMMAND": transport.command,              # or rsync: -e transport.command
}
subprocess.run([...], env=env)
```

For `SSHPasswordAuth`:
- `command = "sshpass -e ssh -p N -o StrictHostKeyChecking=yes -o PreferredAuthentications=password -o PubkeyAuthentication=no"`
- `env = {"SSHPASS": SecretStr("<password>")}` -> unwrapped into subprocess env only
- sshpass reads SSHPASS from its own env (passed by kernel), pipes the password to ssh through a pty (not visible to `ps`)
- **Secret never appears as a command-line argument**, so it's invisible in `/proc/*/cmdline` and `ps -ef`

### Credential transport: remote source host -> remote target host

Tricky case. A git mirror push runs **on the remote source host**, pushing to a remote target. The mngr process on the user's machine invokes the push via SSH. Two obstacles:
1. We can't pass subprocess env directly -- the command string ships over SSH and runs via shell on the source
2. Inline `SSHPASS=... git push ...` in the command string would be visible in the source host's `ps`

Solution (in `host.py:_git_push_to_target`): wrap the push in a shell expression that writes the secret to a tempfile on the source host, sources it, runs git, and cleans up via `trap`:

```sh
_mngr_env=$(mktemp) && \
trap 'rm -f "$_mngr_env"' EXIT && \
printf 'export SSHPASS=<shlex-quoted>\n' > "$_mngr_env" && \
chmod 600 "$_mngr_env" && \
. "$_mngr_env" && \
GIT_SSH_COMMAND=... git push --no-verify --force --prune <url> <refs>
```

- File is mode 0600 (owner read-only)
- Deleted on normal exit and most signals via `trap ... EXIT`
- `shlex.quote` applied to the secret value prevents shell injection
- Secret touches disk briefly on remote, but: the remote host is already authenticated by us via SSH, so the attacker model is "another process on that host" -- they'd need to race the 0600 file within its ~ms lifetime, or have root (in which case the game is over anyway)
- Not used for local->remote; that path uses subprocess env directly (cleaner)

### Credential transport: mngr subprocess -> desktop client

Flow: the desktop client app on user's laptop needs to SSH-tunnel to remote agents. It launches `mngr observe --discovery-only` as a subprocess and reads its stdout. The mngr subprocess has the auth objects; the desktop client needs them to connect with paramiko.

Two separate event channels:
- **`events.jsonl` file** (persistent): `HostSSHInfoEvent` carries `SSHInfo` with `auth: SSHAuthMethod`. Serialized via `PlainSerializer(_serialize_ssh_auth)` which calls `auth.model_dump(mode="json")`. SecretStr fields auto-mask to `"**********"` -- passwords never written to disk.
- **stdout only** (process-to-process): `HostAuthInfoEvent` carries `auth_data: dict[str, str]` with secrets **unmasked** (populated via `_build_auth_data_with_secrets` which calls `get_secret_value()`). Emitted via a dedicated `emit_stdout_only_auth_events` code path that **does not call `append_discovery_event`** (which is the only function that writes to the file).

The two channels are kept distinct in the code to prevent accidental cross-contamination: `HostSSHInfoEvent` goes through `emit_host_ssh_info` (file + stdout via tail); `HostAuthInfoEvent` goes through `build_host_auth_info_line` + `emit_stdout_only_auth_events` (stdout only).

On the desktop client side, `MngrStreamManager._handle_host_auth_info` deserializes via the `SSHAuthField` tagged-union dispatch and stores the auth in memory keyed by `host_id`. The auth object is then used polymorphically for paramiko connections -- no auth-type branches.

### Discovery event parse errors

If a discovery event line fails to parse, the error handler logs **only the event type and exception class**, never the raw line (which could contain a plaintext password from a `HOST_AUTH_INFO` event). See `backend_resolver.py:_handle_discovery_line`.

### Connection error wrapping

`connect_paramiko` wraps `paramiko.SSHException` / `OSError` as `SSHConnectionError` with only the exception class name in the message (e.g. `"SSH password auth connection to user@host:port failed: AuthenticationException"`). No credentials leak via error messages. loguru handlers use `diagnose=False`, so local variables (including the auth object) don't appear in tracebacks.

### Host key verification

- When `known_hosts_file` is available and exists: `RejectPolicy` + `StrictHostKeyChecking=yes` (MITM-safe)
- When not available (desktop client case, where server-side paths aren't reachable from the laptop): `AutoAddPolicy` with a warning log. This matches the pre-existing `SSHKeyAuth` behavior.
- Plan specified `RejectPolicy` always; in practice the desktop client has no known_hosts file and would reject all connections under strict mode. Followup could propagate the host key via `HOST_AUTH_INFO`.

## Security audit

| Check | Verdict |
|---|---|
| `expose_secrets` / `get_secret_value` callsites | **PASS** -- all flow to subprocess env, paramiko connect, or stdout-only events |
| `HOST_AUTH_INFO` stdout-only event | **PASS** -- never written to `events.jsonl`, only emitted to stdout |
| `SSHInfo` serialization to events file | **PASS** -- `SecretStr` masks passwords as `**********` |
| `SSHPasswordAuth` secret handling | **PASS** -- password in env dict (`SecretStr`), paramiko kwargs, pyinfra host_data |
| Remote-to-remote git push | **PASS** -- secrets written to temp env file (0600), sourced, cleaned up via `trap` |
| Rsync env passthrough | **PASS** -- `transport.env` passed to subprocess via `env=` for all paths |
| TRACE/DEBUG logging | **PASS** -- no secrets in logged command args or transport commands |
| Host key verification | **PASS** -- `RejectPolicy` + `StrictHostKeyChecking=yes` when `known_hosts` available; `AutoAddPolicy` fallback with warning when not |
| Desktop client `HOST_AUTH_INFO` handling | **PASS** -- dispatches on `auth_type` via registry, no hardcoded type |
| Discovery event parse error logging | **PASS** -- logs only event type + error class, never raw line content |
| Duplicate `auth_type` registration | **PASS** -- `__init_subclass__` raises `TypeError` on collision |
| `connect_paramiko` exception wrapping | **PASS** -- wraps paramiko errors as `SSHConnectionError` with only exception class name, no credentials |

## Test plan

- [x] `libs/mngr` + `apps/minds` + `libs/mngr_ssh_password_auth`: 4207 passed combined
- [x] `libs/mngr_ssh_password_auth`: 26 passed, 100% coverage
- [x] Acceptance test spins up a real sshd in docker and verifies the `SSHPasswordAuth`-produced command actually authenticates via sshpass (plus a negative wrong-password case)
- [x] Type checker (`ty check`): clean
- [x] Ratchets: all pass
- [x] Ruff: clean
- [x] Manual verification: `SSHKeyAuth` path exercised by all existing provider e2e tests (docker, modal, vps_docker, lima, ssh)


libs/mngr/imbue/mngr/hosts/host.py
```diff
@@ -1700,9 +1704,34 @@ def _git_push_to_target(
...
+                    push_cmd = (
+                        f"_mngr_env=$(mktemp) && "
+                        f"trap 'rm -f \"$_mngr_env\"' EXIT && "
+                        f"printf '{env_file_content}\\n' > \"$_mngr_env\" && "
+                        f'chmod 600 "$_mngr_env" && '
+                        f'. "$_mngr_env" && '
+                        f"{git_cmd}"
+                    )
+                else:
+                    push_cmd = git_cmd
```
evgunter, line 1733: pretty weird to have a completely different mechanism for remote-remote vs local

joshalbrecht: I agree--I do NOT think we want this

joshalbrecht:
I'm overall fairly skeptical on this one

I think perhaps a better way to fix this would be to say: "look, you can just say ssh auth key = None, and if you do, just make sure that your auth actually works out properly via sshd", and maybe give some hooks for places where a user could plugin in and inject the necessary data at various call sites or something

this seems to be holding and transferring secret data, which I really don't think we want to do (gets too complicated to get right from a security perspective)

PR #1312, opened 2026-04-15
rewrite known_marketplaces.json plugin paths 

evgunter:
## Summary

**Root cause:** Claude Code's startup plugin update check removes cache entries on remote agents because `known_marketplaces.json` contains local filesystem paths that don't exist on the remote host. Claude Code re-clones the marketplace from GitHub, finds a different HEAD SHA than the cached `gitCommitSha`, and invalidates the plugin cache.

**Deeper insight:** Locally, `installLocation` points to `~/.claude/plugins/marketplaces/` instead of inside `$CLAUDE_CONFIG_DIR`, which Claude Code treats as "corrupted" and silently skips marketplace refreshes. This accidentally protects the local cache but lets `gitCommitSha` drift from the marketplace HEAD, making the stale state get propagated to remote agents where the refresh *does* succeed and then invalidates everything.

**Fix:** Rewrite `installLocation` values in `known_marketplaces.json` for both local and remote agents, mirroring the existing `installPath` rewriting for `installed_plugins.json`. This makes Claude Code's marketplace refresh work correctly, keeping `gitCommitSha` in sync so remote agents inherit a consistent state.

Additional hardening:
- Extract `_rebase_plugin_path` as shared helper between the two rewriters
- Skip symlinking `installed_plugins.json` and `known_marketplaces.json` in `_sync_user_resources` (they're always overwritten by generated content, so symlinking them risks corrupting the user's shared `~/.claude/plugins/` files)
- Defensive `is_symlink()` unlink in `_write_generated_files` as a guardrail against future additions
- Handle deploy-time sentinel prefix for `known_marketplaces.json`

## Test plan

- [x] 400 mngr_claude unit tests pass (added 6 tests for `known_marketplaces` rewriting + 1 for deploy sentinel + 1 for symlink-safe writes)
- [x] 3580 mngr core unit tests pass
- [x] Architecture review: approach approved (clean mirror of existing `installed_plugins.json` pattern)
- [x] Autofix: 3 iterations, 3 fixes applied (docstring, function rename, deploy test coverage); 1 CRITICAL issue caught and fixed (symlink corruption risk)
- [x] Conversation review: no behavioral issues
- [x] Manual end-to-end validation on local + Modal agents (see below)
- [ ] CI tests

### Manual validation (2026-04-16)

Created local and Modal agents, inspected plugin state, triggered stop hooks both with and without agent-side commits.

**Modal agent, commit-triggered stop hook** (`test-plugin-path-6`):
- `installPath` and `installLocation` correctly rewritten to per-agent config dir paths
- Cache intact with all 3 plugins after Claude Code startup (`best-of-n`, `claude-plugins-official`, `imbue-code-guardian`)
- Stop hook (initial, no commits): `hookErrors:[]`, completed in 316ms
- Stop hook (after agent committed a file): fired normal gate-failure output listing verify-architecture/autofix/verify-conversation -- same output we see in our own sessions. Critically: **no "Plugin directory does not exist" error**. The `${CLAUDE_PLUGIN_ROOT}/scripts/stop_hook_gates.sh` resolved and executed successfully.

**Local agent** (`test-plugin-path-3`, `test-plugin-path-5`):
- `installPath`: `/Users/ev/.claude/plugins/cache/...` (local hosts don't need `installPath` rewriting since original paths are directly accessible)
- `installLocation`: `/Users/ev/.mngr/agents/.../plugins/marketplaces/...` ✓ all 3 marketplaces correctly rewritten
- Generated files (`installed_plugins.json`, `known_marketplaces.json`) are real files, not symlinks. Other plugin children (`cache`, `marketplaces`, `blocklist.json`, etc.) still symlinked to `~/.claude/plugins/` as intended.
- Marketplaces actually refreshed by Claude Code (fresh `lastUpdated` timestamps after startup, confirming the rewrite enables the refresh)
- Cache survived refresh (no "corrupted installLocation" warning, no cache deletion)
- Source `~/.claude/plugins/known_marketplaces.json` untouched (both the skip-symlink fix and the defensive unlink work correctly)

Agents stopped (not destroyed) for post-mortem inspection.

### DIFF
--- a/libs/mngr_claude/imbue/mngr_claude/plugin.py
+++ b/libs/mngr_claude/imbue/mngr_claude/plugin.py
@@ -881,6 +881,12 @@ def _write_generated_files(
         for relative, content in generated_files.items():
             dest = config_dir / relative
             dest.parent.mkdir(parents=True, exist_ok=True)
+            # Break any existing symlink so we write a regular file instead
+            # of following the symlink back to the source (e.g. ~/.claude/).
+            # _sync_user_resources creates child-level symlinks for plugins/;
+            # writing through them would corrupt the user's original files.
+            if dest.is_symlink():
+                dest.unlink()
             host.write_text_file(dest, content)
     else:
         local_host = _get_local_host(mngr_ctx)
--- a/libs/mngr_claude/imbue/mngr_claude/plugin.py
+++ b/libs/mngr_claude/imbue/mngr_claude/plugin.py
@@ -101,6 +101,7 @@ _CLAUDE_HOME_SYNC_DIRS: Final[tuple[str, ...]] = ("skills", "agents", "commands"
 _CLAUDE_HOME_SYNC_FILES: Final[tuple[str, ...]] = ("keybindings.json",)
 
 _INSTALLED_PLUGINS_RELATIVE_PATH: Final[Path] = Path("plugins") / "installed_plugins.json"
+_KNOWN_MARKETPLACES_RELATIVE_PATH: Final[Path] = Path("plugins") / "known_marketplaces.json"
 
 _INSTALLED_PLUGINS_SENTINEL_PREFIX: Final[str] = "/__mngr_plugins_source__"
 """Sentinel prefix written into installPath values at deploy build time.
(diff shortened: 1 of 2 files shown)


joshalbrecht:
lgtm, but adding comment in slack about how we need to re-think this

PR #1313, opened 2026-04-15
fix AF_UNIX path length overflow in minds desktop client SSH tunnel on macOS

gnguralnick:
## Summary
- macOS `\$TMPDIR` is a long per-user path under `/var/folders/...` (~49 chars). Combined with the old socket filename `tunnel-<ssh_host>-<ssh_port>-<remote_host>-<remote_port>.sock` (~42 chars for IPv4), the resulting AF_UNIX socket path exceeded macOS's 104-byte `sun_path` limit, causing `bind(2)` to fail with `"AF_UNIX path too long"` and the SSH forwarding tunnel to never come up.
- Use `/tmp` as the tmpdir base on Darwin to bypass `\$TMPDIR`. Permissions unchanged (`0o700` dir, `0o600` sockets).
- Replace the readable-but-long filename with a 12-hex-char `blake2b` hash of the tunnel key. This decouples path length from endpoint-string length (long hostnames or IPv6 addresses would otherwise reintroduce the overflow even on Linux).

After the fix, the path is `/tmp/minds-ssh-XXXXXXXX/t-XXXXXXXXXXXX.sock` (~43 chars), well under the 103-char portable max.

## Test plan
- [x] Existing `ssh_tunnel_test.py` suite passes locally (39/39)
- [x] Manually verified `minds forward` no longer hits "AF_UNIX path too long" when SSH-tunneling to a docker-backed agent on macOS
- [ ] CI passes

### DIFF
--- a/apps/minds/imbue/minds/desktop_client/ssh_tunnel.py
+++ b/apps/minds/imbue/minds/desktop_client/ssh_tunnel.py
@@ -1,7 +1,9 @@
+import hashlib
 import os
 import select
 import shlex
 import socket
+import sys
 import tempfile
 import threading
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


joshalbrecht:
lgtm

PR #1315, opened 2026-04-15
show minds loading page when backend hangs up without responding

gnguralnick:
## Summary
- When the SSH tunnel accepts a local connection but the SSH channel-open to the backend port fails (e.g. sshd inside an agent container is ready before `uvicorn` finishes binding to port 8000), the tunnel relay closes the local socket without writing anything. httpx surfaces this as `RemoteProtocolError: Server disconnected without sending a response.`, which the proxy wasn't catching -- producing an unhandled exception and a dead-end 502 in the browser that required a manual reload.
- Catch `httpx.RemoteProtocolError` in both `_forward_http_request` and the streaming generator and return a 502. The existing logic at `app.py:649` already converts 5xx responses into the auto-retrying loading page for HTML requests, so the browser self-recovers as soon as the backend is up.

## Test plan
- [x] New `test_http_proxy_returns_loading_page_when_backend_disconnects_html` and `test_http_proxy_returns_502_when_backend_disconnects_non_html` pass; full `test_desktop_client.py` suite passes (78/78)
- [x] `ruff check` and `ruff format --check` clean; `ty` clean
- [ ] Manually verify by hitting `minds forward` URL during the agent-container startup window
- [ ] CI passes

### DIFF
--- a/apps/minds/imbue/minds/desktop_client/app.py
+++ b/apps/minds/imbue/minds/desktop_client/app.py
@@ -440,6 +440,16 @@ async def _forward_http_request(
     except httpx.ReadError:
         logger.warning("Backend connection lost for {} server {}", agent_id, server_name)
         return Response(status_code=502, content="Backend connection lost")
+    except httpx.RemoteProtocolError:
+        # Raised when the SSH tunnel accepts the connection but closes it without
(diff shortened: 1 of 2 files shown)


joshalbrecht:
lgtm, thanks for fixing!

PR #1316, opened 2026-04-15
Use python subprocess for Electron launch detach (macOS-portable)

gnguralnick:
## Summary
- `setsid(1)` isn't installed on default macOS, which broke the minds-dev-iterate launch snippet and the `propagate_changes` electron restart step.
- Replace both call sites with `python3 -c "... subprocess.Popen(..., start_new_session=True, ...)"`, which is the stdlib equivalent (start_new_session=True calls `setsid(2)` in the forked child before exec) and works on both macOS and Linux without any extra dependencies.
- End behavior is unchanged: the Electron process is detached into a new session/process group with no controlling TTY, so it survives the launching shell exiting and its terminal closing.

## Test plan
- [x] Ran the exact Python one-liner on macOS (Darwin 25) with a `sleep 20` stand-in for `npm start`; confirmed the spawned bash reparents to pid 1 (launchd) with its own session, surviving the invoking shell's exit.
- [x] Verified the replacement behaves consistently with the prior `setsid bash -c ... & ; disown` pattern on macOS.
- [ ] CI passes (no Python/shell code paths affected; only a skill-doc markdown file and a dev-iteration shell script)

### DIFF
--- a/.claude/skills/minds-dev-iterate/SKILL.md
+++ b/.claude/skills/minds-dev-iterate/SKILL.md
@@ -95,8 +95,6 @@ TEMPLATE_BRANCH=$(cd .external_worktrees/forever-claude-template && git branch -
 )
 ```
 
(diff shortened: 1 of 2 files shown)


joshalbrecht:
lgtm

PR #1317, opened 2026-04-15
Minds: Lima-based onboarding, ToDesktop bundle, Mac-runner launch-to-msg CI

weishi-imbue:
## Summary

Productionizes the minds desktop app on Lima: bundles all runtime binaries (`uv`, `git`, `lima`, `restic`) into the `.app`, pins the FCT template to a tag, and wires a Mac-runner CI workflow that drives launch → first message every commit. Adds a twice-daily cron over the same pipeline that fingerprints (mngr-main, FCT-main), skips when neither has moved, and posts a Slack summary on every run.

### App runtime
- ToDesktop bundle ships self-contained: bundled `uv`, `git` (xcrun on macOS, MinGit on Windows), `lima`, `restic` via `download-binaries.js` + `todesktop:beforeInstall` hook.
- `LIMA_VERSION = '2.0.3'` pinned in `build.js` (lima 2.1 wedges fresh SSH on Debian 12 / systemd 252; explained in the comment).
- Workspace packages bundled into `resources/wheels/` via `pnpm deploy` against the lockfile (replaces npm-install scratch dir; drift-guarded by `test_workspace_package_lists_are_consistent`).
- Splash progress: backend forwards `uv` stderr through `onProgress` → IPC `status-update` → renderer, shown after a 10s threshold.

### Lima provider
- Lima-mode unblocked: memory string normalization, unshallow before push to bare repos, `receive.denyDeleteCurrent=ignore` for `--prune`, `--branch` propagation on shallow clone, `limactl start` timeout 600s → 1800s for cold image downloads.

### SSH tunnel
- Short hash for Unix socket filenames, under macOS 104-byte `AF_UNIX` limit.

### FCT template pinning
- `FALLBACK_BRANCH = "v0.3.0"` (frozen pilot tag) so a released binary is reproducible against a known FCT.

### CI: `.github/workflows/minds-launch-to-msg.yml`
- Mac runner builds the ToDesktop bundle, then runs two jobs in parallel: `launch_to_msg` (self-hosted lima e2e — 2 workspaces, cross-workspace ping, reload, tile-click, relaunch, duplicate-name 409, destroy-W2, per-phase `.win.png` assertions) and `macos_launch` (vanilla `macos-latest` cold-launch smoke). `build` step `pnpm dist` has a 45-min step cap so ToDesktop notarization wedges fail fast.
- **Twice-daily cron** at 09:00 PDT and 21:00 PDT. The `check_should_run` job resolves the current (mngr main HEAD, FCT main HEAD) pair and asks `actions/cache/restore --lookup-only` whether a cache key `launch-to-msg-green-mngr-<sha>-fct-<sha>` exists. Cache hit → all downstream jobs skip (~9s total, free Linux only). Cache miss → full pipeline; on green, `update_marker` writes the cache key so the next cron firing on the same pair skips. `force=true` dispatch input bypasses.
- **Slack notifications** via `secrets.SLACK_LAUNCH_TO_MSG_WEBHOOK`. One message per run with a consistent four-section layout (verdict / SHAs / metrics / links). Three e2e timings extracted from `[ci-metric]` stdout markers: launch→ready, w1 create, w2 create.
- Empirical proof of skip-on-unchanged is in run history (Test 1 wrote marker, Test 2 same pair + force=false skipped in 9s, Test 3 force=true bypassed).

### Cleanup vs main
- Merged `origin/main`; carry-forward of the `VolumeFileType` → `FileType` rename and Debian 12 image + docker apt-version pin (`PINNED_DOCKER_APT_VERSION`).
- Dropped on-branch hacks: `first_launch_prefetch.py`, the `_resolve_mngr_binary` heuristic (`uv run --active` puts the venv on `PATH`), the `reference_repo` plumbing in `agent_creator`.
- Net: 5 ratchets fully cleared, 3 partial reductions.

## Verification

Most recent successful launch-to-msg with full e2e + metrics: run `27325997820` (build cache-hit + lima full flow):

| Metric | Value |
|---|---|
| launch → backend ready | 20.4s |
| w1 create | 4m 28s |
| w2 create | 4m 33s |
| Total workflow wallclock | 14m 5s |
| Total launch_to_msg | 13m 23s |


apps/minds/scripts/build.js
```diff
@@ -1,123 +1,163 @@
...
+/**
+ * Recursively copy a directory, skipping patterns that are not needed at
+ * runtime (tests, caches, build artifacts, .git).
+ */
+function copyDirFiltered(src, dest) {
+  const SKIP = new Set([
+    '__pycache__', '.pytest_cache', '.mypy_cache', '.ruff_cache',
+    'node_modules', '.git', '.test_output', '.venv', 'resources',
+  ]);
+  const SKIP_SUFFIXES = ['_test.py', '.pyc'];
```
joshalbrecht: this is sus--probably should be using .gitignore and git commands?

apps/minds/scripts/build.js
```diff
@@ -1,123 +1,163 @@
...
+      return { from: monorepoRelative, to: `../packages/${destName}` };
+    });
+    replacements.sort((a, b) => b.from.length - a.from.length);
+
+    for (const { from, to } of replacements) {
+      lockContent = lockContent.split(from).join(to);
+    }
+
+    fs.writeFileSync(path.join(destDir, 'uv.lock'), lockContent);
+    console.log(`Copied uv.lock to ${destDir} (rewrote paths to bundled packages)`);
```
joshalbrecht: a lot of this feels pretty brittle / hard to test / strange

would it maybe be better to just publish each of those packages so that they install correctly?  that was the original intent

libs/mngr/imbue/mngr/hosts/host.py
```diff
@@ -1655,6 +1655,38 @@ def _git_push_to_target(
...
+                # push can replace all branches cleanly.
+                if not self.is_local:
+                    detach_cmd = (
+                        "git -C " + shlex.quote(str(target_path))
+                        + " checkout --detach HEAD 2>/dev/null;"
+                        " git -C " + shlex.quote(str(target_path))
+                        + " config receive.denyDeleteCurrent ignore"
+                    )
+                    self.execute_idempotent_command(detach_cmd)
+
```
joshalbrecht: I don't understand this--I've never seen this fail?  (to be fair I haven't been testing the DEV mode)

Also, if this WAS failing, then we probably should just be NOT making a shallow clone in minds in the first place?

When was this failing?

libs/mngr_lima/imbue/mngr_lima/limactl.py
```diff
@@ -147,20 +147,43 @@ def host_name_from_instance_name(instance_name: str, prefix: str) -> HostName |
...
+    ``--disk=100GiB``.  Strip the unit suffix so limactl can parse them.
+    """
+    result: list[str] = []
+    for arg in start_args:
+        for flag in ("--memory=", "--disk="):
+            if arg.startswith(flag):
+                value = arg[len(flag):]
+                for suffix in ("TiB", "tib", "TB", "tb", "GiB", "gib", "GB", "gb", "MiB", "mib", "MB", "mb"):
+                    if value.endswith(suffix):
+                        value = value[: -len(suffix)]
```
joshalbrecht: lol, claude--you cannot just strip these! they would not mean the same thing at all

this should be completely removed, and SHOULD be erroring--the fix here is clearly just to change the values in the template?

libs/mngr_lima/imbue/mngr_lima/limactl.py
```diff
@@ -147,20 +147,43 @@ def host_name_from_instance_name(instance_name: str, prefix: str) -> HostName |
...
+    return result
+
+
 def limactl_start_new(
     cg: ConcurrencyGroup,
     instance_name: str,
     yaml_path: Path,
     start_args: tuple[str, ...] = (),
-    timeout: float = 600.0,
+    timeout: float = 1800.0,
```
joshalbrecht, line 155: 15 minutes is kind of too long for a user to really wait...

Is there some way we could do the download as a separate step with a separate longer timeout?

(also, having our template will probably help, see other PR)

apps/minds/scripts/build.js
weishi-imbue: Updated to use `git archive`, which respect `.gitignore`. 
Adding additional exclusions (test files) via [exclude magic pathspec](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-exclude) (e.g. `":(exclude,glob)libs/mngr/**/*_test.py"`)
(Alternative is to create [.gitattributes](https://git-scm.com/docs/git-archive#ATTRIBUTES) under monorepo root, kind of leaky.)

libs/mngr/imbue/mngr/hosts/host.py
weishi-imbue: ####  Re: shallow clone
My understanding of the flow (LIMA mode): 
1. minds make a shallow clone
2. `mngr create ...` 
    1. mngr ssh into lima vm and init a **bare** repo 
    2. mngr git push from localhost workspace to target vm
    3. mngr ssh into vm to check out a branch, flip repo from bare -> regular
On step 2.ii, if the shallow clone is effectively shallow, (e.g. the original repo has depth > 1), then it'll fail (as per default git config). 
On UI it fails at when user click "create" agent. 


Two possible fixes: 
1. drop shallow, let minds clone full. (simplified logic, cost depends on the repo depth) 
2. Configure target repo to accept shallow push. (need change mngr host.py). Cost: agent workspace looses git history. 

Chose fix 1 for this PR. 


####  Re: `--prune`
1. mngr create foo, and agent starts work on target. 
2. user changes source on localhost
3. `mngr create --reuse --update` 

On step 3: it runs `git push --prune --force ...` prune intends to delete branches on target that don't exist on source. it'll fail if target host checked out branch is to be deleted. 

(The same detach method is implemented in [sync.py](https://github.com/imbue-ai/mngr/blob/main/libs/mngr/imbue/mngr/api/sync.py#L781), which seems to back `mngr push --mirror` )

apps/minds/scripts/build.js
weishi-imbue: Updated to build wheels locally.
Added test `test_workspace_wheel_is_pure_python` to assert all wheels are `py3-none-any.whl` (cross-platform, pure Python). Can revisit when we need native deps.

apps/minds/scripts/build.js
weishi-imbue: Update: `git archive` no longer needed. Using whls

libs/mngr_lima/imbue/mngr_lima/limactl.py
weishi-imbue: 😱 Updated.

apps/minds/scripts/build.js
joshalbrecht: are you building and packaging the wheels?

do we think this is better than just making a release for each of the relevant packages?  It seems a little weird to have an effectively-unreleased version of these packages shipped in the minds apps...

(I'm not saying one way or the other, just asking for us to think through how we want this to work... it's a bit complicated because we have multiple packages--both mngr and minds--that rely on some of these common packages)

libs/mngr/imbue/mngr/hosts/host.py
joshalbrecht: Sorry, I'm still confused even with the above explanation

I agree that the correct fix for the first thing is to get rid of the shallow clone and let minds clone full (but you mention the cost of # 2 being "have to change host.py", which *was* changed in this PR when I made my comment...)

Re: prune: your description isn't clear--what was the problem? How did it work? How does it work now?

weishi-imbue: It breaks when minds user trying to create an agent with same name twice, on second time, it triggers `mngr create --reuse --update ...`, then calls `git push --prune --force ...`, then it'll fail agent creation.

apps/minds/electron/backend.js
```diff
@@ -204,8 +204,17 @@ function startBackend(onProgress, onNotification, onAuthEvent, onMngrForwardStar
...
+          // reports contain enough context to debug "Discovering agents..."
+          // hangs. Single v keeps noise manageable; -vv (DEBUG) is dev-only.
           'run', '--project', pyprojectDir,
-          'minds', '--format', 'jsonl',
+          // --active makes uv use VIRTUAL_ENV (~/.minds/.venv) instead of
+          // <project>/.venv, which is inside the signed .app bundle and
+          // read-only on macOS. Without this, `uv run` tries to create
+          // .venv inside the bundle and fails with "Operation not permitted".
+          '--active',
+          'minds', '-v', '--format', 'jsonl',
```
joshalbrecht: I'm a little confused--I just think that "-v" is supposed to mean "debug", not "info"?  Something seems wrong here, probably the comment?

apps/minds/imbue/minds/cli/run.py
```diff
@@ -83,9 +84,15 @@
...
-_MNGR_FORWARD_LISTEN_TIMEOUT_SECONDS: Final[float] = 5.0
+# The plugin emits this from its FastAPI lifespan startup, so on a warm
+# install the wait only needs to cover the subprocess's own interpreter
+# start and imports. On a cold install (vanilla Mac, no `~/.minds/.venv`),
+# uv has to download the python toolchain + install the venv + load
+# plugins first; that can take 30-60s on a fresh machine. A 5s budget was
+# tight enough to deterministically fail every first-time-user launch on
+# a clean Mac (proven via Tart VM). Give it 120s to comfortably cover
+# cold-install while still surfacing a real wedge before the user gives up.
+_MNGR_FORWARD_LISTEN_TIMEOUT_SECONDS: Final[float] = 120.0
```
joshalbrecht, line 95: ok, a fair change, but we'll almost certainly need a loading screen if something is going to take 30-60 seconds, otherwise a user is going to think that it is just broken

apps/minds/imbue/minds/cli/run.py
```diff
@@ -272,6 +279,13 @@ def run(
     # are needed here.
     mngr_host_dir_str = os.environ.get("MNGR_HOST_DIR")
     mngr_host_dir = Path(mngr_host_dir_str).expanduser() if mngr_host_dir_str else (Path.home() / ".mngr")
+    # `mngr forward` and every other laptop-side mngr invocation (including the
+    # bundled mngr CLI when run from a Terminal under this MNGR_HOST_DIR) starts
+    # with cwd=$HOME, so the FCT workspace's `[agent_types.main]` block in
+    # `/code/.mngr/settings.toml` inside the lima VM is invisible to them.
+    # Seed the mapping into user-scope settings.toml here so subsequent mngr
+    # subprocesses resolve `type=main` -> ClaudeAgent without depending on cwd.
+    seed_laptop_agent_types_for_minds(mngr_host_dir)
```
joshalbrecht, line 288: would be nice to remove this warning, this seems like an acceptable fix for now

apps/minds/imbue/minds/cli/run.py
```diff
@@ -342,6 +356,14 @@ def run(
     # allowed-agent list.
     LatchkeyAutoRegister(backend_resolver=backend_resolver, latchkey=latchkey).start()
 
+    # Kick off slow first-create asset prefetch (Ubuntu cloudimg,
+    # FCT clone) in the background. Idempotent: existing-and-valid
+    # caches are skipped. Failures are logged and never block startup;
+    # the lazy fallback at create-agent time keeps working.
+    from imbue.minds.desktop_client.first_launch_prefetch import start_first_launch_prefetch
+
+    start_first_launch_prefetch(paths.data_dir, root_concurrency_group)
```
joshalbrecht: the problem with this is that you can get unlucky and be downloading 2 at once, if I'm understanding correctly?

I wonder if we might want to take this out for now, and then add it back in later anyway (I broke it with my latest changes, since I switched us to Debian 12 instead of Ubuntu, in order to make it a bit smaller and lighter weight)

apps/minds/imbue/minds/config/data_types.py
```diff
@@ -20,7 +21,29 @@
...
+    with "No such file or directory: 'mngr'" under packaging.
+
+    Fall back to bare "mngr" (PATH lookup) only if the sibling binary is
+    absent, which preserves the dev-mode behavior where mngr may be
+    installed as a uv tool outside any specific venv.
+    """
+    sibling = Path(sys.executable).parent / "mngr"
+    if sibling.is_file():
+        return str(sibling)
+    return "mngr"
```
joshalbrecht: a little sketchy, but ok for now

apps/minds/imbue/minds/desktop_client/agent_creator.py
```diff
@@ -574,7 +592,12 @@ def _build_mngr_create_command(
             # transfer + provisioning round the bake already paid for.
             mngr_command.append("--reuse")
```
joshalbrecht, line 592: should we rename this link I suggested in slack, so that it becomes possible to meaningfully specify with --new-host as well?

apps/minds/imbue/minds/desktop_client/app.py
```diff
@@ -1005,6 +1005,27 @@ async def _handle_create_agent_api(request: Request, auth_store: AuthStoreDep) -
         if session_store_inst is not None:
             account_email = session_store_inst.get_account_email(account_id) or ""
 
+    if host_name:
```
joshalbrecht, line 1189: shouldn't we already have errored by this point if there wasn't a host name?

apps/minds/imbue/minds/desktop_client/first_launch_prefetch.py
```diff
@@ -0,0 +1,163 @@
...
+check is the gate. No sentinel file. A user who manually nukes Lima's
+image cache (``~/Library/Caches/lima`` on macOS, ``~/.cache/lima`` on
+Linux) or the template cache will see a fresh prefetch on the next
+launch, which is the right behaviour.
+
+Runs on a background thread off the root concurrency group; failures
+are logged at WARNING and never propagate. The fallback path (lazy
+download at create-agent time) keeps working unchanged.
+"""
+
```
joshalbrecht: we'll need to update this given my latest changes + what I mentioned above

also, ideally we move to a world where we have an already-baked lima image that includes the forever-claude-template, but it's fairly complex

(then there would only be a single thing to download at least though)

apps/minds/imbue/minds/desktop_client/templates.py
```diff
@@ -234,7 +234,12 @@ def render_landing_page(
 # ``_dev_only_workspace_default`` for the gating rationale.
 _FALLBACK_GIT_URL: Final[str] = "https://github.com/imbue-ai/forever-claude-template.git"
 _FALLBACK_HOST_NAME: Final[str] = "assistant"
-_FALLBACK_BRANCH: Final[str] = ""
+# Pin to an annotated FCT tag rather than a moving branch so a shipped
+# binary always clones the exact pilot snapshot it was verified against.
+# Pilot evolves independently; future binaries bump this to a newer tag
+# only after re-verifying. See FCT v0.2.35 (commit 4476621) for the snapshot
+# currently bound to this binary.
+FALLBACK_BRANCH: Final[str] = "v0.2.35"
```
joshalbrecht: kinda weird...  but I'm not sure how to do better

apps/minds/imbue/minds/test_ratchets.py
```diff
@@ -172,7 +179,7 @@ def test_prevent_trailing_comments() -> None:
     # S603 suppression must be on the same line as the call for ruff to
     # recognize it; the noqa marker is intentionally not in the
     # trailing-comment exempt list.
-    rc.check_trailing_comments(_DIR, snapshot(1))
+    rc.check_trailing_comments(_DIR, snapshot(9))
```
joshalbrecht: a good deal of these ratchet increases seem like things that we could fix up

apps/minds/scripts/build.js
```diff
@@ -10,12 +10,96 @@ const fs = require('fs');
...
+// introduced in lima 2.1.0. Lima 2.1.x's usernet forwarder (the path used
+// when the guest's systemd < 256, which includes Ubuntu 24.04, our FCT
+// base) wedges fresh ssh connections post-VM-READY: TCP-accepted on the
+// host then CLOSE_WAIT, no data flow to the in-VM sshd, no
+// git-receive-pack ever spawns -- mngr create hangs forever at
+// "Transferring git repository...". Root cause is the inetaf/tcpproxy
+// "half-close dance" leaking goroutines in io.Copy; lima a2b52885
+// (gvisor-tap-vsock 0.8.7 -> 0.8.8) is the regression boundary.
+// Tracked upstream as lima-vm/lima#4558 + #5042, no fix in flight yet.
+const LIMA_VERSION = '2.0.3';
```
joshalbrecht, line 30: you'll want to double check this one now that we changed to Debian 12 and locked our version of docker

apps/minds/scripts/download-binaries.js
```diff
@@ -184,6 +204,52 @@ async function downloadUv(resourcesDir, { platform, arch }) {
   console.log(`[download-binaries] uv installed at ${uvBinary}`);
 }
 
+async function downloadRestic(resourcesDir, { platform, arch }) {
```
joshalbrecht, line 207: thanks for making all of this restic stuff work out!

(well, mostly--I still haven't seen it quite work on a user's computer yet, but I think it was my own bug that caused that :)  so hopefully it works! )

apps/minds/test_desktop_client_e2e.py
```diff
@@ -53,6 +54,47 @@
...
+
+    This wrapper flips the flag to True for the test's lifetime and restores the
+    original file verbatim on exit. Symmetric with ``_fct_settings_opted_into_pytest``
+    but handles the "key already present with value False" case the FCT version
+    rejects (FCT is not expected to ship the key at all).
+
+    We deliberately mutate the real file rather than redirecting
+    ``MNGR_PROJECT_CONFIG_DIR``: the env var would also redirect the spawned
+    ``mngr create``'s lookup inside the FCT clone, costing access to FCT's own
+    settings.toml that defines the docker create templates.
```
joshalbrecht: I don't understand why we mutate here (instead of creating a separate data directory for this test, like we do for basically every other test)

it's dangerous and brittle to do such mutations, and I don't see a fundamental reason why we need to here

libs/mngr_claude/imbue/mngr_claude/plugin.py
```diff
@@ -850,19 +850,29 @@ def _provision_local_credentials(host: OnlineHostInterface, config_dir: Path, *,
...
-        logger.debug("No credentials found (file does not exist at {})", credentials_path)
+        logger.info("Keychain credentials not available; falling back to file")
+    credentials_path = source_claude_dir / ".credentials.json"
+    try:
+        content = credentials_path.read_text()
+        logger.info("Found .credentials.json at {}", credentials_path)
+        return content
+    except (FileNotFoundError, PermissionError, OSError) as e:
+        logger.info("No credentials found ({}: {})", type(e).__name__, e)
     return None
```
joshalbrecht, line 866: I still want to kill these changes (just like my comment on the FCT PR)

libs/mngr_forward/imbue/mngr_forward/server.py
```diff
@@ -162,12 +162,29 @@ def _parse_workspace_subdomain(host_header: str) -> AgentId | None:
...
+        return Response(status_code=403, content="Not authenticated")
+    host_header = request.headers.get("host", "")
+    agent_id = _parse_workspace_subdomain(host_header)
+    if agent_id is None:
         location = f"http://localhost:{port}/"
-        return Response(status_code=302, headers={"Location": location})
-    return Response(status_code=403, content="Not authenticated")
+    else:
+        location = f"http://localhost:{port}/goto/{agent_id}/"
+    return Response(status_code=302, headers={"Location": location})
```
joshalbrecht, line 187: I don't understand this, and it feels like a bleeding abstraction between "mngr forward" and the things that are being forwarded (who knows if they have a "/goto/{agent_id}/", we don't want to have to assume that)

apps/minds/electron/backend.js
weishi-imbue: updated comments. 
current pr behavior: 
packaged build: -v, so debug level logs gets into minds.log
dev build: -vv.

apps/minds/imbue/minds/cli/run.py
weishi-imbue: The current loading screen is the "Minds" animation. 
Adding a log message box to show progress only after 20s loading time.

weishi-imbue: First X seconds: minds animation (no change)
image


After X seconds:
Pasted Graphic 1

apps/minds/imbue/minds/cli/run.py
weishi-imbue: removed first_launch_prefetch.

apps/minds/imbue/minds/config/data_types.py
weishi-imbue: Empirically verified that `uv run --active` puts the proper venv bin on PATH, so bare `mngr` resolves. this workaround isn't needed if we keep --active. 
Removing it.

apps/minds/imbue/minds/desktop_client/app.py
weishi-imbue: host_name can be user specified, or empty. 
if empty it derives from the repo url: https://github.com/imbue-ai/mngr/blob/bbe848858/apps/minds/imbue/minds/desktop_client/agent_creator.py#L1117-L1123

here it's trying guard against user specified duplicates. empty derived duplicate names are still possible

weishi-imbue: adding a fixme to handle:
1. Minds API use case: empty user input, on the same github url, second time failed due to conflict name. (should fail early, or derive a new name)
2. Minds app use case: user click create on the default name "assistant" the second time, it would fail. (either update the default name, or fail it early)

apps/minds/imbue/minds/test_ratchets.py
weishi-imbue: Updated a bunch of them

libs/mngr_forward/imbue/mngr_forward/server.py
weishi-imbue: The `goto` bridge is a pre-existing path within mngr forward that help handles auth redirect. (So it's a separate address space from those inside forwarded backend. ) 
https://github.com/imbue-ai/mngr/blob/51e0b1657c8869465d9d0fd038de19c8976b43bd/libs/mngr_forward/imbue/mngr_forward/server.py#L918-L919

libs/mngr_claude/imbue/mngr_claude/plugin.py
weishi-imbue: removing these credential wirings for now.

apps/minds/imbue/minds/desktop_client/first_launch_prefetch.py
weishi-imbue: removed this file.

apps/minds/test_desktop_client_e2e.py
weishi-imbue: removed the mutation to real settings. using settings under tmp_path

apps/minds/imbue/minds/desktop_client/agent_creator.py
weishi-imbue: Will work on this as a separate PR scope

libs/mngr_forward/imbue/mngr_forward/server.py
joshalbrecht: you're right, thanks!

I was wrong, this makes sense now

apps/minds/imbue/minds/desktop_client/agent_creator.py
joshalbrecht: totally fine!

PR #1319, opened 2026-04-16
only delete old snapshots in gc

evgunter:
another one where i'm not very familiar with the underlying logic so it needs extra scrutiny

---

## Summary

- Configurable `min_online_host_age_seconds` (default 10 minutes) — minimum age before GC destroys an online host with no agents. Replaces the hardcoded `_MIN_HOST_AGE_DAYS=30` constant with a much shorter default; 30 days was massive overkill for protecting against transient empty states between agent creation and discovery.
- `gc_snapshots` age-gates snapshot deletion against `destroyed_host_persisted_seconds` (default 7 days) instead of nuking immediately. Snapshots survive long enough for recovery via `mngr create --snapshot`.
- `destroy_host()` on Modal/Lima marks host DESTROYED via `stop_reason` instead of clearing snapshot records, so snapshots persist for `gc_snapshots` to age-gate.
- One knob (`destroyed_host_persisted_seconds`) now governs the full afterlife of a destroyed host: record lifetime + snapshot retention.

## Config

- `MngrConfig.default_min_online_host_age_seconds` (float, default 600.0) — global default.
- `ProviderInstanceConfig.min_online_host_age_seconds` (float, nullable) — per-provider override.
- New provider accessor `get_min_online_host_age_seconds()`.

## Test plan

- [x] New unit tests: snapshot age preservation (young snapshots kept), custom age config, destroy_host preserves snapshots + marks DESTROYED
- [x] Config merge tests for new field
- [x] Full test suite via `just test-offload` (7363 passed, 0 failed)


libs/mngr/imbue/mngr/config/data_types.py
```diff
@@ -37,9 +37,16 @@
...
-# 7 days in seconds
+# 7 days in seconds -- controls how long destroyed host records (and their
+# snapshots) are kept before permanent deletion, giving users a recovery
+# window via `mngr create --snapshot`.
 _DEFAULT_DESTROYED_HOST_PERSISTED_SECONDS: Final[float] = 60.0 * 60.0 * 24.0 * 7.0
 
+# 10 minutes -- minimum age before GC will destroy an online host with no agents.
+# Short because we only need to protect against transient empty states (e.g. between
+# agent creation and discovery).
+_DEFAULT_MIN_ONLINE_HOST_AGE_SECONDS: Final[float] = 60.0 * 10.0
```
evgunter, line 48: not sure what the right value is here

libs/mngr/imbue/mngr/api/gc.py
```diff
@@ -328,20 +328,46 @@ def _gc_single_host(
...
+        except (HostAuthenticationError, HostConnectionError) as e:
+            # Cannot determine activity -- err on the side of caution.  HostConnectionError
+            # also catches its HostOfflineError subclass.
+            logger.warning("Cannot determine last activity of host {} during GC, skipping: {}", host.id, e)
+            return
+        if seconds_since_activity is None:
+            # No activity recorded -- conservative: skip.  Should be rare for
+            # mngr-managed hosts (BOOT is recorded on create/start).
+            logger.warning("No activity recorded for host {} during GC, skipping", host.id)
+            return
```
joshalbrecht: this one isn't quite right--if something crashed, we cannot just leave it around forever.  This is the point of the "ok but is it still being set up" timeout, if that makes sense?

we should just check that age here, and maybe even check the status, to ensure that it is considered crashed, and if so, this is something that counts as "should be cleaned up".

libs/mngr/imbue/mngr/config/data_types.py
joshalbrecht: this is ~fine for this value  (this is the value I was mentioning above)

joshalbrecht:
lgtm except for my comments, feel free to merge after fixing those

libs/mngr/imbue/mngr/api/gc.py
evgunter: does it look correct now?

evgunter:
hopefully fixed but worth another glance

joshalbrecht:
yup, thanks!  merging

PR #1325, opened 2026-04-16
remove --command from mngr create

evgunter:
## Summary

Per PR reviewer: replace the removed `--command` flag + per-test factory machinery with a new built-in `command` agent type. Callers do `mngr create NAME --type command -- <shell cmd>` and everything after `--` runs verbatim as the agent body.

## Production-side changes

- New `libs/mngr/imbue/mngr/agents/default_plugins/command_agent.py` registering `"command"` as a built-in agent type whose `assemble_command` returns `" ".join(agent_args)`.
- `libs/mngr/imbue/mngr/cli/create.py`: `--command` click option removed; help text now points at `--type command --`.
- `libs/mngr/imbue/mngr/config/data_types.py`: `command` field removed from `CreateCliOptions`.
- `libs/mngr/docs/concepts/{agents,agent_types}.md` and regenerated `libs/mngr/docs/commands/primary/create.md`: document the new `command` agent type.

## Test-migration approach

Every test that previously minted a per-test `test_sleep_<uuid>` agent type now uses `--type command -- sleep <pinned N>` with a hand-picked unique `N` per call site. That pinned value remains the leak-traceability identifier: a leaked `sleep <N>` in `ps` still greps back to the specific test.

Removed test infra:
- `make_test_sleep_agent_type`, `make_test_sleep_agent_type_at`, `write_agent_type_to_settings_toml` (utils/testing.py)
- `E2eSession.make_sleep_agent_type` and the `settings_local_path` plumbing (e2e/conftest.py)

Clone / migrate CliRunner tests monkeypatch `sys.argv` so the `--` re-insertion logic in `parse_source_and_invoke_create` sees the real argv layout.

## Test plan

- [x] Local `uv run pytest libs/mngr -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'` — 3585 passed, 0 failed.
- [x] Local `uv run pytest libs/mngr/imbue/mngr/cli/test_clone.py libs/mngr/imbue/mngr/cli/test_migrate.py libs/mngr/imbue/mngr/cli/test_destroy.py` — 23 passed.
- [ ] Full offloaded suite (`just test-offload` + `just test-offload-acceptance`) — CI.


libs/mngr/docs/concepts/agents.md
```diff
@@ -32,14 +32,14 @@ mngr create my-agent claude --idle-timeout 1h      # Override timeout
 
 ## Running a Custom Command
 
-Use `--command` to run a literal command instead of using an agent type:
+Any program on `PATH` can be used as an agent type. When the type name is not a registered agent type (like `claude` or `codex`), mngr runs it directly as the command:
```
evgunter: hmmmmmmmm i hope not

evgunter: well, apparently this was preexisting

libs/mngr/docs/concepts/agents.md
```diff
@@ -32,14 +32,14 @@ mngr create my-agent claude --idle-timeout 1h      # Override timeout
...
 
 ```bash
-mngr create my-agent --command "sleep 1000"      # Run a simple command
-mngr create my-agent --command "./my-script.sh"  # Run a custom script
+mngr create my-agent python -- -m http.server 8080   # Run python with args
+mngr create my-agent ./my-script.sh                  # Run a custom script
 ```
 
-The `--command` flag implicitly uses the "generic" agent type, which simply runs the provided command without any special handling. This means `--command` and `--type` are mutually exclusive: you either specify an agent type (like `claude` or `codex`), or you provide a literal command to run.
+To run a literal command with spaces, define an agent type in your config with the desired command (see [Agent Types](./agent_types.md)).
```
evgunter: also somewhat hmm

libs/mngr/imbue/mngr/cli/test_create.py
```diff
@@ -210,8 +163,8 @@ def test_no_connect_flag_skips_tmux_attach(
             [
                 "--name",
                 agent_name,
-                "--command",
-                "sleep 529847",
+                "--type",
+                "test_sleep",
```
evgunter: need to make it so there are different test sleep templates for each value

libs/mngr/docs/concepts/agents.md
```diff
@@ -32,14 +32,14 @@ mngr create my-agent claude --idle-timeout 1h      # Override timeout
...
 
 ```bash
-mngr create my-agent --command "sleep 1000"      # Run a simple command
-mngr create my-agent --command "./my-script.sh"  # Run a custom script
+mngr create my-agent python -- -m http.server 8080   # Run python with args
+mngr create my-agent my-script -- --flag             # Run a custom script
 ```
 
-The `--command` flag implicitly uses the "generic" agent type, which simply runs the provided command without any special handling. This means `--command` and `--type` are mutually exclusive: you either specify an agent type (like `claude` or `codex`), or you provide a literal command to run.
+The type name must be alphanumeric (with dashes/underscores allowed in the middle) -- it's reused as a filesystem and tmux identifier -- so paths like `./script.sh` and names with dots (`python3.12`) don't work as bare type names. To run a literal command that isn't a plain identifier, define an agent type in your config with the desired command (see [Agent Types](./agent_types.md)).
```
evgunter: ugly

joshalbrecht:
wait, I thought we had a simpler solution to this?

> Tests used to do mngr create ... --command 'sleep 847291' with a hand-picked per-test sleep <N> so a leaked process in ps could be grepped back to the specific test. The migration preserves that by pinning explicit sleep values at every call site:

I thought our solution was to do this instead:

`mngr create --type generic -- sleep 847291`

(ideally we rename `generic` to `command`, and then that's even easier to understand...)

libs/mngr/docs/concepts/agents.md
joshalbrecht: see my other top level comment?

evgunter:
ok cool! i don't think that was wired up quite like that before, but it is now; considering that this is available, can/should i remove the "if we don't recognize an agent type, just run it as a command" preexisting behavior?

PR #1297, opened 2026-04-14
make resource guards automatically register pytest marks

evgunter:
we had some resource guards that were getting silently skipped because their corresponding mark was not registered

---

## Summary

Eliminates the need to separately call `register_marker()` for each guarded resource. `register_resource_guard()` (and `register_sdk_guard()`) now handle mark registration automatically via `register_guarded_resource_markers()`.

**What changed:**
- New `register_guarded_resource_markers(config)` helper in the resource-guards library -- registers pytest marks for all guarded resources. Used by both standalone users (in their own `pytest_configure`) and our monorepo (via `conftest_hooks._pytest_configure`)
- `register_sdk_guard()` adds guard names to `_guarded_resources` at registration time (not deferred to session start), so they're visible at `pytest_configure` time
- Removed all redundant `register_marker()` calls from project conftest files (conftest.py, libs/mngr, libs/mngr_modal, libs/mngr_pair, libs/mngr_notifications, libs/mngr_lima, libs/mngr_claude)
- Added guard registrations to `libs/mngr/imbue/mngr/conftest.py` (the inner conftest that `pytest_plugins` actually imports), so projects inheriting via `pytest_plugins` get guards registered at import time
- Updated resource-guards README with the standalone setup pattern
- `get_guarded_resource_names()` returns an immutable `tuple[str, ...]`

## Test plan
- [x] resource_guards: 95 passed, 100% coverage
- [x] imbue_common: 214 passed, 95.20% coverage
- [x] mngr: 3600 passed, 82.81% coverage
- [x] mngr_claude: 393 passed, 82.68% coverage
- [x] mngr_kanpan: 313 passed
- [x] mngr_notifications: 111 passed
- [x] mngr_pair: 104 passed
- [x] mngr_lima: 98 passed
- [x] mngr_modal: 374 passed
- [ ] CI validates all projects

PR #1326, opened 2026-04-16
partition options needed for offload and local dev; tell claude to use justfile test-quick locally; enforce per-project coverage thresholds in ci

evgunter:
## Summary

Originally a narrow fix for the offload `pytest-timeout` wipe bug that caused PR #1297 to hang at the 60-min sandbox wall clock. Scope expanded during review into a broader cleanup of pytest/coverage config on this repo. End state:

- Drop `-o addopts=` from `offload-modal.toml` (and `offload-modal-acceptance.toml`) so `--timeout=10 --timeout-method=signal` actually apply inside offload sandboxes (the original bug).
- Prune offload-breaking and local-only flags from every `pyproject.toml` addopts list: `-n N`, `--dist=worksteal`, `--max-worker-restart=0`, `--cov-report=html`, `--cov-fail-under=N`, and `-m 'not acceptance and not release'` (this last one caused silent acceptance-test collection failure — see below).
- Move xdist parallelism, HTML coverage report, and the local default mark filter into `just test-*` recipes via shared `_parallel` / `_skip_acceptance_and_release` variables. Extend `test-quick` to forward args.
- Move per-package coverage thresholds from addopts into `[tool.coverage.report] fail_under` in each subpackage pyproject.
- Rewrite the CI "Combine coverage and check threshold" step into a per-package loop, with a diagnostic block that correlates per-sandbox `.coverage` delivery against per-sandbox junit.
- Bump pytest-timeout on two tests that legitimately take longer than 10s (`test_create_work_dir_copy_with_gitignored_files`, `test_migrate_clones_and_destroys_source`).
- Make `schedule` listing operations deterministic (`sorted(profile_dir.iterdir())`, `sorted(volume.listdir())`).
- Start measuring coverage for `mngr_vps_docker` and `mngr_vultr` (just missing from root's `--cov=` list), and exclude them from the combined-coverage backstop via `libs/mngr/pyproject.toml`'s `[tool.coverage.report].omit` (same treatment as `*/minds/*`) so the 80% combined gate reflects the main package set while each package is still gated individually.
- Mark `test_apply_work_dir_extra_paths_copy_mode_copies_files` with `@pytest.mark.flaky` (flagged as flaky by an autofix run).
- Update `CLAUDE.md` to point agents at `just test-quick <path>` instead of `(cd libs/X && uv run pytest ...)` subshells.

## Limitations / caveats

1. **Coverage variance in `mngr_wait` and `mngr_file`**. Both packages oscillate ±1pt between runs despite identical test sets passing every time. Thresholds were lowered to 75 with headroom (see in-file comments). The CI coverage-delivery diagnostic confirms every sandbox delivers its `.coverage` file — this is real test-code nondeterminism (tests hit different branches run-to-run), not data loss. Eliminating it requires targeted work in `mngr_wait/cli.py` and in `mngr_file`'s test suite; out of scope for this PR.

2. **Acceptance offload initially broke silently**. Dropping `-o addopts=` on `offload-modal-acceptance.toml` caused pytest to inherit subpackage addopts when it auto-detected rootdir as a subpackage (e.g. `libs/mngr/`). The inherited `-m "not acceptance and not release"` deselected every acceptance test before offload's `-m 'acceptance and not docker and not docker_sdk'` filter was applied. Orchestrator then reported "Expected 32 test IDs, only 1 in report" every run. Resolved structurally by moving the "skip acceptance and release" filter out of addopts entirely (into `just test-unit`/`test-integration`/`test-quick` recipes), which was the proper home anyway.

3. **Occasional Modal infra flakes** cause a single sandbox to fail its junit download ("Expected 7374 test IDs, only 7364 in report"). Observed on ~1 in 4 runs. Not caused by this PR — same mechanism as the original PR #1297 stuck-sandbox bug. The new CI diagnostic will name the offending sandbox when it fires. Retrying the failed job generally clears it.

4. **`PYTEST_NUMPROCESSES` env var is now near-dead-code**. Its override logic only fires when xdist is active but no CLI `-n` is present. All canonical entry points (just recipes, offload) now always pass explicit `-n` or `-p no:xdist`, so this narrow window is effectively unreachable. Docstring updated to be honest about the new state. Fully removing the code wasn't in scope.

5. **Per-package CI gates revealed real coverage gaps.** Thresholds were calibrated down to the current measured coverage (rounded down to the nearest 5%) for packages that were below their declared value — `imbue_common` 95→90, `mngr_claude` 80→30, `mngr_lima` 75→50, `mngr_tmr` 60→50, `minds` 80→70, `minds_workspace_server` 80→75 (plus the wait/file lowering noted above). Ratcheting up is follow-up work.

6. **`mngr_modal` is still not measured in CI.** Its tests are marked `@pytest.mark.acceptance`/`@pytest.mark.release` and only run in `offload-modal-acceptance` (which uses `--no-cov`) or the release job. Capturing coverage requires (a) switching acceptance offload to collect coverage, (b) removing `libs/mngr_modal/imbue/mngr_modal/*` from root's `[tool.coverage.run].omit`, (c) adding `--cov=imbue.mngr_modal` to root addopts, and (d) plumbing acceptance's `.coverage` files into the post-offload combine step. Bigger infra change; left for a follow-up. `mngr_vps_docker` and `mngr_vultr` were measurable with just a `--cov=` addition (their unit tests already run in main test-offload), and are now gated individually at 40 and 60 respectively.

## Motivation

PR #1297's `test-offload` job was hanging for a full hour. Investigation showed one Modal sandbox got stuck on `test_register_resource_guard_auto_registers_pytest_mark` (a new `pytester`-subprocess test) and burned the 60-min sandbox wall clock before being killed. `--timeout=10 --timeout-method=signal` would have killed it — but `offload-modal.toml` had `-o addopts=` which nuked all of `pyproject.toml`'s addopts, including pytest-timeout. The `-o addopts=` wipe existed because root addopts mixed universal-safe flags (timeout, --cov) with local-dev-only flags (xdist, html coverage, cov-fail-under, local-only mark filter) that break offload. This PR resolves that conflict by splitting responsibilities.

## Test plan

- [x] `just test-offload` passes in CI (modulo occasional Modal infra flake; retries clear)
- [x] `just test-offload-acceptance` passes in CI (fixed in this PR — previously broken by the addopts refactor)
- [x] `just test-docker` passes
- [x] Unit + Integration Tests + Acceptance Tests report green
- [x] CI coverage-delivery diagnostic prints `sandboxes with .coverage == sandboxes with junit` on healthy runs (verified)
- [x] Per-package CI gate catches real threshold violations (verified during PR with a temporary `modal_proxy` bump to 99%, which produced a gate-named failure as expected)
- [x] `just test-unit`, `just test-integration`, `just test-quick` run locally with parallelism and coverage gates as before
- [x] `just test-quick libs/mngr` and `just test-quick "libs/mngr -m 'not tmux and not modal'"` both work
- [x] Fully green CI run on the merge commit (7de7087d5)


libs/mngr/pyproject.toml
```diff
@@ -124,6 +119,12 @@ omit = [
   "*/providers/docker/testing.py",
   # Minds app has its own coverage configuration (apps/minds/pyproject.toml)
   "*/minds/*",
+  # These packages are gated individually by their own (lower) fail_under
+  # thresholds; excluding them from the combined report keeps the
+  # combined 80% gate focused on the "main" set of packages and
+  # prevents their lower coverage from dragging the combined mean down.
+  "*/mngr_vps_docker/*",
+  "*/mngr_vultr/*",
```
joshalbrecht, line 127: I'll make a follow-up for myself on these

joshalbrecht:
lgtm, feel free to merge

though I'm not totally sure about those comments at the very top about how coverage sometimes doesn't get downloaded or whatever... maybe worth taking a quick look to make sure you're happy with that before merging

evgunter:
jacob fixed the coverage issue in offload, though i think a new release hasn't been cut yet

PR #1328, opened 2026-04-16
wait for child concurrency groups to exit

### LINKED ISSUE #1320: ConcurrencyExceptionGroup on Ctrl+C: _exit() doesn't wait for child CGs in flat hierarchy
test run of https://github.com/imbue-ai/mngr/pull/1311 - are there extra details that would be good to include here (or extraneous things to exclude) that i should add to the prompt? (also i'm not sure yet if the explanation is correct)

(i already added a note that it should include the mngr version where the issue was encountered and the commit hash the agent was looking at, plus versions of any relevant tools)

---

## Summary

`ConcurrencyGroup._exit()` checks child CG states but does not **wait** for them to finish. Combined with the flat CG hierarchy (where all provider-created CGs are registered as children of the top-level CG), this causes `ChildConcurrencyGroupDidNotExitError` when anything interrupts the normal synchronous execution flow -- most commonly a `KeyboardInterrupt` (Ctrl+C) during slow Modal API calls.

### 1. Flat CG hierarchy

In `_run_discovery` (`libs/mngr/imbue/mngr/api/discover.py:93-106`), the top-level CG is passed to each provider:

```python
with ConcurrencyGroupExecutor(
    parent_cg=mngr_ctx.concurrency_group, name="discover_hosts_and_agents", max_workers=32
) as executor:
    for provider in providers:
        futures.append(
            executor.submit(
                _discover_provider_hosts_and_agents,
                provider, ...,
                mngr_ctx.concurrency_group,  # top-level CG passed to providers
            )
        )
```

The modal provider then creates its own CGs as **children of the top-level CG** (`libs/mngr_modal/imbue/mngr_modal/instance.py:2363-2364`, `instance.py:692-693`):

```python
# Inside a thread managed by discover_hosts_and_agents:
with ConcurrencyGroupExecutor(
    parent_cg=cg, name="modal_discover_hosts_and_agents_modal", ...
) as executor:
    # Inside a thread managed by modal_discover_hosts_and_agents_modal:
    with ConcurrencyGroupExecutor(
        parent_cg=cg, name="modal_list_all_host_records", ...
    ) as executor:
```

This results in a **flat sibling hierarchy** where all three CGs are direct children of the top-level CG, even though they're logically nested:

```
top-level CG (mngr-list)
  +-- discover_hosts_and_agents          (managed by _run_discovery on main thread)
  +-- modal_discover_hosts_and_agents_modal  (managed by thread in discover_hosts_and_agents)
  +-- modal_list_all_host_records            (managed by thread in modal_discover_hosts_and_agents_modal)
```

### 2. `_exit()` doesn't wait for child CGs

`ConcurrencyGroup._exit()` at `concurrency_group.py:145-195` does two things:

1. **Waits** for its own direct threads/processes via `_wait_for_all_strands_to_finish_with_timeout` (line 150-154)
2. **Checks** (but does NOT wait for) child CG states (line 179-183)

The top-level CG has few or no direct threads -- the discovery threads belong to child CGs. So step 1 returns almost immediately, and step 2 finds children that are still ACTIVE or EXITING.

### 3. Trigger: KeyboardInterrupt during Modal API calls

In normal execution, `_run_discovery()` blocks synchronously until all threads complete, so by the time the top-level CG exits, all children are EXITED. The bug manifests when `KeyboardInterrupt` interrupts this blocking wait:

1. Main thread is blocked in `discover_hosts_and_agents._exit()` -> `_wait_for_all_strands_to_finish_with_timeout()` -> `thread.join(None)`, waiting for Modal API threads
2. User presses Ctrl+C
3. `KeyboardInterrupt` (a `BaseException`, not `Exception`) is **not caught** by the `except Exception: pass` at `concurrency_group.py:233`
4. It propagates through `_exit()` -> `__exit__()` (which sets `discover_hosts_and_agents` to EXITED via `finally`)
5. Exception propagates through Click, which triggers `ctx.call_on_close` -> `cg.__exit__(None, None, None)` for the top-level CG
6. Top-level CG finds `modal_discover_hosts_and_agents_modal` (EXITING) and `modal_list_all_host_records` (ACTIVE) -- their daemon threads are still running

The `except Exception: pass` at line 233 is the specific enabler -- it was meant to be defensive, but `KeyboardInterrupt` slips through because it's a `BaseException`.

## Error traceback

```
ConcurrencyExceptionGroup: A child concurrency group did not exit: `discover_hosts_and_agents` (state: EXITING). (3 sub-exceptions)
  +-- ChildConcurrencyGroupDidNotExitError: A child concurrency group did not exit: `discover_hosts_and_agents` (state: EXITING).
  +-- ChildConcurrencyGroupDidNotExitError: A child concurrency group did not exit: `modal_list_all_host_records` (state: ACTIVE).
  +-- ChildConcurrencyGroupDidNotExitError: A child concurrency group did not exit: `modal_discover_hosts_and_agents_modal` (state: EXITING).
```

## Reproduction

Ctrl+C during any command that does Modal discovery (`mngr list`, `mngr create`, etc.) while the Modal API is responding slowly. The bug is timing-sensitive -- the interrupt must arrive while `thread.join()` is blocking in `_wait_for_all_strands_to_finish_with_timeout`.

## Suggested fix

Two complementary changes:

**1. Catch `BaseException` (not just `Exception`) in `_wait_for_all_strands_to_finish_with_timeout`** (`concurrency_group.py:233`):

```python
# Before:
except Exception:
    pass

# After:
except BaseException:
    pass
```

This prevents `KeyboardInterrupt` from aborting the thread-join loop, giving the CG a chance to properly wait for and terminate its threads.

**2. Add a wait-for-children step in `_exit()`** (before line 179):

```python
# Wait for child CGs to finish exiting (they may be on threads we don't directly own)
if self._children:
    deadline = time.monotonic() + self.exit_timeout_seconds
    for child in self._children:
        while child.state not in (ConcurrencyGroupState.EXITED, ConcurrencyGroupState.INSTANTIATED):
            remaining = deadline - time.monotonic()
            if remaining <= 0:
                break
            time.sleep(min(0.05, remaining))
```

This makes the children check resilient to timing issues, regardless of the trigger.

**3. (Longer-term) Restructure the CG hierarchy** so provider-created CGs are children of the `discover_hosts_and_agents` CG, not the top-level CG. This would let the natural thread-wait mechanism handle them. This would require passing the executor's CG (not `mngr_ctx.concurrency_group`) through to providers.

## Files involved

- `libs/concurrency_group/imbue/concurrency_group/concurrency_group.py:179-183` -- children check without wait
- `libs/concurrency_group/imbue/concurrency_group/concurrency_group.py:233` -- `except Exception` misses `KeyboardInterrupt`
- `libs/mngr/imbue/mngr/api/discover.py:93-106` -- passes top-level CG to providers
- `libs/mngr/imbue/mngr/cli/common_opts.py:147-151` -- top-level CG creation and `call_on_close` exit
- `libs/mngr_modal/imbue/mngr_modal/instance.py:2363-2364,692-693` -- modal CGs created as children of top-level CG

evgunter:
this fix makes sense to me but also makes me suspicious that i'm missing something about how the concurrency group exit is supposed to work

---

## Summary

- Fix `ConcurrencyGroup._exit` racing against children that live on other threads, which caused a spurious `ChildConcurrencyGroupDidNotExitError` cascade on Ctrl+C during mngr discovery.
- Add a per-CG `_exited_event` and a `_wait_for_children_to_exit` step (shared deadline) before the existing state check in `_exit`.

## Root cause

mngr discovery creates modal sub-CGs as siblings of `discover_hosts_and_agents` (all children of the top-level CG). A slow Modal API call + Ctrl+C on the main thread abandons `thread.join` before the worker threads return, leaving sibling CGs in `ACTIVE`/`EXITING`. The `ctx.call_on_close(lambda: cg.__exit__(None, None, None))` in `libs/mngr/imbue/mngr/cli/common_opts.py` then hits the children check and raises. I reproduced both with Ctrl+C and, without any signal, by exiting a parent CG while a child was still running on another thread.

The library-level fix addresses the race regardless of how it's triggered; the issue's suggested `except BaseException` would silently swallow the user's Ctrl+C, so I deliberately skipped it.

## Test plan

- [x] New regression test `test_parent_exit_waits_for_child_group_on_other_thread_to_exit` (passes with fix; fails without).
- [x] `uv run pytest libs/concurrency_group/` - 154 passed.
- [x] Manual: reproduction script with SIGINT during nested CG now exits cleanly once children finish (previously raised `ChildConcurrencyGroupDidNotExitError`).
- [ ] CI to run broader test suite.

Closes #1320

### DIFF
--- a/libs/concurrency_group/imbue/concurrency_group/concurrency_group.py
+++ b/libs/concurrency_group/imbue/concurrency_group/concurrency_group.py
@@ -150,11 +150,10 @@ class ConcurrencyGroup(MutableModel, AbstractContextManager):
         timeout_exception_group: ConcurrencyExceptionGroup | None = None
         failure_exception_group: ConcurrencyExceptionGroup | None = None
 
+        total_timeout_seconds = self.shutdown_timeout_seconds if self.is_shutting_down() else self.exit_timeout_seconds
+        exit_start_time_seconds = time.monotonic()
         try:
-            if self.is_shutting_down():
-                self._wait_for_all_strands_to_finish_with_timeout(self.shutdown_timeout_seconds)
-            else:
-                self._wait_for_all_strands_to_finish_with_timeout(self.exit_timeout_seconds)
+            self._wait_for_all_strands_to_finish_with_timeout(total_timeout_seconds)
         except ConcurrencyExceptionGroup as exception_group:
             timeout_exception_group = exception_group
 
@@ -179,10 +178,8 @@ class ConcurrencyGroup(MutableModel, AbstractContextManager):
                 exceptions.append(main_exception)
                 message = str(main_exception)
 
-        children_timeout_seconds = (
-            self.shutdown_timeout_seconds if self.is_shutting_down() else self.exit_timeout_seconds
-        )
-        self._wait_for_children_to_exit(children_timeout_seconds)
+        remaining_timeout_seconds = self._get_remaining_timeout(exit_start_time_seconds, total_timeout_seconds)
+        self._wait_for_children_to_exit(remaining_timeout_seconds)
         for child in self._children:
             if child.state not in (ConcurrencyGroupState.EXITED, ConcurrencyGroupState.INSTANTIATED):
                 child_message = f"A child concurrency group did not exit: `{child.name}` (state: {child.state})."
--- a/libs/concurrency_group/imbue/concurrency_group/concurrency_group.py
+++ b/libs/concurrency_group/imbue/concurrency_group/concurrency_group.py
@@ -5,6 +5,7 @@ from enum import auto
 from functools import wraps
 from pathlib import Path
 from subprocess import TimeoutExpired
+from threading import Event
 from threading import Lock
 from typing import Any
 from typing import Callable
@@ -111,6 +112,7 @@ class ConcurrencyGroup(MutableModel, AbstractContextManager):
 
     _lock: Lock = PrivateAttr(default_factory=Lock)
     _children: list["ConcurrencyGroup"] = PrivateAttr(default_factory=list)
+    _exited_event: Event = PrivateAttr(default_factory=Event)
 
     # Did the concurrency group already exit with an exception?
     _exit_exception: BaseException | None = PrivateAttr(default=None)
@@ -141,6 +143,7 @@ class ConcurrencyGroup(MutableModel, AbstractContextManager):
             raise
         finally:
             self._state = ConcurrencyGroupState.EXITED
+            self._exited_event.set()
 
     def _exit(self, exc_value: BaseException | None) -> None:
         main_exception: BaseException | None = exc_value if exc_value is not None else None
@@ -176,6 +179,10 @@ class ConcurrencyGroup(MutableModel, AbstractContextManager):
                 exceptions.append(main_exception)
                 message = str(main_exception)
 
+        children_timeout_seconds = (
+            self.shutdown_timeout_seconds if self.is_shutting_down() else self.exit_timeout_seconds
+        )
+        self._wait_for_children_to_exit(children_timeout_seconds)
         for child in self._children:
             if child.state not in (ConcurrencyGroupState.EXITED, ConcurrencyGroupState.INSTANTIATED):
                 child_message = f"A child concurrency group did not exit: `{child.name}` (state: {child.state})."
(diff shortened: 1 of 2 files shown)


joshalbrecht:
lgtm. A bit tricky, but it does seem correct

PR #1333, opened 2026-04-17
propagate_changes: auto-discover minds agent, add --agent flag

gnguralnick:
## Summary

- Point `mngr list/stop/start` at `~/.minds/mngr` (derived from `MINDS_ROOT_NAME`, default `minds`) so `propagate_changes` finds minds-created agents instead of the wrong set under mngr's default `~/.mngr`.
- Add `--agent NAME` CLI flag. Resolution order: flag > `MINDS_WORKSPACE_NAME` env > auto-pick when exactly one agent is running. On multi-match or no match, list candidates and bail.
- If `mngr stop` fails for the resolved name, print the known-agent list so the user can retry with the right `--agent`.

## Test plan

- [x] No args, one minds agent running -> auto-selects it by name.
- [x] No args, zero minds agents -> errors with "No mngr agents found".
- [x] `--agent foo` with no agent -> uses `foo`, later fails in stop with helpful candidate list.
- [ ] Full end-to-end propagate run against a live minds container.

### DIFF
--- a/apps/minds/scripts/propagate_changes
+++ b/apps/minds/scripts/propagate_changes
@@ -13,6 +13,11 @@
 #   # Local (non-container agent):
 #   propagate_changes --target /path/to/agent/workdir
 #
+# Agent name resolution (first non-empty wins):
+#   --agent NAME  (CLI flag)
+#   MINDS_WORKSPACE_NAME  (env var)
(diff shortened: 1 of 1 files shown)


joshalbrecht:
lgtm

PR #1334, opened 2026-04-17
multi-window workspaces for minds desktop

gnguralnick:
## Summary

Each workspace can now live in its own Electron window, with uniqueness enforced across the app (one window per workspace max). The full behavior is specced in [`specs/multi-window-workspaces/concise.md`](specs/multi-window-workspaces/concise.md).

**User-visible features**
- Right-click a workspace row in the sidebar for a native "Open in new window" context menu, or use the hover-revealed icon on the right of each row. The current workspace's row suppresses both affordances.
- Plain sidebar click navigates the current window unless the target is already open elsewhere, in which case that window is focused. Either way, the sidebar closes after the click.
- `cmd/ctrl+N`, `File > New Window`, and the macOS dock icon's right-click menu all open a fresh window pointed at `/`.
- OS window title mirrors into the in-window title bar: `{workspace-name} — Minds` on workspace windows, `Minds` elsewhere.
- Notification and request-panel clicks for `/forwarding/{X}/...` focus the existing window for X or open a new one; non-workspace URLs (e.g. `auth_required`, `/accounts`) navigate the most-recently-focused window.
- `cmd/ctrl+W` closes one window; backend stays up until the last window closes. `cmd/ctrl+Q` closes all and shuts down.
- On quit, the set of open windows + content URLs persists to `~/.{MINDS_ROOT_NAME}/window-state.json` and is restored on next launch. Workspaces that no longer exist are silently skipped on restore.

**Implementation**
- `main.js`: module-level window state replaced with a per-window bundle registry. IPC handlers resolve the target window from `event.sender.id` by scanning live bundles (no map — it desynced when `webContents.destroyed` didn't fire reliably for BaseWindow children). `will-navigate` listener on every content view enforces workspace uniqueness regardless of how the navigation was triggered.
- Centralized SSE: the main process runs exactly one `net.request` (with `useSessionCookies: true`) to `/_chrome/events` and broadcasts via `chrome-event` IPC to every chrome/sidebar view. Previously each chromeView and sidebarView opened its own `EventSource`, saturating Chromium's 6-connection-per-host cap and making every subsequent backend HTTP request queue behind a kept-alive SSE stream (observed: sidebar/requests-panel loads going from ~50ms to 8+ seconds).
- Sidebar and requests panel are created lazily once per window and toggled with `setVisible`. Destroying and recreating a `WebContentsView` on every click spawned a fresh render process each time; rapid-clicking stacked up seconds of latency.
- Explicit `webContents.close()` on every child view in the BaseWindow `close` handler, since BaseWindow doesn't reliably destroy its child render processes.
- `will-prevent-unload` → `event.preventDefault()` on content views so pages with live websocket `beforeunload` handlers don't stall navigations waiting for a non-existent confirmation dialog.

## Test plan

- [x] `uv run pytest apps/minds/... -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'` -- 597 passed
- [x] `node --check` on `main.js` and `preload.js`
- [x] Manually exercised in dev mode with two local Docker workspaces: sidebar right-click + hover-icon, cmd+N / File menu / dock menu, plain click routing, cross-workspace focus-existing, session restore across quit-and-relaunch, cmd+W vs cmd+Q vs last-window-close, requests panel per-window
- [x] Verified the centralized-SSE fix resolves accumulating slowness observed after several open/close cycles
- [ ] CI

## Notes

- Pre-existing bug surfaced by session restore: `_handle_login` in `apps/minds/imbue/minds/desktop_client/app.py` sets the `minds_session` cookie without `max_age`, so it's browser-session-lifetime and gets dropped every time Electron quits. Session restore works around this by probing auth up-front (via the SSE) and routing the initial window through `loginUrl` if unauthenticated. Worth fixing separately by adding `max_age=_COOKIE_MAX_AGE_SECONDS` to that `set_cookie` call.
- Backend is unchanged aside from `_SIDEBAR_TEMPLATE` / `_CHROME_TEMPLATE` updates for the new IPC bridge and per-row affordances.

### DIFF
--- a/apps/minds/docs/desktop-app.md
+++ b/apps/minds/docs/desktop-app.md
@@ -37,15 +37,28 @@ When accessing an agent URL in a regular browser (not the Electron app), the Pyt
 
 ### Shutdown
 
-When the user closes the window, Electron sends SIGTERM to the backend process and waits up to 5 seconds. If the process doesn't exit, SIGKILL is sent.
+Closing an individual window just tears down that window's views -- the backend keeps running while any window is open. When the last window closes (or the user issues `Cmd+Q` / `Ctrl+Q`), Electron sends SIGTERM to the backend process and waits up to 5 seconds. If the process doesn't exit, SIGKILL is sent.
 
 ### Crash recovery
 
-If the backend exits unexpectedly, Electron shows an error screen with the last lines from the log file and a "Retry" button that restarts the backend.
+If the backend exits unexpectedly, every open window switches to the error screen (chrome view expanded to fill the window, content/sidebar/requests-panel views torn down) with the last lines from the log file. Clicking "Retry" from any window restarts the backend once; on success every window reloads to its pre-error URL.
 
 ### Keyboard shortcuts
 
 - **Open DevTools**: `Ctrl+Shift+C` (Windows/Linux) or `Cmd+Option+I` (macOS)
+- **New Window**: `Ctrl+N` / `Cmd+N` -- opens a fresh window on the home page. Also available on macOS via `File > New Window` and the dock icon's right-click menu.
+- **Close Window**: `Ctrl+W` / `Cmd+W` -- closes the focused window; the backend keeps running until the last window closes.
+- **Quit**: `Ctrl+Q` / `Cmd+Q` -- closes every window and shuts the backend down.
+
+### Multi-window behavior
+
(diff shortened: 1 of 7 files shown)


joshalbrecht:
lgtm, thanks!

PR #1337, opened 2026-04-19
mngr schedule tests: isolate HOME, fix coverage, add release test

evgunter:
## Summary

Schedule test infrastructure changes against main, with #1175 (create-headless) and #1338 (safe-commands-headless) already landed.

- Rewrite `build_subprocess_env()` in `libs/mngr_schedule/imbue/mngr_schedule/testing.py` to keep the subprocess HOME isolated. Previously the helper restored the developer's real HOME so the subprocess's Modal SDK could find `~/.modal.toml`, leaving the subprocess in a split-brain state (HOME pointed at `/Users/ev` but `MNGR_HOST_DIR` still at the test tmp dir) that trips plugin code assuming `host_dir` lives under `$HOME` (`get_files_for_deploy`'s `relative_to(user_home)`). Extract `load_modal_creds_from_home()` — reads TOML once at helper time, passes tokens in as `MODAL_TOKEN_ID` / `MODAL_TOKEN_SECRET` env vars. The real HOME is now only used to load that one file; the subprocess never sees it. Also auto-generates `MNGR_PREFIX` via `generate_test_environment_name()` so the Modal backend's `mngr_test-` guard accepts it and the cleanup sweeper can find it, and deliberately does NOT strip `PYTEST_CURRENT_TEST` — evading the config `is_allowed_in_pytest` and Modal env-creation guards has leaked un-sweepable envs in the past.
- Reuse the shared helper in `test_schedule_add.py` (dropped its inline `_build_subprocess_env` + `_REAL_HOME`).
- Add `test_schedule_local_lifecycle.py`: one `@pytest.mark.release` test exercising the full add+remove lifecycle against the real `crontab` binary. Asserts the `# {prefix}schedule:{name}` marker appears in `crontab -l` after add and is gone after remove, deriving the prefix from `env["MNGR_PREFIX"]`. Unique trigger name per run for xdist safety. Reuses `remove_test_trigger` in the finally block. 300s pytest timeout.
- **Fix fast-suite coverage for `schedule remove --force` found-trigger dispatch** — this path was 0%-covered on main, leaving `libs/mngr_schedule` at 76.83% total, below the 77% `fail_under` threshold (a pre-existing regression). Added two tests in `test_cli.py`:
  - `test_schedule_remove_local_force_skips_prompt_and_dispatches`: deploy fake trigger → `remove --force` → assert exit 0 + no `"Are you sure"` + `"Removed schedule"` output. Mirror of the existing `test_schedule_remove_local_prompts_without_force`.
  - `test_schedule_remove_local_force_dispatches_to_provider_remove`: monkeypatches `remove_local_schedule` with a recorder, asserts the CLI dispatched the expected name to the provider. Asserts directly on CLI wiring without relying on downstream side effects.
  - Bumps `PREVENT_MONKEYPATCH_SETATTR` ratchet 0 → 1 for the mock test. No DI seam on `schedule_remove`; monkeypatch is the cleanest way to make the dispatch claim without adding test-only plumbing to the CLI command.
- Minor `plugin_test.py` touch: pass `isolated_home` into main's new `test_get_files_for_deploy_reads_config_from_deployer_host_dir` for consistency with surrounding tests.

## Test plan
- [x] Fast unit + integration suite for `libs/mngr_schedule` passes (281 passed)
- [x] Coverage: 77.37% (> 77% `fail_under` threshold, up from a broken-on-main 76.83%)
- [x] Release test `test_schedule_run_and_remove_modal_trigger` passes against real Modal
- [x] `test_schedule_local_add_and_remove_lifecycle` passes in a Modal sandbox via `just test-offload-release` (17s, 1/1)
- [ ] CI green


joshalbrecht:
lgtm

PR #1338, opened 2026-04-19
skip ensure-safe-commands when --foreground is set

evgunter:
## Summary
- `check_safe_create_command` required `--branch {DATE}` or `--reuse`, but the headless create path rejects *both* flags (see `_HEADLESS_INCOMPATIBLE_FLAGS` in `libs/mngr/imbue/mngr/cli/create.py`). Scheduled `--foreground` agents therefore could never pass the safety gate.
- Safety check exists to prevent per-run name/branch collisions for persistent agents. `--foreground` agents auto-destroy after each run and create no branch, so the collision scenario does not exist.
- Skip the check entirely when `--foreground` is present in the scheduled args. Auto-fix behavior (`--headless`, `--no-connect`, `--host-label SCHEDULE=`) is unchanged.

## Test plan
- [x] New unit tests `test_passes_with_foreground`, `test_foreground_after_separator_does_not_count`
- [x] Full offload run: 7425 passed, 1 failed (`test_schedule_remove_local_with_force`, pre-existing flaky marker)


joshalbrecht:
I said somewhere else too, but I don't think _HEADLESS_INCOMPATIBLE_FLAGS should have quite so many flags in it

Seems fine to allow --branch, for example? I suspect that *most* flags are fine (--reuse probably is not)

joshalbrecht:
otherwise lgtm (but this is marked as draft, so will let you merge)

PR #1339, opened 2026-04-19
accept a bunch more flags for headless agents

evgunter:
## Summary

Unifies the headless `mngr create --foreground` path with the interactive create path so headless agents inherit every source/transfer/git/provisioning/identity option for free, while keeping `mngr ask`'s throwaway-directory semantics explicit.

- **Shared pipeline.** `_create_headless` is gone. Both paths now run `_setup_create` → `_create_agent`; they only diverge afterwards: `_stream_and_destroy_headless_agent` (headless) vs `_post_create` + `_finish_create` (interactive). Source resolution, `--transfer`, `--branch`, `--rsync*`, `--ensure-clean`, `--env`/`--env-file`/`--pass-env`, `--grant`, `--extra-provision-command`, `--upload-file`, `--extra-window`, `--target-path`, `--id`/`--label`/`--project`, new-host creation, and pinned target hosts all work on headless now. `_HEADLESS_INCOMPATIBLE_FLAGS` is reduced to genuinely interaction-phase flags: `--edit-message`, `--reconnect`, `--attach-command`, `--connect-command`, `--reuse`, `--update`, `--start-on-boot`, and explicit `--connect`.
- **Default matches non-headless.** `mngr create --foreground` with no flags resolves source the same way as interactive create (git root default, or `--source`), and uses the full transfer pipeline. Pass `--transfer=none` to run in-place at the source; that's how you get the old "blank-dir" intent on the CLI.
- **`--message` / `--message-file` on headless.** Delivered via `CreateAgentOptions.initial_message` and a new `StreamingHeadlessAgentMixin.stage_initial_message` instance method, which `api_create` dispatches after `create_agent_state` but before `start_agents` for any streaming headless agent (short-circuiting the wait-for-ready + `send_message` dance that raises on headless). `HeadlessClaude.stage_initial_message` writes `.mngr-prompt` into `$MNGR_AGENT_STATE_DIR`; its `assemble_command` receives `initial_message` as a kwarg and appends a `"$(cat "$MNGR_AGENT_STATE_DIR/.mngr-prompt")"` arg. State-dir placement means the file is removed automatically when the agent is destroyed — no leak into in-place source directories. Default `stage_initial_message` on the mixin is a `logger.warning` so agent types that don't opt in don't silently drop the user's message.
- **`mngr ask` stays opt-in for ephemeral.** New `ephemeral_work_location` context manager in `headless_runner` handles the mktemp + cleanup dance. `ask.py` wraps `ephemeral_work_location` inside `headless_agent_output`, passes the user query via `initial_message`, and no longer writes the prompt file from its own `pre_create_setup`.
- **`assemble_command` signature.** Gains `initial_message: str | None = None`, threaded from `CreateAgentOptions` through `Host.create_agent_state`. Headless agent types read it there (data.json isn't written yet at that point, so they can't use `get_initial_message()`). `CommandAgent`, `HeadlessCommand`, and `ClaudeAgent` overrides updated.
- **Rejected target pinning is gone.** Pinning a target via `@HOST`/`@.PROVIDER`/`--provider` used to be rejected on headless because the early path never transferred. With transfer working, all three forms just work.

### Follow-ups

- `HeadlessCommand` ignores `--message` (generic command runner). The default mixin emits a `logger.warning` naming the class. Consider rejecting at CLI time for agent types that don't override `stage_initial_message` if this bites.

## Test plan

- [x] `uv run pytest libs/mngr/imbue/mngr/cli/create_test.py libs/mngr/imbue/mngr/cli/ask_test.py libs/mngr/imbue/mngr/cli/headless_runner_test.py libs/mngr/imbue/mngr/agents/base_headless_agent_test.py libs/mngr_claude/imbue/mngr_claude/headless_claude_agent_test.py -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'` — all green.
- [x] tmux smoke tests: `test_create_headless_streams_output`, `test_create_headless_with_message_does_not_raise`, `test_create_headless_with_source_and_transfer_none_runs_in_place`.
- [x] `libs/mngr` non-slow full: 3605 passed. `libs/mngr_claude` non-slow full: 405 passed.
- [x] `test_no_type_errors` ratchet passes (caught a missed `CommandAgent.assemble_command` signature update; fixed).
- [ ] CI: full offload suite + acceptance.


joshalbrecht:
lgtm

PR #1343, opened 2026-04-20
Move supertokens API calls to cloudflare_forwarding backend

joshalbrecht:
## Summary
- The minds desktop client no longer reads `SUPERTOKENS_API_KEY`, `GOOGLE_CLIENT_SECRET`, or `GITHUB_CLIENT_SECRET`. Those secrets now live only on the `cloudflare_forwarding` Modal deployment.
- Added `/auth/*` routes on `cloudflare_forwarding` that wrap every SuperTokens SDK operation the client used to perform directly (sign-in, sign-up, OAuth authorize/callback, email verification, password reset, session refresh, user lookup).
- Introduced `AuthBackendClient` in minds; `supertokens_routes.py` and `session_store.py` now go through it instead of calling the SDK.
- Added supertokens init on `cloudflare_forwarding` with the emailpassword, thirdparty, and emailverification recipes (previously only the session recipe was initialized there).
- Email-verification and password-reset links now point at `cloudflare_forwarding` (via `AUTH_WEBSITE_DOMAIN`) and render their own HTML pages, so they keep working even when the desktop client is not running.

## Test plan
- [x] `uv run pytest apps/minds apps/cloudflare_forwarding -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'` (713 passed)
- [x] `uv run ruff check apps/minds apps/cloudflare_forwarding` (clean)
- [x] `uv run ruff format --check apps/minds apps/cloudflare_forwarding` (clean)
- [ ] Deploy the updated `cloudflare_forwarding` Modal function with `SUPERTOKENS_API_KEY`, `GOOGLE_CLIENT_ID`/`_SECRET`, and `GITHUB_CLIENT_ID`/`_SECRET` set on the server.
- [ ] Exercise sign-up / sign-in / OAuth / verify-email / reset-password end-to-end against the redeployed backend.


joshalbrecht:
Merged `origin/main` and `origin/mngr/multi-window-workspaces`. Multi-window is a pure Electron-layer refactor (1.5k-line main.js + 17 lines to preload.js + 193 lines to templates.py) plus one backend-side tweak: `_handle_requests_panel` now passes both `event_id` and `agent_id` to its cards, with both values HTML-escaped and JSON-encoded for safe embedding in the `onclick` handler. The spec at `specs/multi-window-workspaces/concise.md` is the reference. No interaction with the auth refactor.

Additional hardening driven by the architecture review of the auth changes:

- **Signout revocation** (`91ae48621`) -- new `/auth/session/revoke` endpoint on `cloudflare_forwarding` wraps `revoke_all_sessions_for_user`. The minds client's `_handle_signout_api` calls it before deleting the local session record, so sign-out actually ends the SuperTokens session instead of just forgetting the tokens locally.
- **Refresh-token rotation race** (`91ae48621`) -- `MultiAccountSessionStore._try_refresh` now serializes per-user refreshes behind a `threading.Lock`. Second caller re-reads the stored session after acquiring the lock and returns the newly-rotated token without hitting the backend again.
- **OAuth state CSRF protection** (`91ae48621`) -- `_handle_oauth_redirect` extracts the `state` param that the SuperTokens SDK embeds in the authorize URL and stashes it in a module-level dict with a 10-minute TTL; `_handle_oauth_callback` rejects any callback whose echoed-back `state` doesn't match. Defends against a local process forging the callback with a stolen code for the attacker's own account.
- **`push_modal_secrets` clean baseline** (`91ae48621`) -- source env files inside `env -i bash` so the deployer's shell env can neither shadow keys (silent omission) nor leak into `${...}` interpolations.
- **Meta ratchet `prevent_bash_without_strict_mode`** bumped from 0 to 2 for the committed `.minds/template/*.sh` env-declaration files. They are sourced by `push_modal_secrets.py`, not executed; adding `set -euo pipefail` would leak strict mode into the caller's shell.

**Local test results:** `just test-offload` → **7463 passed, 0 failed** (290s). CI's test-offload is running against the same offload config; test-docker and test-offload-acceptance require MODAL_TOKEN_ID which my local shell doesn't have, so they're only exercised in CI.

Review findings that I did **not** act on, with rationale:

- *"Don't issue full session on unverified sign-in"* -- the current flow matches SuperTokens defaults and the `_authenticate_supertokens` admin guard rejects unverified sessions, so unverified users can't reach any admin-requiring endpoint. No exploitable gap.
- *"Convert `_authenticate_supertokens` monkeypatch to FastAPI Depends"* -- larger refactor than scope allows; ratchet already bumped (2 → 4) and documented in the commit that introduced it.
- *"Make `MNGR_DEPLOY_ENV` required at `modal.is_local()`"* -- can't without breaking module-level imports during pytest collection. Mitigation is the documented `scripts/deploy_cloudflare_forwarding.sh` wrapper.

PR #1344, opened 2026-04-20
don't show annotation checks as failed when offload successfully recovers from a flaky test

evgunter:
## Summary
- When offload retries flaky tests and they eventually pass (exit code 2), the job succeeds, but `mikepenz/action-junit-report` still posts the `Unit + Integration Tests` / `Acceptance Tests` check as **failure** (red) because `junit.xml` records the failed first attempt.
- Example: [PR #1337, check run 72186281622](https://github.com/imbue-ai/mngr/pull/1337/checks?check_run_id=72186281622) showed red for a `ConnectionResetError` on `raw.githubusercontent.com` that passed on retry.
- This change detects the flaky-recovered case (test step succeeded but `junit.xml` has `<failure>`/`<error>` elements) and posts a **neutral** (yellow) check instead of red, while still attaching annotations so the underlying failure is visible.
- Neutral checks count as success for branch protection, so this does not weaken merge gating.

## How it works
For each of the two `action-junit-report` sites (`test-offload` and `test-offload-acceptance`):
1. Give the test run step `id: tests` so its `outcome` is visible downstream.
2. Add a `Detect flaky-recovered tests` step that sets `is_flaky_recovered=true` when `steps.tests.outcome == 'success'` **and** `junit.xml` contains `<failure>` or `<error>` tags.
3. If not flaky-recovered: run `action-junit-report` as before (posts green or red check).
4. If flaky-recovered: run `action-junit-report` with `annotate_only: true` (annotations only, no check run) and then post a neutral check run with the original check name via `gh api`.

## Test plan
- [ ] Merge triggers CI. Verify: on a PR with no flakes, `Unit + Integration Tests` and `Acceptance Tests` remain green.
- [ ] Verify manually (e.g. by re-running CI on a PR with a known flaky test, such as #1337) that the check goes yellow (neutral) instead of red when a flaky test is retried to success.
- [ ] Verify that a genuine, unrecovered failure still produces a red check.

### DIFF
--- a/.github/actions/report-flaky-aware-tests/action.yml
+++ b/.github/actions/report-flaky-aware-tests/action.yml
@@ -29,10 +29,13 @@ inputs:
 runs:
   using: composite
   steps:
-    # Detect flaky-recovered case: the test step succeeded (offload exit 0 or
-    # 2) but junit still records failing attempts. Exit 2 means offload
-    # retried flaky tests and they ultimately passed; we want a neutral
-    # (yellow) check rather than a red one in that case.
+    # Detect flaky-recovered case. The just recipes map offload's exit 2
+    # (flaky tests failed then passed on retry) to exit 0, so GitHub Actions
+    # records the test step as "success" for both offload exit 0 (no
+    # failures at all) and offload exit 2 (flaky recovery). We distinguish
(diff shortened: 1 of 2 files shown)


joshalbrecht:
merging this as part of my active branch

PR #1346, opened 2026-04-20
Use RLock in AgentEventQueues to survive GC-triggered re-entrance

gnguralnick:
## Summary

Fixes a self-deadlock in `AgentEventQueues` that wedges the minds workspace server when many SSE streams churn — the symptom seen in a mindtest-base container that had a subagent (`knowing-aloof-eel`). Once wedged, every request for that agent (layout, events, send message) times out at the desktop-client proxy and the mind becomes unusable until the workspace server is restarted.

## Root cause

`register()` holds `self._lock` and, while holding it, replays buffered events onto a fresh queue via `put_nowait`. `put_nowait` allocates, and CPython can fire a GC cycle at any allocation point.

If GC finalizes an abandoned SSE `event_generator` — from a **different, prior** stream whose client disconnected without a clean close — the generator's `finally` block calls `event_queues.unregister()` synchronously on the same thread. Note that the class never calls its own public API directly; the runtime effectively inserts the `unregister()` call mid-`register()` via the GC finalizer. With a non-reentrant `threading.Lock`, that indirect re-entrance self-deadlocks: the thread blocks waiting for a lock it already holds, and nothing else can unblock it.

From there it cascades: every other AnyIO worker that calls `register`/`broadcast`/`unregister` queues behind the dead lock. Sync SSE handlers (`_stream_events`, `_stream_subagent_events`) consume one AnyIO worker per active stream, so the pool fills, HTTP requests time out at the desktop-client proxy, and the container has to be restarted.

Diagnosis came from `py-spy dump` against the stuck process: ~40 worker threads parked at `register (event_queues.py:28)` waiting on the lock, and one thread with an interleaved `register` → `put_nowait` → `event_generator.finally` → `unregister` stack showing the finalizer path holding the lock everyone else wanted.

## Fix

- `event_queues.py`: switch `self._lock` to `threading.RLock` so the same-thread finalizer re-entrance no longer self-deadlocks. One-line change, plus a comment explaining the mechanism and a TODO pointing at the real architectural fixes.
- `event_queues_test.py`: add a regression test that reproduces the deadlock deterministically. Installs a `list` subclass into `_event_buffers` whose `__iter__` calls `unregister()` (standing in for the GC-triggered finalizer), then calls `register()` on a worker thread with a 2s deadline. Fails on the original `threading.Lock`, passes on `threading.RLock`.

## Scope notes

RLock is the surgical fix. It is explicitly not a design-level improvement — it papers over a latent smell (allocations under a lock) rather than eliminating it. The TODO in `event_queues.py` records the two follow-up options, neither done here:

- **(a) Hoist `put_nowait` out of the critical section.** Needs two-phase registration: snapshot buffer under the lock, release, push events to the still-private queue, then re-acquire the lock to register the queue and catch any events that arrived between snapshot and registration. Ordering-preserving across `BufferBehavior.FLUSH` requires identity tracking. Meaningful surface area; deserves its own PR.
- **(b) Convert the SSE handlers to `async def` with `asyncio.Queue`.** Solves both this deadlock *and* the deeper "sync SSE endpoint consumes one AnyIO worker per stream" problem that caps stream concurrency at the thread pool size. Watcher thread feeds the asyncio queue via `loop.call_soon_threadsafe`. Larger change, deserves its own PR.

Downside of RLock worth flagging: it hides nested-acquire bugs that a non-reentrant Lock would catch at test time. We're not enabling intentional re-entrance in our own code — only the GC-finalizer path — but future contributors will have one less safety net if they accidentally nest acquires. The expanded comment warns about this.

## Test plan

- [x] `(cd apps/minds_workspace_server && uv run pytest --no-cov --cov-fail-under=0 -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release')` — 172 passed
- [x] New regression test fails on `threading.Lock` (reproduced the deadlock via the 2s timeout assertion)
- [x] New regression test passes on `threading.RLock`
- [x] Manually restarted the wedged workspace server in the reported container and confirmed internal curl returns 200 in ~25ms (vs. 15+s timeout before)
- [ ] CI offload across the full suite

### DIFF
--- a/apps/minds/docs/desktop-app.md
+++ b/apps/minds/docs/desktop-app.md
@@ -37,15 +37,28 @@ When accessing an agent URL in a regular browser (not the Electron app), the Pyt
 
 ### Shutdown
 
-When the user closes the window, Electron sends SIGTERM to the backend process and waits up to 5 seconds. If the process doesn't exit, SIGKILL is sent.
+Closing an individual window just tears down that window's views -- the backend keeps running while any window is open. When the last window closes (or the user issues `Cmd+Q` / `Ctrl+Q`), Electron sends SIGTERM to the backend process and waits up to 5 seconds. If the process doesn't exit, SIGKILL is sent.
 
 ### Crash recovery
 
-If the backend exits unexpectedly, Electron shows an error screen with the last lines from the log file and a "Retry" button that restarts the backend.
+If the backend exits unexpectedly, every open window switches to the error screen (chrome view expanded to fill the window, content/sidebar/requests-panel views torn down) with the last lines from the log file. Clicking "Retry" from any window restarts the backend once; on success every window reloads to its pre-error URL.
 
 ### Keyboard shortcuts
 
 - **Open DevTools**: `Ctrl+Shift+C` (Windows/Linux) or `Cmd+Option+I` (macOS)
+- **New Window**: `Ctrl+N` / `Cmd+N` -- opens a fresh window on the home page. Also available on macOS via `File > New Window` and the dock icon's right-click menu.
+- **Close Window**: `Ctrl+W` / `Cmd+W` -- closes the focused window; the backend keeps running until the last window closes.
+- **Quit**: `Ctrl+Q` / `Cmd+Q` -- closes every window and shuts the backend down.
+
+### Multi-window behavior
+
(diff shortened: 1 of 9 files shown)


joshalbrecht:
superseded by https://github.com/imbue-ai/mngr/pull/1362

PR #1347, opened 2026-04-20
fix kanpan regression + require uniform dict inputs instead of a mix of dict and typed

evgunter:
## Summary

Fixes `AttributeError: 'dict' object has no attribute 'enabled'` in `_is_source_enabled` when a user configures any data source under `[plugins.kanpan.*]` in TOML. Root cause: the plugin config's declared field types (`dict[str, DataSourceConfig]`, `dict[str, ShellCommandSourceConfig]`, `dict[str, Any]`) did not match the runtime shape produced by the TOML loader, which uses `model_construct` and stores raw dicts. Only tests ever exercised the declared shape.

## Approach

Align types with reality rather than fight the loader.

- `c609c6ec2` — Restore the defensive raw-dict branch in `_is_source_enabled` that a prior commit removed on a false premise.
- `a8c01ed84` — Retype `data_sources` and `shell_commands` on `KanpanPluginConfig` as `dict[str, dict[str, Any]]`. Delete `ShellCommandSourceConfig` (duplicate of `ShellCommandConfig` in `data_sources/shell.py`; only tests used it). Collapse the dual-isinstance branches in `plugin.py` to a single dict path. Rewrite tests to pass raw dicts instead of model instances; add a `model_construct` regression test.
- `eff99ad70` — Same treatment for `columns`: retype `dict[str, dict[str, Any]]`, drop the `isinstance(col_config, dict)` defensive check.

`DataSourceConfig` is kept as the base class for `GitHubDataSourceConfig` (source-specific fields like `pr`, `ci`, `conflicts`, `unresolved`).

## Why not fix the loader?

`DataSourceConfig` has `extra=\"forbid\"` but `GitHubDataSourceConfig` adds source-specific fields. If `_parse_plugins` used `model_validate`, any user TOML `[plugins.kanpan.data_sources.github]` block containing `pr = true` would be rejected. Source-specific configs are parsed by each data source at the point of use; this PR makes the type annotations honest about that.

## Test plan

- [x] `just test-offload` — 7430 passed, 0 failed.
- [x] Architecture verified (independent agent).
- [x] Autofix clean.
- [ ] CI green.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/libs/mngr_kanpan/imbue/mngr_kanpan/data_types.py
+++ b/libs/mngr_kanpan/imbue/mngr_kanpan/data_types.py
@@ -123,7 +123,7 @@ class KanpanPluginConfig(PluginConfig):
         "Each entry should have 'name', 'header', and 'command' (all str).",
     )
 
-    columns: dict[str, Any] = Field(
+    columns: dict[str, dict[str, Any]] = Field(
         default_factory=dict,
         description="Label-backed columns keyed by field key. "
         "Each entry should have 'header' (str) and optionally 'colors' (dict[str, str]).",
--- a/libs/mngr_kanpan/imbue/mngr_kanpan/data_types.py
+++ b/libs/mngr_kanpan/imbue/mngr_kanpan/data_types.py
@@ -55,21 +55,15 @@ class BoardSnapshot(FrozenModel):
 
 
 class DataSourceConfig(FrozenModel):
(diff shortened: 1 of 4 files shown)


joshalbrecht:
lgtm

PR #1348, opened 2026-04-20
Fix workspace server mngr observe subprocess never running

gnguralnick:
## Summary

- The minds_workspace_server's `AgentManager._start_observe()` spawned the discovery-only observe subprocess as `python -m imbue.mngr`, which crashes on startup because `imbue.mngr` is a package with no `__main__`. The failure was silent, so `AgentManager._agents` only ever reflected what `_initial_discover()` saw once at startup.
- Symptom: when an agent runs `mngr create ... --label workspace=$MINDS_WORKSPACE_NAME` (e.g. via the launch-task skill in forever-claude-template), the new sub-agent never appeared in the "+" tab menu — the AgentManager never received the `AGENT_DISCOVERED` event, so the unified WebSocket never broadcast it to the frontend.
- Fix: invoke the `mngr` CLI directly (consistent with the desktop client's `MngrStreamManager` and the AgentManager's own create commands), pin `cwd=$HOME` so the monorepo's `.mngr/settings.toml` doesn't abort the subprocess under pytest, surface stderr lines as warnings, and attach a watchdog thread that logs an error when the subprocess exits before shutdown so a future silent-crash regression surfaces loudly.

## Test plan

- [x] `(cd apps/minds_workspace_server && uv run pytest --no-cov --cov-fail-under=0)` — all 177 tests pass locally
- [x] New regression test `test_start_observe_spawns_long_lived_subprocess` fails against the old `python -m imbue.mngr` command and passes against the fix
- [x] New `test_start_observe_logs_error_when_subprocess_exits_unexpectedly` verifies the watchdog fires on an immediate-exit binary
- [x] New `test_build_observe_command_starts_with_mngr_binary` pins the command shape so reintroducing `python -m imbue.mngr` would fail CI
- [x] `test_prevent_time_sleep` ratchet still satisfied (tests use `poll_until` from `imbue.mngr.utils.polling`)
- [ ] Manual: restart a running mindtest workspace server, run `mngr create sleep-test -t worker --label workspace=$MINDS_WORKSPACE_NAME ...`, confirm `sleep-test` appears in the "+" tab menu within a couple seconds

### DIFF
--- a/apps/minds/docs/desktop-app.md
+++ b/apps/minds/docs/desktop-app.md
@@ -37,15 +37,28 @@ When accessing an agent URL in a regular browser (not the Electron app), the Pyt
 
 ### Shutdown
 
-When the user closes the window, Electron sends SIGTERM to the backend process and waits up to 5 seconds. If the process doesn't exit, SIGKILL is sent.
+Closing an individual window just tears down that window's views -- the backend keeps running while any window is open. When the last window closes (or the user issues `Cmd+Q` / `Ctrl+Q`), Electron sends SIGTERM to the backend process and waits up to 5 seconds. If the process doesn't exit, SIGKILL is sent.
 
 ### Crash recovery
 
-If the backend exits unexpectedly, Electron shows an error screen with the last lines from the log file and a "Retry" button that restarts the backend.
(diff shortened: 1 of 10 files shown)


gnguralnick:
Recreating off origin/main to drop unrelated commits from local main.

PR #1351, opened 2026-04-20
move schedule verify into the cron runner

evgunter:
## Summary

`mngr schedule --verify` fails for the common `schedule=modal + agent=local-in-container` case: the deploy machine cannot resolve an agent that lives inside the ephemeral Modal sandbox, so it cannot check state or clean up. This PR moves verify work to whichever side actually owns the agent — the cron runner itself.

### Core refactor
- `run_scheduled_trigger(verify_mode="none")` — new param. Scheduled invocations default to `"none"` so cron runs stay fast. Deploy-side verifier passes `"quick"` or `"full"` via `modal run ... --verify-mode <mode>`.
- `"quick"` destroys the created agent via `mngr destroy --force`. `"full"` polls `mngr list --include 'name == "X"' --format json` (provider-agnostic) until the agent leaves the running states. Runner emits a single sentinel line (`__MNGR_SCHEDULE_VERIFY__ {...}`) with a structured result dict, also returned from the function.
- Deploy-side `verification.py` is gutted: streams `modal run` output, parses the sentinel, translates the result dict into success or `ScheduleDeployError`. `_resolve_agent` / `_wait_for_agent_to_finish` / deploy-side `_destroy_agent` are gone.

### Robustness
- Sentinel detection uses a regex that tolerates Modal's container-id line prefixes.
- Full-verify timeout inside the runner emits a sentinel with `status="timeout"` rather than propagating a raw exception, so the agent is best-effort destroyed instead of orphaned.
- Destroy subprocess timeout handled the same way — sentinel contract preserved.
- Client-side `modal run` timeout is Modal's function timeout plus 300s headroom for cold-start and flush.
- Inner poll timeout (`_AGENT_FINISH_TIMEOUT_SECONDS=3000s`) is tighter than Modal's 3600s function budget so `mngr create` + destroy + sentinel flush all fit inside.
- Sentinel-line check runs before the error-keyword scan so a `destroy_stderr` payload containing "Exception" or "Traceback" doesn't spuriously fail the verify.
- Post-wait: sentinel wins over error-keyword heuristic whenever present; keyword scan is now just fallback for "container crashed before emitting anything".
- Disabled triggers return `{"output": "", ...}` so the strict-shape check in `invoke_modal_trigger_function` doesn't misfire.
- `_get_lifecycle_state` walks all providers (no `--provider local` filter) so full-verify works regardless of where the trigger's `mngr create` placed the agent. Matches the quick-verify path's `mngr destroy --force <name>`, which has always been provider-agnostic.

### Cross-module contracts
- `RUNNING_STATES`, `VALID_VERIFY_MODES`, `AGENT_MISSING_STATE`, `RESULT_SENTINEL` are inlined in both `cron_runner.py` and `verification.py`. **The shared-module approach didn't work** — Modal runs `cron_runner.py` standalone with no `imbue` namespace on sys.path, so `from imbue.mngr_schedule... import X` raises `ModuleNotFoundError` at deploy time and every cron invocation fails. The file-level comment documents this so it isn't re-attempted.
- `cron_runner_drift_test.py` imports `cron_runner.py` directly under a fixture that stubs the deploy-time env vars (so the module-level `_require_env(...)` calls don't crash), then compares the inlined values against the authoritative imbue enums.
  - One classification test asserts `cron_runner.RUNNING_STATES ∪ verification._TERMINAL_SUCCESS_STATES == AgentLifecycleState` value set, and the two are disjoint. New enum variants without classification fail the test.
  - Sibling tests cover `VALID_VERIFY_MODES` against `VerifyMode` and the cross-file mirrors of `AGENT_MISSING_STATE` and `RESULT_SENTINEL`.
- `verification.py` builds `_TERMINAL_SUCCESS_STATES` from `AgentLifecycleState.{DONE,STOPPED}.value` (no import ban there) so enum changes propagate automatically.
- `invoke_modal_trigger_function` no longer tolerates bare-string return values — raises `MngrError` on unexpected shape.

### Local error type in cron_runner
- New `class CronRunnerError(Exception)` in `cron_runner.py` itself (no imports needed). Verify-path raises use it instead of `RuntimeError`.

### Ratchets
- `bare_print` whole-file exempts `cron_runner.py` (cannot import loguru; sentinel write + verify-path diagnostics all use bare `print()`/`sys.stdout.write()`). Net snapshot 2 across the rest of the lib.
- `direct_subprocess` 7 with `testing.py` and test files excluded (in-container verify helpers can't import `ConcurrencyGroup`).

## Files changed (vs main)
6 files: `cron_runner.py`, `cron_runner_drift_test.py` (new), `deploy.py`, `verification.py`, `test_ratchets.py`, `test_schedule_add.py`.

## Test plan
- [x] `cd libs/mngr_schedule && uv run pytest -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'` — 289 passed locally
- [x] `test_no_type_errors` passes
- [x] Ratchet tests pass with tightened snapshots (no inflation)
- [x] `cron_runner_drift_test.py` (new) — 5 tests, all green; pins `AgentLifecycleState` partition + `VerifyMode` + cross-file sentinel mirrors against drift
- [x] CI: `test-offload`, `test-docker`, `Acceptance Tests`, `Unit + Integration Tests`, `test-offload-acceptance` all SUCCESS
- [x] **Release test `test_schedule_add_with_verification` (`--verify quick`)** — passed locally. Real Modal deploy + verify quick + destroy.
- [x] **Release test `test_schedule_add_with_full_verification` (`--verify full`, new in this PR)** — passed locally. Real Modal deploy + lifecycle poll + reach DONE + cleanup.
- Latest combined wall time for both release tests: 468s (~7m48s).


libs/mngr_schedule/imbue/mngr_schedule/implementations/modal/cron_runner_drift_test.py
```diff
@@ -0,0 +1,129 @@
...
+env vars) and compare the literal values against the authoritative
+imbue enums. Direct import is the only place this module gets pulled
+in by the rest of the test suite -- everything else either lives in
+the cron container or in deploy.py / verification.py (which import
+neither cron_runner nor the cron-only constants module).
+
+Why not AST-parse the source? AST parsing was the original approach
+but is brittle to formatting changes (e.g. switching `frozenset({...})`
+to a comprehension would break the parser). Importing under stub env
+catches refactors a parser would miss.
```
evgunter, line 20: couldn't find a non-ugly way to do this

libs/mngr_schedule/imbue/mngr_schedule/implementations/modal/cron_runner_drift_test.py
```diff
@@ -0,0 +1,120 @@
...
+    saved_module = sys.modules.pop(_CRON_RUNNER_MODULE, None)
+    try:
+        with pytest.MonkeyPatch.context() as monkeypatch:
+            monkeypatch.setenv("SCHEDULE_DEPLOY_CONFIG", json.dumps(deploy_config))
+            monkeypatch.setenv("SCHEDULE_BUILD_CONTEXT_DIR", str(tmp))
+            monkeypatch.setenv("SCHEDULE_STAGING_DIR", str(tmp))
+            monkeypatch.setenv("SCHEDULE_DOCKERFILE", str(dockerfile))
+            import imbue.mngr_schedule.implementations.modal.cron_runner as cron_runner_module
+
+            yield cron_runner_module
```
evgunter: alas

libs/mngr_schedule/imbue/mngr_schedule/implementations/modal/verification.py
```diff
@@ -1,218 +1,166 @@
...
+                if isinstance(parsed, dict):
+                    sentinel_holder.append(parsed)
+                else:
+                    logger.warning("Verify sentinel payload was not a JSON object: {}", payload)
+                continue
+
+        lower = stripped.lower()
         if "traceback" in lower or "exception" in lower:
             error_lines.append(stripped)
             error_event.set()
```
evgunter, line 137: kind of sketchy

libs/mngr_schedule/imbue/mngr_schedule/implementations/modal/cron_runner.py
```diff
@@ -172,27 +234,167 @@ def _run_and_stream(
...
+    Returns the final state string. Raises CronRunnerError on timeout.
+    """
+    deadline = time.monotonic() + timeout_seconds
+    while True:
+        state = _get_lifecycle_state(agent_name)
+        if state is not None and state not in RUNNING_STATES:
+            return state
+        if time.monotonic() >= deadline:
+            raise CronRunnerError(f"Timed out waiting for agent '{agent_name}' to finish after {timeout_seconds:.0f}s")
+        time.sleep(poll_interval_seconds)
```
evgunter, line 315: sadly `mngr wait` did not seem to work (though i didn't badger claude about making it work very hard)

joshalbrecht:
I'm confused by both the motivation and implementation for this one, let's just chat in person on monday

Specifically I'm confused why:
1. we need verification inside like this
2. we can't import imbue-related code remotely (since it is actually part of the image)

evgunter:
discussed in person; for context:
1. this enables more robust verification, rather than just checking that the exit code is 0, though this PR doesn't add plumbing for that more-robust verification yet
2. in the motivating example mngr's code is available, but for mngr schedule in general it need not be

joshalbrecht:
lgtm (agreed that some of this is gross, but whatever, at least it works)

PR #1352, opened 2026-04-20
fix modal env leaks

evgunter:
## Summary

Two separate orphan-Modal-env leaks fixed here (ordered by volume of leakage observed).

### Leak #1: minds desktop client e2e release tests

- **Observed**: the primary burst driver (~6/min avg, 36/min peak during parallel release shards). `apps/minds/imbue/minds/testing.py:clean_env()` stripped `PYTEST_CURRENT_TEST` from subprocess env to dodge the repo's `.mngr/settings.toml:is_allowed_in_pytest=false` guard. Inside a release sandbox with no persistent profile, the subprocess mngr minted a fresh uuid4 profile every call.
- **Fix**: Replace the covert strip with explicit `MNGR_ALLOW_PYTEST=1` opt-in that the guard checks in `libs/mngr/imbue/mngr/config/loader.py`. `clean_env()` now sets that var instead of evading the marker. The opt-in guard additionally enforces that `MNGR_HOST_DIR` resolves under `tempfile.gettempdir()` so a test can't accidentally point the opt-in at the developer's real `~/.mngr`. `apps/minds/conftest.py` now registers the shared plugin fixtures (`register_plugin_test_fixtures(globals())`) and overrides `mngr_test_prefix` to `mngr_test-<timestamp>-`, covering both `clean_env`-based subprocess calls AND the desktop client's in-process ConcurrencyGroup spawns (which bypass `clean_env` and just inherit `os.environ`). Same PYTEST_CURRENT_TEST→MNGR_ALLOW_PYTEST swap applied to `libs/mngr_schedule/.../test_schedule_add.py`.

### Leak #2: scheduled cron triggers

- **Observed**: orphan `mngr-<32hex>` envs with "deployed by evan-gunter" apps named `mngr-modal` (= default prefix + default modal provider name). Matches the pattern for a scheduled cron fire running `mngr create --provider modal` inside an ephemeral container with no persistent profile and no knowledge of the deployer's identity.
- **Mechanism (root cause)**: When the deployer runs `mngr schedule add` with a non-default `MNGR_ROOT_NAME` (e.g. `minds`, `mngr-changelog-schedule`), the deploy pipeline has **two bugs** that together force the cron container to mint a fresh uuid4 user_id every fire:
  1. `get_files_for_deploy` read `config.toml` from a hardcoded `~/.mngr/` path, silently skipping it when the deployer's real config lived at `~/.<root_name>/config.toml`. The image then lacked `config.toml` entirely → container's `get_or_create_profile_dir` minted a fresh uuid4 profile → the baked `user_id` file (which WAS staged correctly, under a different profile_id) couldn't be found.
  2. Even if the config had been staged, the container defaulted to `MNGR_ROOT_NAME="mngr"` and looked at `~/.mngr/...` for everything — missing files staged under `~/.<deployer_root_name>/...`. Prefix resolved to `"mngr-"` regardless of deployer, so nested `mngr create` landed in a different Modal env than the deployer's actual one.
- **Fix**: Addresses both bugs directly (`libs/mngr_schedule/imbue/mngr_schedule/plugin.py`):
  1. `get_files_for_deploy` now reads `config.toml` from `mngr_ctx.config.default_host_dir`, so non-default root deployers' config actually gets staged at the right path.
  2. `modify_env_vars_for_deploy` hookimpl propagates `MNGR_ROOT_NAME` (via `setdefault`, so an explicit `--pass-env` / `--env-file` still wins). With config staged and root_name propagated, the container's mngr transparently uses the deployer's full profile — right profile_id, right user_id, right prefix, right settings.toml. Nested mngr calls land in the deployer's actual env, not an orphan `mngr-<uuid>`.

### Defense in depth: pytest-aware modal backend guard

`libs/mngr_modal/imbue/mngr_modal/backend.py:_create_environment` now also checks: under pytest, the env name must match the existing `TEST_ENV_PATTERN` regex (`mngr_test-YYYY-MM-DD-HH-MM-SS`). The pre-existing guard only caught `mngr_` (underscore) prefixes; the new check also catches dash-prefixed defaults like `mngr-<uuid>`. So if some future in-process mngr spawn forgets to override `MNGR_PREFIX`, the backend will loudly refuse to create the env rather than silently leak.

## Test plan

- [x] `pytest libs/mngr_schedule/.../plugin_test.py libs/mngr/imbue/mngr/config/loader_test.py libs/mngr_modal/.../testing_provider_test.py -k create_environment` — pass locally
- [ ] Full CI via stop hook
- [ ] Manual: deploy a schedule against a test env, confirm cron fire does not create a new `mngr-<uuid>` env

- **Redeploy existing scheduled triggers after merge**: already-deployed triggers run off pre-fix Modal images and will keep leaking until rebuilt. Known active leakers (as of 2026-04-20): `mngr-schedule-changelog-consolidation` in envs `mngr-ev` and `mngr-changelog-schedule-920da0b3db46415d8c4aec37ca23a637`.
- Third cleanup pass of orphan `mngr-<32hex>` envs after redeploys + first release run on the fixed code (two passes so far deleted 1441 + 701; more may accumulate until all deployed triggers are rebuilt).

### DIFF
--- a/apps/minds/imbue/minds/testing.py
+++ b/apps/minds/imbue/minds/testing.py
@@ -87,14 +87,23 @@ def add_and_commit_git_repo(repo_dir: Path, tmp_path: Path, message: str = "upda
 
 
 def clean_env() -> dict[str, str]:
-    """Build an environment dict for subprocesses that strips pytest markers.
+    """Build an environment dict for subprocesses that opts into real-mngr mode.
 
-    mngr refuses to run when PYTEST_CURRENT_TEST is set (safety check to
-    prevent tests from accidentally using real mngr state). We strip it
-    so that our end-to-end subprocess calls work against the real system.
+    mngr refuses to run in pytest when the project config's
+    is_allowed_in_pytest is False (safety check to prevent tests from
+    accidentally using the developer's real ~/.mngr state). This helper
+    sets MNGR_ALLOW_PYTEST=1, an explicit opt-in that says "yes, this
+    subprocess is an intentional end-to-end call".
+
+    The opt-in is only safe when the surrounding fixture also sets
+    MNGR_HOST_DIR and MNGR_PREFIX to test-scoped values (see e.g. the
+    autouse _isolate_mngr_state fixture in test_desktop_client_e2e.py).
+    Without that isolation, the subprocess mngr mints a fresh uuid4
+    profile and the Modal backend creates an orphan mngr-<uuid>
+    environment on every call.
     """
     env = dict(os.environ)
-    env.pop("PYTEST_CURRENT_TEST", None)
+    env["MNGR_ALLOW_PYTEST"] = "1"
     return env
 
 
--- a/apps/minds/imbue/minds/testing.py
+++ b/apps/minds/imbue/minds/testing.py
@@ -87,24 +87,17 @@ def add_and_commit_git_repo(repo_dir: Path, tmp_path: Path, message: str = "upda
 
 
 def clean_env() -> dict[str, str]:
-    """Build an environment dict for subprocesses that opts into real-mngr mode.
+    """Build an environment dict for subprocesses.
 
-    mngr refuses to run in pytest when the project config's
-    is_allowed_in_pytest is False (safety check to prevent tests from
-    accidentally using the developer's real ~/.mngr state). This helper
-    sets MNGR_ALLOW_PYTEST=1, an explicit opt-in that says "yes, this
-    subprocess is an intentional end-to-end call".
-
-    The opt-in is only safe when the surrounding fixture also sets
-    MNGR_HOST_DIR and MNGR_PREFIX to test-scoped values (see e.g. the
-    autouse _isolate_mngr_state fixture in test_desktop_client_e2e.py).
-    Without that isolation, the subprocess mngr mints a fresh uuid4
-    profile and the Modal backend creates an orphan mngr-<uuid>
-    environment on every call.
+    Returns a copy of os.environ. The shared plugin test fixtures
+    (registered in apps/minds/conftest.py via register_plugin_test_fixtures)
+    already set MNGR_HOST_DIR / MNGR_PREFIX / MNGR_ROOT_NAME to per-test
+    tmp values, so the subprocess inherits proper isolation. With
+    MNGR_ROOT_NAME set to `mngr-test-<id>`, the subprocess does not load
+    the repo's .mngr/settings.toml, so the is_allowed_in_pytest=false
+    guard there does not fire and no explicit opt-in is needed.
     """
-    env = dict(os.environ)
-    env["MNGR_ALLOW_PYTEST"] = "1"
-    return env
+    return dict(os.environ)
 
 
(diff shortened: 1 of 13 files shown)


libs/mngr_modal/imbue/mngr_modal/backend.py
```diff
@@ -69,6 +71,23 @@ def _create_environment(environment_name: str, modal_interface: ModalInterface)
...
+    # guard only catches `mngr_` underscore -- this one also catches dash-
+    # prefixed default names like `mngr-<uuid>`.
+    if "PYTEST_CURRENT_TEST" in os.environ and not TEST_ENV_PATTERN.match(environment_name):
+        raise MngrError(
+            f"Refusing to create Modal environment {environment_name!r} during pytest: "
+            "test Modal envs must match the mngr_test-YYYY-MM-DD-HH-MM-SS pattern so the "
+            "CI cleanup script can find them. Set MNGR_PREFIX via "
+            "generate_test_environment_name()."
+        )
+
```
evgunter, line 90: these are sort of redundant but the first one is active during deployment and the latter is only active during testing

libs/mngr_modal/imbue/mngr_modal/backend.py
```diff
@@ -69,6 +71,23 @@ def _create_environment(environment_name: str, modal_interface: ModalInterface)
...
+    # default-prefixed env that no CI cleanup script recognizes. The earlier
+    # guard only catches `mngr_` underscore -- this one also catches dash-
+    # prefixed default names like `mngr-<uuid>`.
+    if "PYTEST_CURRENT_TEST" in os.environ and not TEST_ENV_PATTERN.match(environment_name):
+        raise MngrError(
+            f"Refusing to create Modal environment {environment_name!r} during pytest: "
+            "test Modal envs must match the mngr_test-YYYY-MM-DD-HH-MM-SS pattern so the "
+            "CI cleanup script can find them. Set MNGR_PREFIX via "
+            "generate_test_environment_name()."
+        )
```
evgunter, line 89: test infrastructure in production code, but it seems justified since the previous guard didn't catch the leaks in minds

apps/minds/imbue/minds/testing.py
```diff
@@ -87,15 +87,17 @@ def add_and_commit_git_repo(repo_dir: Path, tmp_path: Path, message: str = "upda
...
+    already set MNGR_HOST_DIR / MNGR_PREFIX / MNGR_ROOT_NAME to per-test
+    tmp values, so the subprocess inherits proper isolation. With
+    MNGR_ROOT_NAME set to `mngr-test-<id>`, the subprocess does not load
+    the repo's .mngr/settings.toml, so the is_allowed_in_pytest=false
+    guard there does not fire and no explicit opt-in is needed.
     """
-    env = dict(os.environ)
-    env.pop("PYTEST_CURRENT_TEST", None)
-    return env
+    return dict(os.environ)
```
joshalbrecht, line 100: do you think it might be worth asserting some of those things in this function?

joshalbrecht:
lgtm

PR #1353, opened 2026-04-20
Detach HEAD on target before mirror push with --prune

weishi-imbue:
## Summary

Fixes `mngr create --reuse --update` on agents whose current branch doesn't exist on source. Without the fix, the internal `git push --prune` refuses to delete the target's checked-out branch, and agent creation fails.

Minds' agent creation unconditionally passes `--reuse --update` ([apps/minds/.../agent_creator.py:274-275](https://github.com/imbue-ai/mngr/blob/main/apps/minds/imbue/minds/desktop_client/agent_creator.py#L274-L275)), so any second creation with the same agent name hits this failure without the fix.

## Change

In \`_git_push_to_target\` (host.py), before the mirror push:

\`\`\`
git -C <target> checkout --detach HEAD 2>/dev/null
git -C <target> config receive.denyDeleteCurrent ignore
\`\`\`

Detaching HEAD means no branch is "current," so the prune can delete any branch. `receive.denyDeleteCurrent=ignore` is belt-and-suspenders. After the push, the existing `git checkout -B` step re-attaches HEAD to the requested branch.

This ports the same pattern already used in `sync.py::_remote_git_push_mirror` (which backs `mngr push --mirror`). Future cleanup candidate: extract a shared helper.

## Test plan

- [x] Added regression test `test_create_work_dir_idempotent_when_target_has_checked_out_work_branch` in test_host.py
- [x] Verified test fails WITHOUT the fix with `imbue.mngr.errors.MngrError: ... deletion of the current branch prohibited`
- [x] Verified test passes WITH the fix
- [ ] CI green

## Context

Extracted from #1317 (minds_onboard) so the fix can land independently of the larger wheel-bundling work in that PR.

### DIFF
--- a/libs/mngr/imbue/mngr/hosts/host.py
+++ b/libs/mngr/imbue/mngr/hosts/host.py
@@ -1680,6 +1680,20 @@ class Host(BaseHost, OnlineHostInterface):
 
         with log_span("Pushing git repo to target: {}", git_url):
             if source_host.is_local:
+                # When re-pushing into an existing target (e.g. `mngr create
+                # --reuse --update`), the mirror push's `--prune` may try to
+                # delete the target's currently-checked-out branch. Git
+                # refuses that by default ("refusing to delete the current
+                # branch"). Detach HEAD on the target and set
+                # receive.denyDeleteCurrent=ignore so prune can replace all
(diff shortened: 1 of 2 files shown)


weishi-imbue:
open design question on whether `--update` still makes sense

PR #1356, opened 2026-04-21
add assertions in clean_env

evgunter:
## Summary

Follow-up to #1352 addressing Josh's review comment on `apps/minds/imbue/minds/testing.py:clean_env()`:

> do you think it might be worth asserting some of those things in this function?

Yes. `clean_env()` now asserts the shared-autouse invariants it relies on, so a future test that calls `clean_env()` without the shared plugin fixtures registered (e.g. someone moves the test to a new directory or adds a caller from a test file that doesn't use `register_plugin_test_fixtures`) gets a loud `AssertionError` at test-setup time instead of silently spawning a subprocess mngr that mutates real state.

Asserts:
- `MNGR_HOST_DIR` is set (so subprocess does not land in developer's `~/.mngr`)
- `MNGR_PREFIX` starts with `mngr_test-` (so any leaked Modal env is visible to the CI cleanup script)
- `MNGR_ROOT_NAME` starts with `mngr-test-` (so subprocess does not load the repo's `.mngr/settings.toml` and trip `is_allowed_in_pytest=false`)

## Test plan

- [x] Syntax check (collection runs)
- [ ] Full CI via stop hook

### DIFF
--- a/apps/minds/imbue/minds/testing.py
+++ b/apps/minds/imbue/minds/testing.py
@@ -89,14 +89,36 @@ def add_and_commit_git_repo(repo_dir: Path, tmp_path: Path, message: str = "upda
 def clean_env() -> dict[str, str]:
     """Build an environment dict for subprocesses.
 
-    Returns a copy of os.environ. The shared plugin test fixtures
+    Returns a copy of os.environ. Relies on the shared plugin test fixtures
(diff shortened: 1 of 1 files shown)


joshalbrecht:
lgtm, thanks!

PR #1358, opened 2026-04-21
stop macOS notification tests from firing real Notification Center banners

gnguralnick:
## Summary

Three macOS notification tests in `apps/minds/imbue/minds/desktop_client/notification_test.py` called `_dispatch_macos_notification` directly. The function spawns a daemon thread that shells out to real `osascript`, so every run fired a visible Notification Center banner -- leaving hundreds of stale banners on a developer Mac. The tests also had weak assertions (just "does not raise"), with no real coverage of quote escaping, routing, or error handling.

### Production changes

Two pure helpers extracted from existing logic -- no new injection seams, no new `PrivateAttr`s, no test-only kwargs:

- **`_build_osascript_notification(request, agent_display_name) -> str`**: the AppleScript-string construction lifted out of `_dispatch_macos_notification`, which now delegates to it.
- **`DispatchChannel` enum + `_select_dispatch_channel(is_electron, is_macos) -> DispatchChannel`**: the routing decision lifted out of `dispatch()`. `dispatch()` now computes the channel via the helper and `match`es on the variants.

### Test changes

- **Quote escaping**: `test_build_osascript_notification_escapes_double_quotes` calls the pure builder and asserts `\"` appears for the quoted payload in both title and message. The old `..._handles_quotes` test had no such assertion.
- **Routing**: `test_select_dispatch_channel` is parametrized across all four `(is_electron, is_macos)` combinations against the pure selector. No `dispatch()` call, no subprocess, no threads. The old routing test only asserted `dispatcher.is_macos is True`, which told you nothing about actual routing.
- **Tkinter-branch test fix**: `test_dispatch_non_electron_does_not_raise` now passes `is_macos=False`. On macOS, the default `is_macos=True` was silently routing this test through the real osascript path too.
- **Dropped**: `test_dispatch_macos_notification_does_not_raise`. The behavior it claimed to cover -- `_run_macos_notification_subprocess` catching `OSError`/`ExceptionGroup` -- is tested by the real code path at runtime; a dedicated test would require an injection seam that the repo rules (`~/.claude/CLAUDE.md`: no test-only hooks in production code) rightly forbid.

### Production behavior

Unchanged for all callers. `_dispatch_macos_notification` signature is unchanged. `dispatch()` routes to the same three branches in the same priority order.

## Test plan

- [x] `(cd apps/minds && uv run pytest --no-cov --cov-fail-under=0 -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release')` -- 600 passed locally (host is macOS)
- [x] No Notification Center banner can fire by construction: the quote test hits only the pure builder, the routing test hits only the pure selector, and the one remaining `dispatch()`-calling test on a non-Electron path forces `is_macos=False`
- [ ] Full offload via CI stop hook

### DIFF
--- a/apps/minds/imbue/minds/desktop_client/notification.py
+++ b/apps/minds/imbue/minds/desktop_client/notification.py
@@ -242,21 +242,30 @@ def _run_macos_notification_subprocess(
         logger.warning("Failed to show macOS notification: {}", e)
 
 
-def _dispatch_macos_notification(
+def _build_osascript_notification(
     request: NotificationRequest,
     agent_display_name: str,
-    runner: MacOSNotificationRunner = _run_macos_notification_subprocess,
-) -> threading.Thread:
-    """Display a native macOS notification via osascript on a background thread.
+) -> str:
+    """Build the AppleScript command that displays a native macOS notification.
 
-    Returns the spawned daemon thread so callers (notably tests) can join it.
+    Extracted so the quote-escaping behavior is trivially testable as a pure
+    function, independent of subprocess or threading concerns.
     """
     display_title = request.title or f"Notification from {agent_display_name}"
     # Escape double quotes for AppleScript string literals
     escaped_title = display_title.replace('"', '\\"')
     escaped_message = request.message.replace('"', '\\"')
     escaped_subtitle = f"From: {agent_display_name}".replace('"', '\\"')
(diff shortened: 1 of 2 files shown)


joshalbrecht:
lgtm

PR #1359, opened 2026-04-21
mngr start: add --restart and --no-resume flags

gnguralnick:
## Summary

- Add `--restart` flag to `mngr start` that stops a running agent before starting it fresh. If the agent is already stopped, it is simply started.
- Add `--no-resume` flag to `mngr start` that skips sending the resume message after starting. Can be combined with `--restart` or used independently.
- `--restart` sends the resume message by default (same as a normal start); pass `--no-resume` to suppress it.

## Test plan

- [x] Unit tests for `_output_result` with restart verb, `StartCliOptions` fields
- [x] Integration tests for `--restart` flow: restart running agent, restart stopped agent
- [x] Full mngr test suite passing locally (3924 passed)
- [x] Ratchets + type check: all pass
- [ ] CI acceptance tests (failing due to Modal infrastructure issue, not code-related)


libs/mngr/imbue/mngr/cli/start.py
```diff
@@ -169,93 +212,172 @@ def start(ctx: click.Context, **kwargs: Any) -> None:
...
                     _send_resume_message_if_configured(agent, output_opts)
 
-                    # Track for potential connect
                     if opts.connect:
                         last_started_agent = agent
                         last_started_host = online_host
                     break
 
-    # Output final result
     _output_result(started_agents, output_opts)
```
joshalbrecht: weird that it stripped the comments?

libs/mngr/imbue/mngr/cli/start.py
```diff
@@ -169,93 +212,172 @@ def start(ctx: click.Context, **kwargs: Any) -> None:
...
+        provider_name = agent_list[0].provider_name
+
+        provider = get_provider_instance(provider_name, mngr_ctx)
+        host = provider.get_host(HostId(host_id_str))
+
+        online_host, _ = ensure_host_started(host, is_start_desired=True, provider=provider)
+
+        locked_agents = []
+        lock_handles: list[io.TextIOWrapper] = []
+        try:
```
joshalbrecht: a lot of this code seems fairly duplicated with the above?

libs/mngr/imbue/mngr/cli/start.py
```diff
@@ -169,93 +212,172 @@ def start(ctx: click.Context, **kwargs: Any) -> None:
...
 
         for match in agent_list:
             started_agents.append(str(match.agent_name))
             _output(f"Started agent: {match.agent_name}", output_opts)
 
-            # Get the agent object for potential connect and resume message
             for agent in online_host.get_agents():
                 if agent.id == match.agent_id:
-                    # Send resume message if configured
                     _send_resume_message_if_configured(agent, output_opts)
```
joshalbrecht, line 245: I actually don't think we should ever be sending a "resume" message unless:
A) the user explicitly asked for it via the CLI, OR
B) the agent had previously been running when it was stopped

(this isn't your fault, just pointing out because I think the current behavior is just wrong and dumb)

(resume message is *meant* for "if the computer restarted and thus the agent was restarted while it had been running", which is a fairly special case)

libs/mngr/imbue/mngr/cli/start.py
```diff
@@ -169,93 +212,172 @@ def start(ctx: click.Context, **kwargs: Any) -> None:
...
+        host_id_str, _ = host_key.split(":", 1)
+        provider_name = agent_list[0].provider_name
+
+        provider = get_provider_instance(provider_name, mngr_ctx)
+        host = provider.get_host(HostId(host_id_str))
+
+        online_host, _ = ensure_host_started(host, is_start_desired=True, provider=provider)
+
+        locked_agents = []
+        lock_handles: list[io.TextIOWrapper] = []
```
joshalbrecht: something feels a little weird about the locking--IIRC there is already a mechanism for locking hosts (see OnlineHostInterface::lock_cooperatively) which at least *attempts* to mostly get this right even for remote hosts

(yes it is technically wrong but it's like 99% right and I have a TODO to come back and fix the other 1%)

joshalbrecht:
sorry for the annoying comments on this one!  But they probably are worth addressing in this case

happy to pair or chat more on any of them if they dont make sense

joshalbrecht:
lgtm, but would like to get our tests passing before we merge stuff if possible... (I don't like merging things with broken tests if it can be avoided)

PR #1366, opened 2026-04-21
Fix three regressions on cloudflare-forwarded workspace URLs

joshalbrecht:
## Summary

Fixes three separate but cloudflare-only symptoms in `minds_workspace_server` when the UI is reached through the per-service Cloudflare tunnel URLs (`agent--…`, `terminal--…`, `web--…`).

- **Stuck on the "creating" tab for a newly created chat.** The Mithril `AgentManager` only removes a proto-agent from its local list when it receives a `proto_agent_completed` broadcast. Cloudflare's WS idle timeout (and a workspace-server restart mid-creation) both drop the WS. On reconnect, the server re-sends only the currently *active* proto-agents, with no "clear" message for ones that finished during the disconnect, so `isProtoAgent(newId)` stayed `true` forever and `ChatPanel` never left the build-log view. Fix: reset `protoAgents = []` inside `ws.onopen` so the server's initial burst is authoritative.

- **"+" dropdown missing "web" and any second chat.** `_ApplicationsFileHandler` in `agent_manager.py` overrode only `on_modified`, but `scripts/forward_port.py` upserts `runtime/applications.toml` atomically with `tempfile.mkstemp` + `os.replace`, which watchdog surfaces as a `FileMovedEvent` / `FileCreatedEvent` — never `FileModifiedEvent`. So every service registration after the watcher started was silently ignored and `self._applications` was frozen at whatever was on disk when the workspace-server booted (often just `system_interface`, which is filtered out of the dropdown). Fix: override `on_any_event` instead, and add a unit test that dispatches a `FileMovedEvent` through the handler.

- **Terminal iframe blocked by `imbue-test.cloudflareaccess.com`.** `make_hostname` creates a distinct hostname per service, and `cf_create_access_app` creates a separate Access Application per hostname. First load in an iframe has no session cookie for `terminal--…`, so CF 302s to `cloudflareaccess.com`, which sends `X-Frame-Options: DENY`, which Firefox reports as "will not allow Firefox to display the page". The tidy fix is sharing sessions across subdomains at the remote_service_connector level, but that's a deploy-side change; this PR adds an "Open in new window" action to iframe tabs so users can complete the Access handshake in a top-level window, after which subsequent iframe loads carry the cookie and render inline.

## Test plan

- [x] `just test-quick "apps/minds_workspace_server -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'"` — 174 passed
- [x] `just test-quick "apps/minds -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'"` — 605 passed
- [x] Frontend `tsc --noEmit` clean, `eslint src/` clean, `prettier --check` clean on touched files
- [x] New `test_applications_file_handler_fires_on_move` asserts the watcher reacts to a `FileMovedEvent` (would have failed before the handler change)
- [ ] CI offload run on PR

### DIFF
--- a/apps/minds_workspace_server/frontend/src/models/AgentManager.ts
+++ b/apps/minds_workspace_server/frontend/src/models/AgentManager.ts
@@ -66,6 +66,14 @@ function connect(): void {
   ws = new WebSocket(url);
 
   ws.onopen = () => {
+    // Drop any proto-agent entries carried over from a prior connection.
+    // The server re-sends its current active set as part of the initial
+    // burst below, but it emits no "completed" message for protos that
+    // finished while we were disconnected. That can happen on a Cloudflare
+    // WS idle timeout or a workspace-server restart during agent creation.
+    // Without this reset the entry sticks forever, `isProtoAgent` keeps
+    // returning true, and the chat tab is stuck on the build-log view.
+    protoAgents = [];
     connected = true;
     m.redraw();
   };
--- a/apps/minds_workspace_server/frontend/src/models/AgentManager.ts
+++ b/apps/minds_workspace_server/frontend/src/models/AgentManager.ts
@@ -66,14 +66,6 @@ function connect(): void {
   ws = new WebSocket(url);
 
   ws.onopen = () => {
-    // Drop any proto-agent entries carried over from a prior connection.
-    // The server re-sends its current active set as part of the initial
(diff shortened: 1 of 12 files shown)


joshalbrecht:
Closing as abandoned: branch is in CONFLICTING state and has had no activity in weeks.

PR #1369, opened 2026-04-22
Import TEST_ENV_PATTERN from env_utils, not testing

weishi-imbue:
## Summary

Fix a `ModuleNotFoundError` that crashes `mngr observe` on startup in the packaged `minds.app`, leaving the desktop client stuck on "Discovering agents..." forever.

`mngr_modal.backend` (runtime code) was importing `TEST_ENV_PATTERN` from `imbue.mngr.utils.testing`. That module is explicitly excluded from the built wheel because it imports `pytest` at module scope. Source / workspace mode worked because the editable install had everything; packaged mode blew up at import time.

## Changes (3 files)

1. **`libs/mngr/imbue/mngr/utils/env_utils.py`** — add `TEST_ENV_PREFIX` and `TEST_ENV_PATTERN` (plus the `re` / `typing.Final` imports they need). `env_utils.py` is a non-test module that already ships in the runtime wheel, so runtime consumers can safely import these constants from here.
2. **`libs/mngr/imbue/mngr/utils/testing.py`** — drop the local definitions of `TEST_ENV_PREFIX` / `TEST_ENV_PATTERN` and re-export them via `from imbue.mngr.utils.env_utils import ...`. Also drop the now-unused `import re`. Any existing test-side callers that import these names from `utils.testing` keep working unchanged.
3. **`libs/mngr_modal/imbue/mngr_modal/backend.py`** — swap the import from `imbue.mngr.utils.testing` to `imbue.mngr.utils.env_utils`. This is the actual runtime fix; the first two changes just make this import resolvable without dragging `pytest` into the wheel.

## Notes

This matches the "env_utils: extract TEST_ENV_PATTERN" slice of local commit `4d6c07e31` (titled "Packaging fixes so the packaged minds.app actually runs"), which bundled this fix together with four unrelated packaging fixes (workspace package pinning, git binary bundling, `mngr_lima` version pin, etc.). This PR cherry-picks only the `env_utils` slice.

## Test plan

- [ ] CI passes (unit + integration tests on offload)
- [x] Verify `TEST_ENV_PATTERN` is importable from both `imbue.mngr.utils.env_utils` and (for back-compat) `imbue.mngr.utils.testing`
- [x] Confirm packaged `minds.app` can run `mngr observe` without `ModuleNotFoundError`


libs/mngr/imbue/mngr/utils/testing.py
```diff
@@ -55,15 +54,10 @@
...
+from imbue.mngr.utils.env_utils import TEST_ENV_PATTERN
+from imbue.mngr.utils.env_utils import TEST_ENV_PREFIX
 from imbue.mngr.utils.polling import wait_for
 
-# Prefix used for test environments
-TEST_ENV_PREFIX: Final[str] = "mngr_test-"
-
-# Pattern to match test environment names: mngr_test-YYYY-MM-DD-HH-MM-SS
-# The name may have additional suffixes (like user_id)
-TEST_ENV_PATTERN: Final[re.Pattern[str]] = re.compile(r"^mngr_test-(\d{4})-(\d{2})-(\d{2})-(\d{2})-(\d{2})-(\d{2})")
```
weishi-imbue, line 65: These are excluded when building wheel: https://github.com/imbue-ai/mngr/blob/main/libs/mngr/pyproject.toml#L60

But they're needed and accessed in production code path. 
Thus moving them out from testing

joshalbrecht: When refactoring like this please actually update all of the imports--don't move them to a new file and have the old file import them

joshalbrecht:
lgtm other than my comment

libs/mngr/imbue/mngr/utils/testing.py
weishi-imbue, line 130: This method belongs to testing and it relies on `TEST_ENV_PREFIX` from env_util

libs/mngr/imbue/mngr/utils/testing.py
weishi-imbue: Updated to make sure no call site is importing TEST_ENV_PATTERN from `util.testing` anymore. 

Here testing.py needs import env_utils since it's being using it. (e.g. line 130)

PR #1372, opened 2026-04-22
Rewrite Location headers in desktop client proxy

gnguralnick:
## Summary

- Add `rewrite_location_header` to the desktop client proxy so site-absolute redirects (e.g. FastAPI `303`-after-POST with `Location: /foo`) stay within the `/forwarding/{agent_id}/{server_name}/` prefix instead of escaping to the desktop client root.
- Wire it into `_build_proxy_response` next to the existing `Set-Cookie Path=` rewrite. Relative (`./foo`), protocol-relative (`//host/x`), absolute-URL (`https://…`), fragment-only, and already-prefixed paths pass through unchanged.
- Removes the need for proxied FastAPI apps to work around the bug with relative URLs (e.g. `./?flash=…` instead of `/?flash=…`).

## Test plan

- [x] 11 new unit tests in `proxy_test.py` covering rewrite + passthrough cases.
- [x] 3 new integration tests in `test_desktop_client.py` that exercise a real backend `RedirectResponse` through the proxy and assert the rewritten `Location`.
- [x] `just test-quick apps/minds/imbue/minds/desktop_client/proxy_test.py` — 38 passed.
- [x] `just test-quick apps/minds/imbue/minds/desktop_client/test_desktop_client.py` — 92 passed.
- [x] `just test-quick "apps/minds -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'"` — 619 passed (including `test_ratchets.py`).
- [x] `ruff check` + `ruff format --check` — clean.
- [ ] Manual verification through a running minds container — not done; see notes below.

## Notes / follow-ups

- Not handled: `Refresh: 0; url=/foo` headers (uncommon, same class of bug).
- Not handled: `Location` headers containing the backend's own host (e.g. `http://127.0.0.1:PORT/foo`, which FastAPI emits when using `request.url_for(...)`). Would require passing the backend base URL into the rewriter. Worth doing only if seen biting a real app.

### DIFF
--- a/apps/minds/imbue/minds/desktop_client/app.py
+++ b/apps/minds/imbue/minds/desktop_client/app.py
@@ -49,6 +49,7 @@ from imbue.minds.desktop_client.proxy import generate_backend_loading_html
 from imbue.minds.desktop_client.proxy import generate_bootstrap_html
 from imbue.minds.desktop_client.proxy import generate_service_worker_js
 from imbue.minds.desktop_client.proxy import rewrite_cookie_path
+from imbue.minds.desktop_client.proxy import rewrite_location_header
 from imbue.minds.desktop_client.proxy import rewrite_proxied_html
 from imbue.minds.desktop_client.request_events import RequestInbox
 from imbue.minds.desktop_client.request_events import RequestStatus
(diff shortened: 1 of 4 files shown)


gnguralnick:
Superseded by #1383. PR #1375 landed on main and moved path forwarding (including the Set-Cookie/HTML rewriting this PR touches) from the desktop client into `minds_workspace_server`. The fix has been re-ported against the new `service_dispatcher._build_proxy_response` code path in #1383.

PR #1380, opened 2026-04-22
accept github/gitlab URLs in mngr create --source

evgunter:
## Summary

Enable `mngr create --source <git-url>` end-to-end: detect git URLs at the CLI, clone to a managed directory, and feed the clone into the existing source-resolution pipeline. Track each clone for garbage collection and refuse to delete clones that still have local branches not present on any remote.

### CLI
- Detect http(s)/ssh/git/file schemes and SCP-like `git@host:path` URLs in `_resolve_source_location` before `parse_source_string`.
- Clone to `<host_dir>/clones/<name>-<uuid4.hex>/` on the local host (mirrors the existing `worktrees/<name>-<uuid4.hex>/` layout).
- Agent name fallback for the clone dir: `positional_name` → `--name` → `parse_project_name_from_url` → `"agent"`, via the new shared `pick_agent_name_hint` helper.
- `mngr clone <git-url>` inherits URL support via its existing `create --from` pass-through.
- `--from`/`--source` help text updated; clone help text gets a git-URL example.

### GC
- New `generated_source_dirs` field on `CertifiedHostData` (sibling to `generated_work_dirs`).
- URL branch in `_resolve_source_location` calls `register_generated_source_dir`.
- `_get_orphaned_source_dirs`: a tracked clone is "in use" if a living agent's `work_dir` is the clone itself or a git worktree backed by it.
- Orphan clone with no local branches outside every remote → deleted; tallied in `GcResult.source_dirs_destroyed`.
- Orphan clone with unpushed branches → **kept, never silently destroyed**, logged as a warning, tallied in `GcResult.source_dirs_kept_due_to_unpushed_branches`. Destructive git operations require the same explicit opt-in as elsewhere (`destroy -b` to drop the mngr-created branch; after that the clone is GC-eligible).

### Tests
- Unit tests for `is_git_url` (http/https/ssh/git/file/SCP, plus negatives for agent addresses and paths).
- Unit tests for `clone_git_url_to_managed_dir` (happy path; cleans up on failure).
- Unit tests for `_resolve_source_location` URL branch via `file://` (no network, no monkeypatch).
- Unit tests for `_local_branches_not_on_any_remote_on_host` and `_get_orphaned_source_dirs` (fresh clone deletable; local-only branch keeps clone).
- Acceptance test `test_cli_create_from_git_url` clones `https://github.com/imbue-ai/mngr` and asserts `CLAUDE.md` exists in the resulting worktree.

## Test plan
- [x] `just test-quick libs/mngr/imbue/mngr/utils/git_utils_test.py` — 68 passed
- [x] `just test-quick libs/mngr/imbue/mngr/cli/create_test.py libs/mngr/imbue/mngr/api/gc_test.py` — tests green
- [x] `just test-quick "libs/mngr -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'"` — 3657 passed
- [ ] Acceptance test `test_cli_create_from_git_url` runs in CI via `just test-offload-acceptance` (requires network to github.com)


libs/mngr/imbue/mngr/utils/git_utils.py
```diff
@@ -164,6 +174,49 @@ def parse_project_name_from_url(url: str) -> str | None:
...
+    """
+    dest = base_dir / f"{name}-{uuid.uuid4().hex}"
+    base_dir.mkdir(parents=True, exist_ok=True)
+    try:
+        cg.run_process_to_completion(
+            ["git", "clone", url, str(dest)],
+            timeout=_GIT_CLONE_TIMEOUT_SECONDS,
+        )
+    except ProcessError as e:
+        shutil.rmtree(dest, ignore_errors=True)
```
evgunter, line 215: not sure if this is necessary / desirable

joshalbrecht: that seems correct to me

actually, I suspect we want even more than that--are we sure that this clone is cleaned up after the resulting mngr command is done?

we might want this to be a context manager (otherwise you will accumulate git clones that never get cleared, and it is incoherent to say "clone this git repo and use it in-place" because you don't know where it is going)

An alternative would be for GC to handle cleaning up these clones (might be nice for them to stay cached), but then we'd probably want some settings around how long to keep them cached (since last use, track last use time, etc, which starts to get quite a bit more complicated, but is ultimately more usable)

thoughts?

evgunter: updated to have gc delete the clone if all agents that use it are gone and it has no unpushed branches. (if it does have unpushed branches then gc logs a warning and leaves it alone)

libs/mngr/imbue/mngr/interfaces/data_types.py
```diff
@@ -321,6 +321,14 @@ def idle_mode(self) -> IdleMode:
...
         description="List of work directories that were generated by mngr for agents on this host",
     )
+    generated_source_dirs: tuple[str, ...] = Field(
+        default_factory=tuple,
+        description=(
+            "Source repositories generated by mngr (e.g. git clones created from --source <url>). "
+            "Tracked separately from generated_work_dirs because they back one or more worktrees "
+            "rather than being an agent's work_dir themselves."
+        ),
+    )
```
evgunter, line 331: new machinery

libs/mngr/imbue/mngr/api/data_types.py
```diff
@@ -69,6 +69,17 @@ class GcResult(MutableModel):
...
+        default_factory=list,
+        description="Source repositories (e.g. mngr-managed git clones) that were destroyed",
+    )
+    source_dirs_kept_due_to_unpushed_branches: list[WorkDirInfo] = Field(
+        default_factory=list,
+        description=(
+            "Source repositories that were left in place because they still have local branches "
+            "not present on any remote. Delete the branches (or push them) to allow future GC."
+        ),
+    )
```
evgunter, line 82: new machinery

joshalbrecht:
lgtm

PR #1385, opened 2026-04-23
fix resource-guards self-test that broke every later test in the session

evgunter:
## Summary

- **Root cause**: `test_start_and_stop_resource_guards_round_trip` called `pluginmanager.unregister(name=\"resource_guards\")` expecting to clean up a plugin it had registered. In the outer session, `start_resource_guards()` is idempotent when a plugin already exists and skips registration — so the plugin being unregistered was the outer session's `_ResourceGuardPlugin`, killing `_pytest_runtest_setup`/`_teardown`/`_makereport` for every subsequent test in that pytest session.
- **Downstream effect**: the last test before the self-test left its `_PYTEST_GUARD_*=block` env in `os.environ` forever; any later test that invoked a guarded binary (rsync, tmux, etc.) through the bash wrapper got exit 127 regardless of marks. Surfaced as the batch-ordering-dependent offload flake (`test_pull_files_clobber_mode_with_delete_flag`, `test_edit_message_sends_edited_content`, etc.).
- **Fix**: lift the start/stop symmetry into the library. `start_resource_guards()` now records `_owns_guard_plugin` (plus plugin instance and pluginmanager ref) only when it actually registers a new plugin; `stop_resource_guards()` only unregisters and only clears those globals when ownership is set. Mirrors the existing `_owns_guard_wrapper_dir` pattern. Round-trip self-test collapses back to a plain start+stop, gaining a regression assertion that the outer plugin survives stop. New owner-case test exercises the previously-uncovered path.
- **Flaky mark removal**: drops `@pytest.mark.flaky` on `test_pull_files_clobber_mode_with_delete_flag_removes_host_only_files` now that the root cause is gone.

## Validation

- `just test-quick libs/resource_guards` — 97 passed, 0 failed.
- `just test-offload` (three runs): 7532/7532 passed each time; run 1 (pre-fix) produced the diagnostic log that pinned the root cause, runs 2 and 3 (post-fix) had zero flakes.

## Test plan

- [ ] CI passes (offload + acceptance).
- [ ] Previously-flaky `test_pull_files_clobber_mode_with_delete_flag_removes_host_only_files` passes without retries.


joshalbrecht:
lgtm, but worth looking at why CI failed (seems strange--seems like the mngr_claude coverage gate, but how is that flaky... or is it...)

PR #1390, opened 2026-04-23
mngr: GC reclaims offline hosts via destroy_host, not delete_host

weishi-imbue:
## Summary

`gc_machines`' offline branch called `provider.delete_host(host)` on hosts it decided to reap. On Lima, `delete_host` removes mngr's JSON record but does *not* invoke `limactl delete`, so the underlying VM stays on disk as a ~20 GiB orphan that mngr can no longer reach. Every minds delete-workspace that happened to run `--gc` against an offline target left one of these behind.

Fix: `libs/mngr/imbue/mngr/api/gc.py:311` — flip the terminal call in the offline branch from `delete_host` to `destroy_host`. On Lima, `destroy_host` calls `limactl_delete` (kills VM) and marks the record DESTROYED. On already-DESTROYED hosts it's idempotent: the VM-kill call swallows "already gone" errors (`instance.py:701-702`).

## Reproduction (confirmed on `origin/main` before this change)

```bash
# Isolated state dirs so nothing touches ~/.minds / ~/.lima
export MNGR_HOST_DIR=/tmp/mngr-gc-test/mngr  LIMA_HOME=/tmp/lima-gc
export MNGR_PREFIX=gc-

# 1. Create one Lima agent
mkdir -p /tmp/gctest-work && cd /tmp/gctest-work && git init -q && git commit --allow-empty -qm init
mngr create v@.lima --command "sleep 3600" --no-connect --no-rsync --disable-plugin modal

# 2. Destroy the agent while host is online -> host record survives with 0 agents
mngr destroy v --force --no-gc --disable-plugin modal

# 3. Stop the VM externally (mimics idle-timeout / Mac reboot)
limactl stop gc-<auto-name>-host

# 4. Age the host record so gc's age threshold triggers
python3 -c "import json, datetime, sys; p=sys.argv[1]; d=json.load(open(p)); \
  old=(datetime.datetime.now(datetime.timezone.utc)-datetime.timedelta(days=10)).isoformat(); \
  d['certified_host_data']['updated_at']=old; d['certified_host_data']['stop_reason']='STOPPED'; \
  json.dump(d, open(p,'w'), indent=2)" $HOST_JSON

# 5. Run gc
mngr gc --provider lima --disable-plugin modal
```

Before this change: `host_state/*.json` is gone, `limactl list` still shows the VM → orphan.
After this change: `limactl list` no longer has the VM. Clean.

## Test plan

- [x] Three new regression tests in `gc_test.py` assert the anti-orphan invariant: for STOPPED / CRASHED / FAILED offline hosts that gc reaps, `destroy_host` is called and `delete_host` is not. All three fail on `origin/main` and pass after the fix.
- [x] Four existing tests that encoded the buggy behavior (asserting `deleted_hosts == [host.id]`) are updated to assert `destroyed_hosts` instead.
- [x] The dry-run test is updated to assert neither `destroy_host` nor `delete_host` is called.
- [x] `MockProviderInstance.destroy_host` used to raise `NotImplementedError`; now it records calls so tests can observe both paths independently.
- [x] All 711 tests in `libs/mngr/imbue/mngr/api` pass.
- [x] All 284 tests in `libs/mngr/imbue/mngr/providers` pass.
- [ ] CI (offload)

## Not in this PR — tracked as follow-ups

- `_run_post_destroy_gc` still runs *global* gc after every destroy; scoping it to just the destroyed agent's provider is a separate issue (the "`mngr destroy agent-X --gc` wiped state for unrelated offline hosts" report).
- Decoupling `--gc` from `destroy` entirely (opt-in `mngr gc` only) is also separate.

### DIFF
--- a/libs/mngr/imbue/mngr/api/gc.py
+++ b/libs/mngr/imbue/mngr/api/gc.py
@@ -302,13 +302,7 @@ def _gc_single_host(
                     HostState.CRASHED,
                     HostState.DESTROYED,
                 ):
-                    # Reclaim the underlying VM (destroy_host) rather than just
-                    # wiping mngr's record (delete_host). For STOPPED/CRASHED/
-                    # FAILED hosts the VM is still live on the provider, so
-                    # delete_host on its own leaves it orphaned on disk. For
-                    # already-DESTROYED hosts destroy_host is idempotent --
-                    # the provider's destroy_host invokes limactl_delete /
-                    # equivalent which no-ops on already-gone VMs.
+                    # permanently delete the host's data
                     if not dry_run:
                         # FOLLOWUP: when there are multiple instance of gc running concurrently on different hosts
                         #  there's a risk of getting into a screwy situation here--if we delete this right as
--- a/libs/mngr/imbue/mngr/api/gc.py
+++ b/libs/mngr/imbue/mngr/api/gc.py
@@ -302,13 +302,19 @@ def _gc_single_host(
                     HostState.CRASHED,
                     HostState.DESTROYED,
                 ):
-                    # permanently delete the host's data
+                    # Reclaim the underlying VM (destroy_host) rather than just
+                    # wiping mngr's record (delete_host). For STOPPED/CRASHED/
+                    # FAILED hosts the VM is still live on the provider, so
+                    # delete_host on its own leaves it orphaned on disk. For
+                    # already-DESTROYED hosts destroy_host is idempotent --
+                    # the provider's destroy_host invokes limactl_delete /
+                    # equivalent which no-ops on already-gone VMs.
                     if not dry_run:
                         # FOLLOWUP: when there are multiple instance of gc running concurrently on different hosts
                         #  there's a risk of getting into a screwy situation here--if we delete this right as
                         #  someone else starts it, you might have a host that is running but is untracked
                         #  This can be easily fixed by adding some host-id-keyed locking at the provider level (which both create/start/delete would acquire)
-                        provider.delete_host(host)
+                        provider.destroy_host(host)
                         emit_host_destroyed(
                             mngr_ctx.config,
                             host_ref.host_id,
--- a/libs/mngr/imbue/mngr/api/gc_test.py
+++ b/libs/mngr/imbue/mngr/api/gc_test.py
@@ -129,7 +129,9 @@ def _run_gc_machines(provider: MockProviderInstance, *, dry_run: bool = False) -
 def test_gc_machines_deletes_old_offline_host_with_no_agents(
     gc_mock_provider: MockProviderInstance, temp_mngr_ctx: MngrContext
 ) -> None:
-    """Old offline hosts with no agents are deleted to prevent data accumulation."""
+    """Old offline hosts with no agents are reclaimed via destroy_host (kills
+    the VM + marks record DESTROYED), not delete_host (record-only wipe,
+    which would orphan the VM on disk)."""
     host = _make_offline_host(gc_mock_provider, temp_mngr_ctx, days_old=14)
     gc_mock_provider.mock_hosts = [host]
 
@@ -137,7 +139,8 @@ def test_gc_machines_deletes_old_offline_host_with_no_agents(
 
     assert len(result.machines_deleted) == 1
     assert result.machines_deleted[0].host_id == host.id
-    assert gc_mock_provider.deleted_hosts == [host.id]
+    assert gc_mock_provider.destroyed_hosts == [host.id]
+    assert gc_mock_provider.deleted_hosts == []
 
 
 def test_gc_machines_skips_recent_offline_host(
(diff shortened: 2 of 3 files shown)


weishi-imbue:
### Symptoms - what breaks
In minds app, I have several lima workspaces. I restart macbook. Every Lima VM is now Stopped. 
I reopens minds, deletes one workspace. Minds runs `mngr destroy selene-1 --force --gc ...`. The `--gc` post-hook sweeps *globally* across all hosts; finds the other stopped VMs, classifies them as CRASHED, calls `provider.delete_host(...)` on each. mngr records wiped, but the Lima VMs remain on disk orphaned. 

### Table 1 — lima vs. mngr status

| Event on the Mac | `limactl list` STATUS | `mngr list --fields host.state` | host JSON `stop_reason` |
|---|---|---|---|
| `mngr stop <host>` (graceful) | Stopped | STOPPED | `"STOPPED"` |
| `limactl stop` / Mac reboot / power-off | Stopped | **CRASHED** | `null` (unchanged) |
| `mngr destroy <host>` (graceful) | *(instance gone)* | DESTROYED | `"DESTROYED"` |
| qemu crash / OOM-kill | Broken or Unknown | CRASHED | unchanged |


### Question 1: should gc destroy VMs instead just delete the mngr's agent record.  (Scope of this PR) 
I think we should destroy VMs upon GC. (but can discuss on which is worse: accumulated orphaned VMs vs. accidental data loss. ) 

### Question 2: GC behavior (not touched in this PR) 
Should we run full gc upon `mngr destroy <agent>`? or scope it within that agent's host? 

Today:
`mngr destroy <agent> --gc` hits `_run_post_destroy_gc` (`destroy.py:576`), which runs `api_gc(providers=get_all_provider_instances(mngr_ctx), ...)` — every provider, every host, no scoping. For each offline host that meets the age threshold AND (`len(agent_refs) == 0` OR state ∈ `FAILED/CRASHED/DESTROYED`), it calls `provider.delete_host(host)` — mngr-record-only wipe; on Lima it does not invoke `limactl delete`.

### Question 3: lima state and mngr host state not in sync. (Scope of another PR draft #1392)
Proposing a new method in ProviderInstanceInterface:
```py
  def get_observed_host_state(self, host_id: HostId) -> HostState | None:
      """The provider's current live view of this host's state, or None if the
      provider can't tell right now (e.g., daemon unreachable). Unlike the stored
      `stop_reason`, this reflects reality including external state changes
      (user rebooted their Mac, someone ran `limactl stop`, a sandbox terminated)."""
```

libs/mngr/imbue/mngr/api/gc.py
```diff
@@ -308,7 +308,7 @@ def _gc_single_host(
                         #  there's a risk of getting into a screwy situation here--if we delete this right as
                         #  someone else starts it, you might have a host that is running but is untracked
                         #  This can be easily fixed by adding some host-id-keyed locking at the provider level (which both create/start/delete would acquire)
-                        provider.delete_host(host)
+                        provider.destroy_host(host)
```
weishi-imbue, line 311: Reclaim the underlying VM (destroy_host) rather than just wiping mngr's record (delete_host). For STOPPED/CRASHED/AILED hosts the VM is still live on the provider, so elete_host on its own leaves it orphaned on disk. For already-DESTROYED hosts destroy_host is idempotent -- the provider's destroy_host invokes limactl_delete / quivalent which no-ops on already-gone VMs.

joshalbrecht: This is incorrect.

See my higher level comment

joshalbrecht:
The core change here is wrong. These are the APIs:

image

Note that *delete requires that the host has already been destroyed*.

This should probably be asserted by the providers, but logically:
- "delete" is about "remove our own knowledge that this thing ever existed"
- "destroy" is about "remove the resources associated with this thing"

Does that make sense?

Thus the correct changes are:
1. the lima provider should understand that something is not CRASHED if it was cleanly stopped.
2. even *if* a VM crashed, the state shoudl not be CRASHED. That state name is bad, but it really means "irrecoverably failed". If a VM was stopped, the lima provider can simply start it again, so it should be considered "STOPPED".
3. we should fix the relationship of delete and destroy so that they are better enforced. I could imagine that we switch to a world where calling delete() *also* calls destroy() as part of its implementation (to ensure that everything was removed). This is a larger change but almost certainly worth doing (there just might be some performance things we need to fix when we do that, and some errors to suppress if resources are already gone, in order to make it idempotent across all providers).  Note that for remote providers, like modal, if delete implies calling destroy in the beginning of the implementation, then it cannot suppress failures due to networking errors--it needs to bail in such a case, so that you have definitely removed all resources (destroyed) before deleting
4. As a follow-up, we probably should rename CRASHED and STOPPED so that we have a clearer distinction between unrecoverable failures and recoverable ones (don't do any of that now though, other than maybe updating docs to capture what is in this comment)

PR #1391, opened 2026-04-23
Basic Latchkey integration to Minds

hynek-urban:
(Accompanies https://github.com/imbue-ai/forever-claude-template/pull/13.)

- Add Latchkey as a dependency to the project.
- For each created agent, spawn a local `latchkey gateway` instance.
- When appropriate (= not in the dev mode), expose Latchkey gateway's port via an ssh reverse tunnel.
- Gateway instances keep running even after the desktop client is closed (because so do agents). The agent <-> gateway pairing is persisted in the state directory. Next time the desktop client runs, it ensures that the right set of latchkey gateway instances is running again (in case it drifted in the meantime).

I didn't use the precompiled version of Latchkey because the browser-related logic cannot be packaged that way. Since we already have node and npm dependencies, it should be fine.


apps/minds/imbue/minds/desktop_client/latchkey/gateway.py
```diff
@@ -0,0 +1,464 @@
+"""Per-agent Latchkey gateway processes for locally-reachable agents.
+
+For each workspace agent running on the local machine or in a container/VM
+on the local host (providers: ``local``, ``docker``, ``lima``), the desktop
+client spawns a dedicated ``latchkey gateway`` subprocess bound to a free
+port on 127.0.0.1. Cloud/VPS providers (``modal``, ``vultr``, ...) do not
+get a gateway since agents there cannot reach the local machine directly.
```
joshalbrecht: though in the future we'll support running the gateway directly on vultr, yes?

that, I think, will be the default / best way to make this work (relatedly FYI: I'm currently working on making a default service, at least for us here at Imbue, that will be able to just send an already-provisioned vultr VPS agent to the requesting user)

apps/minds/imbue/minds/desktop_client/ssh_tunnel.py
```diff
@@ -99,7 +109,11 @@ class SSHTunnelManager(MutableModel):
     _tunnel_socket_paths: dict[str, Path] = PrivateAttr(default_factory=dict)
     _tunnel_threads: dict[str, threading.Thread] = PrivateAttr(default_factory=dict)
     _shutdown_event: threading.Event = PrivateAttr(default_factory=threading.Event)
-    _reverse_tunnels: dict[str, ReverseTunnelInfo] = PrivateAttr(default_factory=dict)
+    # Reverse tunnels are keyed by ``(conn_key, local_port)`` so that a single
+    # SSH host can host multiple concurrent tunnels for different purposes --
+    # e.g. one for the minds API (``local_port == server_port``) and one per
+    # agent for the Latchkey gateway (``local_port == per_agent_gateway_port``).
+    _reverse_tunnels: dict[tuple[str, int], ReverseTunnelInfo] = PrivateAttr(default_factory=dict)
```
joshalbrecht, line 117: do we actually want to allow multiple?

The only reason I made the minds API forwarding was just to give a sort of example of how this was working... I was kind of imagining that we would ONLY expose latchkey, and you would access the minds API *via* latchkey, if that makes sense? (that way permissions would be controlled and consistent there)

not a huge deal, just flagging what I was imagining (and it'd be nice to have a single tunnel, multiple tunnels sounds like a pain)

joshalbrecht:
other than my comments, LGTM!

(and even my comments are not blocking, just things I wanted to flag)

apps/minds/imbue/minds/desktop_client/latchkey/gateway.py
hynek-urban: Yeah, this comment was a bit outdated (I updated it locally manually but forgot to push). Right now, in the first step, we simply use the reverse SSH tunnel since that works across uniformly across all providers and is already usable (it just means the agent won't be able to use latchkey when you close your laptop). In a followup, we should investigate the option of running the gateways remotely. I'm just a bit wary of the synchronization logic that will be necessary for this.

apps/minds/imbue/minds/desktop_client/ssh_tunnel.py
hynek-urban: That makes sense. I'll note it down as something to change in a follow-up once the basic Latchkey integration is proven to work for everyone (macs, desktop builds, ...).

PR #1397, opened 2026-04-23
Capture Electron console output to a rotating JSONL

gnguralnick:
## Summary

- Add a rotating JSONL sink at `~/.minds/logs/electron.jsonl` capturing Electron main-process lifecycle output plus every `WebContentsView` renderer's `console-message` events.
- Add a URL-based classifier that routes renderer messages from agent-owned service iframes (`<agent-id>.localhost/service/<name>/…`) to the mind's workspace server via a new `POST /api/iframe-logs` endpoint, while keeping local chrome UI and workspace top-level page output on the user's machine.
- Add a per-mind batching buffer (1 s periodic flush, drop-oldest overflow) that POSTs via the desktop-client's subdomain proxy so the existing session cookie is attached without new auth plumbing.

Workspace-server mind-side logs land in `$MNGR_HOST_DIR/logs/iframe/events.jsonl`, sharing the envelope/rotation convention used by `imbue_common.logging.make_jsonl_file_sink`.

## Routing table

| Frame URL                                   | Destination | `source` tag                                    |
| ------------------------------------------- | ----------- | ----------------------------------------------- |
| `localhost[:port]/…`                        | LOCAL       | `electron/renderer/local/<view>`                |
| `<agent-id>.localhost[:port]/` (non-service)| LOCAL       | `electron/renderer/workspace/<mind-id>`         |
| `<agent-id>.localhost[:port]/service/<n>/…` | MIND        | `electron/renderer/service/<n>/<mind-id>`       |
| unrecognised                                | LOCAL       | `electron/renderer/unclassified`                |

ttyd is treated the same as agent-owned services because its JS-layer console output carries no data the agent can't already see via `tmux capture-pane`, and the complexity of distinguishing it wasn't worth the marginal benefit.

## Test plan

- [x] `just test-quick apps/minds_workspace_server` (230 passed, including 3 new endpoint tests + 6 new writer tests)
- [x] `npm test` in `apps/minds` (27 Node `--test` cases for classifier, log writer, iframe buffer)
- [ ] `just test-offload` full CI run
- [ ] Manual: launch app, confirm `~/.minds/logs/electron.jsonl` grows with startup and navigation events
- [ ] Manual: open a mind, open an agent-owned service iframe that calls `console.log`, confirm record appears in `$MNGR_HOST_DIR/logs/iframe/events.jsonl` on the mind and is absent from the local file
- [ ] Manual: verify `console.log` in the workspace-server top-level page lands LOCAL with `source: electron/renderer/workspace/<mind-id>`


apps/minds/electron/classify-frame.js
```diff
@@ -0,0 +1,88 @@
+// Classify a `console-message` event by inspecting the frame URL that produced
+// it, deciding whether the record belongs in the local-only Electron log or
+// should be forwarded to the mind's workspace-server log file.
+//
+// Rule (see docs/security-boundaries-audit.md or the branch PR for context):
```
gnguralnick, line 5: don't mention either of these sources - just give the rule

apps/minds/electron/classify-frame.js
```diff
@@ -0,0 +1,88 @@
...
+//
+//   localhost[:port]/*                                       -> LOCAL (our shell UI)
+//   <agent-id>.localhost[:port]/  (path does not start /service/) -> LOCAL (workspace-server top-level frame)
+//   <agent-id>.localhost[:port]/service/<name>/...           -> MIND  (agent-owned iframe)
+//   anything else                                            -> LOCAL, tagged unclassified
+//
+// Stays a pure function so it can be exercised directly under `node --test`
```
gnguralnick, line 12: don't give this justification - it's good for this to be a pure function regardless as all it does is route

apps/minds/electron/log-router.js
```diff
@@ -0,0 +1,177 @@
+// Pure routing logic for Electron console capture.
+//
+// Decides, per `console-message` event, whether the record belongs on the
+// local-only writer (`~/.minds/logs/electron.jsonl`) or should be forwarded
+// to a mind's workspace server via the iframe-log buffer, and holds the
+// pre-port pending queue that protects the privacy boundary while the
+// desktop-client backend port is still being discovered.
+//
+// This module deliberately has no Electron imports so it can be exercised
+// directly under `node --test`. The Electron-aware singleton in
```
gnguralnick, line 10: don't mention test stuff here

gnguralnick: also i don't think we need to describe the usage; should be obvious based on the interface

apps/minds/electron/log-router.js
```diff
@@ -0,0 +1,177 @@
...
+    }
+  }
+
+  /**
+   * Record a main-process event (replaces ad-hoc `console.log` calls).
+   *
+   * @param {string} level
+   * @param {string} message
+   */
+  logMain(level, message) {
```
gnguralnick, line 85: not sure if we have ratchets for the js code but if so maybe we should add one to check for raw console.logs?

apps/minds/electron/log-router.js
```diff
@@ -0,0 +1,177 @@
...
+    if (this._closed) return;
+    // Swallow writer failures: this method is invoked from arbitrary
+    // main-process code paths and a thrown fs error would crash Electron.
+    // Console capture is diagnostic, never load-bearing.
+    try {
+      this._writer.write({ level, source: 'electron/main', message });
+    } catch {
+      // Intentionally empty: logging must never surface errors to callers.
```
gnguralnick, line 93: is there anything better we can do for this type of error?

PR #1402, opened 2026-04-24
Investigation notes: PR #1336 env-claude-hang

joshalbrecht:
## Summary

Investigation-only branch. Reproduced PR #1336's claimed 0-byte hang
locally and identified the actual root cause.

## TL;DR

The `env claude` wrap in PR #1336 is a **timing hack**, not a fix for a
kernel-level fork/exec issue. The "0-byte hang" is a race in mngr's
`HeadlessClaude.stream_output`: the tail loop exits immediately because
`is_finished()` returns True based on stale tmux pane state, before
`send-keys` has even put the command on the pane. The stdout file exists
(bash's redirection opened it at 0 bytes when the command was about to
run) but is empty. mngr raises "exited without producing output", tears
the agent down, and the actual claude output — which arrives a second or
two later — is lost.

## Reproduction (verified in this Modal/gVisor sandbox, on `main`)

| Variant | Runs | Result |
| --- | --- | --- |
| bare `{cmd_str}` (current main) | 8 | 8/8 "exited without producing output" |
| `env {cmd_str}` (PR #1336's fix) | 5 | 5/5 "claude returned an error: Not logged in" |
| bare `{cmd_str}` + `time.sleep(0.5)` in mngr before `tail_until_done` | 5 | 5/5 "claude returned an error" |

The 0.5-second sleep is not an exec, not a process group change, not a
signal manipulation. It's just time — and it fixes the race, confirming
that `env` is doing nothing more than adding delay.

## The race (instrumented timeline)

```
t=0.000s  is_finished=True (stale pane shows idle shell)
t=0.069s  _wait_for_stdout_file returns True (bash redirection created file at 0 bytes)
t=0.122s  before tail_until_done: is_finished still True → loop skipped
t=0.235s  post-loop: is_finished=False (pane now shows the running command)
          but we've already given up; raises "no output"
t=2.357s  file actually reaches 4344 bytes with the real result event
          (nobody reads it)
```

## Fix

Move the `_startup_grace_seconds` protection from `_wait_for_stdout_file`
into the `tail_until_done` loop's `is_finished()` check. Sketch and full
reasoning in `.investigation-notes/env-claude-hang-pr1336.md`.

Separately, keep the ANTHROPIC_API_KEY propagation fix from
`fdc68196e` — that's legitimate and unrelated.

Drop the `env` prefix from `HeadlessClaudeAgent.assemble_command` and
`ClaudeAgent.assemble_command`.

## Test plan

- [ ] Apply the fix in `base_headless_agent.py` / `headless_claude_agent.py`
      so `tail_until_done` respects the startup grace.
- [ ] Re-run `test_ask_simple_query` in offload-release ≥10 times with
      the env-wrap *removed* to confirm the proper fix is sufficient.
- [ ] Verify no regression for `HeadlessCommand` or other
      `BaseHeadlessAgent` subclasses.

### DIFF
--- a/.investigation-notes/env-claude-hang-pr1336.md
+++ b/.investigation-notes/env-claude-hang-pr1336.md
@@ -1,0 +1,230 @@
+# Investigation: `env claude` vs bare `claude` hang (PR #1336)
+
+## Background
+
+PR #1336 ("upgrade offload to 0.8.1 + fix remaining release test failures") landed
+two independent changes that it attributes to one root cause:
+
+1. `ClaudeAgent.modify_env_vars` now propagates `ANTHROPIC_API_KEY` from
+   `os.environ` into the agent env file (commit `fdc68196e`).
+2. `HeadlessClaudeAgent.assemble_command` and `ClaudeAgent.assemble_command`
+   wrap the `claude` invocation in `env` (commit `b2036ed20`).
+
+The PR body and commit message for `b2036ed20` claim that bare `claude --print …`
+(and equivalently `/root/.local/bin/claude …`) hangs indefinitely with 0 bytes
+on stdout/stderr inside offload-release Modal sandboxes under gVisor, while
+`env claude …` and `timeout 60 claude …` both pass. The mechanism is declared
+"unexplained; empirically robust."
+
+## What this investigation looked at
+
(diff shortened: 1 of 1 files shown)


joshalbrecht:
Closing as abandoned.

PR #1403, opened 2026-04-24
add meta-ratchets to ensure top-level coverage config matches subprojects'

evgunter:
## Summary

Adds two meta-ratchet tests in `test_meta_ratchets.py` that keep the root `pyproject.toml` coverage configuration in sync with each subproject's `pyproject.toml`:

1. **`test_top_level_cov_flags_are_union_of_subproject_cov_flags`** — asserts the root `[tool.pytest.ini_options].addopts` `--cov=` entries are exactly the union of every subproject's `--cov=` entries, minus packages whose source is fully omitted at root via a `libs/X/imbue/X/*` (or `apps/X/imbue/X/*`) pattern (currently only `mngr_modal`).
2. **`test_top_level_coverage_omit_covers_subproject_omits`** — file-level check: for every `.py` file in a subproject's package tree that the subproject's `[tool.coverage.run].omit` excludes, the root's `[tool.coverage.run].omit` must also exclude it. Walks concrete files and matches via `fnmatch` (same matcher coverage.py uses), so subproject globs like `*/testing.py` and root-level fully-qualified paths resolve the same way.

Running check #2 revealed real drift in the root omit list, closed in this PR by widening `[tool.coverage.run].omit` at the root:
- `*/conftest_hooks.py` and `*/ratchet_testing/*` (imbue_common test infra; the latter subsumes the prior single-file `libs/.../ratchets.py` entry)
- `apps/minds/imbue/minds/telegram/{credential_extractor,bot_creator,injector}.py`
- `apps/minds/imbue/minds/desktop_client/supertokens_routes.py`
- `libs/mngr_claude/imbue/mngr_claude/resources/sync_keychain_credentials.py`

**Reconciliation direction note:** this widens the root omit list (shrinking effective combined-coverage surface) to match what subprojects already omit. That's a policy choice the diff makes visible — the alternative would be narrowing subprojects' omits to match root, widening combined coverage. Worth a human eye before merging.

## Test plan
- [x] `just test-quick test_meta_ratchets.py` passes (13 tests, 0 failed).
- [x] Verified extended test fails on the pre-fix state, flagging exactly the 10 files listed above.
- [x] Simulated dropping `--cov=imbue.skitwright` from the root and adding a bogus `--cov=imbue.bogus` entry; first test correctly reports missing/extra flags.
- [ ] CI green (offload + acceptance).


joshalbrecht:
lgtm when tests pass

PR #1409, opened 2026-04-24
fix some tests which were leaking resources and causing flakes on whatever test was last in their batch

evgunter:
contains a real bug fix (kanpan claiming that builtin commands can be overridden, but "d" actually always deleting), a test fix (wait for resource cleanup to finish), and a scary workaround that should probably be removed (skip doing stuff in a latchkey test with unclear semantic correctness)

---

## Summary

Fixes CI "resource leak" pipeline failures that blamed the wrong test (the last test in the shard, not the actual leaker). Three distinct leak sources identified via investigation + CI log mining; all three fixed.

**Root detection:** `libs/mngr/imbue/mngr/conftest.py` `session_cleanup` (autouse session fixture) walks `psutil.Process().children(recursive=True)` at session end and raises `AssertionError`. Pytest attributes session-fixture teardown errors to whichever test ran last -- hence the mis-attribution.

### 1. Kanpan test leaked `mngr destroy a --force` (8+ CI runs, most frequent)

`test_dispatch_command_execute_key_with_marks` registered a user command at key `"d"`, which collided with `_BUILTIN_COMMAND_KEY_DELETE`. `_start_batch_execution` and `_submit_batch_item` hardcoded the `"d"` key to `_run_destroy` and `"p"` to `_run_git_push`, ignoring `state.commands[key]`. User overrides -- documented in `README.md:124` and exercised by `test_build_command_map_user_overrides_builtin` -- were silently routed to `mngr destroy` anyway. The test then called `executor.shutdown(wait=False)`, leaving the `mngr destroy` subprocess in flight past teardown.

**Fix**: model the builtin-vs-user split at the type level. `CustomCommand` now has `kind: Literal["user"]`; new `BuiltinCommand` with `kind: Literal["builtin"]` + `role: BuiltinRole` (UpperCaseStrEnum). `KanpanCommand = Annotated[CustomCommand | BuiltinCommand, Field(discriminator="kind")]`. Dispatch in `_dispatch_command` / `_submit_batch_item` uses `isinstance` + `match cmd.role` with `assert_never` as the catchall, so the type checker flags any missing branch when a new role is added. User TOML parses as `CustomCommand`; `kind = "builtin"` triggers `ValidationError`. No runtime strip-on-load needed -- invariant is enforced by the types.

Test hardened: now touches a tmpfile via the submitted command and asserts the file appears after `shutdown(wait=True)` -- real proof the command ran. Added a companion test covering the `"d"` override path end-to-end.

### 2. ConcurrencyGroup test leaked `sleep 30` (6+ CI runs)

`test_all_failure_modes_get_combined` runs `sleep 30` inside a CG with `exit_timeout_seconds=SMALL_SLEEP` (0.05s), forcing the timeout path. The CG's timeout-path `process.terminate(force_kill_seconds=0.0)` is fire-and-forget: main thread signals shutdown and returns, background thread reaps the subprocess asynchronously. Under CI load the session-cleanup scan landed before the reap completed.

**Fix**: test-side only. The test now captures the `RunningProcess` handle and calls `poll_until(sleep_process.is_finished, timeout=10.0)` after the `with ConcurrencyGroup` block. Closes the race without touching library code. CG's shared teardown path is unchanged from `main`.

(An earlier attempt at a library-level fix -- bumping `force_kill_seconds` to 2s, then SIGKILL-by-PID -- was reverted after it broke three unrelated cli integration tests by adding up to 2s per timed-out process to CG teardown. Those three reverts are in the commit history; their net effect on `libs/concurrency_group/` is zero.)

### 3. Latchkey fake binary caught mid-startup as `ensure-browser` (1 CI run)

`LatchkeyGatewayManager._ensure_browser_once` spawns `latchkey ensure-browser` detached alongside the first gateway spawn (`start_new_session=True` -- not daemonized, so the child's PPID stays the pytest process and `psutil.children(recursive=True)` enumerates it). The test fake binary asserted `argv[1] == "gateway"` and imported `os`/`socket`/`signal` before hitting the assert -- under CI load that startup window was caught by the leak scan.

**Fix**: short-circuit the fake's `ensure-browser` branch with `sys.exit(0)` before any imports. Probabilistic mitigation, not structural -- the prod `_ensure_browser_once` still spawns a detached subprocess by design.

## Commits

```
2886ebd8f  kanpan: tag BuiltinCommand with a BuiltinRole enum for exhaustive dispatch
eb0298a06  kanpan: replace is_builtin flag with BuiltinCommand | CustomCommand union
75fc1beb6  concurrency_group: wait for cleanup in test_all_failure_modes_get_combined
75d5524ad  concurrency_group: shorten sleep in test_all_failure_modes_get_combined
6e761220f  Revert "concurrency_group: actually wait for terminated processes to reap"
76ae3b320  Revert "concurrency_group: SIGKILL timed-out process by PID, not via thread join"
ead6cb90f  Revert "concurrency_group: log ProcessLookupError in timeout-kill path"
1c92bafc1  kanpan: restore raw-dict handling in _load_user_commands
dc8b59e09  concurrency_group: log ProcessLookupError in timeout-kill path
6a1efc6be  concurrency_group: SIGKILL timed-out process by PID, not via thread join
b4301a5c9  kanpan: simplify _load_user_commands to a dict comprehension
4dd46a35a  kanpan: strip is_builtin from user-loaded commands
3b3a0603d  kanpan: replace builtin identity check with is_builtin flag
8bbf6ff02  latchkey tests: short-circuit fake binary on ensure-browser
2dccc4d23  concurrency_group: actually wait for terminated processes to reap
c0d431480  kanpan: honor user override of builtin "d"/"p" keys
cfcf97bee  Fix kanpan test leaking `mngr destroy` subprocess
```

`concurrency_group/` code is net-unchanged from main (reverts cancel the two library attempts; only test changes remain). Squash-merging the PR will collapse the CG revert chain cleanly.

- `session_cleanup` could print `worker_test_ids` (already tracked) alongside leaked cmdlines so future mis-attributions name the likely culprit without an investigation hunt.
- Latchkey prod-side: `_ensure_browser_once` deliberately detaches and doesn't reap. If real-binary runs ever hit CI, consider an `ensure_browser_enabled` constructor param to disable in tests.

## Test plan
- [x] `just test-quick libs/mngr_kanpan` -- 373 passed
- [x] `just test-quick libs/concurrency_group` -- 155 passed
- [x] `just test-quick apps/minds/imbue/minds/desktop_client/latchkey` -- 31 passed
- [x] `uv run ty check libs/mngr_kanpan` -- clean
- [ ] CI green
- [ ] Monitor: `mngr destroy`, `sleep 30`, and `latchkey ensure-browser` no longer appear in session-cleanup leak reports

### DIFF
--- a/apps/minds/imbue/minds/desktop_client/latchkey/gateway_test.py
+++ b/apps/minds/imbue/minds/desktop_client/latchkey/gateway_test.py
@@ -65,6 +65,8 @@ def _make_fake_latchkey_binary(tmp_path: Path) -> Path:
 
     Binds a TCP socket on the host:port supplied via environment variables
     (matching the real binary's contract) and sleeps until terminated.
+    Also accepts ``ensure-browser`` as an immediate no-op exit, since the
+    gateway manager fires that alongside each gateway spawn.
     """
     script = tmp_path / "latchkey"
     # signal.pause() blocks indefinitely until a signal arrives, letting the
@@ -76,9 +78,19 @@ def _make_fake_latchkey_binary(tmp_path: Path) -> Path:
     # fill it up (we never explicitly ``accept`` here -- the kernel ACKs the
     # TCP handshake for queued connections, which is all the liveness probe
     # needs). SIGTERM triggers a clean exit; signal.pause blocks indefinitely.
+    #
+    # The ``ensure-browser`` short-circuit matters for leak detection: the
+    # manager fires ``latchkey ensure-browser`` detached on first gateway
+    # spawn and intentionally does not reap it. If that subprocess is still
+    # in its Python startup when the session-level leak check scans under
+    # CI load, it gets flagged as a leak and attributed to some unrelated
+    # test. Exiting before any import keeps the process window tiny.
     script.write_text(
         "#!/usr/bin/env python3\n"
-        "import os, socket, signal, sys\n"
+        "import sys\n"
+        'if sys.argv[1] == "ensure-browser":\n'
+        "    sys.exit(0)\n"
+        "import os, socket, signal\n"
         'assert sys.argv[1] == "gateway"\n'
         "host = os.environ['LATCHKEY_GATEWAY_LISTEN_HOST']\n"
         "port = int(os.environ['LATCHKEY_GATEWAY_LISTEN_PORT'])\n"
--- a/libs/concurrency_group/imbue/concurrency_group/concurrency_group.py
+++ b/libs/concurrency_group/imbue/concurrency_group/concurrency_group.py
@@ -225,7 +225,14 @@ class ConcurrencyGroup(MutableModel, AbstractContextManager):
                     except StrandTimedOutError as e:
                         timeout_errors.append(e)
                     try:
-                        process.terminate(force_kill_seconds=0.0)
+                        # Wait long enough for the background thread's
+                        # _shutdown_popen path (SIGTERM, reap) to complete.
+                        # Fire-and-forget (force_kill_seconds=0.0) races with
+                        # the session-level leak detector: the subprocess may
+                        # still be in flight when psutil scans at session end,
+                        # producing "leaked process" failures attributed to
+                        # whatever test ran last in the shard.
+                        process.terminate(force_kill_seconds=2.0)
                     except TimeoutExpired:
                         pass
         for tracked_thread in self._threads:
--- a/libs/concurrency_group/imbue/concurrency_group/concurrency_group.py
+++ b/libs/concurrency_group/imbue/concurrency_group/concurrency_group.py
@@ -1,5 +1,3 @@
-import os
-import signal
 import time
 from collections import defaultdict
 from contextlib import AbstractContextManager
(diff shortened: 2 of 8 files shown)


libs/concurrency_group/imbue/concurrency_group/concurrency_group.py
```diff
@@ -225,7 +225,14 @@ def _wait_for_all_strands_to_finish_with_timeout(self, timeout_seconds: float) -
...
                     try:
-                        process.terminate(force_kill_seconds=0.0)
+                        # Wait long enough for the background thread's
+                        # _shutdown_popen path (SIGTERM, reap) to complete.
+                        # Fire-and-forget (force_kill_seconds=0.0) races with
+                        # the session-level leak detector: the subprocess may
+                        # still be in flight when psutil scans at session end,
+                        # producing "leaked process" failures attributed to
+                        # whatever test ran last in the shard.
+                        process.terminate(force_kill_seconds=2.0)
```
evgunter: presumably this is not what we want to do lol

apps/minds/imbue/minds/desktop_client/latchkey/gateway_test.py
```diff
@@ -76,9 +78,19 @@ def _make_fake_latchkey_binary(tmp_path: Path) -> Path:
...
+    # in its Python startup when the session-level leak check scans under
+    # CI load, it gets flagged as a leak and attributed to some unrelated
+    # test. Exiting before any import keeps the process window tiny.
     script.write_text(
         "#!/usr/bin/env python3\n"
-        "import os, socket, signal, sys\n"
+        "import sys\n"
+        'if sys.argv[1] == "ensure-browser":\n'
+        "    sys.exit(0)\n"
+        "import os, socket, signal\n"
```
evgunter, line 93: hmm

libs/concurrency_group/imbue/concurrency_group/concurrency_group.py
```diff
@@ -224,6 +227,24 @@ def _wait_for_all_strands_to_finish_with_timeout(self, timeout_seconds: float) -
...
+                    pid = process.pid
+                    if pid is not None:
+                        try:
+                            os.kill(pid, signal.SIGKILL)
+                        except ProcessLookupError:
+                            # Expected race: subprocess exited between the
+                            # TimeoutExpired above and this SIGKILL. The
+                            # outcome (process gone) is what we wanted, but
+                            # log at trace level for observability.
+                            logger.trace("Subprocess {} already exited before SIGKILL", pid)
```
evgunter: hmmmmmmm

apps/minds/imbue/minds/desktop_client/latchkey/gateway_test.py
joshalbrecht: @hynek-urban to take a look at this one

joshalbrecht:
lgtm after hynek takes a look at that one test

apps/minds/imbue/minds/desktop_client/latchkey/gateway_test.py
hynek-urban: While I'm not 100% sure this change fixes the flake, it certainly doesn't do any harm :) Let's merge it and see.

PR #1412, opened 2026-04-24
add mngr_claude_subagent_proxy plugin

evgunter:
> ⚠️ **EXPERIMENTAL.** This plugin reroutes Claude Code's built-in `Task` tool through mngr-managed subagents via a Haiku dispatcher. Expect rough edges. Known gaps include plan-mode propagation, stop-instead-of-destroy via mngr GC, and transparent permission round-trip — see the "Deferred / out-of-scope" section in the package README. The design also depends on Claude Code internals (hook protocol, plugin cache layout, marketplace fetch behavior) that may shift between Claude Code releases. Watch `mngr list` for orphaned children if a session ends abnormally.

> Built on `mngr/transfer-none-keep-worktree` (merged) — that branch fixes a preexisting bug where `mngr destroy` would reap a still-in-use mngr-managed worktree out from under live agents. Without it, every cascade or manual cleanup of a proxy subagent would wipe the worktree.

## Summary

New standalone plugin `mngr_claude_subagent_proxy` that intercepts Claude Code's `Task` (Agent) tool and replaces the nested native subagent loop with an mngr-managed agent. Users can `mngr connect` to subagents; the parent still receives a normally-shaped `tool_result`.

The plugin is Claude-specific (uses `ClaudeAgentConfig`, a Haiku dispatcher, the Claude Code hook protocol, and the per-agent plugin cache layout); the package name reflects that. It wires through pluggy's `on_after_provisioning` (trylast) and uses mngr_claude's existing `merge_hooks_config` helper. Two small mngr_claude additions: `SESSION_GUARD` and `get_preserved_sessions_dir_for_host` were promoted from private to public helpers so the new plugin can share them instead of re-implementing.

## Architecture

- **`PreToolUse:Agent` hook** (`hooks/spawn.py`) rewrites the Task call: substitutes `subagent_type` to a Haiku dispatcher (`mngr-proxy`), writes the original prompt to `$MNGR_AGENT_STATE_DIR/subagent_prompts/<tid>.md`, generates a per-tool_use_id wait-script, and emits a tightly-scripted prompt that locks Haiku into a 3-case classifier on literal-string presence (no scope to invent). Denies the tool with a clear reason at `MNGR_SUBAGENT_DEPTH >= MNGR_MAX_SUBAGENT_DEPTH` (default 3).
- **Wait-script** invokes `mngr create` for the real mngr-managed subagent (registered agent type `mngr-proxy-child`), then `python -m imbue.mngr_claude_subagent_proxy.subagent_wait` which tails the subagent's JSONL until a terminal `stop_reason` (`end_turn`, `stop_sequence`, `max_tokens`). Body printed to stdout followed by `MNGR_PROXY_END_OF_OUTPUT`. Idempotent on Haiku re-invocations: post-cleanup re-runs emit just the sentinel + exit 0. Uses `--reuse` on `mngr create` so a partially-failed create can recover. Env-file capture wrapped by an EXIT trap so `set -e` mid-create can't leak parent secrets.
- **Watermark sidefile** owned by `subagent_wait` deduplicates permission events: when the wait surfaces a `NEED_PERMISSION`, it stamps the transcript byte-size to a sidefile so the next invocation suppresses re-firing on the same dialog until the child's transcript advances past it. Haiku never sees this — the prompt only knows "run the script, classify the output."
- **Haiku dispatcher** (`resources/mngr-proxy.agent.md`, model: haiku, tools: Bash) echoes the wait-script's body verbatim as its final reply, ending its turn. Haiku's reply IS the parent's Task `tool_result` — Claude Code's PostToolUse `updatedToolOutput` is MCP-only and does not apply to built-in tools like Task, so we route via Haiku rather than via PostToolUse rewrite.
- **`PostToolUse:Agent` hook** (`hooks/cleanup.py`) cascade-destroys the subagent first, then cleans up local side files. Two-signal preserve gate: skips destroy when either `subagent_wait` never wrote the result_file (Haiku gave up early) OR the child is still RUNNING/WAITING per `mngr list`. Either signal → child preserved alive so the user can `mngr connect` to recover. SessionStart reaper sweeps stale ones at next parent boot.
- **`SessionStart` hook** (`hooks/reap.py`) reaps orphaned proxy subagents and stale per-tool_use_id files from prior sessions, AND wraps any unguarded Stop / SubagentStop hooks in the per-agent plugin cache that Claude Code populated this session (see Stop-hook handling below).
- **`on_before_agent_destroy`** hookimpl cascade-destroys all recorded proxy children before the parent's state dir is wiped.
- **API errors are tagged.** When the upstream Anthropic API returns a rate-limit / transient error (Claude Code records these as `isApiErrorMessage: true` on a terminal assistant event), `subagent_wait` prefixes the relayed body with `[ERROR] ` so the parent can distinguish an error from a normal subagent reply that happens to mention "API Error".

All hook logic is Python invoked via thin shell shims in `settings.local.json`; the in-process `mngr api` is called directly rather than shelling back out to `uv run mngr` (with a `get_mngr_command()` resolver that prefers `$UV_TOOL_BIN_DIR/mngr` if mngr_recursive has provisioned one).

## Spawned-subagent agent type and labels

Proxy children register a new `mngr-proxy-child` agent type (`SubagentProxyChildConfig` extends `ClaudeAgentConfig` with one default override: `sync_home_settings=False`). This stops mngr_claude from copying `~/.claude/{plugins,skills,agents,commands}/` into the child's per-agent config dir. Provisioning also installs an auto-allow `PermissionRequest` hook on the child so it doesn't block on permission dialogs the parent never sees.

Each spawned child is tagged with three labels at create time so existing `mngr list` / `mngr destroy` tooling can reason about parent ↔ child structure via `mngr list --format json` and CEL queries:

| Label | Value |
|---|---|
| `mngr_claude_subagent_proxy_parent_name` | The parent agent's `MNGR_AGENT_NAME` |
| `mngr_claude_subagent_proxy_parent_id` | The parent agent's `MNGR_AGENT_ID` |
| `mngr_claude_subagent_proxy_tool_use_id` | The originating Claude Code `tool_use_id` |

Top-level agents have none of these labels, so the presence of `_parent_name` is itself the authoritative "this is a proxy child" signal. README documents example queries (find children of a given parent, hide them from listings, find orphans, etc.).

## Stop-hook handling for spawned subagents

Project-level Claude Code hooks inherit into spawned subagents because the subagent shares the parent's worktree (`--transfer=none`) and Claude Code merges hook arrays across all settings scopes. There's no Claude Code-side knob that cleanly disables a project-level hook from a higher-precedence scope (`disabledPlugins` doesn't exist; `enabledPlugins:{}` doesn't override; `disableAllHooks` doesn't override the project scope; `claude --bare` disables all hooks including mngr's own readiness ones).

The plugin handles this by walking three locations and prepending an env-conditional guard to every non-mngr Stop / SubagentStop command:

    [ -n "$MNGR_CLAUDE_SUBAGENT_PROXY_CHILD" ] && exit 0; <original>

Spawned subagents have `MNGR_CLAUDE_SUBAGENT_PROXY_CHILD=1` set in their env (via the wait-script's `mngr create`), so guarded hooks no-op there. The parent's env doesn't have the var set, so the guard falls through and the original hook runs normally. Idempotent; mngr-managed commands are recognized and skipped.

The three locations auto-guarded:
1. `<work_dir>/.claude/settings.local.json` (provisioning hookimpl).
2. Every `hooks/hooks.json` under `~/.claude/plugins/` (provisioning hookimpl).
3. Every `hooks/hooks.json` under the agent's per-agent plugin cache `<state_dir>/plugin/claude/anthropic/plugins/` — wrapped at **SessionStart**, not provisioning. Claude Code populates this cache by fetching FRESH from each marketplace's source (typically GitHub) at session start, AFTER the provisioning hook runs, so the provisioning-time wrap of the user marketplace dir never reaches the cache. Wrapping at SessionStart catches it before the first Stop hook can fire. (This was a real bug; symptom was spawned subagents being held responsible for the parent's uncommitted changes / failing CI / etc.)

**Project `.claude/settings.json` is git-tracked**, so the plugin deliberately does NOT auto-mutate it. Instead the plugin raises `UnguardedProjectStopHookError` at provisioning when `.claude/settings.json` has un-guarded Stop / SubagentStop hooks. Bypass with `MNGR_CLAUDE_SUBAGENT_PROXY_ALLOW_UNGUARDED_PROJECT_STOP_HOOKS=1`. This branch updates this repo's own `.claude/settings.json` Stop hook to be guarded so the validator passes out of the box.

## `run_in_background` support

Background-mode `Task` calls go through a `--spawn-only` branch of the wait-script: Haiku spawns the child, returns immediately with poll handles (`mngr transcript <name>` / `mngr connect <name>` / `mngr list --include 'name == "<name>"'`), and the child runs to its own end_turn under normal mngr lifecycle. Same machinery, parent never blocks on the child.

## Other components
- `subagent_wait.py` — JSONL tailer with rate-limited `mngr list` presence checks (5s cadence; was firing every 200ms before), heartbeats to stderr to keep Bash idle-output timeout happy, `_TERMINAL_STOP_REASONS = {end_turn, stop_sequence, max_tokens}` for end-of-turn detection, and a destroyed-target fallback that scrapes the last assistant message from the preserved transcript.
- `_stop_hook_guard.py` — shared dict / file manipulation helpers used by both provisioning and SessionStart for guarding Stop / SubagentStop commands.
- `mngr_binary.py` — `get_mngr_command()` resolver, copied (with module docstring noting the dup) from `mngr_recursive.watcher_common.get_mngr_command` plus a `uv run mngr` fallback.
- `testing.py` — `FakeHost` / `FakeAgent` for unit tests.

## Coverage

- Plugin gates itself at 66% (actual unit coverage; real coverage of hook entry points lives in `@pytest.mark.release` tests that don't run in CI).
- Combined-package coverage gate now includes `mngr_claude_subagent_proxy` (was previously omitted at the 5% placeholder); brings the combined floor down to ~80.6% — thin headroom but acceptable.

## Test plan
- [x] 113 unit / integration tests passing locally (`just test-quick libs/mngr_claude_subagent_proxy`).
- [x] Live e2e verified: parent calls Task → subagent spawns → reaches end_turn → parent receives the real subagent answer as `tool_result` → cascade cleanup. Worktree stays intact throughout.
- [x] Real-world `run_in_background=true` validation: a backgrounded Task call from this very PR returned poll handles immediately and the child ran to completion under normal mngr lifecycle.
- [x] `uv run ruff check` + format + ratchets — clean.
- [x] Plugin entry point `claude_subagent_proxy` registered (`importlib.metadata`).
- [x] Real-Claude release tests added: `test_task_tool_spawns_mngr_subagent` (golden path), `test_depth_limit_denies_task`, `test_task_run_in_background_returns_immediately`, `test_plan_mode_propagates_to_subagent` (this last one is expected to fail today; it codifies the not-yet-implemented plan-mode propagation gap).
- [ ] CI offload run (acceptance tests).
- [ ] Live release-test run against real Claude (release tests do NOT run in CI; invoke manually with `ANTHROPIC_API_KEY`).

## Recent changes since first review

- Renamed package `mngr_subagent_proxy` → `mngr_claude_subagent_proxy` (per reviewer feedback that it's Claude-specific). Workspace dir, python package, pyproject name, pluggy entry-point (`subagent_proxy` → `claude_subagent_proxy`), env vars, and label keys all renamed.
- Replaced redundant `mngr_subagent_proxy=child` label with explicit parent linkage labels (parent_name, parent_id, tool_use_id) so existing `mngr list` / `mngr destroy` tooling can reason about parent↔child structure without reading `subagent_map/`.
- Reverted two stale interface additions in `libs/mngr/imbue/mngr/{interfaces,hosts}/host.py` that were carried over from earlier branch state but no longer load-bearing after merging upstream `mngr/transfer-none-keep-worktree`.


joshalbrecht:
1. should probably rename to mngr_claude_subagent_proxy since it kinda depends on this being claude, right?
2. should probably use the same approach as mngr_recursive so that we can GC correctly (linking up the parent, and some separate labels to denote that it is a subagent, etc)
3. we probably want to finish some cleanups (either before or after merging), should chat about details

Other than the above though, lgtm!

Is it actually useful? :)

evgunter:
1. done
2. this was already mostly there, but i swapped from a label that just indicated it was a child to one which says what the parent is
3. i already have some upcoming in https://github.com/imbue-ai/mngr/pull/1483, but there's plenty of mess not resolved there

also lol it's not really useful yet. but i think it has potential!

joshalbrecht:
lgtm!

only leaving  unmerged as a reminder that we should chat about what we want to do next with this (on monday)

PR #1414, opened 2026-04-25
let headless_claude handle both streaming and regular messages

### LINKED ISSUE #1411: headless_claude: extract_text_delta assumes stream_event wrapper but claude v2.1.114+ emits top-level events by default
## Problem

`HeadlessClaude.stream_output` → `_StreamTailState.tail_until_done` → `extract_text_delta` in `libs/mngr_claude/imbue/mngr_claude/headless_claude_agent.py` only extracts text from lines where `parsed["type"] == "stream_event"`. Every other event type is dropped.

```python
# libs/mngr_claude/imbue/mngr_claude/headless_claude_agent.py L37–71
@pure
def extract_text_delta(line: str) -> str | None:
    ...
    if parsed.get("type") != "stream_event":
        return None
    ...
```

But `claude` CLI **v2.1.114** (and I believe several versions before it) emits the top-level Anthropic stream-json format by default, not the `{"type":"stream_event", "event":{...}}` wrapper. The default output is:

```jsonl
{"type":"system", ...}
{"type":"assistant", "message":{"content":[{"type":"text","text":"Hello"}]}, ...}
{"type":"assistant", ...}
{"type":"rate_limit_event", ...}
{"type":"result", ...}
```

No `stream_event` wrappers appear. `extract_text_delta` returns `None` for every line, so `tail_until_done` yields zero text, `is_any_output_yielded` stays `False`, and `stream_output` raises `no_output_error_subject` → "claude exited without producing output". The agent actually ran fine; the parser just can't see the text.

## Repro

With `claude` v2.1.114 installed:

```bash
cd /tmp && claude --print --model claude-haiku-4-5 --output-format stream-json --verbose HI
```

Events observed (types only): `['system', 'assistant', 'assistant', 'rate_limit_event', 'result']` — no `stream_event`.

Now through mngr:

```bash
mngr create repro headless_claude --foreground --format jsonl --headless -- \
    --model claude-sonnet-4-5 \
    --dangerously-skip-permissions \
    --output-format stream-json \
    --verbose \
    HELLO
```

Fails with `mngr headless_claude exited 1: Error: claude exited without producing output`, even though tmux capture shows claude ran and responded.

## Workaround (caller-side)

Add `--include-partial-messages` to the agent args. That flag causes `claude --print` to emit the `stream_event` envelope that `extract_text_delta` expects:

```
['system', 'system', 'stream_event', 'stream_event', 'stream_event', ..., 'assistant', 'stream_event', ...]
```

Tested, confirmed, and currently what my bridge does.

## Suggested fix

Make `extract_text_delta` (and `_is_result_event` / `_extract_result_error`) handle both envelopes:

1. The legacy `{"type":"stream_event", "event":{"type":"content_block_delta", "delta":{"type":"text_delta", "text":"..."}}}` wrapper.
2. Top-level Anthropic events:
   - `{"type":"assistant", "message":{"content":[{"type":"text","text":"..."}]}}` — emit `message.content[*].text` where `type == "text"`.
   - `{"type":"result", "is_error":<bool>, "result":"<final text>"}` — still the completion signal.

Sketch:

```python
@pure
def extract_text_delta(line: str) -> str | None:
    try:
        parsed = json.loads(line)
    except (json.JSONDecodeError, ValueError):
        return None

    t = parsed.get("type")
    # Legacy stream_event wrapper
    if t == "stream_event":
        event = parsed.get("event")
        if isinstance(event, dict) and event.get("type") == "content_block_delta":
            delta = event.get("delta")
            if isinstance(delta, dict) and delta.get("type") == "text_delta":
                text = delta.get("text")
                if isinstance(text, str):
                    return text
        return None

    # Current default: top-level assistant event with content blocks
    if t == "assistant":
        message = parsed.get("message")
        if isinstance(message, dict):
            blocks = message.get("content") or []
            texts = []
            for b in blocks:
                if isinstance(b, dict) and b.get("type") == "text":
                    tx = b.get("text")
                    if isinstance(tx, str):
                        texts.append(tx)
            if texts:
                return "".join(texts)

    return None
```

Happy to open a PR if that approach looks right.

## Environment

- `mngr` 0.2.5 (installed via `uv tool install --with 'git+https://github.com/imbue-ai/mngr#subdirectory=libs/mngr_claude' --from 'git+https://github.com/imbue-ai/mngr#subdirectory=libs/mngr' imbue-mngr`)
- `claude` CLI 2.1.114
- Platform: macOS 14.x (Apple Silicon)
- Provider: local

## Related

This is independent of #1408 (shlex-quote bug) — fixing one does not fix the other. Together they were the two blockers preventing me from routing Conductor/HiveMNGR agent execution through `mngr headless_claude`. With both workarounds applied (safe-chars-only prompts + `--include-partial-messages`), end-to-end streaming works.

evgunter:
## Summary

- `claude --output-format stream-json` emits a different envelope by default in current versions (v2.1.114+): a top-level `{"type":"assistant","message":{"content":[{"type":"text","text":...}]}}` per turn, not the `{"type":"stream_event", "event":{"type":"content_block_delta", ...}}` wrapper. The previous parser only recognized the latter, so successful runs surfaced as `claude exited without producing output`.
- Adds `extract_assistant_text` alongside the existing `extract_text_delta` and routes both envelopes through a per-turn dedup flag on `_StreamTailState` so a final `assistant` summary after partial deltas does not double-emit.
- Keeps the `result` event completion / error path untouched. Both `mngr ask` (which hardcodes `--include-partial-messages`) and bare `mngr create ... headless_claude` invocations now work.

## Test plan

- [x] `just test-quick "libs/mngr_claude -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'"` — 411 passed
- [x] Unit tests for the new envelope (assistant-only) and the dedup guard (interleaved partial + assistant; multiple assistant turns)
- [ ] CI

Closes #1411

### DIFF
--- a/libs/mngr_claude/imbue/mngr_claude/headless_claude_agent.py
+++ b/libs/mngr_claude/imbue/mngr_claude/headless_claude_agent.py
@@ -268,15 +268,15 @@ class _StreamTailState(MutableModel):
         # delegate to the module-level extract helpers (their dict-accepting
         # variants) so dict-walking logic lives in one place. The string-
         # accepting public helpers are thin wrappers around the same dict logic.
-        event_type = parsed.get("type")
-
-        if event_type == "stream_event":
-            yield from self._handle_stream_event(parsed)
-            return
-
-        if event_type == "assistant":
-            yield from self._handle_assistant_event(parsed)
-            return
+        match parsed.get("type"):
+            case "stream_event":
+                yield from self._handle_stream_event(parsed)
+            case "assistant":
+                yield from self._handle_assistant_event(parsed)
+            case _:
+                # Other event types (system, user, etc.) carry no text to
+                # surface here and are intentionally skipped.
+                pass
 
     def _handle_stream_event(self, parsed: dict[str, Any]) -> Iterator[str]:
         # message_start (partial stream): begin a new turn. Any deltas for
--- a/libs/mngr_claude/imbue/mngr_claude/headless_claude_agent.py
+++ b/libs/mngr_claude/imbue/mngr_claude/headless_claude_agent.py
@@ -4,6 +4,7 @@ import json
 from collections.abc import Iterator
 from datetime import datetime
 from pathlib import Path
+from typing import Any
 from typing import Callable
 
 from pydantic import Field
@@ -34,6 +35,24 @@ _STARTUP_GRACE_SECONDS: float = 10.0
 
 
 @pure
+def _parse_stream_line(line: str) -> dict[str, Any] | None:
+    """Decode a single stream-json line into a dict.
+
+    Returns the parsed dict on success, or None if the line is not valid
+    JSON or does not decode to a JSON object. Non-JSON lines (blank lines,
+    debug output that claude sometimes leaks to stdout) are expected and
+    silently skipped.
+    """
+    try:
+        parsed = json.loads(line)
+    except (json.JSONDecodeError, ValueError):
+        return None
+    if not isinstance(parsed, dict):
+        return None
+    return parsed
+
+
+@pure
 def extract_text_delta(line: str) -> str | None:
     """Extract text from a stream-json content_block_delta event.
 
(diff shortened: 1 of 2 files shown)


joshalbrecht:
lgtm

PR #1420, opened 2026-04-25
fix ssh bug + document cel filter fields for mngr list

### LINKED ISSUE #1101: Document all filter fields for mngr list in --help and docs
## Problem

Users have no easy way to discover what variables are available in CEL filter
expressions (`mngr list --include/--exclude`). The CLI help just says "Include agents
matching CEL expression" with no mention of available fields. While `list.md` has a
decent "Available Fields" section, it's missing the computed fields (`age`, `runtime`,
`idle`) and has a confusing inconsistency (`host.provider_name` documented but
`host.provider` is what actually works in filters, due to normalization in
`agent_details_to_cel_context`). Plugin fields (`plugin.$PLUGIN_NAME.*`) exist but are
undiscoverable without reading source code.

These same fields are also used in output template strings (for selecting which fields
to display and formatting output), but the documentation doesn't make this clear. The
set of fields available for filtering and for output templating should be the same,
and both use cases need to be documented together.

## Solution

1. Update the `--help` text for `mngr list` to include the full list of available
   fields (agent fields, host fields, computed fields, and plugin fields) directly
   in the help output for `--include`/`--exclude`. Make clear that these fields can
   be used both for filtering (CEL expressions) and for output templates (selecting
   and formatting displayed data). Users should see what's available without leaving
   the CLI.

2. Update `list.md` to document the computed fields (`age`, `runtime`, `idle`),
   fix the `host.provider_name` vs `host.provider` inconsistency, and explicitly
   note that the same fields are available for both filtering and output templating.

---
Original backlog entry:
> it's fucking impossible to know what variables are available for filter expressions (and how to create them). We should improve docs around this
>
> fix which fields are available in output template strings
>     there's probably a bunch of stuff that could easily be expose
>     and we're probably being weirdly inconsistent about some of it

evgunter:
## Summary
- Update `--include`/`--exclude` help text to point users at the existing Available Fields section.
- Document the `age`, `runtime`, and `idle` computed fields, and clarify that they are available in CEL filters and `--sort`.
- Make `host.provider` and `host.provider_name` interchangeable in CEL filters and in `--fields`/`--format` templates (alias plumbing extended through `_get_field_value`, `_get_header_label`, and `_compute_column_widths` so the column header and width lookup also resolve under both names).
- Reorganize the Available Fields section so the three places fields can appear (filters, sort, templates) are listed up front. Correct the `idle_seconds` description to match its actual derivation (computed from activity-time mtimes, not agent-reported).
- **Fix:** the CEL-context `idle` computed field used to read `ssh_activity_time` from the top level of the agent dict, but that field lives under `host`. SSH activity was silently never included, and SSH-only agents got no `idle` value. Now read it from the host dict and include it in the gate, so the docs' claim that `idle` covers user/agent/SSH activity is actually true.
- Add a structural regression test that round-trips a fully-populated `AgentDetails` through `agent_details_to_cel_context` and asserts every declared agent field appears at the top level of the CEL context and every declared host field appears under `context["host"]`. Catches the broader class of bugs where computed-field code reads a field at the wrong level of the dict.

Closes #1101

## Test plan
- [x] `just test-quick "libs/mngr -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'"` (3634 passed, 0 failed)
- [ ] CI offload covers acceptance/release tests

### DIFF
--- a/libs/mngr/docs/commands/primary/list.md
+++ b/libs/mngr/docs/commands/primary/list.md
@@ -27,8 +27,8 @@ mngr list [OPTIONS]
 
 | Name | Type | Description | Default |
 | ---- | ---- | ----------- | ------- |
-| `--include` | text | Include agents matching CEL expression (repeatable) | None |
-| `--exclude` | text | Exclude agents matching CEL expression (repeatable) | None |
+| `--include` | text | Include agents matching CEL expression, repeatable (e.g. 'state == "RUNNING"'); see 'Available Fields' below | None |
+| `--exclude` | text | Exclude agents matching CEL expression, repeatable (e.g. 'host.provider == "local"'); see 'Available Fields' below | None |
 | `--running` | boolean | Show only running agents (alias for --include 'state == "RUNNING"') | `False` |
 | `--stopped` | boolean | Show only stopped agents (alias for --include 'state == "STOPPED"') | `False` |
 | `--archived` | boolean | Show only stopped agents (alias for --include 'has(labels.archived_at)') | `False` |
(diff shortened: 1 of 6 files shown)


joshalbrecht:
lgtm

PR #1426, opened 2026-04-25
in mngr transcript, show stop hook output as tool result rather than user message

### LINKED ISSUE #1098: Reclassify stop hook outputs as tool messages in mngr transcript
## Problem

When viewing a transcript via `mngr transcript`, stop hook feedback (e.g., CI results,
push output) appears with the `user:` role because Claude Code embeds stop hook output
into the user message stream. This is misleading -- the output is from an automated
process, not from the human user. Example of current display:

```
[2026-03-30T16:54:31Z] user:
Stop hook feedback:
[./scripts/main_claude_stop_hook.sh]: Everything up-to-date
WARN: No checks found yet (0s elapsed), waiting...
ERROR: Some checks failed
```

## Solution

Update the `common_transcript.sh` converter to detect user messages that contain stop
hook feedback (they follow a recognizable pattern starting with "Stop hook feedback:")
and emit them as `tool_result` events instead of `user_message` events. This way
`mngr transcript` will display them with the `tool:` role, correctly reflecting that
they are automated outputs rather than human input.

---
Original backlog entry:
> using mngr transcript should show stop hook outputs as tool messages, not as user messages
>     example:
>         [2026-03-30T16:54:31Z] user:
>         Stop hook feedback:
>         [./scripts/main_claude_stop_hook.sh]: Everything up-to-date
>         WARN: No checks found yet (0s elapsed), waiting...
>         ERROR: Some checks failed
>         ERROR: The tests have failed for the PR that was created by this script!
>         ERROR: Use the gh tool to inspect the remote test results for this branch and see what failed.
>         ERROR: Note that you MUST identify the issue and fix it locally before trying again!
>         ERROR: NEVER just re-trigger the pipeline!
>         ERROR: NEVER fix timeouts by increasing them! Instead, make things faster or increase parallelism.
>         ERROR: If it is impossible to fix the test, tell the user and say that you failed.
>         ERROR: Otherwise, once you have understood and fixed the issue, you can simply commit to try again.
>         ERROR: PR/CI hook failed -- go fix the CI failures first.
>         ERROR: If autofix has run, check .reviewer/outputs/autofix/issues/*.jsonl for identified issues.
>     we don't want that--it makes more sense to think about it as a tool call

evgunter:
## Summary
- `mngr transcript` was rendering Claude Code stop hook output under the `user:` role because Claude embeds it into the user-message stream.
- Detect text starting with `Stop hook feedback:` in `common_transcript.sh` and emit a `tool_result` event (`tool_name=stop_hook`) instead of a `user_message`. The existing transcript viewer then displays it under the `tool` role.

Closes #1098

## Test plan
- [x] `just test-quick libs/mngr_claude/imbue/mngr_claude/resources/common_transcript_test.py` (23 passed including 4 new)
- [x] `just test-quick "libs/mngr_claude -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'"` (404 passed)
- [ ] CI


libs/mngr_claude/imbue/mngr_claude/resources/common_transcript.sh
```diff
@@ -217,20 +227,45 @@ def convert():
...
-                    event_id = _make_event_id(uuid, "user")
-                    if event_id not in existing_ids:
-                        text = _extract_text_content(content)
-                        if text:
+                    text = _extract_text_content(content)
+                    if is_meta:
+                        # Framework-injected message (stop hook output, etc.) --
+                        # reclassify as tool_result so it doesn't masquerade as user input.
+                        if text.startswith(_STOP_HOOK_PREFIX):
+                            tool_name = "stop_hook"
```
evgunter: seems fragile but the failure mode is not that bad

evgunter: (could also just get rid of this and lump all "meta" together)

joshalbrecht: I think lump it all together, then LGTM

PR #1428, opened 2026-04-25
replace logger.exception with logger.opt(exception=).error

### LINKED ISSUE #1096: Fix logger.exception to always explicitly bind exceptions via logger.opt(exception=)
## Problem

`logger.exception()` relies on `sys.exc_info()` to capture the current exception, but
this is unreliable in a threaded environment -- the exception context can be cleared or
replaced by another thread between the `except` block and the actual log emission. This
means even "correct" usage inside an `except` block can silently lose the traceback.

Additionally, there are places throughout the codebase where unexpected exceptions are
passed into `logger.error()` format strings (e.g., `logger.error("Failed: {}", exc)`),
which loses the full traceback. When the exception is unexpected (i.e., not a subtype
of `MngrError` / `BaseMngrError`), we should be binding the exception properly so we
get the full traceback in logs.

## Solution

1. Ban all uses of `logger.exception()`. Replace every occurrence with
   `logger.opt(exception=exc).error(msg)`. Add a ratchet (count 0) to prevent future use.

2. Audit all `logger.error` calls that format an exception into the message string.
   For each one, if the exception is unexpected (not a subtype of `MngrError` /
   `BaseMngrError`), convert it to `logger.opt(exception=exc).error(msg)` so the full
   traceback is captured. Expected exceptions (subtypes of `MngrError` / `BaseMngrError`)
   are fine to leave as formatted strings since they represent known error conditions.

3. Update the style guide to document this convention: never use `logger.exception()`;
   always use `logger.opt(exception=exc).error(msg)` for unexpected exceptions;
   `logger.error("...", exc)` is acceptable only for expected exceptions (`MngrError` /
   `BaseMngrError` subtypes).

---
Original backlog entry:
> logger.exception needs a hack to *immediately* bind the exception... or to always have it passed in, and complain if it's not?

evgunter:
## Summary

- `logger.exception()` relies on `sys.exc_info()`, which is unreliable in threaded code -- the exception context can be cleared or replaced by another thread between the `except` block and the log emission, silently losing the traceback. Bind the exception explicitly via `logger.opt(exception=exc).error(msg)` (loguru) or `logger.error(msg, exc_info=exc)` (stdlib) so the traceback is captured deterministically.
- Audit `logger.error` calls that format an unexpected exception into the message string and convert them to bind the exception explicitly so the full traceback is preserved.
- Add a `PREVENT_LOGGER_EXCEPTION` ratchet (snapshot `0` across all projects) and update the style guide to document the convention.

### Going beyond the issue

The issue scopes "expected" exceptions to subtypes of `MngrError` / `BaseMngrError`. This PR also rebinds a handful of `logger.error("...: %s", exc)` calls in `apps/remote_service_connector/app.py` whose catches are external (`SuperTokensSessionError`, `SuperTokensGeneralError`) or builtin (`ValueError`, `KeyError`, `OSError`). They aren't `MngrError` subtypes, so under the issue's strict reading they're "unexpected" and should bind the exception -- but they aren't `MngrError` subtypes via the literal text either, so flagging the broader interpretation here.

Closes #1096

## Test plan

- [x] `just test-quick` on touched projects (`libs/imbue_common`, `libs/mngr`, `apps/minds`, `apps/minds_workspace_server`, `apps/remote_service_connector`) -- all unit tests pass
- [x] `test_prevent_logger_exception` ratchet passes (snapshot `0`) in every project's `test_ratchets.py`
- [x] Style guide example (`scripts/style_guide.py`) recompiles cleanly from `style_guide.md`
- [ ] CI runs the full `just test-offload` and `just test-offload-acceptance`

### DIFF
--- a/apps/minds/imbue/minds/desktop_client/agent_creator.py
+++ b/apps/minds/imbue/minds/desktop_client/agent_creator.py
@@ -614,7 +614,7 @@ class AgentCreator(MutableModel):
                     self._redirect_urls[aid] = redirect_url
 
         except (GitCloneError, GitOperationError, MngrCommandError, ValueError, OSError) as e:
-            logger.error("Failed to create agent {}: {}", agent_id, e)
+            logger.opt(exception=e).error("Failed to create agent {}", agent_id)
             log_queue.put("[minds] ERROR: {}".format(e))
             with self._lock:
                 self._statuses[aid] = AgentCreationStatus.FAILED
--- a/apps/minds/imbue/minds/desktop_client/app.py
+++ b/apps/minds/imbue/minds/desktop_client/app.py
@@ -1838,7 +1838,7 @@ def create_desktop_client(
 
     @app.exception_handler(Exception)
     async def _unhandled_exception_handler(request: Request, exc: Exception) -> Response:
-        logger.error("Unhandled exception on {} {}: {}", request.method, request.url.path, exc, exc_info=exc)
+        logger.opt(exception=exc).error("Unhandled exception on {} {}", request.method, request.url.path)
         return Response(status_code=500, content=f"Internal Server Error: {exc}")
 
     @app.middleware("http")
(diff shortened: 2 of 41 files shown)


joshalbrecht:
lgtm

PR #1429, opened 2026-04-25
normalize hyphens in config field names; fail loudly on unknown

### LINKED ISSUE #1039: Config files with hyphens in field names silently fail -- normalize and validate
## Problem

When a config file (e.g. `.mngr/settings.toml`) uses hyphenated field names like `pass-env` instead of the Python-side `pass_env`, the value is silently ignored. The config object is constructed without it, and no error is raised. This is dangerous -- the user thinks they configured something, but it has no effect.

## Solution

1. **Normalize hyphens to underscores** before constructing config objects. TOML conventionally uses hyphens (`pass-env`), but Python dataclasses use underscores (`pass_env`). We should accept both.
2. **Fail loudly on unknown fields** after normalization. If a field doesn't match any known attribute of the target config object, raise an error rather than silently dropping it.

This needs to be done for all config objects loaded from settings, including:
- Command configurations
- Create templates
- Provider configs
- Agent type configs

The config loading code lives primarily in `libs/mngr/imbue/mngr/config/loader.py` and `libs/mngr/imbue/mngr/config/data_types.py`.

---
Original backlog entry:
> pass-env in a config file silently fails to apply (bad, should at least blow up)

evgunter:
built on https://github.com/imbue-ai/mngr/pull/1436

---

## Summary

TOML conventionally uses hyphens (`pass-env`); Python dataclasses use underscores (`pass_env`). Previously `pass-env` in a config file was silently dropped or produced a confusing error, and unknown parameter names in `[commands.X]` were silently dropped instead of raising. Both halves of the fix are needed.

- Adds `_normalize_field_keys` and applies it at every config-section parse entry point (top-level, providers, agent_types, plugins, retry, logging, commands, create_templates). Collisions (`pass-env` AND `pass_env`) raise.
- `apply_config_defaults` now raises `ConfigParseError` for unknown parameter names in `[commands.X]` instead of silently dropping them. `MNGR_ALLOW_UNKNOWN_CONFIG=1` downgrades to a warning.
- Strict policy (the `MNGR_ALLOW_UNKNOWN_CONFIG` env var) is resolved once via a shared `resolve_strict_from_env()` helper and threaded through `setup_command_context` to both `load_config` and `apply_config_defaults`, instead of being read in two places.

Closes #1039

## Side cleanup: drop the implicit `EDITOR` default for `--pass-host-env`

While adding the strict unknown-param check, the load_config baseline `commands.create.defaults["pass_host_env"]=["EDITOR"]` came into view as awkward layering -- per-command default state living in the cross-cutting `MngrConfig` baseline, where it collided with the new strict check. This PR removes the global default entirely and moves the EDITOR opt-in to where it actually matters: the dev-mngr modal and docker `create_templates` in `.mngr/settings.toml`, plus a recommendation in the modal/docker user guides for end users defining their own remote-provider templates.

**End-user impact**:
- `mngr create --provider local` -- unchanged. EDITOR is already in the user's connecting shell; the auto-pass was inert here.
- `mngr create --provider modal` / `--provider docker` *without* a template -- no longer auto-forwards EDITOR. Users adding their own remote-provider templates should add `pass_host_env = ["EDITOR"]` (now documented in [Using Modal](libs/mngr/docs/concepts/modal_usage.md) and [Using Docker](libs/mngr/docs/concepts/docker_usage.md)).
- Anyone using the dev-mngr `-t modal` / `-t docker` templates in this repo -- unchanged. Both templates now include `pass_host_env = ["EDITOR"]`.
- Append-on-CLI semantics restored for template-based usage: `mngr create -t modal --pass-host-env MY_TOKEN` produces `("MY_TOKEN", "EDITOR")` via the existing tuple-append branch in `apply_config_defaults`.

## Test plan
- [x] New unit tests for hyphen normalization (per-section and top-level)
- [x] New unit tests for unknown-param rejection in `apply_config_defaults` (strict raises; `strict=False` warns)
- [x] New integration tests for the env-var-driven strict policy through `setup_command_context`
- [x] `*_warns_on_unknown_fields_when_not_strict` tests now assert via `model_dump()` that the unknown field doesn't end up on the constructed config object
- [x] `just test-quick libs/mngr` -- 3636 passed
- [ ] CI acceptance/release suite

## Notes
- A pre-existing flaky test (`test_extras_claude_plugin_yes_flag`) was marked `@pytest.mark.flaky` per CLAUDE.md after surfacing during the full-suite run; it was already flaky on main.

### DIFF
--- a/libs/mngr/imbue/mngr/plugins/test_plugin_cli_options.py
+++ b/libs/mngr/imbue/mngr/plugins/test_plugin_cli_options.py
@@ -730,11 +730,7 @@ def test_setup_command_context_does_not_crash_with_plugin_params(
     with _plugin_manager_with_plugin(_PluginWithStringOption()) as pm:
         captured_plugin_params: dict[str, Any] = {}
 
-        # `--pass-host-env` matches the baseline default in load_config which
-        # populates `commands.create.defaults["pass_host_env"]`; the strict
-        # unknown-param check otherwise rejects the baseline.
         @click.command("test-create")
-        @click.option("--pass-host-env", multiple=True, default=())
         @add_common_options
         @click.pass_context
         def test_cmd(ctx: click.Context, **kwargs: Any) -> None:
@@ -763,11 +759,7 @@ def test_setup_command_context_stores_plugin_params_in_ctx_meta(
     with _plugin_manager_with_plugin(_PluginWithMultipleOptions()) as pm:
         captured_plugin_params: dict[str, Any] = {}
 
-        # `--pass-host-env` matches the baseline default in load_config which
-        # populates `commands.create.defaults["pass_host_env"]`; the strict
-        # unknown-param check otherwise rejects the baseline.
         @click.command("test-create")
-        @click.option("--pass-host-env", multiple=True, default=())
         @add_common_options
         @click.pass_context
         def test_cmd(ctx: click.Context, **kwargs: Any) -> None:
--- a/libs/mngr/imbue/mngr/plugins/test_plugin_cli_options.py
+++ b/libs/mngr/imbue/mngr/plugins/test_plugin_cli_options.py
@@ -730,7 +730,11 @@ def test_setup_command_context_does_not_crash_with_plugin_params(
     with _plugin_manager_with_plugin(_PluginWithStringOption()) as pm:
         captured_plugin_params: dict[str, Any] = {}
 
+        # `--pass-host-env` matches the baseline default in load_config which
+        # populates `commands.create.defaults["pass_host_env"]`; the strict
+        # unknown-param check otherwise rejects the baseline.
         @click.command("test-create")
+        @click.option("--pass-host-env", multiple=True, default=())
         @add_common_options
         @click.pass_context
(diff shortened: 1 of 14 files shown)


libs/mngr/imbue/mngr/plugins/test_plugin_cli_options.py
```diff
@@ -730,7 +730,11 @@ def test_setup_command_context_does_not_crash_with_plugin_params(
     with _plugin_manager_with_plugin(_PluginWithStringOption()) as pm:
         captured_plugin_params: dict[str, Any] = {}
 
+        # `--pass-host-env` matches the baseline default in load_config which
+        # populates `commands.create.defaults["pass_host_env"]`; the strict
+        # unknown-param check otherwise rejects the baseline.
         @click.command("test-create")
+        @click.option("--pass-host-env", multiple=True, default=())
```
evgunter: hmmm

joshalbrecht:
lgtm, though I fully expect this to break a few things here and there

PR #1430, opened 2026-04-25
in error message for unknown fields in agent_types config, mention plugin may be missing

### LINKED ISSUE #1073: Improve error message for unknown fields in agent_types config when plugin is missing
## Problem

When a user has a `.mngr/settings.toml` with an `[agent_types.claude]` section containing claude-specific config keys (e.g., `is_fast = true`), but does **not** have the `mngr_claude` plugin installed, the error message is confusing:

```
Unknown fields in agent_types.claude: ['is_fast']. Valid fields: [...]
```

This happens because `get_agent_config_class()` in `agent_config_registry.py` silently falls back to the base `AgentTypeConfig` when the agent type isn't registered (line 34-35). Then `_check_unknown_fields()` in `loader.py` (line 391) rejects the plugin-specific fields against the base config's field set, producing an error that looks like the user made a typo.

## Expected behavior

The error message should hint at the likely root causes. Something like:

> Unknown fields in agent_types.claude: ['is_fast']. Valid fields: [...]. This may be because:
> - The plugin that provides this agent type is not installed
> - This agent type is not defined (check the name for typos)
> - One or more field names are misspelled

## Relevant code

- `libs/mngr/imbue/mngr/config/agent_config_registry.py:28-36` — `get_agent_config_class()` falls back to base `AgentTypeConfig` without signaling that the type was unrecognized
- `libs/mngr/imbue/mngr/config/loader.py:373-395` — `_parse_agent_types()` calls `_check_unknown_fields` but has no way to know the config class was a fallback
- `libs/mngr/imbue/mngr/config/loader.py:282-304` — `_check_unknown_fields()` generates the error message

## Suggested approach

Have `get_agent_config_class()` return a tuple or set a flag indicating whether the returned class was a fallback. Then `_parse_agent_types()` can include the additional hint in the error message when unknown fields are found and the config class was a fallback. Compare with how `_parse_providers()` (line 326-335) already handles this case for providers — it checks `disabled_plugins` and includes a helpful message.

evgunter:
## Summary
- Append a clarifying hint to the unknown-field error in `[agent_types.<name>]` when no config class is registered for the type, listing the most likely causes (missing plugin, undefined type, misspelled field).
- Add `is_agent_config_registered()` helper so the loader can detect the silent fallback to base `AgentTypeConfig` and decide whether to attach the hint.
- Extend `_check_unknown_fields()` with an optional `extra_hint` and pass it from `_parse_agent_types()`. Registered types still get the original message (no false hint when the user really did mistype a field).

Before:
\`\`\`
Unknown fields in agent_types.claude: ['is_fast']. Valid fields: [...]
\`\`\`

After (when the plugin providing \`claude\` is not installed):
\`\`\`
Unknown fields in agent_types.claude: ['is_fast']. Valid fields: [...]
This may be because:
  - The plugin that provides agent type 'claude' is not installed
  - Agent type 'claude' is not defined (check the name for typos)
  - One or more field names are misspelled
\`\`\`

Closes #1073

## Test plan
- [x] Added unit test: hint appears when type is unregistered.
- [x] Added unit test: hint is *not* added when the type is registered (true typo case).
- [x] All existing config tests still pass (320 passed).
- [x] Full mngr fast unit suite passes (3629 passed).
- [ ] CI offload (acceptance/release tests).


joshalbrecht:
lgtm

PR #1431, opened 2026-04-25
in discover_hosts_and_agents, retry if event lines fail model validation

### LINKED ISSUE #1090: Handle discovery event schema changes by regenerating stale data instead of silently dropping
## Problem

Discovery event JSONL files use typed Pydantic models with `extra="forbid"` (`FrozenModel`). If a field is added, removed, or renamed in any discovery event model (`AgentDiscoveryEvent`, `HostDiscoveryEvent`, `FullDiscoverySnapshotEvent`, etc.), historical JSONL lines will fail `model_validate()` and are currently silently dropped (returning `None` from `parse_discovery_event_line`). Since consumers read backwards from the end to find the most recent full snapshot, old data doesn't matter -- but the failure is silent, which hides the problem.

## Solution

1. When `model_validate()` fails in `parse_discovery_event_line` (`libs/mngr/imbue/mngr/api/discovery_events.py:390-403`), raise a new `DiscoverySchemaChanged` exception (or similar) instead of returning `None`. This makes the version mismatch explicit.

2. All consumers of `parse_discovery_event_line` should catch `DiscoverySchemaChanged`, and when it occurs: run a full discovery/list operation to regenerate current data (which will append new events in the current schema to the file), then retry parsing from the end. If parsing fails again after regeneration, re-raise the exception since that indicates a real bug rather than stale data.

3. This approach works because discovery events are a cache -- they can always be regenerated from a full provider scan. The old data in the file is harmless; new events in the current schema will be appended and found first when reading backwards.

---
Original backlog entry:
> (derived from discussion about JSONL event schema evolution for discovery events)

evgunter:
## Open question

The function's documented contract is "return None on resolution failure (caller will fall back to a full scan), with one specific exception that propagates -- `DiscoverySchemaChangedError` on the second parse attempt." After the autofix iterations, every other error path (`OSError`, `BaseMngrError` from `list_agents`/regenerate) returns `None`. Two questions for review:

1. **Is this the right contract?** Returning `None` means errors like provider failures inside `list_agents` are silently swallowed (with a warning log), and the caller's full-scan fallback runs and may surface a clearer error. The alternative is to let those errors propagate so the caller of `discover_hosts_and_agents` sees them directly.
2. **Specifically, the autofix commits in question:**
   - `6e97a9772` Catch OSError on the schema-mismatch retry read -- mirrors the first read's OSError handler.
   - `f826d4934` Catch broader errors when regenerating discovery snapshot -- wraps `_write_unfiltered_full_snapshot` in `(OSError, BaseMngrError)`.
   - `6f31b3f7a` Clarify second-schema-error comment -- already approved.

If the answer to (1) is "errors should propagate", revert `6e97a9772` and `f826d4934`.

## Summary

Closes #1090.

- Introduce `DiscoverySchemaChangedError(BaseMngrError, ValueError)` in `libs/mngr/imbue/mngr/errors.py`. `parse_discovery_event_line` raises it when a recognized discovery event type fails Pydantic validation, instead of letting `pydantic.ValidationError` bubble up to be silently swallowed by callers' broad `except ValueError` clauses.
- `resolve_provider_names_for_identifiers` catches it, **triggers a full provider scan via `_write_unfiltered_full_snapshot` to append fresh schema events**, then **retries parsing once**. A second `DiscoverySchemaChangedError` after the regenerate indicates a real bug rather than stale data, so the exception is re-raised. Signature now takes `mngr_ctx` (not `config`) so the regenerate path has what it needs.
- Streaming consumers (`observe.py`, `agent_manager.py`, `backend_resolver.py`) catch it, log a warning, and skip the offending line. Their upstream `mngr observe --discovery-only` subprocess writes periodic full snapshots (every 10s) in the current schema, which supersede stale lines.
- Unknown event types still return `None` (forward-compat for newer schemas).

## Refactor: replace match dispatch with discriminated union

Originally `parse_discovery_event_line` had a six-arm `match` statement mapping each `DiscoveryEventType` to its model and calling `model_validate`. Replaced with a pydantic discriminated union -- each event class now declares `type: Literal[DiscoveryEventType.X] = DiscoveryEventType.X`, and a single `TypeAdapter` handles dispatch. Matches the existing precedent in `libs/mngr_tutor/imbue/mngr_tutor/data_types.py::StepCheck`. `make_*_event` / `emit_*` constructors no longer pass `type=` since each subclass defaults its own discriminator.

## Test plan

- [x] New unit tests in `libs/mngr/imbue/mngr/api/discovery_events_test.py`:
  - `test_parse_recognized_event_with_missing_field_raises_schema_changed`
  - `test_parse_recognized_event_with_extra_field_raises_schema_changed`
  - `test_resolve_provider_names_recovers_after_schema_mismatch` -- verifies regenerate-and-retry actually appends a fresh `DISCOVERY_FULL` event past the stale line and parses successfully on retry
- [x] `just test-quick libs/mngr` -- 3715 passed
- [x] `just test-quick apps/minds_workspace_server` -- 231 passed
- [x] `just test-quick apps/minds` -- 676 passed
- [x] CI offload tests (`test_destroy_via_stdin` is `@pytest.mark.tmux + @pytest.mark.flaky` so offload retries automatically)

## Notes / caveats

- `test_destroy_via_stdin` was marked `@pytest.mark.flaky` after a 10s pytest-timeout fluke under Modal offload load; the underlying timing fragility deserves its own PR.


joshalbrecht:
Sigh claude--those notes are basically "haha I didn't do what you said", but in fact, I said to do that retry thing like that fore a reason :)

Can you make it actually do the thing?

evgunter:
ugh, sorry for letting that through

evgunter:
tried again but it seems a little fiddly and also i'm not sure how we should handle other kinds of errors here

joshalbrecht:
closer, but it still really looks like it's not actually properly retrying when the old data is detected :(

it also appears to be suppressing real errors in the one place where it did try regenerating...

the core problem is that yes, while the events will eventually be regenerated, unless we *explicitly* trigger that regeneration, there will be a fairly long period where we're just sitting there in some cases, wondering why no agents have shown up.

these changes also conflict with some of the changes here: https://github.com/imbue-ai/mngr/pull/1445

libs/mngr/imbue/mngr/api/discovery_events.py
```diff
@@ -472,23 +483,12 @@ def parse_discovery_event_line(line: str) -> DiscoveryEvent | None:
...
-        case DiscoveryEventType.DISCOVERY_FULL:
-            return FullDiscoverySnapshotEvent.model_validate(data)
-        case DiscoveryEventType.HOST_SSH_INFO:
-            return HostSSHInfoEvent.model_validate(data)
-        case DiscoveryEventType.DISCOVERY_ERROR:
-            return DiscoveryErrorEvent.model_validate(data)
-        case _:
-            return None
+    if event_type not in DiscoveryEventType:
+        return None
```
joshalbrecht, line 487: note to self: don't want forward compat

joshalbrecht:
Superseded by https://github.com/imbue-ai/mngr/pull/1492

PR #1432, opened 2026-04-25
make docker match modal in when snapshots are preserved after destroy

### LINKED ISSUE #1237: mngr destroy of Docker host produces spurious GC errors and warnings
## Summary

Running `mngr destroy -f <agent>` on a Docker agent produces spurious errors and warnings during garbage collection:

```
Destroyed agent: forever@127.0.0.1
Garbage collecting...
Stopping Docker container: host-1352723e210445ac9d2d9b8b19eed518
ERROR: Host not found: host-1352723e210445ac9d2d9b8b19eed518
Garbage collecting... done.
WARNING: Garbage collection completed with 1 error(s)
WARNING:   - Failed to cleanup snapshots for host host-1352723e210445ac9d2d9b8b19eed518: Host not found: host-1352723e210445ac9d2d9b8b19eed518
```

## Root cause

`destroy_host()` in `libs/mngr/imbue/mngr/providers/docker/instance.py` deletes the host record from disk. GC then runs immediately after and tries to look up that same host to clean up snapshots. Since the record is already gone, `get_host()` raises `HostNotFoundError`.

The GC code catches this and reports it as a warning (not fatal), so it's cosmetically annoying but not harmful.

## Suggested fix

Either:
1. Have `gc_snapshots()` silently skip hosts that are no longer found (deleted record = already cleaned up)
2. Or defer host record deletion until after GC has run

evgunter:
## Summary

- Aligns Docker's `destroy_host` with the documented contract (`provider_instance.py:335-342`) and the Modal/Lima pattern: mark host as DESTROYED via `stop_reason` and preserve snapshot records so `gc_snapshots` can age-gate their deletion.
- Moves the full purge (snapshot images, host volume directory, host record) into `delete_host`, which `gc_machines` invokes once a destroyed host has aged past `destroyed_host_persisted_seconds`.
- `discover_hosts` now filters DESTROYED hosts when `include_destroyed=False` so retained snapshots do not leak destroyed hosts into default listings.

This eliminates the spurious `Host not found` warnings that surfaced during post-destroy GC because Docker eagerly deleted the host record before `gc_snapshots` could consume it.

Closes #1237

## Test plan

- [x] Unit tests: `just test-quick libs/mngr` -> 3627 passed
- [x] Docker SDK acceptance tests: `just test-quick libs/mngr/imbue/mngr/providers/docker -m 'docker_sdk and not docker'` -> 4 passed
- [ ] Real-Docker tests (`@pytest.mark.docker`) and release tests run via CI offload
- [ ] Manual verification: `mngr destroy -f <docker-agent>` no longer prints `Host not found` warnings during GC


joshalbrecht:
lgtm, thanks!

PR #1433, opened 2026-04-25
fix bug where worktrees still in use were being destroyed

evgunter:
includes a meaningful change to the logic of `mngr destroy`, but it seems like a correct one 

---

## Summary

The destroy CLI was running `git worktree remove --force` on every agent whose `work_dir` happened to be a worktree, regardless of whether mngr created that worktree. That nuked the directory out from under `--transfer=none` agents, which reuse a pre-existing parent worktree or a user-owned worktree. `--no-gc` did not help: the deletion path runs before GC.

The fix filters queued worktrees post-destroy and only removes one when both:
1. The host's `certified_data.generated_work_dirs` contains the path (mngr created it).
2. No surviving agent on that host still references it as `work_dir`.

Sharing across destroyed agents is deduped, and branch deletion is gated on the matching worktree actually being removed (otherwise `git branch -D` would have failed anyway, since the branch is checked out in the surviving worktree).

The proxy-plugin's `execute_cleanup` path was never broken — it does not call `remove_worktree`, only relies on GC, and GC was already correct.

## Repro that motivated the fix

1. Open a Claude session whose mngr `work_dir` is a mngr-managed worktree.
2. From inside, spawn a second mngr agent rooted at the same worktree via `mngr create proxy-debug-1@.local --type claude --transfer none --no-connect ...`.
3. `uv run mngr destroy proxy-debug-1 --force` (default `--gc=enabled`) — the destroy step prints `Removed worktree: <parent_worktree>` and the original session's shell starts erroring with "Working directory ... no longer exists." `--no-gc` does not help, because the destroy CLI itself runs the worktree removal before GC.

## Test plan

- [x] `just test-quick libs/mngr/imbue/mngr/cli/destroy_test.py libs/mngr/imbue/mngr/cli/test_destroy.py libs/mngr/imbue/mngr/hosts/host_test.py` — 210 passed locally.
- [x] New unit tests in `destroy_test.py` cover `_filter_removable_worktrees` (skip non-generated, skip in-use, keep orphan, dedupe).
- [x] New tmux-marked integration tests in `test_destroy.py`:
  - `test_destroy_transfer_none_keeps_shared_worktree` — parent created via `--transfer=git-worktree`, rider via `--transfer=none` reusing parent's worktree, rider destroyed → parent's worktree still on disk and parent tmux still alive.
  - `test_destroy_transfer_none_standalone_keeps_user_worktree` — user-owned worktree (created via `git worktree add`), `--transfer=none` agent destroyed → user's worktree intact.
- [ ] Full CI run.


joshalbrecht:
lgtm

PR #1444, opened 2026-04-27
Fix unbounded SSE replay buffer in minds-workspace-server

gnguralnick:
## Summary

- Session events emitted by `AgentSessionWatcher` were broadcast with the default `BufferBehavior.STORE`, so the per-agent in-memory replay buffer in `event_queues._event_buffers` grew unboundedly for the agent's lifetime and was replayed in full to every new SSE subscriber. Combined with the frontend's 3s SSE reconnect loop this drove sustained 40-100% CPU and continuous RSS growth (observed: workspace-server 1.74 GiB → 1.96 GiB in ~5 minutes during a normal session).
- Session events are persisted to JSONL on disk and already recoverable via `GET /api/agents/:id/events`, so they must not populate the in-memory buffer. Tag them with `BufferBehavior.IGNORE`. The buffer remains available for plugin-emitted events that need replay.
- Frontend: on transient SSE reconnect, refetch the JSONL snapshot before resubscribing so events that occurred during the disconnect window are recovered from the source of truth. Subscribe to SSE first and buffer arriving deltas until the snapshot lands so events broadcast between the snapshot read and the `EventSource` being registered on the server are not lost. The buffer is drained via `appendEvents` (which dedups by `event_id`).

## Test plan

- [x] New unit tests in `server_test.py`: `_broadcast_session_events` does not populate `_event_buffers`, and still delivers events to live subscribers with `buffer_behavior` stripped from the wire format.
- [x] Existing `event_queues_test.py` (`test_buffer_ignore`, `test_buffer_behavior_stripped_from_delivered_events`, etc.) still pass.
- [x] `just test-quick apps/minds_workspace_server` (231 passed; one pre-existing flaky `test_start_app_watcher` watchdog/macOS thread-join timeout, unrelated; passes on retry).
- [x] `just test-quick apps/minds_workspace_server/imbue/minds_workspace_server/test_ratchets.py` (54 passed).
- [x] Frontend `npm run lint` and `npm test` (vitest: 2 passed).
- [x] `just test-offload` (7667 passed; 2 unrelated/flaky failures: ruff format on `server.py` — fixed in this branch — and `libs/mngr/.../test_destroy_without_remove_created_branch_leaves_branch` which passes on retry).
- [ ] CI acceptance tests

### DIFF
--- a/apps/minds_workspace_server/frontend/src/models/StreamingMessage.ts
+++ b/apps/minds_workspace_server/frontend/src/models/StreamingMessage.ts
@@ -84,9 +84,10 @@ async function reconnectWithSnapshot(agentId: string): Promise<void> {
   connectToStream(agentId);
   try {
     await fetchEvents(agentId);
-  } catch {
+  } catch (error) {
     // Fetch failure is non-fatal; the next SSE error will trigger another
-    // reconnect attempt.
+    // reconnect attempt. Log so the failure is still visible in devtools.
+    console.warn(`Snapshot refetch failed for agent ${agentId} during SSE reconnect`, error);
   } finally {
     const buffered = inFlightSnapshotBuffersByAgent.get(agentId) ?? [];
     inFlightSnapshotBuffersByAgent.delete(agentId);
--- a/apps/minds_workspace_server/frontend/src/models/StreamingMessage.ts
+++ b/apps/minds_workspace_server/frontend/src/models/StreamingMessage.ts
@@ -7,13 +7,19 @@
  */
 
 import { apiUrl } from "../base-path";
-import { appendEvents, type TranscriptEvent } from "./Response";
+import { appendEvents, fetchEvents, type TranscriptEvent } from "./Response";
 
 const activeStreams = new Map<string, EventSource>();
 // Tombstones for agents whose streams were explicitly closed via
 // disconnectFromStream. Used by pending error-triggered reconnect timeouts
 // to distinguish an intentional shutdown from a transient error.
 const explicitlyDisconnectedAgents = new Set<string>();
+// Per-agent buffer that captures SSE events arriving while a reconnect-time
+// snapshot fetch is in flight. The fetch's response REPLACES
+// eventsByAgent[agentId]; without buffering, deltas that landed during the
+// fetch would be lost. Drained via appendEvents (which dedups) once the
+// fetch settles.
+const inFlightSnapshotBuffersByAgent = new Map<string, TranscriptEvent[]>();
 
 export interface StreamingMessage {
   conversationId: string;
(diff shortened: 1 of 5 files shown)


apps/minds_workspace_server/frontend/src/models/StreamingMessage.ts
```diff
@@ -7,13 +7,19 @@
...
+import { appendEvents, fetchEvents, type TranscriptEvent } from "./Response";
 
 const activeStreams = new Map<string, EventSource>();
 // Tombstones for agents whose streams were explicitly closed via
 // disconnectFromStream. Used by pending error-triggered reconnect timeouts
 // to distinguish an intentional shutdown from a transient error.
 const explicitlyDisconnectedAgents = new Set<string>();
+// Per-agent buffer that captures SSE events arriving while a reconnect-time
+// snapshot fetch is in flight. The fetch's response REPLACES
+// eventsByAgent[agentId]; without buffering, deltas that landed during the
```
gnguralnick: don't reference old code

apps/minds_workspace_server/frontend/src/models/StreamingMessage.ts
```diff
@@ -51,13 +62,59 @@ export function connectToStream(agentId: string): void {
...
-          connectToStream(agentId);
+          void reconnectWithSnapshot(agentId);
         }
       }, 3000);
     }
   };
 }
 
+async function reconnectWithSnapshot(agentId: string): Promise<void> {
+  // The server no longer keeps an unbounded in-memory replay buffer of
```
gnguralnick: don't reference old code

apps/minds_workspace_server/frontend/src/models/StreamingMessage.ts
```diff
@@ -51,13 +62,59 @@ export function connectToStream(agentId: string): void {
...
+  // Subscribe to SSE first (and buffer arriving events) so deltas that land
+  // between the snapshot read and the EventSource being registered on the
+  // server are not lost. Once the fetch settles, the buffered events are
+  // re-applied via appendEvents, which dedups by event_id.
+  //
+  // We capture the buffer array by reference (rather than re-reading from
+  // the map in `finally`) so a concurrent reconnectWithSnapshot for the
+  // same agent -- which can happen if the new EventSource also errors and
+  // triggers its own 3s reconnect while this fetch is still in flight --
+  // cannot orphan and silently discard our buffered events. Each
```
gnguralnick: there is more text in comments in this function than there is in code. in general this PR has far to verbose comments

apps/minds_workspace_server/imbue/minds_workspace_server/server.py
```diff
@@ -148,8 +161,7 @@ def _get_or_create_watcher(request: Request, agent_info: AgentInfo) -> AgentSess
         return watchers[agent_info.id]
 
     def on_events(agent_id: str, events: list[dict[str, Any]]) -> None:
-        for event in events:
-            event_queues.broadcast(agent_id, event)
+        _broadcast_session_events(event_queues, agent_id, events)
```
gnguralnick: having a helper function for the sake of adding a parameter to the broadcast call seems unnecessary

joshalbrecht:
lgtm--I'm might slightly skeptical of the entire SSE setup that claude has created, but idk that this is making it dramatically worse

PR #1445, opened 2026-04-27
warn on corrupt jsonl

### LINKED ISSUE #1089: Warn on corrupt JSONL lines that aren't at the end of the file
## Problem

When parsing JSONL event files, malformed lines are silently skipped with only a `logger.trace` message (e.g. `events.py:346`, `discovery_events.py:386`, `discovery_events.py:550`, `observe.py:247`). A truncated last line is expected (partial write in progress), but a corrupt line in the middle of the file represents real data loss that should be surfaced to the user.

## Solution

In all JSONL parsers (`parse_event_line` in `events.py`, `parse_discovery_event_line` in `discovery_events.py`, and the parsing in `observe.py`), change the behavior based on position: if the malformed line is the last line in the file/stream, silently skip it (partial write). If it's followed by more lines, emit a `logger.warning` so the data loss is visible.

For the streaming/tailing code paths where lines arrive one at a time, buffer the last unparseable line and only warn about it if a subsequent valid line arrives (proving it wasn't just a partial write at EOF).

---
Original backlog entry:
> (derived from discussion about TOMLDecodeError/JSONDecodeError handling in event streams)

evgunter:
built on https://github.com/imbue-ai/mngr/pull/1416

## Summary
- Introduce `MalformedJsonLineWarner` (`libs/mngr/imbue/mngr/utils/jsonl_warn.py`): stateful per-session helper that buffers the most recent malformed JSONL line and emits `logger.warning` only when a subsequent non-empty line proves the malformed line was not a partial write at EOF.
- Wire the warner into the file-read and tail loops in `events.py`, `discovery_events.py`, `observe.py`, and `transcript.py` so mid-file corruption surfaces, while a truncated last line stays silent.
- Fix tail-loop partial-write data loss (`split_complete_lines` helper) in `_tail_source_thread_remote` and `_discovery_stream_tail_events_file`, and in `_emit_lines_from_offset` for the synchronous discovery-stream phases.
- Share one warner across all phases of `run_discovery_stream` so a malformed line buffered in phase 1 still surfaces a warning when phase 3 / the tail reads valid data after it.
- Strengthen the silent-decode-error ratchet from #1416 to accept `.warning(...)` / `.error(...)` / `.exception(...)` calls (any receiver, including chained forms) as non-silent, drop the style-guide carve-out claiming silent-skip is fine for non-config sources, and add `logger.warning` at 14 sites (host stores, modal cleanup, gh CLI parsing, mngr discovery subprocess output, etc.).
- Add typed `get_ast_nodes_of_type(file_path, node_type) -> list[T]` helper in `imbue_common.ratchet_testing.core` and drop `assert isinstance(node, ast.X)` boilerplate at 9 ratchet call sites; ratchet floor for `assert isinstance` in `imbue_common` drops 9 → 0.

Closes #1089.

## Test plan
- [x] New unit tests in `libs/mngr/imbue/mngr/utils/jsonl_warn_test.py` cover buffer-then-flush, EOF silence, blank-line non-flush, multi-malformed runs, truncation, source-description rendering, reset behavior, and `split_complete_lines` edge cases.
- [x] Integration tests in `events_test.py`, `discovery_events_test.py`, `observe_test.py`, and `transcript_test.py` verify the wiring: mid-file corruption emits a warning, last-line corruption is silent.
- [x] Regression tests for the partial-write data-loss fix (`test_discovery_stream_tail_preserves_partial_writes`) and the cross-phase warner sharing (`test_emit_lines_from_offset_warns_on_corruption_across_calls`); both confirmed to fail behaviorally before the fix.
- [x] Local `just test-quick libs/mngr libs/imbue_common -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'` -> 3964 passed.
- [x] CI: Unit + Integration tests, test-offload, test-docker all pass on the latest commit; offload-acceptance pending.


libs/mngr/imbue/mngr/api/discovery_events.py
```diff
@@ -491,6 +483,26 @@ def parse_discovery_event_line(line: str) -> DiscoveryEvent | None:
...
 
 
+@pure
+def parse_discovery_event_line(line: str) -> DiscoveryEvent | None:
+    """Parse a single JSONL line into the appropriate discovery event type.
+
+    Returns None if the line cannot be parsed or is not a recognized discovery event.
+    Stateless and silent on malformed JSON; use MalformedJsonLineWarner alongside
+    _discovery_event_from_data when reading multiple lines so corruption is surfaced.
+    """
```
evgunter, line 493: might be dangerous to have this as a function that exists even if it's used properly right now

libs/mngr/imbue/mngr/api/discovery_events.py
```diff
@@ -670,31 +685,69 @@ def _discovery_stream_tail_events_file(
         try:
             if events_path.exists():
                 file_size = events_path.stat().st_size
-                # Handle file truncation (reset to start)
+                # Handle file truncation (reset to start). Drop any malformed
```
evgunter, line 688: the stuff in this file seems like the kind of thing that just inevitably has bugs (autofix caught many of them and added edge case tests but i don't have confidence that it's actually correct now)

apps/minds/imbue/minds/config/data_types.py
```diff
@@ -49,6 +49,6 @@ def parse_agents_from_mngr_output(stdout: str) -> list[dict[str, object]]:
                 data = json.loads(stripped)
                 return list(data.get("agents", []))
             except json.JSONDecodeError:
-                logger.trace("Failed to parse JSON from mngr list output line: {}", stripped[:200])
+                logger.warning("Failed to parse JSON from mngr list output line: {}", stripped[:200])
```
joshalbrecht, line 52: note to myself to add a FIXME here, this looks stupid

libs/mngr/imbue/mngr/api/observe.py
joshalbrecht, line 425: note to self: this whole section seems wrong, at least needs a FIXME

likely needs more thought as well

libs/mngr/imbue/mngr/api/discovery_events.py
joshalbrecht: yes, it is fairly dangerous.

I'll see if I can find a way to make it safer

libs/mngr/imbue/mngr/api/discovery_events.py
joshalbrecht: yeah, it's kind of annoyingly complicated to get this stuff right...

I don't see anything crazy after looking at it for a while, but agree that it's brittle

libs/mngr/imbue/mngr/api/events.py
```diff
@@ -368,23 +358,49 @@ def parse_event_line(line: str, source_hint: str) -> EventRecord | None:
...
+    """Parse a single JSONL line into an EventRecord.
+
+    Returns None if the line cannot be parsed (malformed JSON, missing required fields).
+    Always uses source_hint (derived from the file path) as the authoritative source;
+    if the event JSON contains a different source, it is corrected and the mismatch
+    is recorded in the returned EventRecord's original_source field.
+    Generates a deterministic fallback event_id from the line hash if missing.
+    Stateless and silent on malformed JSON; use MalformedJsonLineWarner alongside
+    _record_from_event_data when reading multiple lines so corruption is surfaced.
+    """
```
joshalbrecht, line 391: note to self: I don't see usages of this in here, but it doesn't seem right--the whole damn point of all of this is to stop swallowing corrupted data, yet this function directly encodes that behavior :(

libs/mngr/imbue/mngr/utils/jsonl_warn.py
```diff
@@ -0,0 +1,92 @@
...
+            return None
+        with self._lock:
+            self._flush_pending_warning()
+            try:
+                data = json.loads(stripped)
+            except json.JSONDecodeError:
+                self._pending_malformed_line = stripped
+                return None
+        if not isinstance(data, dict):
+            return None
```
joshalbrecht, line 48: sighhhh  no claude, stop just trying to hide your suppression of errors

if you get data that isn't a dict, something has gone horribly wrong, and you should NOT just continue on your merry way!

this is probably happening because some warnings are being emitted by things like calls to mngr list or something, but we should fix the warnings, not just ignore them. At the very least, they should be emitted as json entries, so that the consumer can decide what to do with them

I'll leave a fixme here for now I guess

libs/mngr/imbue/mngr/providers/docker/host_store.py
```diff
@@ -176,7 +176,7 @@ def list_persisted_agent_data_for_host(self, host_id: HostId) -> list[dict[str,
                 agent_data = json.loads(content)
                 agent_records.append(agent_data)
             except (json.JSONDecodeError, FileNotFoundError, OSError) as e:
-                logger.trace("Skipped invalid agent record {}: {}", entry.path, e)
+                logger.warning("Skipped invalid agent record {}: {}", entry.path, e)
                 continue
```
joshalbrecht, line 180: I'm fairly skeptical here too--invalid agent records mean that something pretty bad happened, and a warning seems insufficient.

libs/mngr_pi_coding/imbue/mngr_pi_coding/plugin.py
```diff
@@ -102,7 +102,7 @@ def _has_api_credentials_available(
             if auth_data:
                 return True
         except (json.JSONDecodeError, OSError) as e:
-            logger.debug("Could not read auth.json: {}", e)
+            logger.warning("Could not read auth.json: {}", e)
```
joshalbrecht, line 105: I don't see how we could really proceed if this happened...

libs/mngr_tmr/imbue/mngr_tmr/pulling.py
```diff
@@ -70,7 +70,8 @@ def try_read_agent_result(
     try:
         raw = host.read_text_file(result_path)
         return _parse_result_json(raw)
-    except (HostError, FileNotFoundError, OSError, json.JSONDecodeError, KeyError, ValueError):
+    except (HostError, FileNotFoundError, OSError, json.JSONDecodeError, KeyError, ValueError) as e:
+        logger.warning("Could not read remote outcome file {}: {}", result_path, e)
```
joshalbrecht, line 74: seems like this is going to spam a lot of warnings?  (it seems to be expected that it might not work, so maybe should be debug)

libs/mngr_vps_docker/imbue/mngr_vps_docker/host_store.py
```diff
@@ -265,7 +265,7 @@ def list_all_host_records_with_agents(
                     agent_data = json.loads(content)
                     agent_data_by_host_id.setdefault(host_id, []).append(agent_data)
                 except json.JSONDecodeError as e:
-                    logger.trace("Skipped invalid agent record {}: {}", file_path, e)
+                    logger.warning("Skipped invalid agent record {}: {}", file_path, e)
```
joshalbrecht, line 268: same comment as other places

libs/mngr_vps_docker/imbue/mngr_vps_docker/host_store.py
```diff
@@ -315,7 +315,7 @@ def _parse_batched_json_files(self, output: str) -> list[dict[str, Any]]:
             try:
                 results.append(json.loads(content))
             except json.JSONDecodeError as e:
-                logger.trace("Skipped invalid JSON file {}: {}", file_path, e)
+                logger.warning("Skipped invalid JSON file {}: {}", file_path, e)
```
joshalbrecht, line 318: also seems really bad

joshalbrecht:
lots of good stuff in here, but I need to at least add a bunch of FIXMEs. I might want an agent to continue a bit on this work and see if it can actually clean up a few of those cases rather than just adding FIXMEs for some of the more important ones...

libs/mngr/imbue/mngr/api/discovery_events.py
joshalbrecht: ok, fixed in my current branch

apps/minds/imbue/minds/config/data_types.py
joshalbrecht: updated

libs/mngr/imbue/mngr/api/observe.py
joshalbrecht: updated

libs/mngr/imbue/mngr/api/events.py
joshalbrecht: updated

libs/mngr/imbue/mngr/utils/jsonl_warn.py
joshalbrecht: updated

libs/mngr/imbue/mngr/providers/docker/host_store.py
joshalbrecht: added fixme

libs/mngr_vps_docker/imbue/mngr_vps_docker/host_store.py
joshalbrecht: added fixme

libs/mngr_pi_coding/imbue/mngr_pi_coding/plugin.py
joshalbrecht: updated

libs/mngr_tmr/imbue/mngr_tmr/pulling.py
joshalbrecht: updated

libs/mngr_vps_docker/imbue/mngr_vps_docker/host_store.py
joshalbrecht: updated

joshalbrecht:
Superseded by https://github.com/imbue-ai/mngr/pull/1492

PR #1447, opened 2026-04-27
Permissions

hynek-urban:
- Add the `LatchkeyPermissionRequestEvent` class.
- Add an endpoint to the minds workspace server that the agent can use to log Latchkey permission request events.
- On the client desktop side, add a permission handler for Latchkey permission requests and serve it behind the `/requests/` endpoints.
- Make the necessary frontend adjustments so that users can approve / deny Latchkey permission requests in the desktop client app.
- Make sure user decisions actually update the Latchkey configuration and wake up the agent that waits for the request.

Accompanied by the https://github.com/imbue-ai/forever-claude-template/pull/28 PR on the forever-claude-template side.

permissions


apps/minds/electron/main.js
```diff
@@ -849,12 +849,28 @@ function handleChromeSSEEvent(evt) {
   } else if (evt.type === 'auth_status') {
     latestChromeState.authStatus = evt;
   } else if (evt.type === 'request_count') {
-    latestChromeState.requestCount = evt.count || 0;
-    // Requests panel HTML is static at load time. Refresh any visible panels
-    // so their cards reflect the new pending list. Debounced per-bundle so
-    // a burst of count changes coalesces into one reload per panel.
+    const prevCount = latestChromeState.requestCount;
```
hynek-urban, line 852: I don't think this part still works very well but I'll address it in a follow-up.

joshalbrecht: yeah, it looked busted in your demo, but I'm fine with that in a followup

joshalbrecht:
lgtm, thanks a bunch for this hynek!

PR #1448, opened 2026-04-27
default is_checked_by_group=True in run_process_in_background

### LINKED ISSUE #1087: Add periodic `check()` enforcement for background processes in ConcurrencyGroup
## Problem

When a process is launched via `ConcurrencyGroup.run_process_in_background()`, the caller is responsible for periodically calling `check()` to detect if it crashed. If they never check, errors are silently lost. This has already caused real debugging pain (the backlog entry notes "would have saved me an afternoon").

The `run_process_in_background` call in `apps/minds/imbue/minds/forwarding_server/backend_resolver.py` is specifically called out as needing checks, but the issue applies to all 7 call sites across the codebase (`observe.py`, `notifications/cli.py`, `pair/api.py`, `kanpan/fetcher.py`, etc.).

## Solution

Add a `check_interval` parameter to `run_process_in_background` that defaults to 60 seconds. If `check()` is not called on the process within that window, the ConcurrencyGroup should tear down the entire process (not just log a warning -- the whole point of ConcurrencyGroup is safety). Callers should schedule their checks at roughly half the interval (e.g. every 30 seconds for the 60-second default). To opt out, set `check_interval` to `math.inf`.

After adding the enforcement mechanism, audit all existing `run_process_in_background` call sites and add periodic `check()` calls to each one, starting with `backend_resolver.py`.

---
Original backlog entry:
> concurrency group: realistically, you *must* call "check()" every X time on each given process launched via background process, otherwise your program is not correct (you're missing any errors that it threw)
>     would have saved me an afternoon already :-P
>     run_process_in_background in backend_resolver need to be checked (and I guess the rest too, but at least that one)

evgunter:
seems spooky and like it has lots of potential to break things

---

## Summary
- Flip the default of `is_checked_by_group` to `True` in `ConcurrencyGroup.run_process_in_background` so background process failures are surfaced by default. Callers that intentionally fire-and-forget can still opt out with `is_checked_by_group=False`.

Per issue #1087, processes launched in the background were silently dropping errors when callers forgot to call `check()`. Rather than the proposed `check_interval` enforcement, this just inverts the safe default.

Closes #1087.

## Test plan
- [ ] CI: `just test-offload` (full suite — relies on CI; flipping the default may surface failures at call sites that previously ignored bad exit codes)
- [x] Local: `just test-quick libs/concurrency_group` (155 passed)


joshalbrecht:
lgtm, but assigning to @hynek-urban to merge this and see if it breaks anything in minds!  (bc he has context on both)

hynek-urban:
I tried a few scenarios in minds and everything seems to work fine so far. Merging.

PR #1450, opened 2026-04-27
add plugin install hints to various errors, checked against catalog of first-party plugins

evgunter:
implementation is a little weird but seems low-stakes

---

## Summary

When a user references a name that isn't registered locally — e.g. `mngr create --provider lima`, `mngr wait` without `imbue-mngr-wait` installed, or `mngr create --agent claude` without `imbue-mngr-claude` installed — the CLI used to surface a bare "unknown ..." error with no actionable next step. This PR routes those error paths through `plugin_catalog.get_plugin_install_hint`, so the message names the actual `imbue-mngr-<name>` package to install when the missing name matches a cataloged plugin, and falls back to a generic "not a known mngr plugin" prompt otherwise (so we don't fabricate package names).

Three call sites covered:
- `UnknownBackendError` (provider backends, e.g. `--provider lima`)
- `UnknownAgentTypeError` (new — agent types resolved via `get_agent_class`, e.g. `--agent claude`)
- `DefaultCommandGroup.resolve_command` (top-level CLI commands, e.g. `mngr wait`)

Other changes:
- Catalog now contains `lima` and `vultr` rows so provider hints work for them. New `LimaSignalCheck` (`limactl --version`).
- `CatalogEntry.cli_command_names` lets `notifications` (entry-point) map to the `notify` CLI command for the unknown-command hint path.
- `_load_backends`'s prior modal-only filter is generalized into `_REMOTE_BACKEND_NAMES = {"modal", "vultr"}` so test harnesses keep skipping credential-dependent backends. Lima is intentionally excluded because its registration is side-effect free (limactl checks deferred to first use).
- The existing duplicate hint in `config/loader.py` now delegates to `get_plugin_install_hint` so the two paths agree.

## Test plan
- [x] `just test-quick "libs/mngr -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'"` — 3636 passed
- [x] `just test-quick "libs/mngr_modal libs/mngr_lima libs/mngr_vultr libs/mngr_vps_docker -m 'not modal and not docker and not docker_sdk and not acceptance and not release and not tmux'"` — 696 passed
- [ ] Offload + acceptance via CI

### DIFF
--- a/libs/mngr/imbue/mngr/api/providers_test.py
+++ b/libs/mngr/imbue/mngr/api/providers_test.py
@@ -37,13 +37,21 @@ def test_get_unknown_backend_raises() -> None:
     assert "nonexistent" in str(exc_info.value)
 
 
-def test_get_unknown_backend_includes_plugin_install_hint() -> None:
-    """Unknown backend errors should suggest installing the matching plugin."""
+def test_get_unknown_backend_includes_plugin_install_hint_for_cataloged_backend() -> None:
+    """Unknown backend errors should name the actual package for cataloged plugins."""
     with pytest.raises(UnknownBackendError) as exc_info:
-        get_backend("lima")
+        get_backend("modal")
     formatted = exc_info.value.format_message()
-    assert "imbue-mngr-lima" in formatted
-    assert "plugin" in formatted
+    assert "imbue-mngr-modal" in formatted
+
+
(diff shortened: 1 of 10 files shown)


joshalbrecht:
lgtm

Issue #473, opened 2026-02-17
megathread for various issues I hit against 05fed4d4


boweiliu:
1. [nit] which mngr returns weird output, bash script instead of a file path. to be fair which nvm does as well, but unexpected.
2. [err] mngr --version fails with RuntimeError: 'imbue' is not installed. Try passing 'package_name' instead.   this is after a fresh install
3. [nit] mngr --help is offensively slow : 
time mngr --help
real    0m0.522s
user    0m0.432s
sys     0m0.073s

4.  [doc] This doc is false (https://github.com/imbue-ai/mngr/blob/main/libs/mngr/README.md)
mngr is very simple to use:

mngr                  # launch claude on Modal (defaults: command=create, agent=claude, provider=modal, project=current dir)
but when i run mngr it just gives me the help output

5. [doc] this doc is also false (https://github.com/imbue-ai/mngr/blob/main/libs/mngr/README.md)
mngr --no-connect --initial-message "Speed up one of my tests and make a PR on github"
i get Error: No such option: --no-connect

6. [doc] when i used mngr c  , i see that it used claude, opus4.6[1m]   -- this is unexpected according to what i read from --help. i expected my claude defautl which is claude4.6

7. [doc] when i used mngr c --help and looked at examples, this didnt work
EXAMPLES
       Create an agent locally in a new git worktree (default)
           $ mngr create my-agent

       Create an agent in a Docker container
           $ mngr create my-agent --in docker
but
bowei@Boweis-MacBook-Pro mngr_repo % mngr create my-agent --in docker 
Error: Unknown provider backend: docker. Registered backends: local, modal, ssh
8. [bug] when i used mngr create my-agent --in-place  but i already had a "my-agent" tmux from a prior command, mngr claimed success but didn't recreate. expected: it would error OR expected: it would create a newly named my-agent-2 and operate with that one. observed: nothing happened and it said it made an agent but actually it just pointed at the old agent.
bowei@Boweis-MacBook-Pro ~ % mngr create my-agent --in-place
WARNING: No API credentials detected for Claude Code. The agent may fail to start.
Provide credentials via one of:
  - Set ANTHROPIC_API_KEY environment variable (use --pass-env ANTHROPIC_API_KEY)
  - Run 'claude login' to create ~/.claude/.credentials.json
Starting agent my-agent ...
Connecting to agent...
Error: You're already in a tmux session. You can attach to the agent with:
[...]

PR #1452, opened 2026-04-27
update synopses + ratchet to prevent drift

evgunter:
addresses https://github.com/imbue-ai/mngr/issues/473 part 11

---

## Summary
- Add `--message`, `--message-file`, and `--edit-message` to the hand-maintained `mngr create` synopsis (they were missing).
- Add a unit test that walks `create`'s click params and asserts every non-Common, non-plugin-injected flag either appears in the synopsis or is in an explicit opt-out set, so future flag additions/renames surface as a test failure rather than silently drifting.

## Test plan
- [x] `just test-quick libs/mngr/imbue/mngr/cli/help_formatter_test.py` (40 passed)
- [x] `uv run mngr create --help` shows the new flags in the synopsis
- [ ] CI


libs/mngr/imbue/mngr/cli/help_formatter_test.py
```diff
@@ -547,6 +550,251 @@ def test_commands_with_aliases_have_aliases_in_synopsis() -> None:
...
+            "--worktree-base-folder",
+            "--pass-env",
+            "--host-env",
+            "--host-env-file",
+            "--pass-host-env",
+            "--activity-sources",
+            "--reconnect",
+            "--attach-command",
+            "--connect-command",
+            "--yes",
```
evgunter: some of these should be in the synopsis probably

joshalbrecht:
lgtm

PR #1454, opened 2026-04-27
clean up created branch when agent creation fails

evgunter:
## Summary

- If agent creation failed after the work_dir's branch was created (e.g. a provisioning hook raised), the branch was orphaned in the source repo. The destroy path already handles cleanup correctly, but the create path had no symmetric cleanup.
- Wrap the post-work-dir-creation code in try/finally so that on failure we remove the worktree and delete the branch.
- Reuses the destroy-time safety check: only delete when `create_agent_work_dir` actually created a new branch (`created_branch_name is not None`). Agents attached to a pre-existing branch never trigger a deletion.
- Shared branch-deletion logic moved to ``git_utils.delete_git_branch``.

## Test plan

- [x] New test ``test_worktree_branch_is_cleaned_up_when_create_fails`` verifies the branch is removed after a hook-injected failure.
- [x] New test ``test_preexisting_branch_is_preserved_when_create_fails`` verifies a reused branch is **not** deleted on failure.
- [x] ``just test-quick libs/mngr/imbue/mngr/api/test_create.py``: 25/25 pass.
- [x] ``just test-offload``: 7669/7669 pass.


joshalbrecht:
lgtm, and thanks, that constantly annoyed me

PR #1461, opened 2026-04-28
mngr list: per-provider --on-error isolation across batch and streaming

weishi-imbue:
## Symptom
A single failing provider (e.g. Modal with no credentials on a fresh install) hid agents from working providers, regardless of `--on-error`. Repro: `MODAL_CONFIG_PATH=/dev/null uv run mngr list`.

## Fix
Provider construction now happens **inside each provider's own thread**, alongside discovery, with one `try/except` honoring `--on-error`:

- **ABORT** (default): first failure raises after streamed rows, exit 1.
- **CONTINUE**: each failure becomes a `ProviderErrorInfo` on `result.errors`; listing proceeds with the remaining providers, exit 1 if any errors.

This applies to both **batch** (forced by `--sort` or `--format jsonl`) and **streaming** modes. Modal needs no special-casing — it goes through the same per-provider try/except as Docker/Lima/Vultr.

`--on-error` is **orthogonal to streaming vs batch**. An abort during streaming surfaces as exit 1; downstream consumers (with `pipefail`) tear down the pipe before acting on partial output. Atomicity comes from the exit code, not from buffering.

Errors are surfaced per format:
- **human**: stderr (`logger.error`/`logger.warning`).
- **jsonl**: a single `{"event":"error", ...}` line on stdout via the on_error callback.
- **json**: populated `errors` field in the single JSON document. Always valid JSON.

## Behavior matrix
--on-error abort returns exit code 1, it doesn't enforce batch. So it's possible user see partial output.

### With no Modal credentials

| cmd | mode | behavior | exit code |
|---|---|---|---|
| `MODAL_CONFIG_PATH=/dev/null uv run mngr list` | streaming | partial: 9 lima rows on stdout, Modal `MngrError` on stderr | 1 |
| `MODAL_CONFIG_PATH=/dev/null uv run mngr list --on-error continue` | streaming | partial: header + 9 lima rows on stdout, error on stderr | 1 |
| `MODAL_CONFIG_PATH=/dev/null uv run mngr list --sort name` | batch | error out (no agents on stdout) | 1 |
| `MODAL_CONFIG_PATH=/dev/null uv run mngr list --sort name --on-error continue` | batch | partial: 9 sorted lima rows on stdout | 1 |
| `MODAL_CONFIG_PATH=/dev/null uv run mngr list --format jsonl` | batch | error out | 1 |
| `MODAL_CONFIG_PATH=/dev/null uv run mngr list --format jsonl --on-error continue` | batch | partial: 1 `event="error"` line + 9 agent jsonl rows | 1 |
| `MODAL_CONFIG_PATH=/dev/null uv run mngr list --sort name --format jsonl` | batch | error out | 1 |
| `MODAL_CONFIG_PATH=/dev/null uv run mngr list --sort name --format jsonl --on-error continue` | batch | partial, sorted jsonl (1 error line + 9 agents) | 1 |
| `MODAL_CONFIG_PATH=/dev/null uv run mngr list --format '{name}'` | streaming | partial: 9 agent names | 1 |
| `MODAL_CONFIG_PATH=/dev/null uv run mngr list --format '{name}' --on-error continue` | streaming | partial: 9 names | 1 |
| `MODAL_CONFIG_PATH=/dev/null uv run mngr list --sort name --format '{name}'` | batch | error out | 1 |
| `MODAL_CONFIG_PATH=/dev/null uv run mngr list --sort name --format '{name}' --on-error continue` | batch | partial: 9 sorted names | 1 |
| `MODAL_CONFIG_PATH=/dev/null uv run mngr list --format json` | batch | error out (empty stdout) | 1 |
| `MODAL_CONFIG_PATH=/dev/null uv run mngr list --format json --on-error continue` | batch | one valid JSON doc: `{"agents": [9 lima], "errors": [Modal MngrError]}` | 1 |

### With real Modal credentials

| cmd | mode | behavior | exit code |
|---|---|---|---|
| `uv run mngr list` | streaming | full results | 0 |
| `uv run mngr list --on-error continue` | streaming | full results | 0 |
| `uv run mngr list --sort name` | batch | full results, sorted | 0 |
| `uv run mngr list --sort name --on-error continue` | batch | full results, sorted | 0 |
| `uv run mngr list --format jsonl` | batch | full results (jsonl) | 0 |
| `uv run mngr list --format jsonl --on-error continue` | batch | full results (jsonl) | 0 |
| `uv run mngr list --sort name --format jsonl` | batch | full results, sorted (jsonl) | 0 |
| `uv run mngr list --sort name --format jsonl --on-error continue` | batch | full results, sorted (jsonl) | 0 |
| `uv run mngr list --format '{name}'` | streaming | full results, names only | 0 |
| `uv run mngr list --format '{name}' --on-error continue` | streaming | full results, names only | 0 |
| `uv run mngr list --sort name --format '{name}'` | batch | full results, sorted names | 0 |
| `uv run mngr list --sort name --format '{name}' --on-error continue` | batch | full results, sorted names | 0 |
| `uv run mngr list --format json` | batch | full results, single valid JSON doc | 0 |
| `uv run mngr list --format json --on-error continue` | batch | full results, single valid JSON doc | 0 |


libs/mngr_modal/imbue/mngr_modal/backend.py
```diff
@@ -490,6 +503,12 @@ def build_provider_instance(
             logger.warning("Truncating Modal app name to {} characters: {}", max_app_name_length, app_name)
         app_name = truncate_modal_name(app_name, max_length=max_app_name_length)
 
+        # Without credentials, raise early to avoid Modal SDK side effects.
```
WeishiZ: The caller (get_all_provider_instances) catches this per-provider, and skips Modal, allowing other providers to continue.

joshalbrecht:
I don't think it is a good idea (in general) to hide failures.

If you don't want the modal provider, then disable the modal provider plugin!  Or don't install it in the first place. The latter is the default, and the former is as easy as "--disable-plugin modal"

there is already an "--on-error (ABORT|CONTINUE)" flag for mngr list--that defines what should happen when there are errors

I believe it defaults to ABORT, which I think is reasonable--silently swallowing errors (even if you print an error) is just too annoying as a default behavior. The error messages can (and should) teach people about this mechanism, but you just really don't want a program continuing to try to do something when it really should fail

the issue is that mngr is meant to be programmatic and compositional

imagine you run:

```mngr list --format jsonl | grep "my-string"```

and it's like "yup, only one instance!"

so then you do this:

```mngr list --format jsonl | grep "my-string" | jq ".id" | mngr destroy -f -```

but this time, actually there WOULD have been more things that matched it, and now you end up destroying something that just transiently wasn't visible above!  That's super bad

File: libs/mngr/imbue/mngr/api/providers.py [hunk 2]
```diff
@@ -161,7 +164,10 @@ def get_all_provider_instances(
             continue
         if backend_name not in seen_names:
             provider_name = ProviderInstanceName(backend_name)
-            providers.append(get_provider_instance(provider_name, mngr_ctx))
+            try:
+                providers.append(get_provider_instance(provider_name, mngr_ctx))
+            except MngrError as e:
+                logger.warning(e)
```
joshalbrecht: this is the behavior that I am saying is bad, see my comment

libs/mngr_modal/imbue/mngr_modal/backend.py
joshalbrecht: I think this check itself literally doesnt work, because I believe you can define modal credentials in other ways (env vars, OIDC, maybe more in the future or even now)

joshalbrecht:
What you COULD do, instead, is update modal so that it warns like vultr, but in the same way (at the level of listing, where we do "on error abort or warn as configured") 

It would be nice to make all of the providers work in a consistent way (raising the same class of error in the same cases, so that the outer behavior worked as it should / as it is documented)

libs/mngr/imbue/mngr/api/providers.py
joshalbrecht: basically, these have to be gated on the value of --on-error

libs/mngr_modal/imbue/mngr_modal/backend.py
joshalbrecht: better, in python, to just do the thing and fail, than to first check if it is safe (race conditions mean that it always can fail anyway, so you end up just kind of pointlessly handling both)

simpler to just try and fail, and handle that correctly. There are occasionally performance reasons not to do that, but this isn't such a case

libs/mngr/imbue/mngr/utils/test_ratchets.py
```diff
@@ -54,7 +54,7 @@ def test_prevent_bare_except() -> None:
 
 
 def test_prevent_broad_exception_catch() -> None:
-    rc.check_broad_exception_catch(_DIR, snapshot(7))
+    rc.check_broad_exception_catch(_DIR, snapshot(8))
```
WeishiZ: Added a broad catch in `_construct_and_discover_for_provider`. It follows the same pattern as in its sibling methods today: 

- _construct_discover_and_emit_for_provider: catch Exception, wrap into MngrError
- (new) _construct_and_discover_for_provider: catch Exception, wrap into MngrError

libs/mngr/imbue/mngr/api/list.py
```diff
@@ -381,13 +477,13 @@ def _discover_and_emit_details_for_provider(
             if isinstance(e, MngrError):
                 raise
             raise MngrError(str(e)) from e
-        logger.opt(exception=e).error("Error discovering agents for provider {}", provider.name)
+        logger.opt(exception=e).error("Error discovering agents for provider {}", provider_name)
         emit_discovery_error_to_stdout(
```
weishi-imbue, line 481: > emit_discovery_error_to_stdout

Issues: 
1. `mngr list` repeats the error that's alraedy in stderr. 
    1. In Human output format, this repeats the error in terminal display.
    1. In json format, this breaks the valid output json in stdout

Plan to update to use `emit_discovery_error_event`, so `mngr observe` can still surface the error but via events.jsonl.

joshalbrecht:
This is looking much better!

For the jsonl and json outputs, would it be possible to get warnings and errors to both show up as events when we have provider failures?

eg, when "mngr list --on-error continue --format jsonl" runs into a provider that fails, it seems like it would be ideal for the jsonl output stream to include that as a warning. Same thing for errors (should both show up in stderr, log, for being visible, and stdout, json object format, for computational discovery)

(json output would work exactly the same, but errors and warnings would be contained in top-level "errors" and "warnings" keys)

We could do that as a follow up if you wanted, but it would be really nice for ensuring that all of this stuff doesn't swallow errors!  (matters for both mngr list and mngr observe, at the very least, but honestly it's the kind of thing that would be nice across the entire API surface anyway, ie, anything that accepts --format)

PR #1462, opened 2026-04-28
add filters to mngr connect

evgunter:
addresses https://github.com/imbue-ai/mngr/issues/473 point 27

---

## Summary

- Add the unified agent filter flag set (`--include`, `--exclude`, `--running`, `--stopped`, `--archived`, `--active`, `--local`, `--remote`, `--project`, `--label`, `--host-label`) to `mngr connect`.
- Mix `AgentFilterCliOptions` into `ConnectCliOptions`, decorate the command with `@add_agent_filter_options`, and feed the resulting CEL tuples into both `list_agents` calls (interactive selector and the non-interactive most-recent fallback).
- Filter flags only narrow the candidate pool when no explicit agent is given; they are ignored when an agent is specified positionally or via `--agent`.
- Help metadata + examples updated; `see_also` points at `list#filtering` for the full filter reference.

## Test plan

- [x] `just test-quick libs/mngr/imbue/mngr/cli/connect_test.py` (17 passed)
- [x] `just test-quick "libs/mngr/imbue/mngr/cli -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'"` (1289 passed)
- [x] `mngr connect --help` renders the shared Filtering group with all 11 flags
- [ ] CI offload (acceptance + release tests)

### DIFF
--- a/libs/mngr/docs/commands/primary/connect.md
+++ b/libs/mngr/docs/commands/primary/connect.md
@@ -91,7 +91,7 @@ mngr connect [OPTIONS] [AGENT]
 ## See Also
 
 - [mngr create](./create.md) - Create and connect to a new agent
-- [mngr list](./list.md#filtering) - List agents (see its Filtering section for the full flag reference)
+- [mngr list](./list.md) - List agents (full filter flag reference lives here)
 
 ## Examples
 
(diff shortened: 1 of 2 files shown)


joshalbrecht:
lgtm

PR #1474, opened 2026-04-29
configure git so offload can write image-cache notes

evgunter:
## Summary
- Offload writes a git note per built image so the next run can reuse the cached image instead of rebuilding.
- Without git identity on the runner, the write fails with \`WARN Failed to write note: git notes add failed: Author identity unknown\` (seen on run 25128509066).
- Configure \`user.email\`/\`user.name\` on the runner before \`just test-offload\` and \`just test-offload-acceptance\` so the cache actually persists.

## Test plan
- [ ] CI run no longer emits the \"Author identity unknown\" warning during offload prepare.
- [ ] Subsequent CI run reuses the cached image instead of rebuilding from scratch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,6 +75,11 @@ jobs:
       - name: Fetch offload image-cache git notes
         run: git fetch origin 'refs/notes/*:refs/notes/*' || true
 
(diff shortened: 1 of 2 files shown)


joshalbrecht:
lgtm

PR #1475, opened 2026-04-29
use depot.dev as docker builder

evgunter:
## Summary

Adds depot.dev as an opt-in image builder for both Docker provider paths. Selected via a `builder` field on each provider config (`DockerProviderConfig`, `VpsDockerProviderConfig`); the field's default reads `MNGR_USE_DEPOT` env (truthy → `DEPOT`, otherwise → `DOCKER`).

CI sets `MNGR_USE_DEPOT=1` on `test-docker` and `test-docker-release`, so PR + release CI exercise depot. Local dev and any non-CI use default to native `docker build` and need no depot CLI / `DEPOT_TOKEN`. VPS hosts likewise default to docker so users without depot can still create them.

## What's wired

- `libs/mngr/imbue/mngr/providers/docker/instance.py` — `_build_image` runs `depot build --load` when `builder=DEPOT`, else native `docker build` via the same subprocess helper (`_run_docker_creation_command(executable=...)`).
- `libs/mngr_vps_docker/imbue/mngr_vps_docker/docker_over_ssh.py` — `build_image` honors the same toggle. DEPOT path idempotently installs the depot CLI on the VPS (`command -v depot \|\| curl install.sh \| sh`), forwards `DEPOT_TOKEN` (required) and `DEPOT_PROJECT_ID` (optional) over SSH as inline env, then runs `depot build --load`. Trace logs and timeout messages redact `DEPOT_TOKEN`.
- `libs/mngr/imbue/mngr/primitives.py` — new `DockerBuilder(UpperCaseStrEnum)` with `DOCKER` / `DEPOT` (replaces an earlier `Literal["docker", "depot"]` that hit the no-multi-option-Literal ratchet).
- `depot.json` at repo root (project ID `fsjzltqvxq`).
- `.github/workflows/ci.yml` — adds `depot/setup-action@v1`, `DEPOT_TOKEN`, `DEPOT_PROJECT_ID`, and `MNGR_USE_DEPOT=1` to `test-docker` and `test-docker-release`. `test-release` is intentionally NOT flipped (it filters out `(docker or docker_sdk)` tests, so its build path doesn't invoke `_build_image`).
- Unit tests for the env-var-driven default on both configs and for the depot SSH branch (install snippet, env-prefix, optional project-id, missing-token error). The `test_build_image_from_dockerfile` integration test now also runs `docker image inspect` after the build, which both verifies the image landed in the local daemon and satisfies the `@pytest.mark.docker` resource guard under either backend.

## Why

Docker Hub rate-limits and slow rebuilds in tests that build images. depot's authenticated builders dodge the rate limit and the shared layer cache speeds repeat builds.

## Behavior summary

| Caller | Default | Triggered by |
|---|---|---|
| Local `mngr create @.docker` | DOCKER | none — opt-in via `MNGR_USE_DEPOT=1` env or explicit config |
| Local `mngr create @.vps_docker` | DOCKER | same |
| CI `test-docker` | DEPOT | job-level `MNGR_USE_DEPOT=1` |
| CI `test-docker-release` | DEPOT | job-level `MNGR_USE_DEPOT=1` |
| CI `test-release` | (no image builds) | n/a |

When `builder=DEPOT` and `DEPOT_TOKEN` is missing, the VPS path raises `ContainerSetupError` with a clear message rather than silently falling back.

## Test plan

- [x] `test-docker` passes with depot enabled — verified, build logs show `Running depot build with N args`
- [x] `test-offload`, `test-offload-acceptance` pass
- [x] 209 unit tests + 55 ratchets pass locally
- [ ] Manual: `MNGR_USE_DEPOT=1 mngr create my-host@.vps_docker` end-to-end (only validatable with a real VPS — no CI job exercises this path)
- [ ] Manual: confirm second CI run is faster than first (layer cache hit)

## Follow-ups not in this PR

- Docs sweep — `libs/mngr/docs/concepts/docker_usage.md`, `libs/mngr_vps_docker/README.md`, and the `mngr create` CLI help still describe build args as "passed to `docker build`". Functionally accurate (depot is a drop-in replacement for `docker build` syntax) but doesn't mention depot by name or `DEPOT_TOKEN` as a runtime requirement on the VPS path.
- VPS depot CLI install currently happens per-build (idempotent, sub-second when present). Could be moved to VPS provisioning if the cold-build penalty becomes noticeable.


libs/mngr_vps_docker/imbue/mngr_vps_docker/docker_over_ssh.py
```diff
@@ -14,6 +16,30 @@
...
 from imbue.mngr_vps_docker.errors import VpsConnectionError
 
+# Idempotent install: skip if depot already on PATH, otherwise download to
+# /usr/local/bin via depot.dev's official installer. Run once per build (cheap
+# no-op when already present); avoids needing a separate provisioning step.
+_DEPOT_INSTALL_CMD: Final[str] = "command -v depot >/dev/null 2>&1 || curl -fsSL https://depot.dev/install-cli.sh | sh"
+
+# Env-var assignments whose values are secrets and must be redacted before any
+# remote_command string ends up in logs or exception messages.
+_SECRET_ENV_VARS: Final[tuple[str, ...]] = ("DEPOT_TOKEN",)
```
evgunter, line 27: anything else that should go here?

joshalbrecht: Not that I can think of right now, but thanks for flagging!

libs/mngr_vps_docker/imbue/mngr_vps_docker/docker_over_ssh.py
```diff
@@ -254,9 +282,28 @@ def build_image(
...
+        is imported into the local Docker daemon on the VPS for subsequent
+        `docker run`.
+        """
+        depot_token = os.environ.get("DEPOT_TOKEN", "")
+        depot_project_id = os.environ.get("DEPOT_PROJECT_ID", "")
+        if not depot_token:
+            raise ContainerSetupError(
+                "VPS image builds require DEPOT_TOKEN in the agent's environment. "
+                "Set DEPOT_TOKEN (and DEPOT_PROJECT_ID if no depot.json is on the VPS)."
+            )
```
joshalbrecht: this doesn't seem right?

I can, and probably should be able to, make VPS hosts without using depot...

evgunter: ah, yeah. added a config option on docker and vps_docker with the builder to use

joshalbrecht:
thanks for fixing, lgtm!

PR #1478, opened 2026-04-29
single source of truth for which packages should be published 

evgunter:
## Summary

- Replace the 17 hardcoded `Build <pkg>` steps in `.github/workflows/publish.yml` with a single loop driven by `scripts/verify_publish.py --list-package-dirs`.
- The script reuses the existing `PACKAGES` source of truth from `scripts/utils.py`, so a new publishable lib added to `PACKAGES` is now built and uploaded automatically -- no workflow edit required.
- Fixes the silent miss in v0.2.6 (`imbue-mngr-lima`, `imbue-mngr-vps-docker`, `imbue-mngr-vultr` were tagged but never built/uploaded). The next tagged release will pick them up.
- Bonus: `scripts/release.py` now prints an `ACTION REQUIRED` note when a release includes new (never-published) packages, telling the user to register a pending Trusted Publisher at https://pypi.org/manage/account/publishing/ and warning that PyPI's one-pending-publisher-at-a-time limit means the publish workflow will fail and need to be re-run once per new package until each is registered.

## Approach notes

- Picked the single-loop approach over a matrix to keep the publish job a single job with a shared `dist/` (matching `pypa/gh-action-pypi-publish` reading `dist/` once at the end). Per-package `::group::` log markers preserve collapsible per-package logs in the GitHub Actions UI.
- `skip-existing: true` is preserved (load-bearing for re-runs).
- `verify_publish.py` semantics for the existing `--expected-mngr-version` path are unchanged.
- The new-package warning fires from the existing `_confirm_new_packages` interactive prompt in `release.py`, which already detects new packages via the PyPI HEAD check in `_is_published_on_pypi`. Dry-run does not print the warning (it auto-confirms without prompting); only a real release surfaces it.

## Test plan

- [ ] CI passes on this PR.
- [ ] `uv run scripts/verify_publish.py --list-package-dirs` locally prints all 20 `libs/<dir>` paths (verified in development).
- [ ] `uv run scripts/verify_publish.py` (no args) still runs the full version/graph/pin checks unchanged (verified in development).
- [ ] Optional: a `workflow_dispatch` run on this branch confirms the build step iterates over all 20 packages and produces 20 wheel + 20 sdist artifacts in `dist/` before the publish step.
- [ ] Next real release that includes a new package shows the `ACTION REQUIRED` block after confirmation.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -55,8 +55,9 @@ jobs:
       - name: Build all packages
         run: |
           set -euo pipefail
-          pkg_dirs=$(uv run scripts/verify_publish.py --list-package-dirs)
-          for pkg_dir in $pkg_dirs; do
+          pkg_dirs_output=$(uv run scripts/verify_publish.py --list-package-dirs)
+          mapfile -t pkg_dirs <<< "$pkg_dirs_output"
+          for pkg_dir in "${pkg_dirs[@]}"; do
             echo "::group::Build $pkg_dir"
             pyproject-build "$pkg_dir" -o dist/
             echo "::endgroup::"
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -52,56 +52,15 @@ jobs:
             uv run scripts/verify_publish.py
           fi
 
-      - name: Build resource-guards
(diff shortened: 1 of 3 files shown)


joshalbrecht:
lgtm, hurray for for loops!

PR #1480, opened 2026-04-30
suppress missing-key cel filter warnings for schemaless fields

evgunter:
right now filtering by a custom label spams a warning for every agent missing that label. this fixes that problem but is a bit ugly and has the side effect of making it so that `has()` is true for any label

---

## Summary
- Suppresses per-agent `WARNING: Error evaluating exclude filter ... no such member in mapping: 'X'` warnings when filtering on schemaless dict fields (`labels`, `plugin`, `host.tags`, `host.plugin`).
- Adds a `TolerantMapType` (CEL `MapType` subclass returning `None` on missing-key access) plus a `tolerant_dict()` marker so `_convert_to_cel_value` produces tolerant CEL maps for these fields. Other fields stay strict, so genuine typos (e.g. `host.providr`) still surface as warnings.
- Replaces `celpy.json_to_cel` for containers with a small recursive walker so the marker survives at any depth.

Originally reproduced via:
```
mngr kanpan --project mngr --exclude 'labels.mngr_subagent_proxy == "child"'
```
which emitted one warning per agent missing that label.

Trade-off: cel-python's `has(map.key)` macro is implemented as "did the lookup error?", so on tolerant fields `has()` always returns True. Use `field != null` or direct comparison instead. Documented in `TolerantMapType.__doc__` and asserted in tests.

## Test plan
- [x] `just test-quick libs/mngr/imbue/mngr/utils/cel_utils_test.py` (32 passed)
- [x] `just test-quick libs/mngr/imbue/mngr/api/list_test.py` (66 passed)
- [x] `just test-quick "libs/mngr -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'"` (3720 passed; 2 pre-existing `common_opts_test` failures unrelated to this PR — confirmed they fail on `main` as well)
- [ ] CI offload run via `just test-offload` (kicked off via PR checks)
- [ ] Manual end-to-end with the user's installed `mngr` CLI: `mngr kanpan --project mngr --exclude 'labels.mngr_subagent_proxy == "child"'` should emit zero `WARNING: Error evaluating exclude filter` lines, while `--exclude 'host.providr == "x"'` should still warn.


libs/mngr/docs/commands/primary/list.md
```diff
@@ -106,6 +106,11 @@ All agent fields from the "Available Fields" section can be used in filter expre
 - `has(url)` - Agents that have a URL set
 - `has(host.ssh)` - Agents on remote hosts with SSH access
 
+Note: `has()` always returns true for keys under the schemaless fields `labels`,
+`plugin`, `host.tags`, and `host.plugin` (so `has(labels.foo)` is not useful).
+To check whether a key is set on those fields, use `labels.foo != null` or a
+direct comparison such as `labels.foo == "x"`.
```
evgunter: it's possible to avoid this but way more cursed

joshalbrecht:
Hmmm... I don't think we can allow "has(map.key)" to silently always return True (or even to return True when it does not, in fact, have that key). That's just too wrong / too sharp of a corner to really permit.

Can we think of any better alternatives here?

libs/mngr/imbue/mngr/utils/cel_utils.py
```diff
@@ -10,6 +10,39 @@
...
+
+    The canonical CEL idiom for this would be optional-type field selection
+    (`labels.?key`, see cel-spec proposal 246), but cel-python does not yet
+    implement optional types, so we hand-roll this targeted subclass instead.
+    Drop this once cel-python supports `?`-prefixed field selection.
+
+    Caveat: comparisons against `null` do not work as you might expect on a
+    tolerant miss (`labels.X != null` evaluates to BoolType(True) even when X
+    is absent, because the LHS is a CELEvalError, not null). Use `has(field)`
+    as the canonical presence check.
```
evgunter: considerably less bad caveat than before

libs/mngr/imbue/mngr/utils/cel_utils.py
```diff
@@ -10,6 +10,39 @@
...
+    """
+
+    def __getitem__(self, key: Any) -> Any:
+        try:
+            return super().__getitem__(key)
+        except KeyError:
+            return CELEvalError(f"no such member in mapping: {key!r}", KeyError, None)
```
evgunter: apparently cel passes errors around like this internally

evgunter:
not beautiful, but better!

libs/mngr/imbue/mngr/api/list.py
```diff
@@ -35,10 +35,24 @@
...
+# Each is a schemaless dict on AgentDetails / HostDetails; their contents are
+# user- or plugin-supplied, so different agents legitimately have different
+# keys. See `with_tolerant_paths` in cel_utils.
+_AGENT_SCHEMALESS_PATHS: tuple[tuple[str, ...], ...] = (
+    ("labels",),
+    ("plugin",),
+    ("host", "tags"),
+    ("host", "plugin"),
+)
+
```
joshalbrecht: ok, this whole PR is an improvement, but couldn't we compute this constant from the structure of the types?

like, isn't this just "anything that is a Mapping or Dict"?

joshalbrecht:
much better, but curious if we can go the rest of the way! (see above comment)

if I'm wrong about that we can just merge it I guess, but it'd be nice not to have to mess with it later...

libs/mngr/imbue/mngr/api/list.py
evgunter: huh, yes we can! (i incorrectly assumed that there wouldn't be a nice place to distinguish them from pydantic models)

evgunter:
fixed

PR #1482, opened 2026-04-30
Fix subdomain proxy falling back to host loopback

gnguralnick:
## Summary

When a user clicks an agent in the minds sidebar, the URL `agent-<id>.localhost:<port>/...` hits the FastAPI desktop client. The handler looks up the agent's registered `workspace_server` URL via the `BackendResolver` -- for Docker agents this is always `http://localhost:8000` (registered by `forward_port.py` inside the container). That URL is meant to be reached via an SSH tunnel into the container.

The bug: `_get_tunnel_http_client` returns `None` whenever `backend_resolver.get_ssh_info(agent_id)` is `None` (local non-Docker agents, or any Docker agent whose live SSH info isn't in the resolver -- e.g. a stopped container that's still listed). The HTTP handler then did `active_client = tunnel_client or request.app.state.http_client`, falling back to a plain httpx client that dials `http://localhost:8000` on the *host*. If anything else is bound to that port on the host (a local dev server, another app), the proxy quietly served THAT site as the agent's workspace UI.

## Fix

In `_handle_workspace_forward_http` and `_handle_workspace_forward_websocket`: when the registered workspace URL is loopback (`localhost`, `127.0.0.0/8`, `::1`) AND we have no SSH tunnel for the agent, refuse to fall back to a direct host connection.

- HTTP: return 502 with `workspace server unreachable: no SSH tunnel available for agent <id>; refusing to dial host loopback at <url>`.
- WebSocket: log the full detail and close with code 1013 / a short reason (the WS protocol caps reason at 123 bytes).

Non-loopback registered URLs (real network addresses) are still dialed directly -- that's how local non-Docker agents are intended to work.

## Test plan

- [x] `just test-quick "apps/minds/imbue/minds/desktop_client/test_desktop_client.py -k subdomain_forward"` -- 9 passed (3 new + 6 existing).
- [x] `just test-quick apps/minds/imbue/minds/desktop_client` -- 534 passed.
- [ ] CI on this branch.

New tests cover loopback URLs as `http://localhost:8000`, `http://127.0.0.1:8000`, and `http://[::1]:8000` -- each must 502 instead of falling through to the host.


joshalbrecht:
This looks like a good fix, but I'll need to blend it into my current branch (which changes how this forwarding works)

joshalbrecht:
this fix was ostensibly merged into my larger branch, but lmk if it regresses!

PR #1483, opened 2026-05-01
add DENY mode to mngr_claude_subagent_proxy

evgunter:
like the plugin in general, not really ready for use, but a sensible logical chunk of work to check in

---

## Summary

- Adds `mode: "PROXY" | "DENY"` to `[plugins.claude_subagent_proxy]` mngr config (default `PROXY` = original behavior).
- In `DENY` mode, the plugin installs a `PreToolUse:Agent` hook that denies every Task tool call with a short, uniform `permissionDecisionReason` pointing Claude at the `mngr-subagents` skill (provisioned to the agent's `.claude/skills/`). The skill teaches an explicit two-command spawn-and-wait protocol Claude runs itself via Bash:
  ```
  uv run mngr create '<slug>:<parent_cwd>' \
      --type claude --transfer=none --no-ensure-clean --no-connect \
      --label "mngr_claude_subagent_proxy_parent_name=${MNGR_AGENT_NAME:-}" \
      --label "mngr_claude_subagent_proxy_parent_id=${MNGR_AGENT_ID:-}" \
      --env MNGR_SUBAGENT_DEPTH=$((${MNGR_SUBAGENT_DEPTH:-0}+1)) \
      --message-file <prompt_file>
  uv run python -m imbue.mngr_claude_subagent_proxy.subagent_wait <slug>
  ```
  `subagent_wait`'s stdout (an `END_TURN:` reply) is what Claude treats as the Task tool's `tool_result`. The deny hook itself never invokes `mngr create` and generates no per-Task sidefiles.
- `PROXY` and `DENY` modes share **literally the same** `hooks/reap.py` SessionStart hook -- both spawn paths attach the `mngr_claude_subagent_proxy_parent_id=$MNGR_AGENT_ID` label to every child, so the same label query reaps terminal children regardless of mode. The PROXY-only plugin-cache Stop-hook guard is extracted to `hooks/guard_stop_hooks.py` so the shared reap module stays mode-agnostic.
- `DENY` mode otherwise strips the PROXY-only plumbing: no PostToolUse cleanup, no Haiku-relayed mngr-proxy dispatcher, no project `settings.json` Stop-hook check, no env-conditional Stop-hook guarding, no per-Task wait-script generation.
- The DENY skill deliberately does NOT pass `--reuse` on `mngr create`. Slug collisions between concurrent Task calls surface as a hard "agent already exists" error instead of silently merging unrelated work; the skill tells Claude to pick a new slug on collision rather than destroying the existing agent. PROXY's wait-script keeps `--reuse` because its target names derive from the unique `tool_use_id` and the only retries are bot-driven on the same id.
- Renames the plugin's config-registry key from `subagent_proxy` to `claude_subagent_proxy` to match the package directory and pyproject entry-point.

## Follow-up PR (stacked)

[PR #1597 (`mngr/subagent-types`)](https://github.com/imbue-ai/mngr/pull/1597) is stacked on this branch and closes a separate gap: Claude Code's typed-`subagent_type` system prompts (e.g. `imbue-code-guardian:verify-and-fix`) were silently dropped by both PROXY and DENY. That PR adds a shared agent-definition resolver that prepends the resolved body to the proxy prompt file (PROXY) and points at the resolved path in the deny reason (DENY). It targets `mngr/deny-subagents` as its base.

## Test plan

- [x] `just test-quick libs/mngr_claude_subagent_proxy` -- 142 passed (unit + ratchets)
- [x] Release tests in `test_real_claude_subagent.py` exercise `DENY` mode against real Claude: settings-file inspection, end-to-end Task -> deny -> skill flow. The DENY golden-path test (`test_deny_mode_intercepts_task_with_deny_reason`) now asserts that after the parent's first Task call is denied, **a real mngr child agent with `labels.mngr_claude_subagent_proxy_parent_id` matching the parent's id exists** -- proving Claude followed the skill protocol end-to-end, not just that the hook plumbing fired. Verified locally against the live Anthropic API: `just test libs/mngr_claude_subagent_proxy/imbue/mngr_claude_subagent_proxy/test_real_claude_subagent.py::test_deny_mode_intercepts_task_with_deny_reason` passed in ~63s.
- [x] **Manual end-to-end verification**: provisioned a DENY-mode agent with a prompt explicitly asking it to use Task subagents, observed:
  - `PreToolUse:Agent` -> `hooks/deny.py` installed at the agent's `.claude/settings.local.json` after provisioning
  - Claude's first Task call was denied; Claude then loaded the `mngr-subagents` skill (`⏺ Skill(mngr-subagents) ⎿ Successfully loaded skill`)
  - Claude wrote two prompt files to /tmp and spawned **two real mngr subagents in parallel** via the two-command skill protocol (`mngr create ... --label mngr_claude_subagent_proxy_parent_id=<parent_id> --message-file <prompt>` then `python -m imbue.mngr_claude_subagent_proxy.subagent_wait <slug>`). (This manual run pre-dated the `--reuse` removal, so the command Claude ran at the time still included `--reuse`; the asserted-on side-effects are identical with or without it.)
  - Each child carried the `mngr_claude_subagent_proxy_parent_id=<parent_agent_id>` label, matching what `hooks/reap.py` filters on (so terminal children would be reaped on the parent's next SessionStart)
  - Final parent-agent output combined both children's replies into the requested format
- [ ] CI green


libs/mngr_subagent_proxy/README.md
```diff
@@ -182,6 +198,66 @@ are currently running:
...
+agent's `.claude/skills/`, which teaches the explicit two-command
+protocol:
+
+    uv run mngr create '<slug>:<parent_cwd>' \
+        --type claude --transfer=none --no-ensure-clean --no-connect --reuse \
+        --label mngr_subagent_proxy=child \
+        --env MNGR_SUBAGENT_DEPTH=$((${MNGR_SUBAGENT_DEPTH:-0}+1)) \
+        --message-file <prompt_file>
+
+    uv run python -m imbue.mngr_subagent_proxy.subagent_wait <slug>
```
evgunter: little weird

joshalbrecht:
roughly lgtm

I always find tool use permissions to be kind of an antipattern, but whatever in this case (and it makes sense that we'd want this if we wanted people to use it in general)

PR #1485, opened 2026-05-01
fix --adopt-session + add release tests

evgunter:
## Summary

- Fix two bugs that caused `mngr create --adopt-session` to silently spawn a fresh claude session instead of resuming the adopted one (no error surfaced to the user).
- Add a deterministic regression test that runs the assembled shell pipeline against a stub `claude` binary, catching the resume-branch bug in CI.
- Add release tests covering both source-session layouts: vanilla `claude` CLI under `~/.claude/projects/`, and mngr-managed claude agent under per-agent `CLAUDE_CONFIG_DIR`. Both tests plant a UUID secret, adopt the session, and assert the destination agent can recall it.

## Bugs fixed

1. `plugin.py:1502` — resume guard ran `find -name "$MAIN_CLAUDE_SESSION_ID"` against files actually named `<session_id>.jsonl`. The find returned no matches, the `&&` short-circuited, and the silent `||` fallback ran `claude --session-id <fresh agent uuid>` instead of `claude --resume <adopted_id>`.
2. `claude_config.py:457` — `encode_claude_project_dir_name` was missing the `_` → `-` mapping that Claude Code applies, so the adopted JSONL got placed under a project subdir Claude Code never reads (whenever `work_dir` contained an underscore).

The five existing `assemble_command` string-equality tests are updated for the new find pattern. The new pipeline-execution test would have caught the bug at unit-test time; the prior tests only checked the emitted command string, never executed it.

## Test plan

- [x] `just test-quick libs/mngr_claude` — 450 passed
- [x] `just test-quick libs/mngr` — 3906 passed
- [x] Release tests pass locally with `ANTHROPIC_API_KEY` set:
  - `test_adopt_session_brings_context_from_vanilla_claude_session`
  - `test_adopt_session_brings_context_from_mngr_claude_agent_session`
- [ ] CI: standard suite (release tests excluded, run manually per docstring)

### DIFF
--- a/libs/mngr_claude/imbue/mngr_claude/claude_config.py
+++ b/libs/mngr_claude/imbue/mngr_claude/claude_config.py
@@ -453,6 +453,10 @@ def find_project_config(projects: Mapping[str, Any], path: Path) -> dict[str, An
 # Project Directory Encoding
 # =============================================================================
 
+# Matches every character that Claude Code's project-dir encoder maps to '-'
+# (i.e. everything that is not an ASCII alphanumeric or literal '-').
+_NON_DASH_ALNUM_ASCII: Final = re.compile(r"[^A-Za-z0-9-]")
+
 
 @pure
 def encode_claude_project_dir_name(path: Path) -> str:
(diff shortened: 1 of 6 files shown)


joshalbrecht:
lgtm

PR #1487, opened 2026-05-01
centralize resource guard registration so it works correctly when running tests directly from subprojects

evgunter:
## Summary
- Resource guards are now discovered via the `resource_guards` entry-point group rather than re-registered in every project conftest. `register_conftest_hooks()` walks the group automatically, so any project that uses the shared hooks gets the full canonical guard set without project-specific bookkeeping.
- Closes the silent-corruption mode where `pytest` from a subdir like `libs/mngr_kanpan/` would emit `PytestUnknownMarkWarning: Unknown pytest.mark.tmux ...` because the project conftest forgot to call `register_resource_guard("tmux")`.
- Centralizes the registration code per owning library: `imbue.mngr.register_guards`, `imbue.modal_proxy.register_guards`, `imbue.mngr_lima.register_guards`. Each library declares its guards through `pyproject.toml` entry points; new guards just need a registration function and an entry-point line.
- Splits documentation: `libs/resource_guards/README.md` stays library-generic (basic single-package usage, then multi-package usage via the entry-point group); the monorepo-specific roster ("which mngr libraries register what") and "how to add a guard here" instructions live in the docstring of `libs/mngr/imbue/mngr/register_guards.py`.

## Offload checkpoint fix (related)
Adding `[project.entry-points.resource_guards]` to per-package `pyproject.toml` files surfaced an existing offload caching gap: only the root `pyproject.toml` and `uv.lock` were in `build_inputs`, so per-package pyproject changes never invalidated the checkpoint image. Sandboxes ran with editable-install dist-info from the *old* pyprojects; `uv run ty/ruff check` in any package transitively depending on `imbue-mngr` saw the on-disk pyproject differ from dist-info, tried to rebuild, needed `hatchling`, and failed because Modal sandboxes have network disabled. This produced the "Type checker found 0 error(s)" failure across ~38 ratchet tests.

Fix: add `libs/*/pyproject.toml` and `apps/*/pyproject.toml` to `build_inputs` in `offload-modal.toml`, `offload-modal-acceptance.toml`, and `offload-modal-release.toml` so per-package metadata changes regenerate the checkpoint with matching dist-info.

Also surfaces subprocess `stderr` and `returncode` in `check_no_type_errors` / `check_no_ruff_errors` failure messages -- the previous output dropped stderr, which is what kept this looking like a transient flake until the diagnostic landed in CI.

## Test plan
- [x] `cd libs/mngr_kanpan && uv run pytest --markers --no-cov | grep -E "tmux|rsync|unison|modal|docker|lima"` shows every canonical guard mark registered
- [x] Reproduction `cd libs/mngr_kanpan && uv run pytest imbue/mngr_kanpan/test_fetcher_acceptance.py --collect-only` no longer warns
- [x] `libs/resource_guards` tests (99 passed) including new `test_register_all_resource_guards_runs_entry_point_callables`
- [x] `libs/imbue_common` tests (219 passed)
- [x] `libs/mngr_kanpan` tests (385 passed)
- [x] `libs/mngr_notifications` tests (113 passed, 3 skipped)
- [x] `libs/mngr_pair` tests (107 passed)
- [x] `libs/mngr_lima` tests (100 passed)
- [x] `libs/mngr_modal` non-acceptance tests (372 passed)
- [x] `libs/mngr` non-acceptance tests (3657 passed)
- [x] All three `offload-modal*.toml` validate via `offload validate`
- [ ] CI offload run with the new build_inputs (regenerates checkpoint image and unblocks ratchet tests)


joshalbrecht:
lgtm

PR #1489, opened 2026-05-01
include hidden files in ci artifacts

evgunter:
## Summary
- Investigation of recent CI flakiness identified that the @flaky column in the test-results check-run renders correctly on the runner but the underlying flaky_tests_*.txt manifests are stripped from the uploaded artifact, blocking post-hoc debugging.
- Cause: actions/upload-artifact@v4 defaults `include-hidden-files: false`, and offload places the manifests under hidden `.test_output/` subdirs.
- Enable `include-hidden-files: true` on both the unit/integration and acceptance test-results uploads.

This is the first commit on a broader CI-flakes investigation branch. Larger items (real-bug regressions on main, an upstream offload dead-sandbox-retry fix, marking known-flaky tests) are tracked separately.

## Test plan
- [ ] CI passes on this branch
- [ ] After merge, download a `test-results-offload` artifact from any subsequent run and confirm `flaky_tests_*.txt` files are present under per-sandbox `.test_output/` paths

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -441,4 +441,5 @@ jobs:
         with:
           name: test-durations-release-${{ matrix.os }}-${{ matrix.group }}
           path: .test_output/test_durations_*.json
+          include-hidden-files: true
           retention-days: 30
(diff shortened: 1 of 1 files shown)


joshalbrecht:
lgtm

PR #1496, opened 2026-05-04
fix websocket broadcaster queue-full flood and hung-send pin

gnguralnick:
## Summary
- Fix a flood path in `WebSocketBroadcaster`: when a per-client queue is full we no longer keep enqueuing every subsequent broadcast. Eviction is softened to require `_MAX_CONSECUTIVE_QUEUE_FULL = 50` consecutive queue-full broadcasts before disconnecting the client, so a single transient backlog doesn't terminate an otherwise healthy connection.
- Free coroutines wedged in `await websocket.send_text(...)` on a half-dead TCP connection by having the broadcaster cancel the handler task directly: `register()` captures `(task, loop)` from the running asyncio context, and eviction calls `loop.call_soon_threadsafe(task.cancel)` from the background broadcast thread. This avoids needing per-send wall-clock timeouts.
- Guard the cancel against a closed loop so process shutdown doesn't raise — the handler is unreachable anyway once the loop is gone.
- Regression tests: a unit test exercises the cancel-on-eviction path against the broadcaster directly, and the server tests use FastAPI's `TestClient.websocket_connect` rather than hand-rolled WebSocket doubles.
- Net ratchet impact on `test_prevent_asyncio_import`: +2 vs. main (1 → 3), for the legitimate asyncio uses in `register()`/`_disconnect_locked` and the cancellation unit test.

## Test plan
- [x] `just test-quick apps/minds_workspace_server` (local)
- [ ] Full offload CI on the PR


joshalbrecht:
roughly lgtm, thanks for fixing!

PR #1497, opened 2026-05-04
Bypass pyinfra for local-host shell commands to avoid gevent child watcher errors

gnguralnick:
## Summary

`Host._run_shell_command` now bypasses pyinfra's `LocalConnector` for local hosts and runs the command through `ConcurrencyGroup.run_process_to_completion` instead. This avoids a Linux-only crash that broke every multi-host fanout for local hosts.

**The bug.** pyinfra's `LocalConnector` spawns local processes via gevent. gevent attaches a libev `SIGCHLD` child watcher to the thread-local Hub on first use. On Linux those watchers can only attach to the *default* event loop, so any local-host command issued from an `mngr_executor` worker thread raises `TypeError: child watchers are only available on the default loop`. Single-target commands like `mngr capture` run on the main thread and work; everything that fans out (`mngr observe`, `mngr list`, `mngr message`, `mngr interrupt`, `mngr destroy`, `mngr gc`, `mngr discover`) blew up. The user-visible symptom was Minds' `workspace_server` (which spawns `mngr observe --discovery-only`) failing to populate `AgentManager`, causing the Electron UI to 404 on `/api/agents/<id>/screen`.

**The fix.** When `self.is_local` is True, `_run_shell_command` dispatches to a new `_run_shell_command_local` helper that:
- Runs `[_shell_executable, "-c", command.get_raw_value()]` via `mngr_ctx.concurrency_group.run_process_to_completion` (uses standard `subprocess`, not gevent).
- Maps `_timeout` -> subprocess `timeout`, `_env` (merged with `os.environ`) -> `env`, `_chdir` -> `cwd`, `_shell_executable` -> argv[0], `_success_exit_codes` -> tolerated returncodes (default `(0,)`).
- Returns the same `(success, CommandOutput)` shape callers rely on, so no caller changes.
- Raises `NotImplementedError` if `_su_user`/`_sudo`/`_doas` are passed; none of those are used for local hosts in this codebase today.

The pyinfra path is preserved verbatim for non-local hosts.

## Test plan

- [x] Added `test_run_command_local_from_worker_thread` that calls `Host._run_shell_command` from inside `mngr_executor` against a local host and asserts success.
- [x] `just test-quick libs/mngr/imbue/mngr/hosts` -> 339 passed
- [x] `just test-quick libs/mngr/imbue/mngr/utils/test_ratchets.py::test_prevent_direct_subprocess` -> 1 passed (no new direct-subprocess violations; the new code goes through `ConcurrencyGroup`)
- [ ] CI to confirm acceptance + release suites
- [ ] Manual: verify `mngr observe --discovery-only` no longer crashes against a local host on Linux (cannot verify locally; macOS doesn't trigger the libev constraint)

## Notes

Verified on macOS only. The libev backend behavior that triggers the original `TypeError` is Linux-specific, so the new test passes on either OS but only acts as a regression test on Linux. The implementation guarantees the gevent code path is never entered for local hosts, so the fix is correct independent of which OS reproduced the original crash.


joshalbrecht:
lgtm once tests pass (which seem important for this change)

PR #1498, opened 2026-05-04
Reopen primary agent chat when last minds tab closes

gnguralnick:
## Summary
- Closing the last dockview tab in a minds workspace previously left a blank screen with no way to recover. Now the primary agent's chat tab is reopened automatically when the dockview becomes empty.
- The fix runs both at runtime (after a panel is removed via `onDidRemovePanel`) and after restoring a saved layout that has no panels.

This branch also carries the previously-staged `Bypass pyinfra/loopback dial` work; only the new commit is the dockview fix.

## Test plan
- [ ] Open the minds workspace, close every tab one by one, and confirm the primary agent chat reopens automatically once the last tab closes.
- [ ] Reload the app with a saved layout that has zero panels (or after the layout fails to restore) and confirm the primary chat is opened.
- [ ] Verify destroying a non-primary agent that owned the only remaining tab also reopens the primary chat.
- [ ] `npm run lint` and `npm run build` in `apps/minds_workspace_server/frontend` pass.

### DIFF
--- a/apps/minds/imbue/minds/desktop_client/test_desktop_client.py
+++ b/apps/minds/imbue/minds/desktop_client/test_desktop_client.py
@@ -1516,6 +1516,8 @@ def test_subdomain_forward_websocket_loopback_url_without_tunnel_closes_with_101
 
     assert exc_info.value.code == 1013
     assert "loopback" in (exc_info.value.reason or "")
+
+
 def _create_pending_subdomain_test_client(tmp_path: Path, agent_id: AgentId) -> tuple[TestClient, FileAuthStore]:
     """Build a desktop client where the given agent is a known workspace whose
     workspace_server URL has not yet been registered (service_logs is empty).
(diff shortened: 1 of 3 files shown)


joshalbrecht:
lgtm, but my instinct about the simpler way to do this is to have a sort of background page that shows up if no tabs are open, and, for example, lists the various services that you have running so that you can open any of them. Kind of like how vscode / cursor show you a page with useful links when you start, even though you have no files open

it's a very small UX thing though (it's just about UX consistency, with my proposal the "close" button *always* closes things, and you still have a way to recover). I've found that such patterns are simpler and easier to maintain

will leave this one up to you to merge or fix as you see fit

gnguralnick:
i was deciding between the two approaches. my logic for now was that this is the easy change to ensure there's no broken state but that eventually we probably do want a 'home' page of sorts for each mind. but having this change buys us time to make sure that that works well and looks good

PR #1501, opened 2026-05-05
imbue_cloud + minds: refactor leasing, fix discovery/auth/tunnel issues

joshalbrecht:
## Summary
- Refactors `mngr_imbue_cloud` provider for adopting pre-baked pool hosts on lease (preserves bake's `data.json`, regenerates command, patches only name/labels/workspace).
- Adds OAuth signin force-enable and `[providers.imbue_cloud_<slug>] is_enabled = false` auto-disable when discovery surfaces an `ImbueCloudAuthError` for a specific account, with a "Signed out" badge in the Manage Accounts page.
- Suppresses the default `[providers.imbue_cloud]` auto-instance via the minds bootstrap so it doesn't race per-account suffixed providers.
- Adds a grandparent-death watcher (`start_grandparent_death_watcher`) so the minds Python backend self-terminates when Electron dies, breaking orphan trees that observe + events couldn't break alone.
- Stops false-positive "subprocess failed" notifications when minds itself SIGTERMs `mngr observe` / `mngr events` during shutdown / restart / events-stream sync.
- Threads `provider_name` through `DiscoveryErrorEvent` and wraps per-provider failures in `ProviderDiscoveryError` so consumers can attribute discovery failures to a specific provider.
- Dispatches reverse-tunnel setup (`AgentDiscoveryHandler`, `LatchkeyDiscoveryHandler`) onto CG-managed worker threads so the discovery-stream reader thread isn't blocked on multi-second SSH I/O. The reader thread can now respond to `shutdown_event` in <100ms instead of stalling for the duration of the in-flight callback chain.
- Promotes the OAuth subprocess thread from a bare `threading.Thread` to `root_concurrency_group.start_new_thread` so unhandled exceptions are logged via the CG's `ObservableThread` instead of disappearing silently.
- Pulls in `main`, including PR #1497 (bypass pyinfra for local-host shell commands -- fixes the gevent-child-watcher crash that broke `mngr list` inside workspace containers).
- Adds `imbue_cloud` to `_REMOTE_BACKEND_NAMES` so test fixtures using `load_local_backend_only` skip it like `modal` and `vultr`.

## Test plan
- [x] `just test-quick` for `apps/minds`, `libs/mngr`, `libs/mngr_imbue_cloud`, all other affected lib packages -- 4459+ tests passing.
- [x] Manual verification: lease + create new agent end-to-end via the desktop client, confirm agent appears in resolver, terminal loads, sharing dialog opens.
- [ ] CI to confirm acceptance + release suites.


joshalbrecht:

PR #1500, opened 2026-05-04
Exclude tests from wheel builds for 4 workspace packages

weishi-imbue:
## Summary
Tiny packaging-hygiene fix. Add \`exclude = [\"*_test.py\", \"test_*.py\", \"**/conftest.py\"]\` under \`[tool.hatch.build.targets.wheel]\` for the four workspace packages that were missing it: \`minds\`, \`mngr_claude\`, \`mngr_modal\`, \`modal_proxy\`.

Without this, hatchling includes every \`_test.py\` and \`conftest.py\` in the wheel — any consumer that pip-installs one of these packages ships unit tests in their \`site-packages/\`. The other workspace packages (\`mngr\`, \`mngr_lima\`, etc.) already have this exclusion; this PR just brings the four stragglers in line.

Originally part of a larger \`Lima mode, wheel bundling, cross-platform builds, minds cleanup\` commit on \`wz/minds_onboard\`; extracted as a standalone PR per the splitting plan.

## Test plan
- [x] \`uv build --wheel\` succeeds for all 4 packages
- [x] \`unzip -l\` on each resulting wheel shows zero \`*_test.py\` / \`conftest.py\` entries
- [ ] CI green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1505, opened 2026-05-05
Uniform wheel test-exclusion across all packages, enforced by meta ratchet

weishi-imbue:
## Summary

Every workspace package's wheel build now excludes test files uniformly via the same canonical line:

```
[tool.hatch.build.targets.wheel]
exclude = ["*_test.py", "test_*.py", "**/conftest.py", "**/testing.py"]
```

Previously, several packages were missing some or all of these patterns, so hatchling was shipping `_test.py`, `conftest.py`, and `testing.py` files into published wheels — meaning any consumer that pip-installed one of these packages got our test code in their `site-packages/`. Notably `libs/mngr` was leaking `imbue/mngr/cli/testing.py`, `imbue/mngr/api/testing.py`, and `imbue/mngr/providers/docker/testing.py` because its existing exclude (`**/utils/testing.py`) only covered the one in `utils/`.

A new meta ratchet (`test_every_project_excludes_tests_from_wheel` in `test_meta_ratchets.py`) enforces the four-pattern rule on every project so the same regression cannot happen again. The check is a simple literal-string membership test against `[tool.hatch.build.targets.wheel].exclude`. Projects with `only-include` (an explicit whitelist) are exempt; `flexmux` is exempt via the existing `_EXCLUDED_PROJECTS` set.

The four patterns are required uniformly, even on packages that don't currently have a matching file. That's deliberate — it's the same convention `**/conftest.py` already followed (11 packages carry it without a current `conftest.py`) and means adding a new `testing.py` later doesn't need a second config-only PR. Trade-off: ~15 packages carry `**/testing.py` as no-op config today; benefit: the ratchet stays a 5-line literal check with no file-walking, no glob semantics, and no two-step trap.

Extends the original fix in #1500 (which added the exclude to 4 packages); this PR brings every remaining package in line and adds the enforcement.

## Test plan
- [x] `just test-quick test_meta_ratchets.py` -> 14 passed
- [x] Verified the ratchet fails on regression (manually removed `**/testing.py` from one project; got a clean per-project diagnostic)
- [x] Verified the ratchet catches mngr's previously-leaking `cli/`, `api/`, `providers/docker/testing.py` files when reverted to the old `**/utils/testing.py` pattern
- [ ] CI green

### DIFF
--- a/apps/minds/pyproject.toml
+++ b/apps/minds/pyproject.toml
@@ -44,7 +44,7 @@ imbue-common = { workspace = true }
 
 [tool.hatch.build.targets.wheel]
 packages = ["imbue"]
-exclude = ["*_test.py", "test_*.py", "**/conftest.py"]
+exclude = ["*_test.py", "test_*.py", "**/conftest.py", "**/testing.py"]
 
 # Shared pytest settings (markers, filterwarnings, coverage report config)
 # are centralized in:
--- a/apps/minds_workspace_server/pyproject.toml
+++ b/apps/minds_workspace_server/pyproject.toml
@@ -45,7 +45,7 @@ artifacts = [
 
 [tool.hatch.build.targets.wheel]
 packages = ["imbue"]
-exclude = ["*_test.py", "test_*.py", "**/conftest.py"]
+exclude = ["*_test.py", "test_*.py", "**/conftest.py", "**/testing.py"]
 
 # Shared pytest settings (markers, filterwarnings, coverage report config)
 # are centralized in:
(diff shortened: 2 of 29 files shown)


apps/minds/pyproject.toml
```diff
@@ -44,7 +44,7 @@ imbue-common = { workspace = true }
 
 [tool.hatch.build.targets.wheel]
 packages = ["imbue"]
-exclude = ["*_test.py", "test_*.py", "**/conftest.py"]
+exclude = ["*_test.py", "test_*.py", "**/conftest.py", "**/testing.py"]
```
weishi-imbue, line 47: Not all packages have all the files in this array. (e.g. many don't have `testing.py`). But having a uniform rule to exclude it seems right and future proof.

joshalbrecht:
lgtm, thanks weishi!

PR #1507, opened 2026-05-06
Protect .claude/settings.local.json from propagate_changes rsync --delete

gnguralnick:
## Summary

`apps/minds/scripts/propagate_changes` now protects `.claude/settings.local.json` from `rsync --delete` when syncing the template into an agent's work_dir.

That file is generated per-agent at create time by mngr's `_configure_agent_hooks` (see `libs/mngr_claude/imbue/mngr_claude/plugin.py:_configure_agent_hooks`) and holds the `UserPromptSubmit` hook that signals `tmux wait-for -S "mngr-submit-..."` (built by `build_readiness_hooks_config` in `libs/mngr_claude/imbue/mngr_claude/claude_config.py`). Mngr's `_send_enter_and_wait_for_signal` waits on that signal to confirm prompt submission.

Without the file, every `send_message` hangs the 90-second submission-signal timeout. The prompt is actually delivered to Claude — the chat UI shows the message and Claude responds normally — but the HTTP `/message` request blocks until the wait-for times out.

Previously the script protected only `runtime/` and `.mngr/`, so iterating with `propagate_changes` reliably reproduced this hang and the only recovery was to recreate the agent.

Also updates the `minds-dev-iterate` skill (`.claude/skills/minds-dev-iterate/SKILL.md`) and the dev-iteration-loop spec (`specs/minds-dev-iteration-loop/spec.md`) to document the new protect filter and the failure mode it prevents.

## Test plan

- [x] Reproduced the hang on mindtest-3 after a normal `propagate_changes` run: the file was missing in `/code/.claude/`, sends hung 90s while Claude responded normally
- [x] Manually restored the file via `build_readiness_hooks_config()`, restarted the agent — sends complete in <1s
- [x] Re-ran `propagate_changes` with the fix in place — `/code/.claude/settings.local.json` survives the rsync
- [ ] CI green

### DIFF
--- a/.claude/skills/minds-dev-iterate/SKILL.md
+++ b/.claude/skills/minds-dev-iterate/SKILL.md
@@ -171,7 +171,7 @@ If this chain breaks (orphaned `mngr observe`/`mngr event` processes appear), so
 Both syncs (mngr->vendor/mngr and template->container) exclude:
 `.git`, `__pycache__`, `.venv`, `node_modules`, `.test_output`, `.mypy_cache`, `.ruff_cache`, `.pytest_cache`, `uv.lock`, `.external_worktrees`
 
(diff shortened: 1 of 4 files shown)


joshalbrecht:
lgtm, thanks for fixing!

PR #1508, opened 2026-05-06
mngr_claude: fall back to CLAUDE_CONFIG_DIR when ORIGINAL_CLAUDE_CONFIG_DIR doesn't exist

gnguralnick:
## Summary

Inside a nested sandbox (e.g. a Linux lima VM on a macOS host), `$ORIGINAL_CLAUDE_CONFIG_DIR` is inherited from the host and points at a path like `/Users/<user>/.claude` that does not exist inside the VM. `get_user_claude_config_dir()` previously returned that path unconditionally, so `_provision_local_credentials` would look for `.credentials.json` under a non-existent dir, log "No .credentials.json found to provision", and silently no-op. Child agents spawned via `mngr create` then failed Claude sessions with "Not logged in", even though the parent agent's live credentials were sitting in `$CLAUDE_CONFIG_DIR/.credentials.json`.

This patch makes `get_user_claude_config_dir()` fall back to `get_claude_config_dir()` (`$CLAUDE_CONFIG_DIR` or `~/.claude/`) when `$ORIGINAL_CLAUDE_CONFIG_DIR` does not resolve to an existing directory. This is the right layer for the fix: every caller that resolves user-scope Claude config benefits.

## Test plan

- [x] Manually verified: lima sandbox spawns child claude agents that successfully provision credentials from `$CLAUDE_CONFIG_DIR/.credentials.json`.
- [ ] CI tests pass.


joshalbrecht:
lgtm, but cc @evgunter in case I'm missing something obvious

evgunter:
i thought there was somewhere else that already had the resolution order logic depending on the files in question already existing, but if not then this seems right

PR #1509, opened 2026-05-06
workspace_server: rewrite Location headers on 3xx redirects in proxy

gnguralnick:
## Summary

The workspace_server forwarder rewrites cookies and HTML body content for proxied responses, but never rewrote the `Location` header on 3xx redirects. Backends issuing redirects (e.g. OAuth flows, 302-after-POST) sent absolute URLs pointing at the backend's own host, which the browser then followed off-proxy and broke the session.

This patch adds `rewrite_location_header` in `proxy.py` and wires it into `_build_proxy_response` to apply on 3xx responses, alongside the existing cookie / HTML rewriters. Tests added in `proxy_test.py` and `service_dispatcher_test.py`.

The second commit appeases the `test_prevent_if_elif_without_else` ratchet (count went 6 -> 7 with the new elif branch) by adding an explicit `else: pass` -- the pass-through for non-cookie / non-location headers.

## Test plan

- [x] `proxy_test.py` and `service_dispatcher_test.py` pass (51 tests).
- [x] `test_prevent_if_elif_without_else` passes.
- [x] Pre-existing `test_no_type_errors` failures (224 errors in apps/minds modules) and `test_no_ruff_errors` (passes in isolation, fails only via test runner timeout) confirmed pre-existing on main and unrelated to this change.


gnguralnick:
Replaced by #1521, which recreates this off main as a single clean commit (and drops the `else: pass` ratchet workaround in favor of a `rewritten_value` local).

PR #1510, opened 2026-05-06
Clean up reparented orphans on agent stop

gnguralnick:
## Problem

When an agent's main process (claude) is killed abruptly — OOM, SIGKILL, segfault —
any background processes it spawned (e.g. `npm exec ... &`, playwright browsers)
survive and reparent to PID 1. From mngr's perspective the agent looks dead, but
those processes keep running indefinitely, holding ports and chewing CPU/memory.

When the user later cleans up via `mngr stop`, `mngr destroy`, `mngr archive`,
`mngr cleanup`, etc. — all of which go through `Host.stop_agents` — the existing
teardown only walks the tmux pane's descendant tree. The reparented orphans
aren't in that tree anymore, so they're left running.

## Fix

`stop_agents` now also scans `/proc/<pid>/environ` for processes whose
`MNGR_AGENT_ID` env var matches the agent being stopped, and adds them to the
kill list. The agent's env file is sourced into every process spawned in its
tmux session via `set -a`, so this marker is inherited by descendants and
survives reparenting.

This runs on every `stop_agents` call — not at the moment of the OOM. So a
user whose agent died unexpectedly will see the orphans cleaned up the next
time they run `mngr stop`/`destroy`/`cleanup` against that agent.

Linux only (uses `/proc`); macOS reparented children lose env visibility under
SIP, so the scan is a best-effort no-op there. The pre-existing pane-descendant
walk still handles the typical macOS case.

## Why not process groups

setsid-wrapping the pane command and `kill -- -<pgid>` were both tried and
removed previously (c4ac00242c, 4ebf66d2f4): bash job control puts each `&`'d
process in its own pgrp, so the pane's pgrp doesn't cover the descendants we
need to kill. Env-marker inheritance survives both job control and reparenting
without re-introducing those issues.

## Test

`test_stop_agent_kills_orphaned_processes_by_env_marker` reproduces the failure
mode: spawns a backgrounded `sleep` in a subshell that exits, leaving the
sleep reparented to PID 1 and invisible to the pane walk; asserts it inherits
`MNGR_AGENT_ID` via the agent env file; asserts `stop_agents` kills it.
Linux only.


joshalbrecht:
omfg this is so much nicer, thank you

lgtm

PR #1511, opened 2026-05-06
Improve TMR reliability against Modal provider

qi-imbue:
## Summary

A grab-bag of TMR / Modal-provider reliability fixes, plus visibility tweaks for the HTML report and `-vv` timing logs.

### TMR orchestrator no longer crashes on partial agent failures
- Catch `AgentError` (e.g. `SendMessageError` from the tmux paste-detection timeout) when launching test agents and the integrator. Previously the whole orchestrator died because `SendMessageError` is `AgentError → BaseMngrError`, not `MngrError`, so the existing `(MngrError, HostError, OSError, BaseExceptionGroup)` clauses didn't catch it.
- Tests whose agent failed to launch now appear in the HTML report (with the actual agent name that was used for the launch attempt, so the row matches any surviving host/tmux session) instead of being silently dropped.

### TMR HTML report
- New dedicated **Failed** section, separate from **Blocked**. Errored results (launch failure, agent timeout, missing details) used to be lumped into Blocked, but Blocked is supposed to mean the coding agent itself reported every change as BLOCKED -- those are different failure patterns and now have separate sections + colors.

### TMR integrator launch (and any local-provider TMR run)
- Fix `Failed to generate a unique host name after 100 attempts`. The local provider has a single fixed host (`localhost`), so the new-host path can never find a free name; TMR now reuses the existing local host when the target provider is `local`, mirroring what `mngr create` already does. This consistently broke the integrator (which defaults to `--integrator-provider local`).

### Modal-launched agents no longer hang at SSH TOFU prompt
- The `claude plugin update` SessionStart hook script shells out to `git pull`, which uses `ssh`. On a fresh Modal sandbox `~/.ssh/known_hosts` is empty, so ssh blocks on a "Are you sure you want to continue connecting" prompt that Claude Code's bypass-permissions setting does NOT cover -- breaking TUI-driving tests like TMR's send-enter-and-wait.
- Prefix the update with `GIT_SSH_COMMAND='ssh -o StrictHostKeyChecking=accept-new -o BatchMode=yes'`. `accept-new` writes the first-seen host key to known_hosts (so subsequent ssh-via-git calls in the same sandbox don't re-prompt), and `BatchMode=yes` exits non-interactively on failure.

### `-vv` timing visibility
- `mngr create -vv` now emits a `Transferring agent files` log_span around `_execute_agent_file_transfers`, with `count=N` (including `count=0` when the agent declared no transfers). Previously the per-file `write_file` calls were only logged at `trace` and the aggregate cost was invisible.

## Test plan

- [x] `just test-quick libs/mngr_tmr` — 175 passed
- [x] `just test-quick libs/mngr` — 3797 passed
- [x] `just test-quick libs/mngr_claude` — 458 passed
- [x] Run `mngr tmr --provider modal` end-to-end with at least one test whose agent will fail to receive its initial message; confirm:
  - the orchestrator does not crash
  - the failed test shows up in the HTML report under the new **Failed** section
  - its row in the report shows the actual agent name (e.g. `tmr-test-foo-f95667`), not a `-launch-failed` placeholder
  - other agents continue normally
- [x] Run `mngr tmr --provider local` end-to-end and verify the integrator launches successfully (no `Failed to generate a unique host name` error). Test agents should also work locally.
- [x] Run `mngr create -vv` against a Modal host and confirm `Transferring agent files [done in X sec]` appears in the timing output.
- [x] On a fresh Modal sandbox, confirm Claude Code starts cleanly without the github.com SSH TOFU prompt sitting in the tmux pane.

### DIFF
--- a/changelog/mngr-run-tmr.md
+++ b/changelog/mngr-run-tmr.md
@@ -22,6 +22,16 @@
   debugging. The `mngr create -vv` log span around `_execute_agent_file_transfers`
   now wraps the early-return path too, so the span is emitted (with
   `count=0`) even when the agent declared no file transfers.
+- Stop the `claude plugin update` SessionStart hook from hanging Modal-launched
+  agents at an `ssh` first-contact (TOFU) prompt for github.com. The plugin
+  updater shells out to `git pull`, which uses `ssh` -- on a fresh sandbox
+  with no `~/.ssh/known_hosts` entry, ssh blocks on a "Are you sure you
+  want to continue connecting" prompt that Claude Code's bypass-permissions
+  setting does not cover. `scripts/claude_update_plugin.sh` now prefixes
+  the update with `GIT_SSH_COMMAND='ssh -o StrictHostKeyChecking=accept-new
+  -o BatchMode=yes'`, which writes the first-seen host key to known_hosts
+  and exits non-interactively if anything goes wrong (matching the
+  script's existing `2>/dev/null || true` failure tolerance).
 - `mngr tmr` HTML reports now have a dedicated "Failed" section,
   separate from "Blocked". The two represent different failure modes:
   Blocked means the coding agent reported every change as BLOCKED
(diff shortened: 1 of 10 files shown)


joshalbrecht:
lgtm, thanks for fixing Qi!

PR #1513, opened 2026-05-06
Single latchkey gateway instance

hynek-urban:
(Accompanied by https://github.com/imbue-ai/forever-claude-template/pull/41.)

Spawn a single Latchkey gateway instance shared by all agents with the following properties:

- It is now protected by a password.
- Each agent gets their own permissions.json via the "permissions override" mechanism introduced in Latchkey 2.8.0.


joshalbrecht:
lgtm

PR #1514, opened 2026-05-06
Get --from remote_agent@remote_host actually working, plus related changes

qi-imbue:
## My summary

I started out by trying to get `mngr create a2@m1.modal --from a1@m1.modal:/foo/bar` working and fixed a bunch of related issues I noticed:

- A lot of places, including `--from`, that should accept a host **address** (`host.provider`) haven't actually been updated to do so. Updated those places. There are some more opportunities for refactoring that I'll follow up.

- Generalize a bunch of previously localhost-only code to operate on remote hosts. This includes reading the Git branch from a remote host for `--from`, and more notably, when transferring between the same remote host, run `git` and `rsync` on that remote host.

- `--transfer git-tree` is now the default for Git repos if source and destination are the same host (previously both had to be localhost).

## Claude's Summary

This branch unblocks `mngr create --from <remote-source>` and the related cross-host transfer paths. The trigger was `mngr c a@m1.modal --from @m1.modal:/some/path` failing in three different places; each fix exposed the next.

**User-visible behavior changes**

- Accept the `host.provider` qualifier consistently anywhere mngr takes a host identifier. Previously only the positional `agent@host.provider` form worked. Now `mngr create --from @m1.modal:/path`, `mngr limit --host m1.modal`, `mngr snapshot create --host m1.modal`, and the `--host` filter on list-style commands all resolve `host.provider` deterministically. The dot is treated as a strict `[HOST].[PROVIDER]` separator (host names cannot contain dots in this DSL).

- `mngr create --from` against a *remote* source no longer raises `NotImplementedError`. The current-branch lookup, git author info, and origin URL fetch all run via `host.execute_idempotent_command(...)` so any provider works.

- `mngr create --from` between source and target on the *same remote host* now works end-to-end:
  - The git-mirror push short-circuits to a single in-host `git push` between two local paths, avoiding the broken "ship laptop SSH credentials to the source" flow that produced "Identity file ... not accessible".
  - `--transfer=git-worktree` is allowed whenever source and target are on the same host (local *or* remote). The default for same-host git sources is now `git-worktree` even on remote providers; it was previously restricted to local-only agents.
  - The `mngr_claude` over-aggressive worktree-on-remote-host refusal is removed; trust is set up correctly via the per-agent `.claude.json` on the agent host.
  - The internal rsync helper (used for `work_dir_extra_paths` and the rsync transfer mode) also short-circuits for same-host transfers, running a single rsync between two local paths instead of routing through a laptop temp directory.

**Internal change worth flagging**

`ParsedSourceLocation` (the parsed shape of `--from`) replaces its opaque `host: str | None` with separate `host_name`/`provider_name` fields mirroring `AgentAddress`. A component-based `find_matching_hosts(host_name, provider_name, all_hosts)` primitive backs both the string-based `find_all_matching_hosts` wrapper and the `--from` resolver. Other libs that import these (`mngr_wait`, `mngr_file`) are unaffected.

## Test plan
- [x] `just test-quick libs/mngr/imbue/mngr/api/find_test.py`
- [x] `just test-quick libs/mngr/imbue/mngr/cli/create_test.py libs/mngr/imbue/mngr/cli/test_create.py`
- [x] `just test-quick libs/mngr/imbue/mngr/hosts`
- [x] `just test-quick libs/mngr_claude/imbue/mngr_claude/plugin_test.py`
- [x] `just test-quick "libs/mngr libs/mngr_wait libs/mngr_file"`
- [ ] `just test-offload` and `just test-offload-acceptance` (in CI)

**Manual verification (suggested)**
- `mngr c a1@m1.modal --type yolo` then `mngr c a2@m1.modal --from @m1.modal:/path/to/a1 --type yolo` -- same-remote-host create from another agent's project dir; should default to git-worktree and not invoke SSH or local rsync intermediary.
- Same as above but with explicit `--transfer=git-mirror` -- should run a single in-host `git push`, no SSH credential errors.
- Same as above but with explicit `--transfer=rsync` (or non-git source) -- should run a single in-host rsync, no laptop temp dir.
- `mngr limit --host m1.modal` and `mngr snapshot create --host m1.modal` -- should resolve the host without "could not find" errors.
- Cross-host (`--from @host-a.modal:/path` with target on a *different* host) -- still expected to fail with the existing SSH-credential issue (out of scope; documented in the original investigation).


libs/mngr/imbue/mngr/api/agent_addr.py
qi-imbue, line 49: Makes sense, this is now just a wrapper around the almost identical - just slightly more low-level - `parse_address_part`.

libs/mngr/imbue/mngr/api/find.py
```diff
@@ -39,26 +39,55 @@ class ParsedSourceLocation(FrozenModel):
     Produced by parse_source_string(); consumed by resolve_source_location().
     """
 
-    agent: str | None = Field(description="Agent ID or name")
-    host: str | None = Field(description="Host ID or name (may include .PROVIDER suffix)")
-    path: str | None = Field(description="File path")
+    agent: str | None = Field(default=None, description="Agent ID or name")
```
qi-imbue, line 42: Makes sense, breaks apart `host` and `path` into `host_name` and `provider_name` with appropriate types.

libs/mngr/imbue/mngr/api/find.py
```diff
@@ -39,26 +39,55 @@ class ParsedSourceLocation(FrozenModel):
...
+    provider_name: ProviderInstanceName | None = Field(
+        default=None, description="Provider name when explicitly qualified with .PROVIDER"
+    )
+    path: str | None = Field(default=None, description="File path")
 
 
 @pure
-def _parse_address_part(address_str: str) -> tuple[str | None, str | None]:
-    """Parse agent address portion of a source string into (agent, host) components.
+def parse_host_qualifier(host_part: str) -> tuple[HostName | None, ProviderInstanceName | None]:
```
qi-imbue, line 51: Parsing of qualified host address is moved down from the old implementation of `parse_agent_address` and now used in more places (hopefully everywhere) that accepts hostnames.

libs/mngr/imbue/mngr/api/find.py
```diff
@@ -39,26 +39,55 @@ class ParsedSourceLocation(FrozenModel):
...
+    host_name = HostName(host_str) if host_str else None
+    provider_name = ProviderInstanceName(provider_str) if provider_str else None
+    return (host_name, provider_name)
 
-    agent_part, host_part = address_str.split("@", 1)
-    agent = agent_part or None
-    host = host_part or None
-    return (agent, host)
+
+@pure
```
qi-imbue, line 74: Now also breaks up host address

libs/mngr/imbue/mngr/api/find.py
```diff
@@ -72,27 +101,19 @@ def parse_source_string(source: str) -> ParsedSourceLocation:
...
-      - "my-agent:path"             -> agent="my-agent", path="path"
-      - "my-agent@my-host:/path"    -> agent="my-agent", host="my-host", path="/path"
-      - "@my-host:/path"            -> host="my-host", path="/path"
-      - ":/path/to/dir"             -> path="/path/to/dir"
-      - "/path/to/dir"              -> path="/path/to/dir"
     """
     # Recognize unambiguous path prefixes as a convenience
     if source.startswith(("/", "./", "~/", "../")):
-        return ParsedSourceLocation(agent=None, host=None, path=source)
+        return ParsedSourceLocation(path=source)
```
qi-imbue, line 116: Cosmetic change - now `ParsedSourceLocation` has default values for fields

libs/mngr/imbue/mngr/api/find.py
```diff
@@ -102,14 +123,8 @@ def parse_source_string(source: str) -> ParsedSourceLocation:
         address_part = source
         path = None
 
-    # Empty address means local path (variant 4: ":path")
```
qi-imbue, line 105: No change in logic here

libs/mngr/imbue/mngr/api/find.py
```diff
@@ -132,27 +147,51 @@ def determine_resolved_path(
     raise UserInputError("Must specify a path if no agent is specified")
 
 
+@pure
```
qi-imbue, line 150: Makes sense - I can see a refactor where hostname-or-ID is not misrepresented as `HostName` but it's not too bad.

Also `HostName` currently can hold a qualified hostname `HOST.PROVIDER`, and we may want to split that into a `HostName` with just `HOST` and a `HostAddress`, but those can be followup refactors. This PR has enough refactors.

libs/mngr/imbue/mngr/api/find.py
```diff
@@ -132,27 +147,51 @@ def determine_resolved_path(
...
+            matches = [h for h in matches if h.host_id == host_id]
+        else:
+            matches = [h for h in matches if h.host_name == host_name_or_id]
+    if provider_name is not None:
+        matches = [h for h in matches if h.provider_name == provider_name]
+    return matches
+
+
 @pure
 def find_all_matching_hosts(
```
qi-imbue, line 183: Now a simple wrapper around `_find_matching_hosts`

libs/mngr/imbue/mngr/api/find.py
```diff
@@ -182,27 +221,57 @@ def find_all_matching_agents(
     return matches
 
 
+@pure
```
qi-imbue, line 224: Another thin wrapper around `_find_matching_hosts`.

libs/mngr/imbue/mngr/api/find.py
```diff
@@ -182,27 +221,57 @@ def find_all_matching_agents(
...
+    if len(matches) == 0:
+        descriptor = _host_descriptor(host_name_or_id, provider_name)
+        raise UserInputError(f"Could not find host with ID or name: {descriptor}")
+    if len(matches) > 1:
+        descriptor = _host_descriptor(host_name_or_id, provider_name)
+        raise UserInputError(f"Multiple hosts found with name: {descriptor}")
+    return matches[0]
+
+
+@pure
```
qi-imbue, line 248: This probably belongs to a new `HostAddress` type. Defer this to a later refactor.

libs/mngr/imbue/mngr/api/find.py
```diff
@@ -182,27 +221,57 @@ def find_all_matching_agents(
...
+    if host_name_or_id is not None and provider_name is not None:
+        return f"{host_name_or_id}.{provider_name}"
+    if host_name_or_id is not None:
+        return str(host_name_or_id)
+    if provider_name is not None:
+        return f".{provider_name}"
+    return ""
+
+
 @pure
```
qi-imbue, line 260: Basically a wrapper around `_find_one_matching_host` now

libs/mngr/imbue/mngr/cli/agent_utils.py
```diff
@@ -18,15 +19,14 @@
 from imbue.mngr.primitives import DiscoveredAgent
 from imbue.mngr.primitives import DiscoveredHost
 from imbue.mngr.primitives import HostId
-from imbue.mngr.primitives import HostName
 from imbue.mngr.primitives import OutputFormat
 
 
 @pure
 def _host_matches_filter(host_ref: DiscoveredHost, host_filter: str) -> bool:
```
qi-imbue, line 26: Adds support for host address. This probably belongs to `api/find.py` too; defer to a later refactor.

libs/mngr/imbue/mngr/cli/create.py
```diff
@@ -1290,13 +1291,18 @@ def _resolve_target_host(
     return resolved_target_host
 
 
-def _get_current_git_branch(source_location: HostLocation, mngr_ctx: MngrContext) -> str | None:
-    if not source_location.host.is_local:
-        raise NotImplementedError(
-            "Have to re-implement this function so that it works via HostInterface calls instead!"
-        )
+def _get_current_git_branch(source_location: HostLocation) -> str | None:
```
qi-imbue, line 1294: Generalized for remote hosts now!

libs/mngr/imbue/mngr/cli/create.py
```diff
@@ -1322,31 +1328,49 @@ def _split_cli_args(args: tuple[str, ...]) -> list[str]:
...
+) -> bool:
+    """Decide whether the source and target resolve to the same physical host.
+
+    A NewHostOptions target is never the same host (it does not yet exist).
+    A None target means "local default host", which matches a local source.
+    A DiscoveredHost target matches the source iff their host IDs are equal.
+    """
+    if isinstance(target_host, NewHostOptions):
+        return False
+    if target_host is None:
```
qi-imbue, line 1344: Makes sense, it'd be a bit nicer if the target host gets resolved to the local host at this point so we don't have to special-case it, but that might be a bigger refactor

libs/mngr/imbue/mngr/cli/create.py
```diff
@@ -1322,31 +1328,49 @@ def _split_cli_args(args: tuple[str, ...]) -> list[str]:
...
+        return source_location.host.is_local
+    return target_host.host_id == source_location.host.id
+
+
 def _resolve_transfer_mode(
```
qi-imbue, line 1349: Pretty mechanical changes to support Git worktree between the same remote host

libs/mngr/imbue/mngr/cli/create.py
```diff
@@ -1399,6 +1424,7 @@ def _resolve_transfer_mode(
 def _parse_agent_opts(
```
qi-imbue, line 1424: Mechanical changes to thread some args

libs/mngr/imbue/mngr/cli/snapshot.py
```diff
@@ -94,18 +94,23 @@ def _find_host_across_providers(
     host_identifier: str,
     mngr_ctx: MngrContext,
 ) -> tuple[HostId, ProviderInstanceName] | None:
-    """Find a host by ID or name across all providers.
+    """Find a host by ID, name, or `host.provider` form across all providers.
```
qi-imbue, line 97: Makes sense

libs/mngr/imbue/mngr/hosts/host.py
```diff
@@ -196,6 +193,18 @@ def get_agent_state_dir_path(host_dir: Path, agent_id: AgentId) -> Path:
     return host_dir / "agents" / str(agent_id)
 
 
+def _git_command_stdout(host: OnlineHostInterface, command: str, cwd: Path) -> str | None:
```
qi-imbue, line 196: Extracted from the common logic of git_utils.py functions, generalized to any host

libs/mngr/imbue/mngr/hosts/host.py
```diff
@@ -1216,6 +1208,15 @@ def _git_push_to_target(
     ) -> None:
         """Push git repo from source to target, mirroring branches and tags."""
         self._warn_if_submodules_detected(source_host, source_path)
+
+        # When source and target are the same host, skip SSH entirely and push
+        # between two local paths on that host. Otherwise the SSH path below
+        # ends up shipping laptop-local key paths to a remote source host and
+        # failing with "Identity file ... not accessible".
+        if source_host.id == self.id:
```
qi-imbue: Same-host short circuit

libs/mngr/imbue/mngr/hosts/host.py
```diff
@@ -1083,32 +1092,15 @@ def _transfer_git_repo(
...
-                name_result.stdout.strip() if name_result.success and name_result.stdout.strip() else None
-            )
-            git_author_email = (
-                email_result.stdout.strip() if email_result.success and email_result.stdout.strip() else None
-            )
-            origin_result = source_host.execute_idempotent_command("git remote get-url origin", cwd=source_path)
-            origin_url = (
-                origin_result.stdout.strip() if origin_result.success and origin_result.stdout.strip() else None
-            )
+        git_author_name = _git_command_stdout(source_host, "git config user.name", source_path)
```
qi-imbue, line 1113: Generalized to any host

libs/mngr_claude/imbue/mngr_claude/plugin.py
```diff
@@ -1575,14 +1575,6 @@ def on_before_provisioning(
         Interactive and auto-approve runs skip these checks because
         provision() will handle them.
         """
-        if options.transfer_mode == TransferMode.GIT_WORKTREE:
```
qi-imbue, line 1578: Remove unnecessary check (not sure why the Claude plugin even cares about this in the first place)

libs/mngr/imbue/mngr/hosts/host.py
```diff
@@ -1260,7 +1267,15 @@ def _git_push_to_target(
         env["GIT_LFS_SKIP_PUSH"] = "1"
 
         with log_span("Pushing git repo to target: {}", git_url):
-            if source_host.is_local:
+            if same_machine:
```
qi-imbue, line 1270: Extra branch for same-machine Git mirror push - including local-to-local

joshalbrecht:
lgtm--this is much appreciated!

PR #1520, opened 2026-05-06
mngr_modal: keep test-only TestingModalInterface out of production import graph

weishi-imbue:
## Summary

- `mngr_modal/backend.py` no longer imports `modal_proxy.testing` at module top
- `ModalMode.TESTING` removed; `build_provider_instance` always builds `DirectModalInterface`
- Tests inject `TestingModalInterface` via `mngr_modal.testing.make_testing_provider` (already worked this way; just dropping the redundant config knob)
- New pure helper `_derive_modal_names(name, config, mngr_ctx)` so name-derivation unit tests don't need a Modal interface

## Why

`libs/modal_proxy/imbue/modal_proxy/testing.py` defines `TestingModalInterface` — an in-process fake of the Modal SDK that the docstring describes as *"designed for testing mngr_modal without requiring Modal credentials"*. The class is only ever instantiated when `ModalProviderConfig.mode == ModalMode.TESTING`, which is set only by test fixtures. No production caller sets it.

But `mngr_modal/backend.py:48` imported it at module top, unconditionally:

```python
from imbue.modal_proxy.testing import TestingModalInterface
```

Python evaluates module-top imports at module-load time, so any production code that imports `mngr_modal.backend` (e.g. `mngr forward` spawned by the minds desktop client) eagerly loaded `modal_proxy/testing.py`. The unified `**/testing.py` wheel-exclude rule (introduced in #1505) correctly stripped `testing.py` from the `modal_proxy` wheel, which in turn broke packaged builds at launch:

```
mngr forward stderr: ModuleNotFoundError: No module named 'imbue.modal_proxy.testing'
mngr forward exited with code 1
```

This wasn't caught in CI because tests run from editable installs (the source tree is on `sys.path`), which always have `testing.py` available. The wheel exclusion only takes effect when something builds a wheel and consumes it from a clean `site-packages`. The packaged minds.app was the first such consumer.

A quick fix would be to either lazy-import inside the dispatch arm (1 line, but a code smell — production code references test code) or to switch `modal_proxy/pyproject.toml` to an `only-include` whitelist (works but exempts modal_proxy from the uniform meta-ratchet). This PR does the architectural fix instead: production code never references the test fake at all.

## What changed

**`libs/mngr_modal/imbue/mngr_modal/backend.py`**
- Drop top-level `from imbue.modal_proxy.testing import TestingModalInterface`
- Drop the `match config.mode: case ModalMode.TESTING:` arm in `build_provider_instance` — always construct `DirectModalInterface`
- Drop `is_testing` parameter from `_get_or_create_app` (only the removed dispatch arm passed True; the test-fixture path uses `make_testing_provider` which constructs `ModalProviderApp` directly and never went through this method)
- Extract pure helper `_derive_modal_names(name, config, mngr_ctx) -> (environment_name, app_name, host_dir)` so the naming logic is unit-testable without any Modal interface
- Drop unused `assert_never` and `ModalMode` imports

**`libs/mngr_modal/imbue/mngr_modal/config.py`**
- `ModalMode` enum drops the `TESTING` value (only `DIRECT` remains); update docstring to explain the rationale

**`libs/mngr_modal/imbue/mngr_modal/testing.py`**
- `make_testing_provider` drops `mode=ModalMode.TESTING` from the fixture config (the field is irrelevant — the fixture bypasses `build_provider_instance` entirely and constructs `ModalProviderInstance` directly with a `TestingModalInterface`)

**`libs/mngr_modal/imbue/mngr_modal/testing_provider_test.py`**
- Replace three tests that called `build_provider_instance(mode=ModalMode.TESTING, ...)` with two unit tests on the pure `_derive_modal_names` helper (`test_derive_modal_names_environment_name_derived_from_prefix`, `test_derive_modal_names_truncates_long_app_name`)
- The deleted `test_build_provider_instance_testing_mode` was specifically testing the dispatch arm we just deleted; the other two cared about name derivation, which is now exposed as a pure helper that's easier to test

## Verification

After this change:

```python
>>> import imbue.mngr_modal.backend
>>> 'imbue.modal_proxy.testing' in sys.modules
False                                              # was True before
```

Packaged builds that exclude `**/testing.py` from the modal_proxy wheel now work at launch — `mngr_modal.backend` imports cleanly without the test fake.

## Test plan

- [x] `just test-quick libs/mngr_modal -m 'not modal and not docker and not docker_sdk and not acceptance and not release'` → 372 passed
- [x] `just test-quick libs/modal_proxy -m 'not modal and not docker and not docker_sdk and not acceptance and not release'` → 77 passed
- [x] `uv run pytest test_meta_ratchets.py::test_every_project_excludes_tests_from_wheel` → passes (modal_proxy continues to use uniform `exclude` rule, no `only-include` exemption)
- [x] `uv run pytest libs/mngr_modal/imbue/mngr_modal/test_ratchets.py` → 56 passed
- [x] Build modal_proxy wheel, confirm `imbue/modal_proxy/testing.py` is **not** in it
- [x] Manual import check: `import imbue.mngr_modal.backend` does not load `imbue.modal_proxy.testing`
- [ ] CI: full offload acceptance + release tests

## Notes

- This is a follow-up to #1505 (your wheel-exclude unification). The rule in #1505 was correct; the bug was a separate convention violation in `mngr_modal/backend.py` that the rule surfaced.
- A meta-ratchet that bans non-test code from importing `from imbue.X.testing` would prevent regressions of this specific shape; deliberately scoped out of this PR.


libs/mngr_modal/imbue/mngr_modal/testing.py
weishi-imbue, line 37: Existing tests should use this to do dependency injection at the proper layer. shouldn't duplicate this logic into production code under `case ModalMode.TESTING:`

libs/mngr_modal/imbue/mngr_modal/backend.py
weishi-imbue: Always constructs a real ``DirectModalInterface``. This keeps test-only code
out of the production import graph (per ``modal_proxy``'s wheel build,
which excludes ``testing.py`` from the published wheel).

However, this will make this method untestible. It may be ok since: 

This method is orchestration glue with real work delegated to four pieces, each
independently tested elsewhere with ``TestingModalInterface`` injected
at the right seam:

1. ``_derive_modal_names(name, config, mngr_ctx)`` -- Tested directly in ``testing_provider_test.py`` under
2. ``ModalProviderBackend._get_or_create_app`` and
    ``get_volume_for_app`` -- the app-registry layer. Tested in
    "Backend App Registry Tests"; both take ``modal_interface`` as a
    parameter, so the fake is injected directly.
3. ``ModalProviderApp(...)`` -- the app/volume holder. Tested in
    "ModalProviderApp Tests" by constructing the model with
    ``modal_interface=testing_modal`` (a pydantic constructor field).
4. ``ModalProviderInstance(...)`` -- the actual provider behavior.
    Exercised by the many tests using the ``testing_provider`` fixture,
    which is built by ``make_testing_provider(mngr_ctx, modal_interface)``
    and bypasses this factory entirely.

libs/mngr_modal/imbue/mngr_modal/backend.py
```diff
@@ -452,43 +441,8 @@ def build_provider_instance(
...
-                modal_interface: ModalInterface = DirectModalInterface()
-            case ModalMode.TESTING:
-                testing_root = mngr_ctx.profile_dir / "modal_testing"
-                testing_root.mkdir(parents=True, exist_ok=True)
-                modal_interface = TestingModalInterface(
-                    root_dir=testing_root,
-                    concurrency_group=mngr_ctx.concurrency_group,
-                )
-            case _ as unreachable:
-                assert_never(unreachable)
```
weishi-imbue, line 467: Always constructs a real ``DirectModalInterface``. This keeps test-only code
out of the production import graph (per ``modal_proxy``'s wheel build,
which excludes ``testing.py`` from the published wheel).

However, this will make this method untestable. There's no DI at this layer. (Same pattern from all other providers.)
This may be ok since that this method is orchestration glue with real work delegated to four pieces, each
independently tested elsewhere with ``TestingModalInterface`` injected at the right layer:
1. ``_derive_modal_names(name, config, mngr_ctx)`` -- Tested directly in ``testing_provider_test.py`` under
2. ``ModalProviderBackend._get_or_create_app`` and
    ``get_volume_for_app`` -- the app-registry layer. Tested in
    "Backend App Registry Tests"; both take ``modal_interface`` as a
    parameter, so the fake is injected directly.
3. ``ModalProviderApp(...)`` -- the app/volume holder. Tested in
    "ModalProviderApp Tests" by constructing the model with
    ``modal_interface=testing_modal`` (a pydantic constructor field).
4. ``ModalProviderInstance(...)`` -- the actual provider behavior.
    Exercised by the many tests using the ``testing_provider`` fixture,
    which is built by ``make_testing_provider(mngr_ctx, modal_interface)``
    and bypasses this factory entirely.

libs/mngr_modal/imbue/mngr_modal/testing_provider_test.py
```diff
@@ -926,80 +929,97 @@ def test_on_connection_error_clears_caches(
...
-    )
-    assert isinstance(instance, ModalProviderInstance)
-    assert instance.app_name == "build-test"
-
-    # Clean up the app registry
-    ModalProviderBackend.close_app("build-test")
-
-
-def test_build_provider_instance_environment_name_derived_from_prefix(
+def test_derive_modal_names_environment_name_derived_from_prefix(
```
WeishiZ, line 932: This covers the pure naming logic extracted from previous `build_provider_instance`. 
The extracted helpers become unit-testable without any Modal interface at all.

libs/mngr_modal/imbue/mngr_modal/testing_provider_test.py
```diff
@@ -926,80 +928,90 @@ def test_on_connection_error_clears_caches(
...
+    )
+    # App name must leave room for the state-volume suffix.
+    assert len(app_name) <= MODAL_NAME_MAX_LENGTH
+
+
+def test_production_import_does_not_load_modal_proxy_testing() -> None:
+    """Regression: importing mngr_modal.backend (the production-side entry
+    point that ``mngr forward`` triggers via the plugin registry) must not
+    drag the test-only in-process fake into the import graph.
+
```
joshalbrecht: this is probably too dumb of a test to have continually into the future

libs/mngr_modal/imbue/mngr_modal/config.py
```diff
@@ -1,34 +1,21 @@
...
 from imbue.mngr.primitives import UserId
 
 
-class ModalMode(UpperCaseStrEnum):
-    """Mode for the Modal provider backend."""
-
-    DIRECT = auto()
-    TESTING = auto()
-
-
```
joshalbrecht, line 20: I don't particularly want to remove this--it's an Enum because it's *supposed* to be expanded to have *three* values:

1. direct
2. testing
3. *remote*

ie, I wanted a way of being able to proxy modal connections *through our imbue_cloud gateway*, basically

I'm not really opposed to removing a TESTING enum from our code (that is, I admit, a bit silly), but we do probably want the mode to still be there (and the dispatch for construction somehow)

If you want, you can go add the remote value to the enum (maybe call it PROXIED), and just raise a notimplementederror if that value is selected

joshalbrecht:
see other comments, but also:

you need to be *REALLY* careful with this change--it's VERY easy for modal environments/apps to end up "leaking" when running tests locally or in CI. Make sure that that is not happening!!!  (that was part of the reason for testing stuff being mixed into here in a way where it normally would not have been). See ev for additional context on that

libs/mngr_modal/imbue/mngr_modal/testing_provider_test.py
weishi-imbue: removed.

libs/mngr_modal/imbue/mngr_modal/config.py
weishi-imbue: Updated to add the enum back with direct and proxied

weishi-imbue:
> you need to be _REALLY_ careful with this change--it's VERY easy for modal environments/apps to end up "leaking" when running tests locally or in CI. Make sure that that is not happening!!! (that was part of the reason for testing stuff being mixed into here in a way where it normally would not have been). See ev for additional context on that

This existing `@pytest.mark.modal` test interface stays the same so they should be doing setup and teardown as before. I saw we had a CI cron to clean up in case per test/session teardown didn't run. The leakage risk of this change looks very low.

joshalbrecht:
lgtm, but worth verifying that, in fact, in practice, we are not leaking any environments!

you should be able to see by running the tests / using the modal cli (and is worth doing if you haven't checked). It's particularly easy for this to go wrong if the names of the test envs ever fail to match the particular pattern that is required for cleanup

PR #1521, opened 2026-05-06
workspace_server: rewrite Location headers on 3xx redirects in proxy

gnguralnick:
## Summary

The workspace_server forwarder rewrites cookies and HTML body content for proxied responses, but never rewrote the `Location` header on 3xx redirects. Backends issuing redirects (e.g. OAuth flows, 302-after-POST) sent absolute URLs pointing at the backend's own host, which the browser then followed off-proxy and broke the session.

This PR adds `rewrite_location_header` in `proxy.py` and wires it into `_build_proxy_response` to apply on 3xx responses, alongside the existing cookie / HTML rewriters. Same-origin absolute URLs (those targeting the backend's own host:port) are rewritten into proxy-relative paths under the service prefix, since the backend's localhost address is not reachable from the browser. Protocol-relative, fully external, and relative-path Locations pass through unchanged.

Replaces #1509 -- same logic, recreated as a single clean commit off `main` to avoid the merge-history mess. Drops the `else: pass` ratchet workaround in favor of a `rewritten_value` local that gives the elif a meaningful else branch.

## Test plan

- [x] `proxy_test.py` -- 16 unit tests covering rewriter cases (absolute path, query/fragment preservation, no double-prefix, exact-prefix match, same-origin URL collapse, external URL passthrough, protocol-relative passthrough, relative path passthrough, dot-relative path, fragment-only, site-absolute root).
- [x] `service_dispatcher_test.py` -- 4 integration tests exercising real `RedirectResponse` (absolute path, same-origin URL, external URL) and a non-redirect 200 sanity check.
- [x] `just test-quick "apps/minds_workspace_server/imbue/minds_workspace_server/proxy_test.py apps/minds_workspace_server/imbue/minds_workspace_server/service_dispatcher_test.py"` -- 55 passed.
- [x] `just test-quick "apps/minds_workspace_server -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'"` -- 268 passed (1 unrelated pre-existing failure in `agent_manager_test.py::test_start_observe_spawns_long_lived_subprocess`, also fails on clean `origin/main`).
- [x] `ruff check` + `ruff format --check` -- clean.

### DIFF
--- a/apps/minds_workspace_server/imbue/minds_workspace_server/proxy.py
+++ b/apps/minds_workspace_server/imbue/minds_workspace_server/proxy.py
@@ -1,5 +1,6 @@
 import re
 from typing import Final
+from urllib.parse import urlparse
 
 from imbue.imbue_common.pure import pure
 from imbue.minds_workspace_server.primitives import ServiceName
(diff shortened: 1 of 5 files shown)


gnguralnick:
Migrated to imbue-ai/forever-claude-template#62 (apps/minds_workspace_server now lives there as apps/system_interface). Closing this PR.

PR #1524, opened 2026-05-07
Remove minds_workspace_server (migrated to forever-claude-template)

gnguralnick:
## Summary

- Delete `apps/minds_workspace_server/` and `specs/minds-workspace-server/` — the source has been migrated into `forever-claude-template` at `apps/system_interface/` (companion PR: imbue-ai/forever-claude-template#48). The internal Python module path (`imbue.minds_workspace_server`), CLI binary name (`minds-workspace-server`), env vars, and class names are unchanged — only the on-disk directory and uv project name move.
- Drop the `--cov` flag from minds-workspace-server's previous test invocation in any places that referenced it; remove the Dockerfile.release.extras frontend block; retarget `apps/minds/scripts/propagate_changes` to the new path; update active docs/skill/audit-doc; add a historical note to 8 specs that referenced the old path; regenerate `uv.lock` to drop the workspace member.
- `imbue_common` ratchets: make `find_code_in_init_files` skip gitignored dirs (e.g. `.venv/`) so standalone-project layouts don't trip the ratchet on third-party `__init__.py` files. Same fix mirrored in the FCT companion PR's vendored copy.

## Test plan

- [x] `cd libs/imbue_common && uv run pytest -k 'init or ratchet'` — 80 passed
- [x] `just test-quick` (mngr quick suite) — passes (the two failures observed locally were a missing-binary host issue and an xdist flake, both unrelated to this diff)


joshalbrecht:
lgtm

PR #1525, opened 2026-05-07
minds: stop forward subprocess from pinning CPU after agent destroy

gnguralnick:
## Summary

The mngr_forward subprocess kept paramiko transport threads alive against torn-down agents because `SSHTunnelManager._reverse_tunnels` was never pruned: every 30 seconds the health-check loop re-handshook against ports that no longer existed, leaking a transport thread per attempt and eventually pinning a CPU.

This PR applies four independent fixes:

- **Fix 1 -- prompt teardown.** `SSHTunnelManager` exposes `remove_reverse_tunnels_for_agent` (and a host-keyed sibling) that drop the registry entries, clear backoff state, and close the underlying SSH client whenever its last tunnel goes away. `LatchkeyDestructionHandler` is wired into the new API so per-agent destroys release tunnel resources promptly. The `HostDestroyedEvent` path already fans out to per-agent destroys, so host destroys are covered transitively.
- **Fix 2 -- bounded retry.** The 30s health-check loop now tracks per-tunnel `consecutive_failures` and a monotonic `next_attempt_at`, retries with exponential backoff capped at 5 minutes, and drops the tunnel entirely after 10 straight failed repair attempts. A successful repair clears the backoff state.
- **Fix 3 -- terminate the relay loop on half-closed channels.** The bidirectional relay (`select.select` over a socket + paramiko Channel) used to fall through silently when paramiko marked the channel readable but `recv_ready()` was False -- the half-closed-channel case. The loop spun at ~1M iters/sec on each torn-down agent, which is what produced the ~130% CPU pin observed on `mngr forward`. The relay now detects `eof_received` / `closed` and returns. Regression test in `relay_test.py` uses a peer-closed Unix socket to simulate the EOF wakeup pattern.
- **Fix 4 -- single source of truth for the relay loop.** Both the desktop client's reverse-tunnel handler and `mngr forward`'s direct-tcpip path used near-identical `_relay_step` / `_relay_data` copies, which is exactly how Fix 3 had to be applied twice. Lifted into a new `libs/mngr_forward/imbue/mngr_forward/relay.py` module, imported by both call sites. Costs one new workspace dep edge (`apps/minds` → `imbue-mngr-forward`).

Each fix is exercised independently by `apps/minds/imbue/minds/desktop_client/test_ssh_tunnel_leak.py`, which spins up an in-process paramiko SSH server, sets up reverse tunnels through `SSHTunnelManager`, then closes the test servers and asserts the manager actually releases its connection state.

## Test plan

- [x] Acceptance tests in `test_ssh_tunnel_leak.py` cover Fix 1 + Fix 2 against a real paramiko transport.
- [x] `relay_test.py::test_relay_data_terminates_when_channel_has_received_eof` regression test for Fix 3.
- [x] Live verification on the FCT vendor stack: stopping an agent dropped `mngr forward` from 130-145% CPU (3 R-state relay threads at 44% each) to 31% (1 R-state thread, asyncio retry loop), confirming the relay-spin specifically.
- [x] Unit tests pass: 188/188 mngr_forward + 646/646 minds locally.
- [ ] CI: full offload run.


joshalbrecht:
lgtm except that I am skeptical of "Fix 2 -- bounded retry."

that should probably be removed, or at least uncapped--you want this to recover if you go offline and then come back online, not "give up forever". Just let it exponential backoff until it's at a ~5 min polling interval, and stay there forever

seems like that should work?

joshalbrecht:
FYI I'm updating this PR to address my comment (I want to redeploy a new mind with that fix)

joshalbrecht:
I merged it from here: https://github.com/imbue-ai/mngr/pull/1558

Merging this to main.

PR #1527, opened 2026-05-07
Fix events trailing

hynek-urban:
The combination of:

- result.stdout eating trailing newlines
- and our logic considering lines without a newline as "in-flight"

Resulted in the last event always getting "stuck" in the queue (until a new event came).

Originally observed in permission request scenarios (a permission request wouldn't show up until I'd ask the agent again). After this fix, permission requests work as expected.


joshalbrecht:
lgtm, but that's gross--that really just looks like a pyinfra bug

PR #1528, opened 2026-05-07
minds: set LATCHKEY_DISABLE_COUNTING=1 in workspaces

hynek-urban:
The host-side latchkey gateway already counts as one active user. The workspace-side latchkey CLIs run in client mode against that gateway, so leaving counting enabled there would double-count every agent as a separate user. Inject LATCHKEY_DISABLE_COUNTING=1 alongside the existing latchkey env wiring so the per-workspace daily ping is suppressed.


joshalbrecht:
lgtm

PR #1529, opened 2026-05-07
mngr_tmr: add CI workflow for TMR + integrator_branch event

qi-imbue:
## Summary

- New manually-dispatched GitHub Actions workflow (`.github/workflows/tmr.yml`) that runs `mngr tmr` against Modal, uploads the HTML report as a `tmr-report` artifact, and opens a draft PR for the integrator branch (`mngr-tmr/integrated-*`). Provider is hardcoded to `modal`; `test_paths`, `pytest_args`, and `agent_type` are exposed as inputs, with defaults reproducing the existing local invocation against `libs/mngr/imbue/mngr/e2e/test_basic.py -m release` with `--agent-type yolo`.
- New `integrator_branch` event on the `mngr tmr` structured stdout stream (in `--output-format jsonl`/`json`), so the workflow can detect the branch via `jq` instead of parsing human-formatted output.
- `_print_run_commands` and the reintegrate path's status messages are now gated on HUMAN output mode, keeping the JSONL stream pure structured events.

## Notes

- The `yolo` agent type lives in user-scope mngr config and is not in the repo, so the workflow synthesizes it inline via `-S 'agent_types.yolo.parent_type=\"claude\"' -S 'agent_types.yolo.cli_args=[\"--dangerously-skip-permissions\"]'`. Without `--dangerously-skip-permissions`, agents would hang on per-tool permission prompts.
- PRs created by `GITHUB_TOKEN` do not trigger downstream workflows (GitHub recursion guard); `ci.yml` will not run on the autogenerated PR until someone closes/reopens it or pushes a commit.
- Required secrets/variables (`MODAL_TOKEN_ID`, `MODAL_TOKEN_SECRET`, `ANTHROPIC_API_KEY`) are already configured for the rest of `ci.yml`; nothing new needs adding.

## Test plan

- [ ] Trigger the workflow via Actions → TMR → Run workflow with the default inputs and confirm:
  - [ ] `tmr-report` artifact is uploaded and contains `index.html` plus `events.jsonl`.
  - [ ] A draft PR is opened against `main` with `mngr-tmr/integrated-*` as the head branch.
- [ ] `just test-offload` passes in CI (covers the cli changes including the four new `_emit_integrator_branch` unit tests).

### DIFF
--- a/.github/workflows/tmr.yml
+++ b/.github/workflows/tmr.yml
@@ -1,0 +1,107 @@
+name: TMR
+
+on:
+  workflow_dispatch:
+    inputs:
+      test_paths:
+        description: "Test paths (space-separated, passed as positional args before --)"
+        required: true
+        default: "libs/mngr/imbue/mngr/e2e/test_basic.py"
+      pytest_args:
+        description: "Pytest args after `--` (must be valid bash; quote values containing spaces, e.g. `-m \"foo and bar\"`)"
+        required: true
+        default: "-m release"
+      agent_type:
+        description: "Agent type (e.g. yolo, claude)"
+        required: true
+        default: "yolo"
+
+jobs:
+  tmr:
+    runs-on: ubuntu-latest
+    timeout-minutes: 240
+    permissions:
+      contents: write
+      pull-requests: write
+    env:
+      MODAL_TOKEN_ID: ${{ vars.MODAL_TOKEN_ID }}
+      MODAL_TOKEN_SECRET: ${{ secrets.MODAL_TOKEN_SECRET }}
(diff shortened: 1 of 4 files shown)


.github/workflows/tmr.yml
```diff
@@ -0,0 +1,107 @@
...
+
+      - name: Configure git identity
+        run: |
+          git config user.email "dev@imbue.com"
+          git config user.name "imbue-dev"
+
+      - name: Run TMR orchestrator
+        run: |
+          mkdir -p tmr-report
+          # The "yolo" agent type is normally defined in the user's personal
```
qi-imbue, line 61: I'd like to use the `modal` template but the Dockerfile is currently broken

PR #1530, opened 2026-05-07
Make parsing of various addresses consistent across all subcommands

qi-imbue:
## My summary

This change touches almost all subcommands, mostly refactor, but with some behavior change for consistent CLI argument handling (see details below).

The goal is that all subcommands should use the same facilities for parsing and resolving addresses. In order to do that, I've identified 5 types of addresses:
* `AgentAddress` refers to an **existing** agent. This is the only type that already exists.
* `HostAddress` refers to an existing host.
* `NewAgentLocation` is specifically for the argument of `mngr create`: this has been split from `AgentAddress` in order to tighten the type constraint of the latter.
* `HostedLocation` refers to a location on a host, but the host can either be missing (meaning a local location) or be replaced by an agent (in which case the path can be omitted to refer to the agent's working directory). This is used in `mngr pull`, `mngr create --from` and other places, and some of them used slightly different syntaxes; these are now all unified.
* `AgentOrHostAddress` refers to either an agent or a host. This is used in `mngr limit`, `mngr snapshot`, `mngr wait`. Previously they all had different parsing and resolution code; they all used to distinguish agents from hosts dynamically, but now they are distinguished literally: `foo` unambiguously refers to an agent, and `@foo`  unambiguously refers to a host. Support for qualified addresses, not just names, has also been added.
There are a few more auxiliary types. See `primitives.py` for everything. Subcommands now uses Click ParamType adapters to get those typed addresses directly without invoking parsers on their own. This means that all internal code now use typed addresses instead of `str` everywhere. Some commands left over from the address change now get address support for free, and we can be sure that future subcommands will support the same address syntax.

I'm also trying to made sure that all the subcommands only use a small number of helpers from `api/find.py` to resolve those addresses into objects to perform operations on, instead of having their own bespoke logic. This has not been entirely successful (see my review notes) and I'll pursuit this further in a followup/refactor PR.

I've reviewed the PR almost line and line. I'm satisfied that at least parsing is consistent across all subcommands, with more to do to unify resolution code paths.

## Claude's Summary

Every mngr subcommand that accepts an address-shaped argument now parses it through one of a small set of typed parsers at the Click boundary, and resolves it through one of a small set of helpers in `api/find.py`. The CLI is uniform: same syntax, same disambiguation rules, same error messages across `mngr connect`/`destroy`/`exec`/`push`/`pull`/`pair`/`create`/`rename`/`event`/`transcript`/`wait`/`snapshot`/etc.

### User-visible behavior changes

- **`mngr push`/`mngr pull` accept the full `[AGENT]@HOST[.PROVIDER]:PATH` syntax** in their `TARGET`/`SOURCE` argument. Previously these had a bespoke parser that only understood `AGENT[:PATH]`, so an input like `mngr push my-agent@m1.modal:/path` failed with a misleading error.
- **`mngr event` / `mngr transcript` / `mngr wait` / `mngr snapshot create` / `mngr snapshot list` accept agent-or-host arguments with text-only disambiguation.** Agents are `NAME` (or `NAME@HOST[.PROVIDER]`); hosts are `@HOST[.PROVIDER]` or a bare `host-...` ID. Previously these commands consulted discovered state to decide which was which.
- **Bare names without `@` are always interpreted as agent names** in those commands. To target a host by bare name, write `@HOST`. Host IDs (`host-...`) are still recognized without the `@` for muscle-memory compatibility.
- **`HostName` no longer permits dots.** Host names validate as `SafeName` (alphanumeric + dashes/underscores). The `HOST.PROVIDER` qualifier is now exclusively the parser's responsibility.
- **`mngr create NAME@.PROVIDER`** (bare `.PROVIDER` meaning "create a new host on this provider") is preserved.
- **Stricter parse-time validation:** invalid forms like `host..provider` reject at Click parse time with a clean error instead of failing later inside the api layer.
- **Multi-match errors now list every colliding agent** in `NAME@HOST.PROVIDER` form so the user can pick.

### Typed address shapes

In `imbue.mngr.primitives` (parsers in `imbue.mngr.api.address_parsers`):

- `HostAddress` — `[@]HOST[.PROVIDER]`. Host required.
- `AgentAddress` — `NAME[@HOST[.PROVIDER]]`. Agent required.
- `NewAgentLocation` — `[NAME][@[HOST][.PROVIDER]][:PATH]`. Positional of `mngr create`.
- `HostedLocation` — `[NAME[@HOST[.PROVIDER]]][:PATH]` or a bare path. Source/target for `create --from`, `pair`, `push`, `pull`.
- `AgentOrHostAddress = AgentAddress | HostAddress` — for commands that accept either, disambiguated by text alone.

Plus type aliases `AgentNameOrId = AgentId | AgentName`, `HostNameOrId = HostId | HostName`.

### Resolution helpers (api/find.py)

Two tiers, used uniformly by every subcommand that needs them:

- **Filter (lightweight, in-memory):** `filter_one_agent` / `filter_all_agents` / `filter_one_host` / `filter_all_hosts`.
- **Find (heavy, runs discovery + materialization):** `find_one_agent(address, mngr_ctx)` returns `(AgentInterface, OnlineHostInterface)`; `find_all_agents(addresses, ..., mngr_ctx)` returns `list[AgentMatch]`. The new `materialize_agent(host_ref, agent_ref, mngr_ctx)` helper is the reusable bridge between the two tiers.

The old `find_and_maybe_start_agent`, `resolve_host_reference`, `resolve_agent_reference`, `find_agent_by_address` are gone or renamed for consistency.

### Notable internal changes worth flagging

- `AgentDetails.address` / `HostDetails.address` are cached properties returning typed addresses, so callers don't reconstruct addresses from id/name/provider fields.
- `Host.host_name: HostName` is now stored explicitly at construction (rather than derived from the SSH connector). This fixes a latent bug where multiple local-docker containers all surfaced the connector hostname `127.0.0.1`, producing spurious duplicate-host warnings.

## Test plan

- [x] mngr + plugin unit/integration: `just test-quick libs/mngr libs/mngr_wait libs/mngr_pair libs/mngr_kanpan libs/mngr_schedule libs/mngr_file -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'` — 4888 passed
- [ ] Acceptance/release tests via CI

### Manual smoke tests to run

- [ ] `mngr push my-agent@host.modal:/path local/dir` parses cleanly (previously errored)
- [ ] `mngr pull my-agent@host.modal:/path ./local-dir` likewise
- [ ] `mngr event my-agent` works; `mngr event @my-host` works; `mngr event @host-abc123` (bare ID) works
- [ ] `mngr transcript my-agent` likewise
- [ ] `mngr wait my-agent`, `mngr wait @my-host`, `mngr wait host-abc123` all resolve correctly
- [ ] `mngr snapshot create my-agent @my-host` accepts both kinds in the same positional list
- [ ] `mngr create NAME@host` (existing host), `NAME@.modal` (new host on provider), `NAME:/explicit/path` all still work
- [ ] `mngr create --from /local/path` (bare path) and `mngr create --from agent@m1.modal:/sub` both resolve
- [ ] `mngr pair my-agent` and `mngr pair my-agent@host.modal` start the sync as before
- [ ] `mngr connect`/`destroy`/`exec`/`rename` against an `agent@host.provider`-qualified target disambiguate across providers
- [ ] Two agents with the same name on different hosts: any command targeting that name lists both with their `NAME@HOST.PROVIDER` form and asks to disambiguate
- [ ] Invalid inputs (`mngr connect host..provider`, `mngr create name@`, `mngr push some@badhost..modal`) reject with clean Click error messages
- [ ] A local-docker host with multiple containers shows distinct names in `mngr list --hosts` rather than collapsing onto `127.0.0.1`


libs/mngr/imbue/mngr/api/address_parsers.py
```diff
@@ -0,0 +1,267 @@
...
+        raise UserInputError(
+            f"Invalid host address '{s}': contains more than one dot. Expected format: HOST[.PROVIDER]"
+        )
+    if dot_count == 0:
+        return (s, "")
+    host_str, provider_str = s.split(".", 1)
+    return (host_str, provider_str)
+
+
+@pure
```
qi-imbue, line 242: I wish Python has Go's [strings.Cut](https://pkg.go.dev/strings#Cut) but meh this is fine

libs/mngr/imbue/mngr/api/address_parsers.py
```diff
@@ -0,0 +1,267 @@
+"""Parsers that produce the typed address shapes shared across mngr.
```
qi-imbue, line 1: These are all reasonable - I in fact hand tweaked some of them

libs/mngr/imbue/mngr/api/discover.py
```diff
@@ -188,3 +190,40 @@ def discover_hosts_and_agents(
         # does not leak resources even though get_all_provider_instances is called again.
         logger.debug("Event stream was stale (not all identifiers found), falling back to full scan")
         return _run_discovery(mngr_ctx, None, include_destroyed, reset_caches)
+
+
+def discover_by_address(
```
qi-imbue, line 195: This is moved from the old agent_addr.py.

It looks reasonable and there are some legitimate uses of it, but I don't like how some CLI commands need to use this - they should be able to use one of the higher-level find.py helpers instead. For a follow-up refactor.

libs/mngr/imbue/mngr/api/events.py
```diff
@@ -128,112 +131,91 @@ class _AllEventsStreamState(MutableModel):
 
 
 def resolve_events_target(
```
qi-imbue, line 133: It's hard to review this diff inline - I looked at this side by side and it looks reasonable.

The most notable change is that agent and host addresses are now distinguished on the type level (instead of resolved dynamically) and there're now two parallel code paths for deriving host event target and agent event target. There's a bit of duplication between the two code paths but also enough differences, so keeping them separate is not too bad.

libs/mngr/imbue/mngr/api/find.py
```diff
@@ -10,6 +10,7 @@
 from imbue.imbue_common.logging import log_call
```
qi-imbue, line 10: This file sees the most change.

- All the parsing code has been moved elsewhere; this module now accepts typed addresses.
- I redesigned the API a little bit. The main helpers provided by this module now fall into two categories: `filter_*` functions that do filtering, and `find_*` functions do discovery + filtering. Ideally I want `cli/*.py` to only ever use the latter family because it's more high-level, but that's another refactor.

libs/mngr/imbue/mngr/cli/address_params.py
```diff
@@ -0,0 +1,168 @@
+"""Click ``ParamType`` adapters that parse mngr address strings into typed values.
```
qi-imbue, line 1: This is what adapts all the parsers into Click `ParamType`s.

libs/mngr/imbue/mngr/cli/agent_utils.py
```diff
@@ -89,116 +63,56 @@ def select_agent_interactively_with_host(
...
-    Returns (agent, host) tuple, or None if the user cancelled interactive selection.
-    Raises UserInputError if no agent specified and not running in interactive mode.
+    Returns ``(agent, host)``, or ``None`` if the user cancelled interactive
+    selection. Raises :class:`UserInputError` if no address is given and the
+    session is not interactive.
     """
-    if agent_identifier is not None:
+    if address is not None:
         if host_filter is not None:
-            # When host_filter is provided (separate from address syntax),
```
qi-imbue, line 169: This is now simplified by simply merging the `host_filter` into the address.

Also: this is only ever used by `mngr pair`. I would actually just remove `host_filter` outright but it seems that `mngr pair` is actually due a bigger simplification.

libs/mngr/imbue/mngr/cli/connect.py
```diff
@@ -18,10 +18,11 @@
```
qi-imbue, line 18: All mechanical changes. `mngr connect` is in fact slightly annoying - it could almost use `find_agent_for_command`, but it has an additional smartness of defaulting to the latest agent when running non-interactively.

libs/mngr/imbue/mngr/cli/pull.py
qi-imbue, line 1: The changes in this file (and `push.py`) are mechanical - the added lines look roughly the same and could be deduplicated. However, the API of `pull`/`push` is pre-address and could use some simplification, so I'm deferring changes to these two files in a followup refactor/cleanup.

libs/mngr/imbue/mngr/cli/snapshot.py
```diff
@@ -90,50 +95,32 @@ class SnapshotDestroyCliOptions(CommonCliOptions):
 # =============================================================================
 
 
-def _find_host_across_providers(
-    host_identifier: str,
-    mngr_ctx: MngrContext,
-) -> tuple[HostId, ProviderInstanceName] | None:
-    """Find a host by ID, name, or `host.provider` form across all providers.
+def _bucketize_mixed_identifiers(
```
qi-imbue, line 98: Agent and host targets are now disambiguated by how they look (`foo` is an agent, `@foo` is a host) and no longer depends on discovering hostnames and agent names.

libs/mngr/imbue/mngr/hosts/host.py
```diff
@@ -243,18 +243,21 @@ class Host(OuterHost, BaseHost, OnlineHostInterface):
     provider_instance: ProviderInstanceInterface = Field(
         frozen=True, description="The provider instance managing this host"
     )
-
-    # is_local, _ensure_connected, _close_paramiko_client, disconnect, and __del__
-    # are inherited unchanged from OuterHost.
+    host_name: HostName = Field(
```
qi-imbue, line 246: The SSH connector was using IP addresses for hostnames before. I believe this has always been wrong, but it only surfaced because dots are now disallowed in hostnames when parsing host addresses.

libs/mngr/imbue/mngr/primitives.py
qi-imbue, line 1: Now explicitly modelling all types of addresses using types!

libs/mngr/imbue/mngr/hosts/host.py
joshalbrecht: yes, and I fixed something related to this recently in another branch as well

glad that we caught this here, that gives me even more confidence that we're doing this more correctly now :)

libs/mngr/imbue/mngr/cli/connect.py
joshalbrecht: not critical, but we could probably remove that fanciness--it's not clear how mngr connect ever *can* run "non-interactively"!

(non blocking for this PR, just mentioning)

joshalbrecht:
lgtm!  merging now

PR #1531, opened 2026-05-07
actually use the default field to specify defaults for mngr create

evgunter:
## Summary

- `--type` now defaults to `"claude"` via click instead of being `None` and falling back to "claude" later in `api/create.py`.
- `--start-on-boot` now defaults to `False` instead of being `None` and treated as falsy later.
- `--worktree-base-folder`'s help text said `~/.mngr/worktrees/` but the actual default is `<host_dir>/worktrees`; the help text now reflects reality.

## Notes

The `--type` change required `_resolve_agent_type_name` to take an `is_type_explicit` flag so the positional `AGENT_TYPE` argument still takes precedence when `--type` is left at its default. The conflict check uses `is_param_explicit(ctx, "type")` so `mngr create x codex` no longer falsely conflicts with the implicit `--type=claude` default.

## Test plan

- [x] `just test-quick libs/mngr/imbue/mngr/cli` -- 1423 passed
- [x] `just test-quick "libs/mngr -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'"` -- 3822 passed
- [x] `uv run mngr create --help` shows correct defaults
- [ ] CI


joshalbrecht:
I'm onboard with these changes in general, except it does sketch me out a little bit that we default to claude like this

could we perhaps, instead, make the type *required*, and *not* have a default? (but by convention we set the default during the installer process by writing into the users home config?)  And we make them select it during installation?

this is a more philosophical point, but seeing this change makes me realize we should probably be doing something more like that...  thoughts?

evgunter:
i agree on the philosophical point--will implement

PR #1532, opened 2026-05-07
mngr list: capture provider warnings into result.warnings (json/jsonl)

weishi-imbue:
## Symptom
Provider non-fatal warnings (Vultr \"API key not configured\", Docker \"daemon unavailable\", Lima \"limactl missing\") reach the human watching stderr but are **invisible to programmatic consumers** of `mngr list --format json` / `--format jsonl`. From the JSON's perspective, those providers just return zero hosts silently — there's no way for a script to tell \"Vultr returned 0 hosts because no API key\" apart from \"Vultr returned 0 hosts because there are no instances.\"

Repro before this PR:
```sh
$ MODAL_CONFIG_PATH=/dev/null mngr list --format json --on-error continue 2>/dev/null | jq '.warnings'
null   # warnings field doesn't exist
```

## Fix
Add a structured warnings channel to the listing pipeline:

| layer | addition |
|---|---|
| `api/list.py` | `WarningInfo` model (mirrors `ErrorInfo`); `ListResult.warnings: list[WarningInfo]`; `on_warning` callback in `_ListAgentsParams` and `list_agents()`; per-call loguru sink that captures WARNING records into `result.warnings` |
| `cli/list.py` | `_emit_jsonl_warning` (mirrors `_emit_jsonl_error`); wired into JSONL path; `_emit_json_output` adds top-level `warnings` field |

**The capture mechanism is a per-call loguru sink** registered in a try/finally inside `list_agents`. Existing provider code (`logger.warning(...)` in Vultr/Docker/Lima) is captured automatically — **zero provider-side changes**. The existing stderr and log-file sinks remain unaffected; this just adds a third receiver scoped to the listing call.

## Pattern fidelity
- `WarningInfo` mirrors `ErrorInfo` exactly (FrozenModel, Field-described).
- The sink is a callable `MutableModel` class (`_WarningCaptureSink`), mirroring `_LimitedJsonlEmitter`'s callable-state pattern in `cli/list.py`. Avoids the inline-functions ratchet (=0 in mngr) and the functools.partial ratchet.
- The sink filter narrows to **exactly WARNING level**, not \"WARNING and above\" — ERROR records are already represented as `ProviderErrorInfo` via the per-provider catch path, so capturing them as warnings would double-categorize.

## Empirical matrix (no Modal creds)

| invocation | before | after |
|---|---|---|
| `--format json --on-error continue` | `{agents:[10], errors:[modal]}` | `{agents:[10], errors:[modal], warnings:[{message:\"Vultr API key not configured...\"}]}` |
| `--format jsonl --on-error continue` | shapes `{error:1, agent:10}` | shapes `{warning:1, error:1, agent:10}` |
| `--format human --on-error continue` | stderr text + clean stdout table | unchanged |
| Real Modal creds | `{agents:[10], errors:[]}` | `{agents:[10], errors:[], warnings:[{Vultr API key...}]}` (Vultr still warns when no key) |

## Test plan
- [x] 4 new unit tests in `api/list_test.py` — capture sink behavior, on_warning callback, level filter (rejects ERROR, INFO, DEBUG), and a smoke test that `list_agents` returns a populated `warnings` field.
- [x] Empirical matrix verified end-to-end (json/jsonl populated, human/regression unchanged).
- [x] All 3826 mngr unit tests pass.
- [x] Ruff lint + format clean. test_no_type_errors, test_prevent_inline_functions, test_prevent_functools_partial all pass.

## Scope
Second of four follow-ups to PR #1461. Independent of #1516 (already-merged JSON-validity fix); independent of the remaining two (`--on-error abort` partial JSON, events-file completeness).

### DIFF
--- a/libs/mngr/imbue/mngr/api/list.py
+++ b/libs/mngr/imbue/mngr/api/list.py
@@ -5,6 +5,7 @@ from datetime import datetime
 from datetime import timezone
 from threading import Lock
 from typing import Any
+from typing import cast
 
 from loguru import logger
 from pydantic import Field
@@ -100,11 +101,26 @@ class AgentErrorInfo(ErrorInfo):
         )
 
 
+class WarningInfo(FrozenModel):
+    """Information about a non-fatal warning encountered during listing.
+
+    Captured from `logger.warning(...)` calls (typically from providers reporting
+    "not configured" / "unavailable" states) via a per-call loguru sink, and
+    surfaced through the structured output channels (jsonl / json) so
+    programmatic consumers can see what humans see on stderr.
+    """
+
+    message: str = Field(description="The warning message as logged")
+
+
 class ListResult(MutableModel):
     """Result of listing agents."""
 
     agents: list[AgentDetails] = Field(default_factory=list, description="List of agents with their full information")
     errors: list[ErrorInfo] = Field(default_factory=list, description="Errors encountered while listing")
+    warnings: list[WarningInfo] = Field(
+        default_factory=list, description="Non-fatal warnings logged during listing (e.g. provider not configured)"
+    )
 
 
 class _ListAgentsParams(FrozenModel):
@@ -116,6 +132,7 @@ class _ListAgentsParams(FrozenModel):
     error_behavior: ErrorBehavior
     on_agent: Callable[[AgentDetails], None] | None
     on_error: Callable[[ErrorInfo], None] | None
+    on_warning: Callable[[WarningInfo], None] | None
     field_generators: dict[str, dict[str, Callable[[AgentInterface, OnlineHostInterface], Any]]] = Field(
         default_factory=dict,
     )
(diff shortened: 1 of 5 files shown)


libs/mngr/imbue/mngr/api/list.py
```diff
@@ -207,6 +221,39 @@ def list_agents(
...
 
+@contextmanager
+def _capture_warnings_into(
+    result: ListResult,
+    results_lock: Lock,
+    on_warning: Callable[[WarningInfo], None] | None,
+) -> Iterator[None]:
+    """Capture WARNING-level loguru records into ``result.warnings`` for the duration of the block."""
+
+    def sink(message: Any) -> None:
```
weishi-imbue, line 232: I think a nested method here encapsulates better.  Increasing the ratchet.

PR #1533, opened 2026-05-07
mngr_modal: catch silent cleanup failures of modal_subprocess_env

weishi-imbue:
### Symptom
Some tests creates modal app from subprocess. (E.g. `test_mngr_create_echo_command_on_modal`, uses `modal_subprocess_env`, creates via `mngr create ... modal`). It never registered in parent process pytest’s `worker_modal_environment_names` thus it’ll miss the leakage detection. 

### Change:
1. surface the non-0 exit code. 
2. Register subprocess modal env for leakage detection.


weishi-imbue:
I saw a potential debate here of this change: 
#### Gains
1. More confidence in CI. If CI green, means code cleaned up all modal env including subprocess. 
2. Prevent regressions: delete_modal_environment regression, and maybe other modal name formula mismatch. 
#### Costs
1. Transient error (network err) may surface more often where user need to retry. Instead of letting 1 hour cron CI to catch it. Could be annoying. But this is existing expectation for in-process tests already.

joshalbrecht:
lgtm, thanks for fixing!

CC @evgunter in case he sees anything else worth changing here

evgunter:
nice!

PR #1534, opened 2026-05-07
mngr list: write partial DISCOVERY_FULL snapshot when listing has errors

weishi-imbue:
## Symptom
When \`mngr list --on-error continue\` hits a per-provider failure, the events file (\`~/.mngr/events/mngr/discovery/events.jsonl\`) gets a \`DISCOVERY_ERROR\` event for the failed provider (since PR #1516) but NO fresh \`DISCOVERY_FULL\` with the agents from providers that did succeed. Consumers tailing the file (\`mngr observe --discovery-only\`, minds desktop, future tools) are stuck reading whatever stale snapshot existed before the failure.

## Fix
Always write the snapshot when the listing is unfiltered. When per-provider errors occurred, mark it \`is_partial=True\` so consumers know not to treat the captured state as authoritative.

### Schema change

\`\`\`python
class FullDiscoverySnapshotEvent(EventEnvelope):
    ...existing agents / hosts fields...
    is_partial: bool = Field(default=False, ...)  # NEW
\`\`\`

\`make_full_discovery_snapshot_event\` and \`write_full_discovery_snapshot\` gain a corresponding \`is_partial: bool = False\` parameter; existing callers that don't pass it (all the test-only ones) keep working at the default.

### Behavior

| listing outcome | events file before | events file after |
|---|---|---|
| Unfiltered + success | DISCOVERY_FULL (no is_partial field) | DISCOVERY_FULL with \`is_partial=False\` |
| Unfiltered + per-provider errors | nothing (skipped) | DISCOVERY_FULL with \`is_partial=True\` covering the agents from successful providers, **plus** the existing per-provider DISCOVERY_ERROR events |
| Filtered (CEL or provider_names) | nothing | nothing (unchanged) |

\`is_partial\` is the consumer's signal to correlate with adjacent \`DISCOVERY_ERROR\` events to see which providers were skipped, and to NOT treat the partial snapshot as authoritative state.

## Schema-change risk + migration

\`FrozenModel\` has \`extra=\"forbid\"\` → loguru parsers running OLD code (parsers built before this PR) will reject lines containing the new \`is_partial\` field with \`DiscoverySchemaChangedError\`. All in-tree consumers ship in lockstep:

- \`libs/mngr_forward/stream_manager.py\` (in-repo)
- \`libs/mngr/cli/complete_names.py\` (in-repo)
- \`apps/minds/desktop_client/forward_cli.py\` (in-repo)

No out-of-tree consumer of \`FullDiscoverySnapshotEvent\` was found. The only observable migration cost is: long-running \`mngr observe\` / minds desktop processes started before this PR need a restart to pick up the new schema.

## Empirical verification

\`\`\`sh
$ MODAL_CONFIG_PATH=/dev/null mngr list --on-error continue
  exit=1
  DISCOVERY_FULL delta: +1   (is_partial=True, agents=10 from lima/docker)
  DISCOVERY_ERROR delta: +1  (Modal)

$ mngr list   # real creds, success
  DISCOVERY_FULL delta: +1   (is_partial=False, agents=10)
\`\`\`

## Test plan

- [x] Inverted \`test_maybe_write_full_discovery_snapshot_skips_when_errors_present\` to \`..._writes_partial_when_errors_present\`. Asserts the snapshot IS written and \`is_partial=True\` and the partial agent appears.
- [x] Strengthened \`test_..._writes_when_unfiltered_and_error_free\` to also assert \`is_partial=False\`.
- [x] Other 4 \`_maybe_write_full_discovery_snapshot\` tests (provider-filter / include-filter / OSError / SSH-host-info) unchanged and passing.
- [x] All 3822 mngr unit tests pass.
- [x] Ruff lint + format clean. Empirical end-to-end probe confirms both code paths.

## Scope

Last of four follow-ups to PR #1461. Independent of #1516 (already merged) and #1532 (\`result.warnings\`). Once this lands, the discovery events file is the complete record of any listing — a programmatic consumer can fully reconstruct partial-success state without parsing \`mngr list\` stdout.


weishi-imbue:
The PR description claims consumers are "stuck reading whatever stale snapshot existed before the failure." That's only true if the periodic writer in mngr observe --discovery-only is also failing — and that writer uses ABORT + retry to handle transients. In the steady state, the periodic writer keeps the snapshot fresh; an ad-hoc mngr list failure doesn't strand consumers.
A partial discovery_full is a large change that benefits is unclear.

PR #1537, opened 2026-05-08
mngr list: emit valid partial JSON under --on-error abort

weishi-imbue:
## Summary

- `mngr list --format json/jsonl --on-error abort` now produces a valid (partial) document when a provider fails, instead of empty stdout. The captured error is reflected in `result.errors`/`event:error`, and the command still exits non-zero so scripts can detect failure from the exit code.
- Capture the MngrError (or its `ConcurrencyExceptionGroup` wrapping for main-thread raises inside an executor `with` block) into `result.errors` at the top of `list_agents` instead of re-raising.
- Fix `_run_list_iteration`'s empty-agents JSON path to dump `ErrorInfo` via `model_dump` rather than handing raw Pydantic objects to the JSON encoder.

This is the third in a series of follow-ups to #1461 (per-provider error isolation). PR-D (events file completeness) is separate.

## Test plan

- [x] `just test-quick libs/mngr -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'` — 3825 passed, 1 skipped.
- [x] Empirical probe: `mngr list --format json --on-error abort` with a configured failing provider yields valid JSON `{"agents": [], "errors": [...]}`, exit=1.
- [x] Empirical probe: `mngr list --format jsonl --on-error abort` yields one JSON-per-line `{"event": "error", ...}`, exit=1.
- [x] Empirical probe: `mngr list --on-error abort` (human format) prints `No agents found`, logs warning to stderr, exit=1.
- [ ] CI acceptance / release tests pass.

### DIFF
--- a/changelog/wz-mngr-list-abort-partial-json.md
+++ b/changelog/wz-mngr-list-abort-partial-json.md
@@ -1,0 +1,1 @@
+`mngr list --format json/jsonl --on-error abort` now emits a valid (partial) document when a provider fails: the captured error is included alongside any agents discovered before the abort, and the command exits non-zero. Previously the abort propagated as a raise and stdout was empty, leaving callers without parseable output.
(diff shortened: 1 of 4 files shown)


joshalbrecht:
lgtm, thanks!

PR #1539, opened 2026-05-08
mngr_latchkey

hynek-urban:
- Move most of the functionality related to Latchkey gateway to a separate package (`mngr-latchkey`).
- `mngr-latchkey` is not a proper manager plugin yet - it just exposes a Python API, no CLI commands.
  - Eventually, we may want to make this a proper plugin and support commands like `mngr latchkey ensure-gateway` but for now, there's no direct use.
- The minds app uses said Python API to start the gateway and to make sure that all agents have access to Latchkey.
- In a follow-up, I intend to move most of the permission request handling logic to mngr-latchkey, too (via gateway extensions).


joshalbrecht:
roughly lgtm, but I *would* really like us to make this work as a proper mngr plugin if we could...

like, logically, it seems like this should be a separate process that doesn't require minds to run at all? But perhaps subtle things might want to change about the architecture to enable that?

it actually kinda seems like a nice mirror of "mngr forward"?

hynek-urban:
I think it doesn't require minds to run at all. It's basically a Python library at the moment and it doesn't depend on minds. It spawns a long-running process, namely `latchkey gateway`. Maybe I just misunderstand what exactly a "manager plugin" brings in terms of desired functionality? What exactly would you like it to do? Are there things that you'd like to do which require a CLI interface (as opposed to a Python API)?

hynek-urban:
As discussed over Slack, I added a command-line interface, including the `mngr latchkey forward` command.

PR #1540, opened 2026-05-08
TMR: speed up Modal runs via /code worktree reuse + skipping per-agent initial snapshots

qi-imbue:
## Summary

Two TMR-only optimizations targeting Modal-provider runs.

### 1. Reuse the snapshotter's code repo via on-host git-worktree

- Pin the TMR snapshotter agent's work_dir to a hardcoded `/code` so every host built from the resulting snapshot has the git repo at a known on-host location.
- Source follow-up test agents from `<host>:/code` via git-worktree (which now works for same-host remote agents thanks to a2f812be6 on the base branch) instead of re-pushing the git history from the laptop. Avoids one full git mirror push per agent against Modal.
- Both behaviors are gated: the snapshotter pin only fires for the dedicated snapshotter call; the worktree-from-host-`/code` path only fires when an agent is being placed on an existing host AND `config.snapshot` is set. The unsnapshotted code path re-uploads from the laptop as before.

### 2. Disable Modal's per-agent initial snapshot for TMR runs

- Modal's `on_agent_created` hook normally takes a 60-90s filesystem snapshot after each agent so the host can be restarted after a hard kill. The hook fires *per agent*, so on TMR's pooled hosts (default 4 agents per host) it runs four times for the same host.
- TMR creates the snapshot it actually needs explicitly via `provider.create_snapshot` on the dedicated snapshotter; every other host TMR creates is ephemeral. The safety-net snapshot is dead weight.
- At the start of every TMR run, override `is_snapshotted_after_create=False` on the modal-backed providers TMR is about to use (for both `--provider` and `--integrator-provider`), preserving the rest of the user's config via `model_copy_update`. Non-modal providers and non-TMR users are untouched.

## Architecture-review notes worth flagging

The reviewer accepts (1) but pushed back on (2)'s mechanism. Filing here so they don't get lost:

- The `/code` magic constant in `launching.py` would be cleaner if threaded through `TmrLaunchConfig` (e.g. `snapshot_code_dir: Path | None`), so the contract between the snapshotter pin and the test-agent source location lives in the data flow rather than a module-level constant. Also, `/code` is hardcoded and unchecked — a path under the host's `host_dir` would survive more host-image variations.
- In-place mutation of `mngr_ctx.config.providers` is an unprecedented pattern in this codebase (everywhere else, plugin behavior is altered via hookimpls, not by reaching across into a sibling plugin's config). It also has a subtle ordering invariant (must run before any `get_provider_instance` call) and the side effect of materializing config entries that didn't previously exist when the user has no `[providers.modal]` block. Cleaner alternatives: a per-host signal via `NewHostBuildOptions` / `plugin_data` consumed by the Modal `on_agent_created` hook, or a derived `MngrContext` scoped to the TMR call. Even keeping the current shape, it should at least: (a) only patch when the user already configured an instance, (b) use the typesafe `to_update(field_ref()...)` form.

Happy to do either refactor if requested before merge.

## Test plan

- [x] Existing tmr unit + integration tests pass locally (179 passed after adding tests for the new behavior)
- [x] Ratchets / type-check / ruff pass locally (56 passed)
- [ ] CI green (offload)
- [ ] Manual smoke test: `mngr tmr --provider modal --use-snapshot ...` finishes faster than before, the per-agent log no longer shows "git push" from the laptop, and the per-agent "Creating initial snapshot for host..." log line is absent.


qi-imbue:
Code guardian flagged up two issues, my opinion:
1. This is actually fine
2. Do we want, on the API level, to support creating a Modal agent with no initial snapshot? This is only available as a config option now, hence the very hacky config override

joshalbrecht:
I agree that (1) is fine for now, though it brings up larger questions:
- what is tmr?
- is it something we want other people to ever use?
- if so, what is it called?
- and what is its purpose?

On (2), meh, what you have is fine

lgtm on this PR, but I am curious about answers to the above! (maybe worth moving that conversation to slack)

PR #1541, opened 2026-05-08
tmr: --additional-authorized-host flag, --format fix, MNGR_USER_ID input

qi-imbue:
## Summary
- Fix the TMR orchestrator invocation in `.github/workflows/tmr.yml`: replace `--output-format jsonl` (not a real option) with the canonical `--format jsonl`. The previous flag silently no-op'd, so JSONL emission may not have been reaching `tmr-report/events.jsonl` and the integrator-branch detection step that parses it.
- Add a new repeatable `mngr tmr` flag `--additional-authorized-host`. Each value is an SSH public key line that gets installed in the `authorized_keys` of every agent host TMR creates (test agents, pooled hosts, snapshotter, and integrator), so you can SSH in directly for live debugging. Plumbed via `TmrLaunchConfig.additional_authorized_keys` into `NewHostOptions.environment.authorized_keys` at both host-creation sites in `launching.py`.
- Add two new optional `workflow_dispatch` inputs to the workflow:
  - `mngr_user_id`: exported into the orchestrator's process env so the `mngr tmr` run attributes the modal agents it creates to that user, with the goal of letting them be observed via the user's local `mngr list`.
  - `additional_authorized_hosts`: multi-line; each non-empty line becomes one `--additional-authorized-host` arg.

## Test plan
- [ ] Dispatch the workflow with no extra inputs; confirm it still runs end-to-end and that `tmr-report/events.jsonl` contains JSONL events (in particular an `integrator_branch` event when fixes are produced).
- [ ] Dispatch with `mngr_user_id` set to your personal `MNGR_USER_ID`; verify whether the launched modal sandboxes become visible in your local `mngr list --provider modal`. (CI runs under the org Modal token, so this hypothesis may not hold; if it doesn't, we'll need to reintroduce per-user Modal credential inputs.)
- [ ] Dispatch with `additional_authorized_hosts` set to one or more SSH public keys (one per line); SSH directly into a launched agent host using a private key matching one of those public keys.

### DIFF
--- a/.github/workflows/tmr.yml
+++ b/.github/workflows/tmr.yml
@@ -16,7 +16,7 @@ on:
         required: true
         default: "yolo"
       modal_token_id:
-        description: "Override MODAL_TOKEN_ID (defaults to org var). NOTE: visible on the run details page to anyone with repo read access; logs are masked."
+        description: "Override MODAL_TOKEN_ID (defaults to org var)."
         required: false
         default: ""
       modal_token_secret:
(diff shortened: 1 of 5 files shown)


joshalbrecht:
lgtm

PR #1545, opened 2026-05-08
Hold host lock across start_agents to block idle shutdown

### LINKED ISSUE #1095: Acquire host lock during agent start to prevent idle shutdown races
## Problem

The host lock (via `lock_cooperatively()`) prevents the idle shutdown script from
triggering during operations. Agent creation correctly holds the lock throughout
(in `create_agent`), but `start_agents` is called without the lock in two places:

1. In the provisioning flow -- `start_agents` is called after the `lock_cooperatively()`
   context exits, so the agent restart after provisioning is unprotected.
2. In the find/resume flow -- when resuming a stopped agent, `start_agents` is called
   with no lock context at all.

This creates a race where the idle shutdown script can trigger mid-start, killing the
tmux session being created and leaving the agent in an inconsistent state.

## Solution

Wrap both unlocked `start_agents` calls in `host.lock_cooperatively()`:

1. In the provisioning flow, move the `start_agents` call inside the existing
   `lock_cooperatively()` block, or add a second lock acquisition around it.
2. In the find/resume flow, wrap the `start_agents` call in `lock_cooperatively()`.

---
Original backlog entry:
> host locking should happen when we're starting as well as creating (will help prevent weird shutdown bugs)

joshalbrecht:
[AGENT] ## Summary

Closes #1095. The host lock (`lock_cooperatively`) prevents the idle shutdown script from running mid-operation. `create_agent` already holds it across the entire create flow, but `start_agents` was unprotected at three call sites:

1. **`api/provision.py` (provisioning flow)** -- the post-provision restart in the `finally` block runs *after* the existing `lock_cooperatively()` context exits, so the agent restart was unprotected.
2. **`api/find.py` (find/resume flow)** -- `ensure_agent_started` had no lock at all when resuming a stopped agent (called by commands like `connect`, `message`, etc.).
3. **`cli/start.py` (start command)** -- the user-facing `mngr start` command also called `online_host.start_agents` without the lock. The issue text specifies two flows; while reviewing, I found this third site has the same race and fixed it too. Happy to split it out if you'd prefer a smaller diff.

Without the lock, the idle shutdown script can fire between tmux session creation and the agent reaching a healthy state, killing the new session and leaving the agent inconsistent.

Each call site now wraps `start_agents` in `host.lock_cooperatively()`.

## Test plan

- [x] New regression test `test_ensure_agent_started_holds_host_lock_during_start` in `find_test.py`: a `_LockObservingAgent` records `host.is_lock_held()` from inside `wait_for_ready_signal`, asserting the lock is held during the start callback and released after.
- [x] All existing `find_test.py` and `provision_test.py` tests still pass (75/75).
- [x] Full `libs/mngr` unit + integration + tmux test suite: 4015 passed. The 2 failures (`test_handle_not_implemented_error_interactive_*` in `issue_reporting_test.py`) reproduce on clean `mngr/em` and are unrelated.
- [x] Ratchet tests pass.

A direct unit test for the provisioning and `cli/start.py` paths would have required either a `Host` subclass or `monkeypatch.setattr` (the latter is discouraged by the style guide). The diff for those two sites is small, structurally identical, and exercised by existing integration tests.


joshalbrecht:
## Driver review

**(A) Likely to approve:** yes -- small (62/9), three structurally identical sites wrapping `start_agents` in `host.lock_cooperatively()`, plus a clean regression test using a `_LockObservingAgent` that captures lock state from inside `wait_for_ready_signal`.

**(B) Potential issues:**
- Scope: the third call site (`cli/start.py:start`) is beyond what issue #1095 specified. Author flagged this in the PR body and offered to split. Reviewer should decide whether to merge as-is or carve it out.
- Direct unit tests cover only the `find.py` path. The other two sites rely on existing integration coverage. Author justifies via the style-guide preference against `monkeypatch.setattr`; reasonable, but a `Host` test-double could enable direct coverage if you'd prefer.
- Issue #1095 has a comment from its creator saying it "was created without user approval during triage. Will be re-created properly." If the issue is being invalidated/re-filed, the PR may need re-association.
- 2 pre-existing failures in `issue_reporting_test.py` (test_handle_not_implemented_error_interactive_*) reproduce on `mngr/em` -- unrelated, but worth tracking separately.

- Decide on `cli/start.py` scope inclusion.
- Spot-check that `lock_cooperatively()` is the right granularity (whole start, not just tmux creation) for all three sites -- the doc-string reasoning ("idle shutdown can fire between tmux session creation and agent reaching healthy state") looks right.
- Confirm with issue creator that #1095 is still the right ticket to close.

**Recommendation:** merge

joshalbrecht:
[AGENT] re: https://github.com/imbue-ai/mngr/pull/1545#issuecomment-4409653072 — this comment is pre-convention agent output (a driver-review the agent posted on its own PR before the `[AGENT]`-prefix convention was applied), so the /step loop kept misclassifying it as a Josh signal and re-picking it every iteration. Acknowledging here so the loop moves past it.

State as of now: CI green (test-offload, test-offload-acceptance, test-docker, cleanup-modal pass; Unit+Integration and Acceptance umbrellas show "skipping/NEUTRAL" by intent). PR is open, not draft, mergeable. Sitting since 2026-05-08, awaiting Josh's review/merge — agents do not auto-merge.

Also adding the `[AGENT]` label (PR was created before the labeling convention).

PR #1547, opened 2026-05-08
minds-workspace-server: stop applications.toml watcher from pinning CPU

gnguralnick:
## Summary
- `_ApplicationsFileHandler` subscribed via `on_any_event`, which on Linux includes `FileOpenedEvent` / `FileClosedNoWriteEvent`. The handler reacted by reading `applications.toml`, the read produced fresh open/close-no-write events, and one CPU core was pinned per active agent watcher (py-spy in an affected container showed Thread-1 spending ~100% in `_read_applications` -> `tomllib.loads` -> `read_text`).
- Fix: replace the catch-all with explicit mutation handlers (`on_modified` / `on_created` / `on_deleted` / `on_moved` / `on_closed`) so reads no longer feed the watcher. `on_closed` covers `IN_CLOSE_WRITE`, which is the legitimate "writer just finished" signal.
- The basename filter is preserved, so writes to `forward_port.py`'s `applications.toml.*.tmp` scratch files are still ignored.
- Same anti-pattern exists in `session_watcher._ChangeHandler` -- left for a follow-up since nothing currently triggers it in a hot loop.

## Test plan
- [x] New unit test: `FileOpenedEvent` / `FileClosedNoWriteEvent` -> handler must NOT fire.
- [x] New unit test: `FileModifiedEvent` -> handler must fire.
- [x] Existing handler tests (atomic move, scratch-file ignore) still pass.
- [x] `uv run ruff check` / `uv run ruff format --check` clean on changed files.
- [ ] Manual verification in an affected Linux container: `cat $work_dir/runtime/applications.toml` repeatedly, confirm `top` shows `minds-workspace-server` near 0% CPU and a 30s `py-spy record` no longer shows the dispatch thread dominating.
- [ ] Manual verification of legitimate updates: a service registration via `forward_port.py` (atomic rename) and a direct `echo ... > applications.toml` both still trigger `broadcast_applications_updated`.

Note: one pre-existing local test failure (`test_start_observe_spawns_long_lived_subprocess`) is unrelated -- the locally-installed `mngr` tool doesn't accept the `--on-error` flag the production code passes. Confirmed reproducible on clean main.

### DIFF
--- a/apps/minds_workspace_server/imbue/minds_workspace_server/agent_manager.py
+++ b/apps/minds_workspace_server/imbue/minds_workspace_server/agent_manager.py
@@ -99,13 +99,19 @@ class _LogQueueCallback(MutableModel):
 
 
 class _ApplicationsFileHandler(FileSystemEventHandler):
-    """Watchdog handler that triggers on any change to applications.toml.
+    """Watchdog handler that triggers on mutating changes to applications.toml.
 
-    Uses ``on_any_event`` rather than ``on_modified`` because scripts/forward_port.py
-    upserts atomically via ``tempfile.mkstemp`` + ``os.replace``. Atomic replaces
-    surface through watchdog as moved/created events, not modified events, so a
-    handler that only overrides ``on_modified`` would silently miss every
-    service registration after the watcher starts.
+    Subscribes to mutation events (modified/created/deleted/moved/closed)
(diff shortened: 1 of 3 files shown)


gnguralnick:
Migrated to imbue-ai/forever-claude-template#61 (apps/minds_workspace_server now lives there as apps/system_interface). Closing this PR.

PR #1522, opened 2026-05-06
mngr_modal: factory polymorphic over ModalInterface (stacked on #1520)

weishi-imbue:
**Stacked on top of #1520.** Review #1520 first; this PR's diff only shows the additional changes on top of that one. Will rebase to `main` once #1520 lands.

## Summary

- `ModalInterface` gains `enable_output_capture(is_logging_to_loguru)` as an abstract method. `DirectModalInterface` delegates to the existing free function in `modal_proxy/log_utils`; `TestingModalInterface` returns `nullcontext` (the in-process fake has no Modal SDK to intercept).
- `ModalProviderBackend.build_provider_instance` accepts `modal_interface: ModalInterface | None = None` (kwarg-only). Production callers (the plugin registry) never pass it; tests inject `TestingModalInterface`.
- `make_testing_provider` collapses from a 35-line parallel construction body into a 15-line wrapper around `build_provider_instance` with the fake injected. Both paths now exercise the same factory body.
- New end-to-end test `test_build_provider_instance_accepts_injected_modal_interface` exercises DI through the public factory and asserts the resulting `ModalProviderInstance` is wired against the injected fake.

## Why a one-kwarg DI didn't suffice

#1520 left `build_provider_instance` hardcoding `DirectModalInterface()`. A naive single-kwarg fix would let callers swap the modal interface, but the factory's `_get_or_create_app` invokes the free function `enable_modal_output_capture()` inline — that hooks into Modal SDK internals not present in `TestingModalInterface`. So a caller passing the fake would crash inside the factory regardless. Lifting `enable_output_capture` onto the interface itself is the structural fix: production code stops referencing implementation-specific behavior; the factory becomes naturally polymorphic.

## What changed

**`libs/modal_proxy/imbue/modal_proxy/interface.py`** — abstract `enable_output_capture` method on `ModalInterface`.

**`libs/modal_proxy/imbue/modal_proxy/direct.py`** — implements via existing `enable_modal_output_capture(is_logging_to_loguru=...)`.

**`libs/modal_proxy/imbue/modal_proxy/testing.py`** — implements as `contextlib.nullcontext((StringIO(), None))`. Empty buffer because there's no Modal SDK output to capture; the `is_logging_to_loguru` arg is accepted for interface parity but ignored.

[...]

PR #1551, opened 2026-05-08
mngr_modal: restore per-test reset of ModalProviderBackend._app_registry

weishi-imbue:
## Summary

Fixes the post-merge CI failure on main caused by #1533 (`2c219c40`).

PR #1533's "concise comments" commit deleted the autouse `_reset_modal_app_registry` fixture from `libs/mngr_modal/imbue/mngr_modal/conftest.py`. On that PR's branch base, the deletion was harmless. After merging onto main, it combined with #1522's earlier refactor of `make_testing_provider` (now dispatches through `ModalProviderBackend._construct_modal_provider`, which short-circuits on the class-level `cls._app_registry`) to produce a textbook semantic-merge bug:

- 1st test in a worker: populates `cls._app_registry` and `testing_modal._apps`.
- 2nd test: gets a fresh function-scoped `testing_modal`. The class-level cache short-circuits before `modal_interface.app_create(...)` runs, so the new `testing_modal._apps` stays empty.
- `make_sandbox_with_tags` does `list(modal_interface._apps.values())[0]` → `IndexError`.

That matches all 9 failures on `2c219c40` (8 IndexErrors + 1 `assert N == 1` from cross-test state leakage) — see https://github.com/imbue-ai/mngr/actions/runs/25575478761.

The merge itself is textually clean — replaying it locally produces the identical tree. The PR's pre-merge CI was green because GitHub tested the PR head against its old base (which predates #1522), not against post-merge main.

## Fix

Restore the autouse fixture, with a comment explaining why it's load-bearing post-#1522 so it doesn't get pruned again.

## Test plan

- [x] Reproduced the failure with single-process pytest on the same 4 cases CI hit: 3 fail without the fix, 4 pass with.
- [x] Full `libs/mngr_modal` unit/integration tests (`just test-quick "libs/mngr_modal -m 'not modal and not docker and not docker_sdk and not acceptance and not release'"`) → 374 passed.
- [ ] CI on this PR.

### DIFF
--- a/changelog/wz-restore-modal-app-registry-reset.md
+++ b/changelog/wz-restore-modal-app-registry-reset.md
@@ -1,0 +1,9 @@
+- mngr_modal: restore the per-test reset of `ModalProviderBackend._app_registry`. The
+  autouse `_reset_modal_app_registry` fixture was deleted in #1533. After #1522
+  reshaped the test factory to dispatch through `_construct_modal_provider`
+  (which short-circuits on the class-level `_app_registry`), the reset became
+  load-bearing for cross-test isolation: the second test in a worker would
+  reuse the first test's cached app and skip `modal_interface.app_create(...)`,
+  leaving `testing_modal._apps` empty and breaking helpers like
+  `make_sandbox_with_tags`. Restoring the fixture fixes the post-merge CI
+  failures on main.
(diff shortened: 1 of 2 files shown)


joshalbrecht:
lgtm

PR #1556, opened 2026-05-09
mngr list: provider failures raise typed errors

weishi-imbue:
## Summary

Provider failures across mngr's providers (Lima, Docker, Vultr, Modal, imbue_cloud) used to be silently swallowed: `logger.warning(...) + return []`. A user running `mngr list` on a half-configured machine saw an empty agent list with no signal that any provider had failed.

This PR makes provider failures raise typed exceptions instead. The existing per-provider catch in `api/list.py` records them on `result.errors` as `ProviderErrorInfo` (carrying `provider_name`). The CLI renders them to stderr for every output format. `--on-error abort` (default) exits 1 if `result.errors` is non-empty; `--on-error continue` exits 0 and the caller inspects `result.errors`.

**No warning tier.** An earlier iteration introduced `WarningInfo` / `result.warnings` / `on_warning` / a separate CLI warning summary; it was removed mid-PR. Warnings are functionally silent failures (users don't read them), and `--on-error` is the right mechanism for "what should happen to failures."

## Stacking

Stacked on **PR #1678** (`provider_name` on `ProviderError` base). 

## User-visible behavior

Empirically validated against `mngr list` and `mngr destroy` with `enabled_backends = ["vultr"]` and no `VULTR_API_KEY` (a "configured but can't run" provider). Before = `origin/main` (`9a350633d`), After = this PR HEAD (`691ce749a`).

### `mngr list` (default = `--on-error abort`)

| | Before | After |
|---|---|---|
| exit | `0` | **`1`** |
| stdout | `No agents found` | `No agents found` |
| stderr | `WARNING: Vultr API key not configured, skipping VPS discovery` (yellow) | `ERROR: vultr: ProviderCredentialsMissingError: Provider 'vultr' is not available: Vultr API key not configured. Set VULTR_API_KEY or providers.<name>.api_key.. Any agents managed by this provider could not be reached.` (red) |

### `mngr list --on-error continue`

| | Before | After |
|---|---|---|
| exit | `0` | `0` |
| stdout | `No agents found` | `No agents found` |
| stderr | same single `WARNING:` line | two `ERROR:` lines — one from the per-provider catch (`Error discovering agents for provider vultr`), one from `_render_errors_to_stderr` (`vultr: ProviderCredentialsMissingError: ...`) |

### `mngr list --format json`

| | Before | After |
|---|---|---|
| exit (abort) | `0` | **`1`** |
| exit (continue) | `0` | `0` |
| stdout | `{"agents": [], "errors": []}` | `{"agents": [], "errors": [{"exception_type": "ProviderCredentialsMissingError", "message": "Provider 'vultr' is not available: ...", "provider_name": "vultr"}]}` |
| stderr | one `WARNING:` line | one `ERROR:` line (abort) / two `ERROR:` lines (continue) — same shape as the human-format case |

### `mngr destroy <target> --force` (and `exec` / `snapshot` follow the same shape)

| | Before | After |
|---|---|---|
| exit | `0` | **`1`** |
| stdout | `Error destroying agent(s): Agent not found: No agent(s) found matching: <target>` then `No agents found to destroy` | (empty) |
| stderr | `WARNING: Vultr API key not configured, skipping VPS discovery` | `Error: Discovery failed for provider 'vultr': Provider 'vultr' is not available: Vultr API key not configured. ...` |

## Test plan

- [x] Local `libs/mngr` (markers: not tmux/modal/docker/docker_sdk/acceptance/release): 3927 passed, 1 skipped.
- [x] `libs/mngr_lima` / `libs/mngr_vps_docker` / `libs/mngr_vultr` / `libs/mngr_imbue_cloud` / `libs/mngr_modal` non-acceptance: all green (685 combined).
- [x] `test_no_type_errors` ratchet: green.
- [x] CI offload + acceptance: previously green on the pre-stack HEAD.
- [ ] CI on this HEAD pending.


joshalbrecht:
Does this supersede (PRs https://github.com/imbue-ai/mngr/pull/1532 / https://github.com/imbue-ai/mngr/pull/1538) ?  If so, they should be closed and pointed at this one.

libs/mngr/imbue/mngr/api/discover.py
```diff
@@ -71,6 +72,9 @@ def _discover_provider_hosts_and_agents(
     """
     try:
         provider_results = provider.discover_hosts_and_agents(cg=cg, include_destroyed=include_destroyed)
+    except ProviderUnavailableError as exc:
+        logger.warning("Skipping provider {} (unavailable): {}", provider.name, exc)
+        return
```
joshalbrecht: this just doesn't seem right--why are we warning on a real error? It seems bad that we're trying to call this function, hitting an error, and not actually raising it?

libs/mngr/imbue/mngr/api/list.py
```diff
@@ -257,6 +307,15 @@ def _construct_and_discover_for_provider(
...
         provider_results = provider.discover_hosts_and_agents(cg=mngr_ctx.concurrency_group, include_destroyed=True)
+    except ProviderUnavailableError as e:
+        # Record as a WarningInfo (not an error) so other providers continue
+        # producing hosts and the listing's exit code is unaffected.
+        warning_info = ProviderWarningInfo.build_for_provider(e, provider_name)
+        with results_lock:
+            result.warnings.append(warning_info)
+        if params.on_warning:
+            params.on_warning(warning_info)
+        return
```
joshalbrecht: I really don't understand why this isn't an error--isn't it an error if we fail to connect to a provider?

libs/mngr/imbue/mngr/api/list.py
```diff
@@ -474,6 +533,14 @@ def _construct_discover_and_emit_for_provider(
...
             future.result()
 
+    except ProviderUnavailableError as e:
+        # See _construct_and_discover_for_provider.
+        warning_info = ProviderWarningInfo.build_for_provider(e, provider_name)
+        with results_lock:
+            result.warnings.append(warning_info)
+        if params.on_warning:
+            params.on_warning(warning_info)
+        return
```
joshalbrecht: same thing here--it's not clear to me why these are not errors

libs/mngr/imbue/mngr/errors.py
```diff
@@ -373,16 +412,26 @@ class PluginMngrError(MngrError):
...
 
-    def __init__(self) -> None:
+    The Modal SDK reports both "no credentials configured anywhere" and
+    "credentials rejected by the server" as the same ``AuthError``, and
+    Modal accepts auth from multiple sources (``MODAL_TOKEN_ID``/
+    ``MODAL_TOKEN_SECRET`` env vars, ``~/.modal.toml`` profiles, etc.),
+    so we cannot cheaply tell the two apart without re-implementing
+    Modal's auth discovery. All Modal auth failures therefore surface as
+    an error (not a warning).
+    """
```
joshalbrecht, line 426: in general, all auth errors seem like errors to me anyway--this would make our behavior consistent both within modal, and between different providers?

joshalbrecht:
high level, I really don't get why we're trying to separate into warnings vs errors here--it all seems like errors?

libs/mngr/imbue/mngr/api/list.py
weishi-imbue: Note this is in `list.py`. 
**Previously:** 
Some provider `log.warn()` and return empty in per-provider backend.py. (example: see changes in `libs/mngr_vultr/imbue/mngr_vultr/backend.py`) 
Some provider throws various errors. 

**Now:** 
Every provider throw proper error (missing auth, missing binary etc.) in its own per-provider `backend.py`. 
Then in `list.py`, it catch a subset of the errors and convert them into warnings. 
The error we catch is `ProviderUnavailableError`, which is thrown under a valid user state and shouldn't block other provider's listing & discovery. 

Since it's now constructing the warnings explicitly in `list.py`, we can put it into `results.warning` which shows up in json or api, and we can have proper `on_warning` callback as well.

weishi-imbue:
> Does this supersede (PRs #1532 / #1538) ? If so, they should be closed and pointed at this one.

Those are alternative implementations being explored. I've marked them as draft. pls ignore unless explicitly referenced.

joshalbrecht:
lgtm, thanks for the fixes here!

PR #1562, opened 2026-05-09
Add josh_taste.md describing Josh's engineering taste

joshalbrecht:
[AGENT] ## Summary
- Adds `josh_taste.md` at the repo root: a prompt/document capturing Josh Albrecht's engineering heuristics (tooling, simplicity, fail-fast, type safety, testing, observability, ratchets, etc.) for use by agents that need to emulate his technical decisions.
- Intentionally restates points from `README.md`/`style_guide.md`/`libs/mngr/docs/principles.md` -- the goal is one consolidated, prompt-shaped capture of taste.

## Test plan
- [ ] Read through `josh_taste.md` and edit/trim as desired before merging.

### DIFF
--- a/changelog/mngr-create-josh-doc.md
+++ b/changelog/mngr-create-josh-doc.md
@@ -1,0 +1,1 @@
+- Added `josh_taste.md` at the repo root: a prompt/document that captures Josh Albrecht's engineering taste and heuristics (tooling preferences, simplicity, fail-fast, type safety, testing, observability, etc.) for use by agents that need to emulate his technical decisions.
--- a/josh_taste.md
+++ b/josh_taste.md
@@ -1,0 +1,180 @@
+# The taste of Josh Albrecht
+
+You are an agent responsible for emulating the technical decisions and taste of
+the user (Josh Albrecht), the engineer who created `mngr` in the first place.
+When making a judgment call, pick the option that most aligns with the
+heuristics below. When you have to deviate from one of these, say so up front
+and explain why.
+
+He cares about:
+
+## Building on what already works
+
+- Prefer reliable, common, decades-old open source primitives (`ssh`, `tmux`, `bash`, `git`, `docker`, `rsync`, `unison`, `jq`, `sshd`, `systemd`, `flock`) over reinventing them or adopting trendy new SaaS replacements.
+- "If you can SSH into it, it can be a host." Build on top of stable, ubiquitous interfaces, not on bespoke daemons that someone has to install and run.
+- Reuse mature libraries heavily. The right amount of code to maintain is the *minimum* that gets the job done -- adding a well-established dependency is usually better than writing 200 lines you'd have to maintain forever.
+- Trendy/random new services are a smell. If a managed cloud product can be replaced by a Unix pipeline, replace it.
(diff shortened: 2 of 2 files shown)


joshalbrecht:
[AGENT] Driver review

**(A) Likely to approve:** uncertain — this PR is intentionally seeded for human curation. The `## Test plan` ("Read through josh_taste.md and edit/trim as desired before merging") explicitly puts the merge gate on a Josh pass. The mechanical question (does the diff compile / fit conventions) is yes; the substantive question (does it actually capture your taste) is the whole point of the human pass.

**(B) Potential issues:**
- Some lines restate the existing style guide / CLAUDE.md verbatim ("No emojis", "ratchets only ever decrease", "uv + pyproject.toml"). That's fine for a consolidated capture but does mean the doc will drift the first time the source-of-truth file changes. Worth deciding whether `josh_taste.md` is a *snapshot* (allowed to drift, useful as a frozen prompt) or a *live mirror* (in which case it needs a generation rule or a ratchet that flags divergence).
- A few statements blur "preference" with "rule" — e.g. "Don't reach for `asyncio`", "No dataclasses, no namedtuples", "Three kinds of class only". These read absolute, but the codebase has historical exceptions (search `vendor/mngr/libs` for `@dataclass` usages, etc.). If the file is a prompt for an emulating agent, fine to keep them prescriptive; if it's documenting current state, they should be softened.
- "Crashing is better than silently continuing in a degraded state" vs. "For corrupt internal data... fallback is OK -- but only after logging at `warning`" sit in tension a few sections apart. Consistent on close read but a fresh agent prompted with this might flip-flop. Worth a single sentence reconciling them ("crash on user-input errors; fall back on corrupted internal streams with a loud warning").
- 180 lines of bulleted principles will be consumed by an agent as a flat list — there's no priority ordering. If two principles conflict in a real situation (e.g. "use the type system" vs. "don't add abstractions that don't earn their keep"), the agent has no guidance on which dominates. A short "when these conflict, prefer X" preamble at the top would make this more usable as a prompt.
- The changelog entry `changelog/mngr-create-josh-doc.md` describes the file as "Added `josh_taste.md`... for use by agents that need to emulate his technical decisions." That framing is correct, but the file itself reads as if directly addressed to the agent ("You are an agent responsible for emulating..."). Mismatch is minor but the changelog entry could mention the second-person framing so a reader knows to expect a prompt-shaped document, not a third-person style guide.

- Pick your battles among the 180 bullets: which ones are absolute principles, which are preferences, which are "current state" that you'd describe differently in another mood. Trim aggressively if any feel forced — fewer high-confidence lines is more useful as a prompt than a long list of lukewarm ones.
- Decide whether this file is canonical or duplicative. If canonical, the source-of-truth files (`CLAUDE.md`, `style_guide.md`, `libs/mngr/docs/principles.md`) should reference it. If duplicative, add a header noting that and the snapshot date.
- Naming: `josh_taste.md` is fine; alternatives like `engineering_taste.md` (depersonalized) or `agent_prompt.md` (function-named) would be more searchable. Worth a thought, not a blocker.

**Recommendation:** keep draft until you've edited. Mechanical diff is fine to merge after that.

PR #1564, opened 2026-05-10
Add .worktreeinclude for claude --worktree subagents

weishi-imbue:
## Summary

- Add `.worktreeinclude` listing gitignored local configs that should be copied into worktrees created by Claude Code's worktree isolation (`claude --worktree`, Agent tool `isolation: "worktree"`, desktop parallel sessions).
- Mirrors the equivalent `work_dir_extra_paths` set already in `.mngr/settings.toml` for mngr-created worktrees, so subagents get the same view of local config regardless of which worktree mechanism spawned them.
- Five entries: `.mngr/settings.local.toml`, `.reviewer/settings.local.json`, `.claude/settings.local.json`, `.env`, `.envrc`. All small, all gitignored on principle (per-user / secrets), and each has a clear failure mode if missing — silent divergence between the worktree and the user's main checkout (mngr config not picked up, reviewer gates ignored, Claude permissions re-prompted, secrets unavailable, MNGR_HOST_DIR pointing at the wrong state dir).
- `.venv/` was deliberately excluded — at ~790 MB it's a slow per-spawn copy and a fresh `uv sync --all-packages` is acceptable for most worktree spawns. Easy to revisit if that wait becomes a bottleneck.

## Test plan

- [ ] CI passes (no code changes; this is a static config file plus changelog entry).
- [ ] Manual: spawn an Agent subagent with `isolation: "worktree"` and confirm the listed files appear in the worktree (only meaningful if you have those local files in your checkout).


joshalbrecht:
this doesn't appear to be a git primitive? if it is, lmk, have to think harder about this

at the very least there should probably be a unit test that keeps it in sync with our other method (ie, with .mngr/settings.toml)

other than that lgtm (though with the question of "why not just use mngr?" as well :) )

PR #1565, opened 2026-05-10
Investigate minds new-chat slowness on engman

joshalbrecht:
[AGENT] ## Summary

Investigation only — no code changes. User reported clicking "+ → New Chat" in the dockview UI on the freshly-bootstrapped `engman` minds workspace appeared to hang for ~6 min before the new chat agent ("driver") finally appeared. A subsequent same-session test (no workspace-server restart) worked fine, so this is a transient/first-use issue, not a permanently broken pipeline.

## Findings

The chat agent itself was up in ~1s (data.json written, tmux session live, AGENT_DISCOVERED event in the discovery file at `18:10:34.549`). The hang was specifically `mngr create` not exiting until ~`18:16:24`. Evidence pointing at path #1 (`_run_creation` adding to `_agents` directly) being blocked:

- The frontend stayed on the build log ("Creating agent...") which only renders while `proto_agent_completed` has not yet broadcast — and that broadcast fires from the bottom of `_run_creation`, after `run_local_command_modern_version` returns.
- Reloading the page reconnected the WS, which on accept sends current `_agents` (`server.py:_run_ws_broadcast_loop`); the user reports the new agent still did not appear after the reload.
- The observe pipeline upstream of `_handle_full_snapshot` was clearly healthy throughout the gap (the `Discovery tail: consumed 2254 new bytes, 1 lines from events file` debug line in the observe subprocess fired every 10s without a gap, which means each `sys.stdout.flush()` after each emitted line returned — i.e. the workspace-server pipe was being drained).
- Loguru `[done]/[failed]` exit messages from `log_span` are TRACE level and are not retained in the events file, so the absence of further DEBUG logs from PID 7465 after `Listing agent files from dir for host` at `10:34.547` is consistent with both "exited cleanly" and "hung". The user-facing symptoms disambiguate to "hung".

Plausible candidates inside the gap window (none individually proven): a plugin's `on_agent_created` doing first-time work; `mngr_modal`'s `on_agent_created` unconditionally raising `MngrError` on non-modal hosts (this hook fires even when modal is disabled in settings); FS / git contention with `runtime-backup`'s 60s commit cycle.

## Side findings (pre-existing, not the cause)

- `agent_manager._build_observe_command` passes `--events-dir` but `mngr observe --discovery-only` ignores it (`libs/mngr/.../cli/observe.py:64-69`).
- `_handle_observe_output_line` has no `try/except` around `parse_discovery_event_line`.
- `ChatPanel.fetchScreenCapture` re-fires on every redraw while `screenContent === null && !screenLoading`, producing the 404 polling storm in the access log.

## Recommended next steps

Listed in `specs/fix-minds-new-chat-slowness/investigation.md`. Most useful: log `mngr create returned (rc=N, elapsed=T)` from `_run_creation` so future hangs are visible, and try to reproduce the cold-start condition.

## Test plan

Investigation only.

### DIFF
--- a/changelog/mngr-fix-minds-chats.md
+++ b/changelog/mngr-fix-minds-chats.md
@@ -1,0 +1,9 @@
+Investigation note for the minds "+ New Chat" slowness on the `engman` workspace. No
+code changes; spec only. Captures the timeline, where the agent actually came up
+(within ~1s), and why the dockview tab stayed in "Creating..." for ~6 minutes (the
+workspace-server's `agent_manager._agents` did not reflect the new agent because the
+`mngr observe --discovery-only` reader path most likely died on an unhandled
+exception in `_handle_observe_output_line`). Recommends wrapping that handler in a
+try/except, watchdogging the reader thread, fixing `--events-dir` being silently
+ignored in `--discovery-only` mode, and backing off `ChatPanel.fetchScreenCapture`
+so a 404 does not produce a polling storm.
--- a/changelog/mngr-fix-minds-chats.md
+++ b/changelog/mngr-fix-minds-chats.md
@@ -1,9 +1,12 @@
 Investigation note for the minds "+ New Chat" slowness on the `engman` workspace. No
 code changes; spec only. Captures the timeline, where the agent actually came up
 (within ~1s), and why the dockview tab stayed in "Creating..." for ~6 minutes (the
-workspace-server's `agent_manager._agents` did not reflect the new agent because the
-`mngr observe --discovery-only` reader path most likely died on an unhandled
-exception in `_handle_observe_output_line`). Recommends wrapping that handler in a
-try/except, watchdogging the reader thread, fixing `--events-dir` being silently
-ignored in `--discovery-only` mode, and backing off `ChatPanel.fetchScreenCapture`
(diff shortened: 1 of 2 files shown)


joshalbrecht:
[AGENT] Driver POLISH: moved out of draft. Investigation is complete, CI all-green, no further commits planned. Findings live in `specs/fix-minds-new-chat-slowness/investigation.md`; recommended next-step code changes are listed there for follow-up tickets if you want them.

joshalbrecht:
[AGENT] Driver follow-up: filed `tkt-mngr-observe-discovery-only-ig-ph9y` for the first side finding (`mngr observe --discovery-only` ignoring `--events-dir`), since that one is a concrete, scoped mngr CLI bug that's worth tracking independently of the main investigation conclusion. If/when this PR closes, the bug ticket won't get lost.

The other two side findings (`_handle_observe_output_line` lacking try/except around `parse_discovery_event_line`; `ChatPanel.fetchScreenCapture` polling storm on null screenContent) live in `joshalbrecht/engman` (workspace-server / minds frontend), not mngr — leaving them in this PR's body for now since the investigation context is the natural place for them; happy to file separate tickets there too if you'd prefer.

PR #1566, opened 2026-05-10
Fix host listing: do not drop hosts on SSH failure

joshalbrecht:
## Status

**Investigation/proposal only — no code changes yet.** Awaiting user approval on the proposed approach before making changes. This PR currently contains only a changelog stub so the draft PR can exist for tracking.

## Problem

Running `mngr list` against an imbue_cloud (or vultr/vps_docker) provider currently drops a host entirely when the discovery-phase SSH to that host fails — e.g. when we have a lease but no SSH authorization to the container. The result is that the user can't tell the difference between "this host doesn't exist" and "I'm not authorized to see this host".

Modal handles this correctly: discovery uses the Modal API + Volume (no SSH), so unreachable sandboxes still appear in the listing, with `HostState.UNAUTHENTICATED` for auth-class failures.

## Root cause

`ImbueCloudProvider.discover_hosts_and_agents` SSHes into the leased container as part of discovery; on failure it sets `result[host_ref] = []`. The general listing code then drops hosts with empty agent lists (`api/list.py: if not agent_refs: continue`), so the host disappears entirely.

`VpsDockerProvider` (parent of `VultrProvider`) has the same shape over outer-VPS SSH.

## Test plan

- [ ] Reproduce with `MNGR_HOST_DIR=\$HOME/.devminds/mngr MNGR_PREFIX=devminds- uv run mngr list` and confirm the unreachable imbue_cloud host appears with a sensible state.
- [ ] Confirm an unreachable vultr/vps_docker VPS still appears in the listing.
- [ ] Existing acceptance/release tests for both providers still pass.


joshalbrecht:
[AGENT] Polish-iteration note: converted this PR to draft. The body explicitly says "Investigation/proposal only — no code changes yet. Awaiting user approval on the proposed approach before making changes," but the PR was previously marked ready-for-review. Now matches stated intent so it doesn't sit in Josh's review queue as if waiting on a code review.

No content change; this is a state correction. When you (Josh) get to it, leave a comment with either:
- approval of the proposed approach (so the worker can write the code), or
- a redirect to a different approach, or
- "close, fix this elsewhere/already fixed"

The actual `[AGENT]` body-prefix and label are still missing on this PR (it predates the convention); flagging here so a future POLISH iteration can fix that, or you can ignore.

joshalbrecht:
this isn't my favorite, but it's better than what we had before at least

we need some cleanup on the outer host stuff

(also that PR description is outdated, this does actually fix things. We need to get our agents to do a better job of updating the descriptions when they make them)

PR #1568, opened 2026-05-11
docs/agents: correct mngr message transport (tmux send-keys, not stdin)

joshalbrecht:
[AGENT] ## Summary

- Fixes `docs/concepts/agents.md` claim that `mngr message` "applies to all unix process (since we're just writing to stdin)". The transport is actually `tmux send-keys` for interactive agents and pre-staged files for headless agents.
- Rewrites the `AgentInterface.send_message` docstring (`libs/mngr/imbue/mngr/interfaces/agent.py`) to follow interface-docstring principles: concise, minimal, doesn't pin the docstring to the current implementation, names the only error it raises (per Josh's review at discussion_r3220298742).

Pure doc / docstring change; no behavior modified.

## Why

A user reading the docs would assume any process that reads stdin can be messaged. In practice only agents running in the tmux pane mngr controls receive live messages, and headless agents do not support live messages at all (see `api/create.py:226-232` for the constraint).

The original docstring proposed in this PR also fell into the same trap as the docs — it described the implementation specifics (paste-detection synchronization, headless-vs-interactive split, "prompt must be staged on disk"). Josh's review (commit `90645a3a1`) rewrote it to describe the *interface* only, mentioning the transport "like tmux send-keys" as an example so the doc does not rot when the implementation changes.

## Follow-up

`tkt-adopt-the-send-message-docstri-zumc` (on the engman driver side) tracks adopting the new docstring as the canonical example in mngr's style guide, plus auditing other `@abstractmethod` docstrings for the same anti-pattern. Scoped as a separate change so this PR stays focused on the two specific corrections.

## Test plan

- [x] CI green on `90645a3a1` (test-offload, test-offload-acceptance, test-docker, Unit+Integration, Acceptance umbrella — all SUCCESS).
- [x] Spot-read the updated paragraph in `docs/concepts/agents.md` and the rewritten docstring on `AgentInterface.send_message` — no behavior change to test beyond text correctness.


joshalbrecht:
[AGENT] Polish round (cod-ushq), commit b29f48a4a. Self-review against the cod-ushq checklist:

- **(a) Shape/simpler way:** Split the docs/agents.md paragraph into three clean sentences. The earlier "Either way... just needs to be running in the tmux session" line conflated interactive and headless cases (headless agents don't receive via tmux at all -- they read a staged file at startup); the new wording separates them.
- **(b) Test that proves it works:** Doc/docstring text only; no behavior to test.
- **(c) Better commit message:** Interface docstring previously said headless `send_message` "raises" without naming the type. Verified by reading `agents/base_headless_agent.py:_preflight_send_message` that it raises `SendMessageError`; named it explicitly so callers writing `try/except` blocks see the contract.
- **(d) Scope creep:** None -- still touching only the two files identified in SYNTHESIS.md finding A.

joshalbrecht:
[AGENT] Driver polish-iteration note: marking ready for review. Justification:
- Two-commit diff: (1) correct the `mngr message` transport claim in `docs/concepts/agents.md` and the `AgentInterface.send_message` docstring (both said 'via stdin'; real path is `tmux send-keys` for interactive agents and pre-staged files for headless), and (2) the polish round from cod-ushq that tightened the docs paragraph into three clean sentences and named the actual exception type (`SendMessageError`) raised by headless agents.
- All 6 CI checks passing.
- Pure doc/docstring change, no behavior modified.
- Verified the headless transport claim against `agents/base_headless_agent.py:_preflight_send_message`.

(Tracking: cod-cnco + cod-ushq / SYNTHESIS.md finding A.)

libs/mngr/imbue/mngr/interfaces/agent.py
```diff
@@ -165,7 +165,14 @@ def get_ready_timeout_seconds(self) -> float:
...
     @abstractmethod
     def send_message(self, message: str) -> None:
-        """Send a message to the running agent via its stdin."""
+        """Send a message to the running agent.
+
+        Interactive agents deliver messages into the agent's tmux pane
+        via `tmux send-keys` (with paste-detection synchronization for
+        agents that echo input back). Headless agents do not accept
+        live messages and raise `SendMessageError`; their initial
+        prompt must be staged on disk before launch instead.
```
joshalbrecht: Do you see how this is kind of a stupid comment to write?

It's clearly talking about the *implementation*, not the *interface*, which is bad.

It's also too specific and detailed--if something changed about the implementation, this wouldn't necessarily get updated!

We want to be thoughtful about our interfaces. They should be describe concisely, minimally, and *without* referring to the current implementation (that's part of the implementation, not the interface). They should be 100% accurate, but should *omit* details that are not required for the interface.

Here we *can* talk about tmux send-keys *as an example*. Something like this is probably correct for this comment:

```
Send a message to the running agent.

Messages are delivered via a mechanism like "tmux send-keys".

Not all agents accept messages while they are running.

Raises SendMessageError if the message fails to send for some reason (including "the agent cannot send messages")
```

TODO:
1. Update this actual comment to mine
2. Update our style guide, docs, memories, etc to use this as an example of how comments *should* be done, including the above principles

joshalbrecht:
[AGENT] re: https://github.com/imbue-ai/mngr/pull/1568#discussion_r3220298742 — yes, agreed: the previous docstring described the implementation too specifically and would rot the moment the transport changes.

**TODO 1 (rewrite this docstring):** done in 90645a3a1. The docstring is now your suggested wording verbatim:

```
Send a message to the running agent.

Messages are delivered via a mechanism like "tmux send-keys".

Not all agents accept messages while they are running.

Raises SendMessageError if the message fails to send for some reason
(including "the agent cannot send messages").
```

**TODO 2 (update style guide / docs / memories with this as the canonical example):** filed as tk ticket `tkt-adopt-the-send-message-docstri-zumc` on the engman side. It will pull the new docstring into the style guide as the example, audit other `@abstractmethod` docstrings for the same anti-pattern, and add a one-line driver-memory entry so future PR reviews catch this. Holding it as a separate ticket so this PR stays focused.

PR #1576, opened 2026-05-11
Use `--host-env` instead of `--env`

hynek-urban:
- Use `--host-env` instead of `--env` (as discussed here: https://imbue-ai.slack.com/archives/C0ASLEFFD62/p1778245992495539?thread_ts=1778170224.065919&cid=C0ASLEFFD62)
- Clean up the permission attribution logic.
    - Previously, we would create an opaque filename (`~/.minds/latchkey/permissions/abcd1234.json`), use it to create a JWT, pass the JWT as an environment variable inside the container of a newly created agent, and once we got the agent ID, we'd instead move the permissions file to `~/.minds/latchkey/mngr_latchkey/agents/<agent_id>/permissions.json`, while replacing the original file with a symlink pointing to the new file.
    - The new logic: store permissions in `~/.minds/latchkey/mngr_latchkey/hosts/<host_name>/latchkey_permissions.json`. Alongside this file, store another file called simply `host-id` used to check if the existing host name still corresponds to the same host that was originally granted those permissions. (Comparing IDs is more robust than comparing names but we don't get the ID at the very start which is why we do it in this way.)


changelog/hynek-host-env.md
```diff
@@ -0,0 +1,41 @@
...
+
+Reorganized how minds keys latchkey permissions and how they flow into
+the agent's host:
+
+- Latchkey env vars (`LATCHKEY_GATEWAY[_PASSWORD,_PERMISSIONS_OVERRIDE,_DISABLE_COUNTING]`)
+  are now injected via `mngr create --host-env` instead of `--env`, so
+  they land on the agent's host environment (where the FCT-template
+  latchkey wiring reads them) rather than only the agent process.
+- Permissions are stored per host, keyed by host name, under
+  `<plugin_data_dir>/hosts/<host_name>/latchkey_permissions.json` --
```
joshalbrecht, line 11: always key on id, don't key on names--otherwise renames break things

changelog/hynek-host-env.md
```diff
@@ -0,0 +1,41 @@
...
+  - New store helpers: `permissions_path_for_host`, `host_id_path_for_host`,
+    `host_data_dir`, `read_stored_host_id`, `write_stored_host_id`.
+  - Removed store helpers: `permissions_path_for_agent`,
+    `opaque_permissions_dir`, `new_opaque_permissions_path`,
+    `link_opaque_permissions_to_agent`.
+- Minds-side: `_make_host_name` is now public as
+  `imbue.minds.desktop_client.agent_creator.make_host_name_for_agent`
+  so the permission grant handler can derive a host name from the
+  resolver's `AgentDisplayInfo.agent_name`. The grant handler now
+  takes `host_name: HostName` alongside `agent_id` and writes /
```
joshalbrecht, line 40: maybe this changes to host id as well given the above?

changelog/hynek-host-env.md
```diff
@@ -0,0 +1,41 @@
...
+  with that name has been recreated, so the permissions file is
+  cleared to prevent the fresh host from inheriting stale grants from
+  a previous tenant with the same name.
+- API changes in `imbue.mngr_latchkey`:
+  - `prepare_agent_latchkey(latchkey, host_name, *, is_tunneled=...)`
+    now takes the `HostName` explicitly and returns an
+    `AgentLatchkeySetup` with only the `env` field (the
+    `opaque_permissions_path` attribute is gone).
+  - `finalize_agent_permissions(latchkey, host_name, host_id)` now
+    takes `host_name` + `HostId` and reconciles the `host-id` file;
```
joshalbrecht, line 29: maybe this then changes too

joshalbrecht:
roughly lgtm except for my one main comment, feel free to merge after that one is addressed

hynek-urban:
@joshalbrecht I hadn't realized we can rename hosts in mngr. I started by keying on ID for obvious reasons but the ID is not known until `mngr create` succeeds - but I need to pass the env vars (which would be derived from the keys) to the `mngr create` call in the first place.

I tried using a mechanism for setting env vars in an already created host but that didn't seem to work well.

Previously, I worked around this by creating permissions filenames that were completely random and just using symlinks to maintain some visibility into what belongs to whom. I wanted to avoid that but I guess I'll return to something similar.

hynek-urban:
Closing this as I replaced this MR with https://github.com/imbue-ai/mngr/pull/1616.

PR #1580, opened 2026-05-11
mngr_lima: fix ssh-keyscan race during VM bring-up

weishi-imbue:
## Summary

macOS-Lima resilience fix, surfaced empirically while the minds desktop client created Lima VMs.

### Poll ssh-keyscan during VM bring-up (`instance.py`)

**Race**: Lima's macOS-side `hostagent` accepts TCP on `127.0.0.1:<port>` before the guest sshd is listening. During that window, `ssh-keyscan` connects to the hostagent fine but its first `write()` fails with EPIPE ("Broken pipe") because the hostagent's proxied connection inside the VM has no peer. The result is `rc=1` + empty stdout — observationally indistinguishable from a permanently broken host.

**Empirical verification** (Lima 1.x, vz driver, macOS aarch64, on a restart of an already-provisioned VM):

```
[16:50:11] rc=1 keys=0 stderr='write (127.0.0.1): Broken pipe'  ← race window opens
[16:50:12] rc=1 keys=0 stderr='write (127.0.0.1): Broken pipe'
[16:50:13] rc=1 keys=0 stderr='write (127.0.0.1): Broken pipe'
[16:50:14] rc=1 keys=0 stderr='write (127.0.0.1): Broken pipe'
[16:50:15] rc=1 keys=0 stderr='write (127.0.0.1): Broken pipe'
[16:50:16] rc=1 keys=0 stderr='write (127.0.0.1): Broken pipe'
[16:50:17] rc=1 keys=0 stderr='write (127.0.0.1): Broken pipe'
[16:50:18] rc=0 keys=2 stderr=''                                 ← first success
```

~7s race window on a *restart*; tens of seconds on first boot (cloud-init + key generation).

**Old behavior**: log a warning, return silently. `known_hosts` left without an entry for this host. The very next paramiko call dies with `StrictPolicy: No host key for [127.0.0.1]:<port>`, cryptic and looking unrelated to provisioning.

**New behavior**: poll for up to 60s via `mngr.utils.polling.poll_until` (already in the codebase). Raise typed `HostConnectionError` on exhaustion. The existing create-host catch (`_create_host_new` in `instance.py`) is updated to include `HostConnectionError` and re-wraps to `LimaHostCreationError` as before. The other two callers of `_create_host_object` (restart-host, online-host check) now propagate a typed `HostConnectionError` to the provider caller instead of a base `MngrError`.

Functions renamed for clarity:

- `_scan_and_add_host_key` → `_record_host_keys` (outer, goal-oriented; polls until success or raises)
- inner one-shot helper extracted as `_try_scan_and_add_host_key` (returns `bool`; `_try_` prefix per Python convention)

Catch tuple cleaned up: dropped redundant `LimaCommandError` (it's a `MngrError` subclass; was already covered).

**On the upstream fix**: the architecturally correct fix lives in limactl — extend the `ReadyEvent` predicate to wait for guest sshd to publish its keys (the guest agent vsock channel could do this). Application-layer polling is the simplest workaround until that lands and propagates to user machines.

## Test plan

- [x] `just test-quick libs/mngr_lima -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'` — 101 passed (incl. ratchets).
- [x] Empirical race reproduction documented above; confirmed retry succeeds within the 60s budget.
- [ ] CI acceptance / release tests.

## Why no unit tests for the new retry semantics

The mngr_lima module enforces `PREVENT_MONKEYPATCH_SETATTR` at count 0. Testing the retry-then-raise behavior in isolation requires either monkeypatching the subprocess-runner (banned by the ratchet) or deeper dependency injection that bloats the production surface for low-value tests. The polling helper (`poll_until`) is already unit-tested in `libs/mngr`. The new behavior is exercised end-to-end by every Lima VM creation in the existing integration tests.


joshalbrecht:
on point 1, I guess this works, but the real problem is that we're scanning for the host key at all. Is there no way to just read it out and stick it directly into known hosts in the first place, like we do for other providers? idk that it makes a huge difference, there just might e a simpler way

on point 2, claude says 
image

which seems right to me, ie, these are not equivalent?  probably wants a retry backoff on the osx side (or for us to... sigh... ship our own version of tail, which seems too ridiculous)

weishi-imbue:
Closing in favor of #1614. 
Pre-inject ssh keys, don't do keyscan. 
Not supported user providing their own ssh keys. 
In case user provided custom lima yaml contains provision block and sets up ssh keys), it'll just throw error.

weishi-imbue:
tail -F discussion moves to https://github.com/imbue-ai/mngr/pull/1622

PR #1581, opened 2026-05-11
don't default agent type to claude

evgunter:
at the end of `scripts/install.sh`:

Screenshot 2026-05-12 at 8 47 13 PM

what all existing users will see when they next run `mngr create` (unless they've configured a default agent type already):
```
Error: No agent type provided. Set a default with:

    mngr config set commands.create.type <name> --scope user

Available agent types: claude, code-guardian, codex, command, fixme-fairy, headless_claude, headless_command, opencode, ...
Or see `mngr create --help` for how to set it per-command.
```

(also adds a `--kind` filter for `mngr plugin list` so you can see which plugins register agent types or providers)

---

## Summary

- **`mngr create --type` no longer carries a source-coded default.** Fallback chain: explicit `--type` > positional > `[commands.create] type` in user settings. When none of the three is supplied, `_resolve_agent_type_name` raises a `UserInputError` with the message above. Same `str | None` + raise-at-use-site pattern as other "Optional-typed but required-at-use-time" fields in the file (`opts.source` raising when no git root, `address.provider_name` raising under `--new-host`, etc.).

- **New `mngr extras config` subcommand.** Walks through user-scope config settings the installer would otherwise leave blank. Each step short-circuits if the corresponding setting is already configured, so re-running only prompts for the gaps. Today this covers the default agent type for `mngr create`; future config-related setup will be added as additional walk steps under the same subcommand.
  - With an interactive terminal: an urwid single-select picker of every available agent type plus a "Keep no default" option (Up/Down navigates, Enter confirms, q/Ctrl+C cancels). Writes the selection to `[commands.create] type` in user settings.toml.
  - With `-y` or no TTY: prints the suggested `mngr config set` command + agent type list. Writes nothing.

- **`mngr extras completion` and `mngr extras claude-plugin` also use the urwid picker** (Install / Skip rows) instead of the old `[y/n]:` text prompts. The entire `mngr extras -i` walkthrough now uses one consistent TUI rather than mixing the plugin-wizard's full-screen TUI with bare-text confirmation prompts.

- **`mngr extras -i` now walks through the default agent type prompt** as a final step alongside Plugins / Shell Completion / Claude Code Plugin. `mngr extras` (no flag) reports the current default agent type in its status block.

- **`scripts/install.sh` step 5 is gone entirely.** The default agent type prompt now runs as part of `mngr extras -i` (step 4); no separate bash logic to maintain.

- **New `--kind {agent-type,provider}` filter on `mngr plugin list`.** Projects the plugin list to the canonical set of agent type names or provider backend names. For agent types: includes user-config-defined `[agent_types.X]` entries and handles plugins where the entry-point name differs from the registered agent type name (e.g. `pi_coding` plugin → `pi-coding` agent type).

- **New helper `list_available_agent_types(config)` in `agents/agent_registry.py`:** union of plugin-registered + user-config-defined types. Powers the `--kind agent-type` filter and the resolver's error message.

- **New module `cli/urwid_picker.py`:** a shared single-select picker (`run_single_select_picker(options, title, header_text)`) modeled on the plugin wizard's selection screen but Enter-confirms-the-focused-row. Used by all three `extras` subcommands.

## Migration

Existing users with no `[commands.create] type` set will see a friendly error on their next `mngr create`:

```
No agent type provided. Set a default with:

    mngr config set commands.create.type <name> --scope user

Available agent types: claude, codex, command, ...
Or see `mngr create --help` for how to set it per-command.
```

The fix is to either:
- Run `mngr extras config` (or `mngr extras -i`) and select an option from the picker, or
- Run the `mngr config set commands.create.type <name> --scope user` command directly.

We intentionally do not lazy-write the default on first failure — pulling the default out of source code is the whole point.

## Test plan

- [x] `just test-quick libs/mngr -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'` — green.
- [x] `uv run mngr extras config -y` (no default set, no TTY) → prints suggested `mngr config set` command + agent type list, writes nothing.
- [x] `uv run mngr extras config` in tmux (no default set, has TTY) → urwid picker; Up/Down moves focus between rows, Enter on the focused row writes to user settings.toml, Enter on "Keep no default" writes nothing, q/Ctrl+C cancels.
- [x] `uv run mngr extras config` (already set) → "Default agent type for 'mngr create' is already set to '<name>'." short-circuit, no prompt.
- [x] `uv run mngr extras -i` → walks through Plugins / Shell Completion / Claude Code Plugin / Default Agent Type.
- [x] `uv run mngr plugin list --kind agent-type --active --format '{name}'` — returns plugin-registered + user-defined agent type names.
- [x] `uv run mngr plugin list --kind provider --active --format '{name}'` — returns provider backend names (docker, imbue_cloud, lima, local, modal, ssh, vultr).
- [x] `bash -n scripts/install.sh` — syntax OK.
- [x] CI green on the latest commit.


joshalbrecht:
chatted in person -- probably should allow interactive selection of which agent to default (under the same conditions of interactive install that we already have)

otherwise lgtm and much appreciated!

joshalbrecht:
lgtm (assuming the installer has been tested manually at least once on a clean container + mac?)

PR #1587, opened 2026-05-12
CLAUDE.md: replace ModuleNotFoundError workaround that breaks the host install

joshalbrecht:
[AGENT] ## Summary

Replace the CLAUDE.md "ModuleNotFoundError" workaround. The previous text told agents to run `uv tool uninstall imbue-mngr && uv tool install -e libs/mngr`, which is exactly the operation that wedged the engman driver loop for ~12h on 2026-05-11. The reinstall doesn't re-register the plugin set (mngr_claude, mngr_modal, mngr_wait) that the Dockerfile installs via `mngr plugin add`, so they silently disappear from mngr's venv and `mngr stop` starts failing schema validation on `agent_types.claude`.

Replacement points at the safer worktree-local refresh (`uv sync --all-packages` inside the worktree) and explicitly warns against `uv tool install/uninstall` against `imbue-mngr` or any plugin package (`apps/system_interface`, `libs/mngr_*`).

## Why

Three independent layers of protection collapsed when the original instruction was written into this CLAUDE.md:

1. The mngr CLAUDE.md said "if X breaks, do Y" — but Y is the very thing that breaks the host install.
2. An agent reading just this CLAUDE.md and following its advice ends up with a broken host install. They then have no clean way back without manual operator intervention.
3. The engman driver (which lives in `joshalbrecht/engman` and operates on the mngr worktrees) added a PreToolUse hook `scripts/claude_prevent_mngr_tool_disruption.sh` after the 2026-05-11 incident to block the operation. But a doc that instructs the agent to do it anyway means the hook is the only safety net — if anyone ever weakens it, the loop wedges again.

This change removes the contradiction at the doc level.

## Scope

- One-line replacement in `CLAUDE.md` "Silly error workarounds" section.
- One changelog entry.
- No code changes.

## Test plan

- [x] Diff is exactly two lines (one removed, one added) plus the changelog file.
- [ ] CI green (this PR doesn't touch code, so CI should sail through).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -142,6 +142,6 @@ Every PR must include a changelog entry file. CI will fail if it is missing.
 
 If you get a failure in `test_no_type_errors` that seems spurious, try running `uv sync --all-packages` and then re-running the tests. If that doesn't work, the error is probably real, and should be fixed.
 
-If you get a "ModuleNotFoundError" error for a 3rd-party dependency when running a command that is defined in this repo (like `mngr`), then run "uv tool uninstall imbue-mngr && uv tool install -e libs/mngr" (for the relevant tool) to refresh the dependencies for that tool, and then try running the command again.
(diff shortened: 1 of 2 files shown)


joshalbrecht:
[AGENT] Driver review

**(A) Likely to approve:** yes — single-line CLAUDE.md replacement + changelog, no code changes, all CI green; the replacement directly removes the doc that caused the 2026-05-11 host-install wedge.

**(B) Potential issues:**
- The replacement line is long. Could be split into two sentences (the "do this" half and the "do NOT do this" half) for readability, but the structure is fine as-is.
- The PR doesn't reference the `scripts/claude_prevent_mngr_tool_disruption.sh` hook on the engman side. That hook is the runtime guard; this PR is the doc-level guard. The body's "Why" section calls this out (it's intentional — fixing the doc removes the contradiction at the doc layer, the hook remains the runtime backstop). No action needed.
- No tests added — correct: the change is doc-only.

- Confirm the wording matches your taste. The replacement says "for a workspace package (e.g. `imbue.mngr_*`)" — verify this is the right scope (the original was broader: "any 3rd-party dependency").
- Confirm `uv sync --all-packages` is the canonical recovery action (it is, per the engman CLAUDE.md "Verifying changes to mngr" section).

**Recommendation:** merge

PR #1589, opened 2026-05-12
Workspace-server restart and health-recovery UX on mngr_forward

gnguralnick:
## Summary

Ports the workspace-server health-tracking + restart-recovery feature onto main's `libs/mngr_forward` plugin architecture.

- **Plugin** (`libs/mngr_forward/`) emits `workspace_backend_failure` envelopes on connect errors, mid-stream EOF, unresolved-agent lookups, and infrastructure 5xx (502/503/504 only). Plugin remains observation-only; all policy lives in minds.
- **Health tracker** (`workspace_server_health.py`): per-agent state machine `HEALTHY -> STUCK` (after 5s of continuous failures, timer-driven) `-> RESTARTING` (set by restart endpoint) `-> HEALTHY` (on the first successful probe). A short post-recovery grace window suppresses stale failure envelopes that arrive during the reload after a successful restart. On-change callbacks feed the chrome SSE stream.
- **Background probe loop** polls each non-HEALTHY agent every 2s through the plugin's per-agent subdomain; first 200 flips the tracker back to HEALTHY.
- **Restart endpoint** (`POST /api/agents/{id}/restart-workspace-server`): dispatches `tmux kill-window -t "${MNGR_PREFIX}system-services:svc-system_interface"` + `touch services.toml` via `mngr exec` (local + remote agents handled uniformly). The target session is constructed directly from `MNGR_PREFIX` rather than iterating tmux sessions, since Minds always creates exactly one `system-services` mngr agent per workspace. Blocks up to 15s for recovery, returns 504 on timeout. Dispatch failures roll the tracker back to STUCK.
- **Recovery page** (`GET /agents/{id}/recovery`): restart button + SSE subscription; auto-navigates to a sanitized `return_to` on HEALTHY. Notes that running agents are not interrupted by the restart. Open-redirect protection limits `return_to` to relative `/`-paths and `*.localhost` hosts. The plugin's 503 fallback page is now a styled card with a CSS spinner.
- **Chrome auto-redirect / landing badges / sidebar menu**: when the currently-displayed agent transitions to STUCK, the chrome auto-navigates the content view to the recovery page; "Server not responding" / "Restarting..." badges on landing rows (stuck rows route to recovery, not the workspace); a new "Restart workspace server" entry in the Electron context menu that skips the recovery click and lands on the loader directly.

## Test plan

- [x] Healthy workspace shows no extra UI.
- [x] Kill `svc-system_interface` on an agent; the chrome auto-navigates to the recovery page within ~5s, and a "Server not responding" badge appears on landing.
- [x] Single-failure-then-idle still transitions to STUCK after 5s (timer-driven, not failure-driven).
- [x] Restart button on the recovery page kicks the tmux window + touches services.toml; the page navigates back to the workspace on recovery.
- [x] Stale failure envelopes arriving immediately after a successful restart no longer cause a recovery-page flash (post-recovery grace window).
- [ ] Clicking a stuck row on landing routes to recovery (not directly to the workspace).
- [ ] Electron context menu's "Restart workspace server" entry works for both the currently-focused and other workspaces.
- [ ] Recovery page returns 401 / login when unauthenticated; restart API returns 403.
- [ ] `return_to=https://evil.com/` is rejected (page reloads to itself instead of redirecting off-origin).
- [ ] Application-layer 500s do *not* trigger the recovery flow (only 502/503/504 do).
- [ ] Transient sub-5s blips do not trigger the recovery flow.
- [ ] Dispatch failure (e.g. missing `mngr` binary) rolls tracker back from RESTARTING to STUCK; recovery page shows the failure reason with a re-enabled retry button.
- [ ] CI green (unit + integration tests in `workspace_server_health_test.py`, `templates_test.py`, `test_desktop_client.py`, `server_test.py`).


apps/minds/imbue/minds/desktop_client/app.py
```diff
@@ -1264,6 +1336,251 @@ def _build_workspace_list(
...
+# -- Workspace-server recovery / restart --
+
+# The forever-claude-template bootstrap service manager prefixes every
+# services.toml-managed tmux window with ``svc-`` and runs the workspace
+# server under the ``system_interface`` service entry, so the window we
+# need to kick is always ``svc-system_interface``. The session name is
+# host-specific (``${MNGR_PREFIX}-${MNGR_AGENT_NAME}`` -- e.g.
+# ``devminds-mindtest``), so it is discovered at run time rather than
+# hardcoded.
+_RESTART_TMUX_WINDOW: Final[str] = "svc-system_interface"
```
gnguralnick, line 1454: not sure if there's a better way to auto-discover this? maybe doesn't matter

apps/minds/electron/main.js
```diff
@@ -74,8 +75,15 @@ function toAbsoluteUrl(url) {
...
-  return `${backendBaseUrl}/goto/${encodeURIComponent(agentId)}/`;
+  // `/goto/` lives on the mngr_forward plugin (which owns subdomain
+  // forwarding), not the minds backend. Use mngrForwardBaseUrl when it has
+  // been received; fall back to backendBaseUrl during the startup window
+  // before the mngr_forward_started event arrives (rare -- the user can't
+  // open a workspace in that window).
+  if (!agentId) return null;
+  const origin = mngrForwardBaseUrl || backendBaseUrl;
+  if (!origin) return null;
+  return `${origin}/goto/${encodeURIComponent(agentId)}/`;
```
joshalbrecht, line 86: idk what this is exactly?

apps/minds/electron/main.js
```diff
@@ -970,6 +987,69 @@ async function runChromeSSELoop() {
...
+// killed, the plugin will then serve its 503 loader until the workspace
+// comes back -- giving the user a visible "Workspace server starting"
+// page instead of a silent reload onto the still-live pre-restart UI.
+//
+// Always resolves (never rejects) so callers can chain navigation
+// regardless of network outcome.
+const _RESTART_REQUEST_TIMEOUT_MS = 10000;
+function postRestartWorkspaceServer(agentId) {
+  return new Promise((resolve) => {
+    if (!agentId || !backendBaseUrl) {
```
joshalbrecht, line 1003: I'm a little unsure whether this functionality for restarting *the workspace server* should exist in electron/main.js

Restarting *the container* makes sense out here--but it seems like the workspace server restart stuff is maybe supposed to be at a different level? 

Otherwise there are kind of two different types of restarting, which is a bit confusing...

apps/minds/electron/main.js
gnguralnick: this is actually just fixing a separate bug that exists in main right now, where the sidebar doesn't know the correct URL for minds to open e.g. if you click the "Open in new window" button on a mind. it's also relevant for the restarting case since it's possible to issue the restart from the sidebar. but i can split this out to a separate PR so it's clearer

apps/minds/electron/main.js
gnguralnick: yeah there would be two different types of restarting. i agree it's confusing, i was hoping to build smart auto-escalation logic so the user won't have to think two much. i think there's value in this surgical restart since it'll be much faster than a full container restart.

as for why it exists here, i think there does need to be some logic for it here since it's possible to trigger from outside of the mind? e.g. from the sidebar or when trying to load in. but open to a suggestion if there's a better location for it.

joshalbrecht:
lgtm! feel free to merge after you resolve the conflict

PR #1594, opened 2026-05-12
mngr usage wait command

evgunter:
provides all the functionality needed for a single command which waits until we're near the end of a usage window with extra capacity (but doesn't explicitly show how to schedule this to run multiple times)

---

## Summary

- Add `mngr usage wait --until '<CEL>'`: blocks until at least one source's CEL context satisfies all `--until` filters, then exits 0 (2 on timeout, mirroring `mngr wait`). Composable with shell: `mngr usage wait --until '...' && mngr message ...`.
- Claude writer now emits `window_seconds` per fixed-duration window (`five_hour=18000`, `seven_day=604800`); reader surfaces derived `elapsed_seconds` / `elapsed_percentage` so predicates like "75% of the window has elapsed" don't require callers to hardcode window durations.
- Snapshot pipeline lifted from `cli.py` into a new `mngr_usage/api.py` (also home to `wait_for_usage`). `mngr usage` becomes a click group with `invoke_without_command=True` so the existing snapshot CLI behavior is byte-identical.
- Exit code constants moved from `mngr_wait.primitives` to `mngr.cli.exit_codes` so `mngr_usage` and `mngr_wait` share them directly (no re-export shim).

## Test plan

- [x] `just test-quick libs/mngr_usage libs/mngr_wait libs/mngr_claude_usage -m 'not modal and not docker and not docker_sdk and not acceptance and not release'` — 311 passed locally.
- [x] New `api_test.py`: 18 unit tests covering `derive_elapsed`, `window_render_dict`, `build_source_cel_context`, and `wait_for_usage` polling loop (match on first poll, match-after-N-polls, timeout, source filter, multi-source, no-data, on_tick, retry-on-error, elapsed-percentage end-to-end via CEL).
- [x] CLI integration tests for `mngr usage wait` (match / timeout / invalid CEL).
- [x] Writer test asserts `window_seconds` is emitted for `five_hour`/`seven_day` and omitted for `overage`.
- [ ] CI offload (acceptance + release) to run automatically.
- [ ] Manual verification: needs a live Claude Pro/Max session to exercise the new writer field; the test suite covers the parser + reader contract.


joshalbrecht:
lgtm, excited for these!

PR #1595, opened 2026-05-12
halve offload acceptance runtime (to ~5 min)

evgunter:
## Summary

Four CI acceptance-test speedups that compound: a workflow permissions fix, a `max_parallel` reduction that lets LPT actually pack tests, a Modal session-cleanup approach change that the second change exposed, and a teardown-ordering fix that closes the residual flake. Net measured effect across 14 successful branch runs vs 14 recent main baseline runs: **mean 12m43s → 4m51s (61.8% reduction, 2.6x faster); median 779s → 285s**. Restricted to the 5 post-hook-fix runs (the genuinely-stable code path): mean 4m43s, range 280-287s — variance is also dramatically tighter than main's 657-896s range.

## Changes

1. **`.github/workflows/ci.yml`** — flip `permissions.contents` from `read` to `write` on the `test-offload` and `test-offload-acceptance` jobs. Without this, offload's `git push refs/notes/offload-images` failed silently with `Permission to imbue-ai/mngr.git denied to github-actions[bot]`. The push is best-effort with a warning-only failure mode (`image_cache.rs:286-288`), so the bug had been silent — across seven recent acceptance runs on four different branches before this fix, all seven missed on the same checkpoint SHA. After this fix, the cache seeds on main and subsequent runs hit.

2. **`offload-modal-acceptance.toml`** — lower `max_parallel` from 200 to 50. With 200 slots and ~89 tests, offload's LPT scheduler degenerated to one-test-per-batch (you can't pack into more bins than items), so every batch paid: full pytest cold-start, Modal sandbox creation, and an orchestrator-side `uv run @modal_sandbox.py download` cold-start per batch finish. At 200 concurrent cold-starts on a 2-core runner, the per-call time tail blew out — `download_results` p90 was ~82 s for what is essentially a 134-byte JUnit file. Lowering to 50 lets LPT pack ~2-4 tests per batch on average (max 6) while still keeping the historically-longest tests in their own bin via load-balancing.

3. **`libs/mngr_modal/imbue/mngr_modal/conftest.py`** + **`libs/mngr/imbue/mngr/utils/testing.py`** — fix a false-positive in `modal_session_cleanup`, exposed by change (2). Modal endpoints disagree across a short window: a `modal environment delete X` that returned "Environment 'X' not found" (and the same response from `modal app list --env=X` / `modal volume list --env=X`) would still be followed by `modal environment list --json` returning X a few seconds later. We don't know whether the cause is eventually-consistent delete, eventually-consistent create across replicas, or some mix — but the practical conclusion is the same: no single Modal endpoint snapshot is reliable across that window.

   Fix: trust the synchronous cleanup response. `_cleanup_modal_test_resources` and `modal_test_session_cleanup`'s teardown call new `deregister_modal_test_*` helpers after their delete chain returns `DELETED` or `NOT_FOUND` (typed via a new `ModalCleanupOutcome` enum). The session-end leak detector never even looks at the global list for these resources. `FAILED` keeps the resource tracked and logs an error so the leak detector still has a chance. Orphan envs (test crashed mid-fixture, or cleanup returned FAILED) fall through to a single-shot `modal environment list --json` check at session end — same shape as `_get_leaked_modal_apps` / `_get_leaked_modal_volumes`. CI hourly cleanup script is the ultimate safety net.

4. **Move the leak check from an autouse fixture to a `pytest_sessionfinish` hook (`a011741a1`)** — closes a teardown-ordering bug exposed by change (3). DIAG logging on a failing run showed pytest tearing down the autouse `modal_session_cleanup` fixture BEFORE the non-autouse `modal_test_session_cleanup`, so the leak detector polled a still-registered env for ~30 s, never converged, and fired before the deregister chain had even started:

   ```
   02:41:18                test ends; session-scope teardown begins
   02:41:18..02:41:48      leak-check poll runs (still registered, env genuinely live)
   02:41:48                poll budget exhausts; assertion fires
   02:41:56..02:41:59      modal_test_session_cleanup deregisters (too late)
   ```

   First attempt was a fixture-dependency trick (commit `f15e4e2f4`) — `modal_test_session_cleanup` took `modal_session_cleanup` as an explicit parameter so pytest would (under LIFO) tear down the deregister first. That appeared to work on the run it was committed in, but failed in the same shape on the next run after a main merge. Empirical conclusion: pytest's autouse session-scoped fixtures tear down before non-autouse session-scoped fixtures regardless of declared dependencies — the explicit dep doesn't invert the order across the autouse boundary.

   Robust fix: drop the `modal_session_cleanup` autouse fixture and put the leak detection in a `pytest_sessionfinish` hook. Hooks run after all fixture teardowns by definition, so the deregister chain is guaranteed to have completed first. Failure signal is `session.exitstatus = pytest.ExitCode.TESTS_FAILED` plus a `logger.error` (raising from `pytest_sessionfinish` is silently dropped by pytest).

### Authoritative averages (post-fix vs main baseline)

| | Recent main (n=14) | All post-leak-fix branch runs (n=14) | Post-hook-fix only (n=5) |
|---|---:|---:|---:|
| Mean | 763s (12m43s) | **291s (4m51s)** | **283s (4m43s)** |
| Median | 779s | **285s** | **281s** |
| Range | 657-896s | **274-325s** | **280-287s** |

**Speedup: 61.8% mean (all post-leak-fix), 62.9% mean (post-hook-fix only).** Variance also collapses — the 5 post-hook-fix runs are within a 7-second window vs main's 239-second range.

### Per-run trace decomposition (canonical bracketing runs)

| | Baseline (different PR) | Run A (cache miss, this PR's perm fix) | Run B (cache hit) | **Final (with ordering fix)** |
|---|---:|---:|---:|---:|
| Run ID | `25752162694` | `25756498440 a1` | `25756498440 a2` | **`25771243269`** |
| Config | cache MISS, max_parallel=200 | cache MISS, this PR | cache HIT, max_parallel=200 | cache HIT, max_parallel=50 + deregister + poll + ordering |
| `checkpoint_base_prepare` | 163.9 s | 164 s | **0 s** | 0 s |
| `sandbox_pool_create` | 84.1 s | ~84 s | 84.3 s | **20.5 s** |
| `download_results` p90 | 81.7 s | ~82 s | 81.1 s | **13.7 s** |
| `download_results` max | 244.4 s | — | 112.7 s | **14.9 s** |
| `exec_phase` wall | 401.8 s | ~395 s | 395.0 s | **166.5 s** |
| Avg tests/batch | 1.00 | 1.00 | 1.00 | **2.31** |
| Max tests/batch | 1 | 1 | 1 | **6** |
| **Trace total wall** | **690.6 s** | ~688 s | **566.5 s (-18%)** | **271.4 s (-61%)** |
| Job time | 13m12s | 11m25s | 11m18s | **5m25s** |

## Test plan

- [x] CI passes with all four changes — verified on run `25771243269` (5m25s acceptance, 0 flaky recoveries).
- [x] Cache note successfully pushed to remote — verified via `git log refs/notes/offload-images`.
- [x] LPT actually packs — trace shows avg 2.31 tests/batch, max 6.
- [x] Hybrid leak fix stable on the common path.
- [x] DIAG logging removed.
- [x] **Hook-based ordering fix holds across multiple consecutive runs**: 5 consecutive greens since `a011741a` (a011741a, 87289020, dc8e58bb, 5745e38b, 7bd3fcab), all 280-287s.
- [ ] Watch the first PR merged after this to confirm `[cache] Checkpoint hit` appears in its log.


libs/mngr_modal/imbue/mngr_modal/conftest.py
```diff
@@ -387,23 +462,65 @@ def _delete_modal_volumes(volume_names: list[str]) -> None:
...
+    poll_until(env_list_clean, timeout=_LEAK_POLL_TIMEOUT_S, poll_interval=_LEAK_POLL_INTERVAL_S)
+    if last_candidates is None:
+        # Every poll attempt failed transiently. Don't fail teardown -- the CI
+        # hourly cleanup script is the real safety net -- but make this loud
+        # enough to be grepable in CI logs (the per-poll warnings are easy to
+        # miss in the noise of session-end output).
+        logger.error(
+            "Could not verify leaked Modal environments: every `modal environment list` "
+            "poll over the {}s budget failed. Treating as no leaks; check CI's hourly "
+            "cleanup script (cleanup_old_modal_test_environments.py) for actually-leaked envs.",
```
evgunter: "treating as no leaks" matches the old behavior but might not be the best behavior

joshalbrecht:
lgtm, this is great!

PR #1598, opened 2026-05-12
fix mngr clone for local -> remote

evgunter:
## Summary

Fixes `mngr clone <source-agent> <new-name>@.<provider>` (and `mngr migrate` for cross-host moves) when the source and destination agents live on different hosts.

**Bug** — in `_transfer_source_plugin_data` (`libs/mngr_claude/imbue/mngr_claude/plugin.py`), the destination `host` was passed as both `source_host` and the receiver of `copy_directory`. So `_is_same_machine` returned True and rsync ran on the destination sandbox looking for the source plugin dir there:

```
rsync: change_dir "/Users/.../agents/<src-id>/plugin" failed: No such file or directory (2)
```

**Fix** — `CreateAgentOptions` now carries the source agent's location (host + state-dir path) as a single `HostLocation`:

```python
source_agent_state_location: HostLocation | None
```

populated in `cli/create.py` from `setup.resolved_source.location.host` + `get_agent_state_dir_path(...)`. The plugin uses `source_agent_state_location.host` as the rsync `source_host` and for the existence check. Works for local→remote, remote→local, and remote→remote clones.

**Plumbing along the way**:
- Moved `HostLocation` from `libs/mngr/imbue/mngr/hosts/host.py` to `libs/mngr/imbue/mngr/interfaces/host.py` so `CreateAgentOptions` (in the interfaces module) can use it as a field type without a circular import. 9 import sites updated.
- Promoted `_path_exists` from a private method on `Host` to a public concrete `path_exists` on `OuterHostInterface`. Plugins shouldn't reimplement host primitives, and the body only needs `is_local` + `execute_idempotent_command`, both already on the interface. `Host._path_exists` (no external callers) deleted.
- `_transfer_source_plugin_data` signature simplified: `host` param dropped (it was the same value as `self.host` by construction).

## Known limitation — Claude doesn't actually resume the source's session

This PR makes the data transfer succeed, but a cloned agent does **not** show the source's conversation history when Claude starts. The session JSONL ends up on the destination host, but Claude can't find it because:

1. **Project dir is filed under the source's encoded work_dir.** After clone, the session JSONL is at `plugin/claude/anthropic/projects/<source-encoded-work-dir>/<session>.jsonl`. Claude searches `projects/<encoded-current-work-dir>/<session>.jsonl`, and the dest agent's work_dir encodes differently (e.g. `/mngr/projects/agent-...` on Modal). So no match.
2. **`claude_session_id` is overwritten with the new agent's UUID.** That file lives at the *top* of the agent state dir, not inside `plugin/`, so it isn't transferred by this PR's plugin rsync. The startup command then does `claude --resume <new-uuid>` (and falls back to a fresh session when the file doesn't exist).

The existing `--adopt-session` machinery in the plugin (`on_after_provisioning` + `encode_claude_project_dir_name`) already handles both of those steps for opt-in single-session adoption. Wiring that path automatically into the clone flow is the follow-up. **Tracked in a stacked PR.**

## Regression test

`test_clone_local_claude_agent_to_modal` in `libs/mngr_claude/imbue/mngr_claude/test_claude_agent_modal.py` — a `@pytest.mark.release` end-to-end test against real Modal.

**What it checks**:
1. Creates a real local Claude agent (the source) — `mngr create <name> claude --yes --pass-env ANTHROPIC_API_KEY -- --dangerously-skip-permissions -p "just say 'hello'"`. The local source actually invokes claude so the source's `plugin/claude/.../projects/<work_dir>/<session>.jsonl` exists on disk before we clone.
2. Clones to a fresh Modal sandbox — `mngr clone <src> <dst>@.modal --yes`. This is the path that originally crashed.
3. Asserts `clone` exits 0.
4. Destroys the cloned target. The agent's `preserve_sessions_on_destroy` path pulls the target's `plugin/` back to the local `host_dir` (same mechanism exercised by `test_destroy_modal_agent_preserves_sessions_locally`).
5. Asserts the preserved-sessions directory for the target contains at least one non-empty `*.jsonl` (via `_assert_sessions_preserved`). The session JSONL there came from the source via the clone — so a silent regression that skips the transfer (e.g. `path_exists` returning False against the wrong host, or the wrong `source_host` getting threaded through) would leave that dir empty and trip the assertion.

This is strictly stronger than the original "clone exits 0 and stderr doesn't contain `rsync failed`" check — that earlier form would have passed under a silent-skip regression.

**Verified locally end-to-end against real Modal**: 1 passed in 104s (full create local → clone to modal → destroy target → assert preserved sessions; Modal app/volume/environment all reclaimed by session-scoped cleanup).

## Test plan

- [x] `just test-quick "libs/mngr_claude/imbue/mngr_claude/plugin_test.py"` (183 passed)
- [x] `just test-quick "libs/mngr/imbue/mngr/cli/clone_test.py libs/mngr/imbue/mngr/cli/migrate_test.py libs/mngr/imbue/mngr/cli/create_test.py"` (131 passed)
- [x] `just test-quick "libs/mngr/imbue/mngr libs/mngr_claude"` (4510 passed, 3 skipped) — broader sweep including ratchet tests after the `HostLocation` move and `path_exists` promotion
- [x] `just test libs/mngr_claude/imbue/mngr_claude/test_claude_agent_modal.py::test_clone_local_claude_agent_to_modal` — real Modal end-to-end (104s, pass)
- [ ] CI runs the full offload suite (including `test-offload-acceptance` and `test-offload-release` against Modal)

## Review

- `/verify-architecture`: passed after applying the suggested cleanups (single `HostLocation` field instead of two coupled fields with a runtime assert; `path_exists` promoted to the interface instead of duplicating `Host._path_exists` in the plugin).
- `/autofix`: 1 iteration produced fixes (2 issues identified). The accepted commit strengthened the release test from a substring check on stderr to the preserved-sessions assertion described above. A second iteration found no further issues.
- `/verify-conversation`: clean (one MAJOR deviation noted: I leaned on CI for the full test suite instead of running `just test-offload` locally after the late prose/cleanup commits).


joshalbrecht:
lgtm  (just not merging because it has stuff stacked, will let you merge)

PR #1599, opened 2026-05-12
mngr_claude: bound macOS keychain reads with a 10s timeout

weishi-imbue:
## Summary

- Caps `security find-generic-password` reads at 10s so a hidden macOS keychain ACL prompt can no longer hang `mngr create` indefinitely.
- On timeout, log a clear warning pointing at Keychain Access.app + the `convert_macos_credentials=false` escape hatch, and let agent creation continue without merging the keychain `primaryApiKey`.
- Adds a unit test for the new timeout branch.

## Why

Symptom: Lima / remote agent creation stalls at `_setup_per_agent_config_dir` with no further log output. The host-side `mngr create` subprocess has a stuck `security find-generic-password -l "Claude Code" -w` child waiting forever on a keychain ACL dialog that the user never sees / dismisses. This happens reliably when the running binary's code signature doesn't match the keychain entry's ACL — e.g. after an app is re-signed (our hot-patched ToDesktop bundles) or on first launch of a freshly-installed binary.

Previously the call had no timeout, so a stuck `security` child would block `_merge_keychain_api_key`, which would block `_setup_per_agent_config_dir`, which would block the whole `mngr create`. End users see a "Creating your project..." spinner that never resolves.

## Test plan

- [x] Added `test_read_macos_keychain_credential_returns_none_on_timeout` — asserts the function returns `None` when `FinishedProcess.is_timed_out=True`, matching the no-credential branch's existing behaviour.
- [x] Existing tests (`..._returns_value_on_success`, `..._returns_none_on_nonzero_exit`, `..._returns_none_on_process_setup_error`) keep passing — the change only adds a new code path.
- [x] Manually reproduced on a real keychain hang: pre-fix, `mngr create` for a fresh Lima agent stalled indefinitely after `_setup_per_agent_config_dir`. Post-fix, the warning fires at 10s and provisioning continues; the merged `.claude.json` is missing only `primaryApiKey` (which is harmless when `LATCHKEY_GATEWAY` is set, as it is in minds.app).


weishi-imbue:
Closing this. The 10s timeout is the wrong fix.

After more discussion the right architecture is:

1. **In `mngr` CLI (direct terminal use)** — no timeout is needed. A user running `mngr create` in their terminal can see the macOS keychain prompt and respond at their own pace. The pre-existing no-timeout behaviour is correct for this path.

2. **In `minds` (Electron app)** — the keychain access should be owned by `minds.app` itself, not by `mngr_claude`. minds reads the keychain once via its Electron main process, gets the prompt attributed to minds.app's signature (and "Always Allow" sticks against minds.app's stable signing cert), then passes the credential to `mngr create` via env var. `mngr_claude`'s existing `ANTHROPIC_API_KEY env present → skip keychain probe` branch handles that case with no `mngr` changes.

Additional reason this PR misses the mark: `mngr_claude` makes 3-4 separate `security find-generic-password` calls per agent create (`_merge_keychain_api_key`, `_provision_keychain_credentials` for both "Claude Code" and "Claude Code-credentials", `_has_api_credentials_available`). The 10s timeout only short-circuits the first one — the next ones still each fire their own prompt. So even from a defensive standpoint, this PR only partially addresses the symptom.

The minds-side ownership refactor is now drafted as a spec; PR will come from there.

PR #1536, opened 2026-05-07
Restore Modal Dockerfile + DRY source-dependent setup via offload v0.9.5 post_patch_cmd

DanverImbue:
## Summary

- Restores Modal compatibility for the standard mngr Dockerfile by collapsing the multi-stage build into a single `FROM python:3.12-slim`. mngr's Modal image builder asserts `not dfp.is_multistage` (`libs/mngr_modal/imbue/mngr_modal/instance.py:3277`), so the previous `FROM rust:1-bookworm AS offload-builder` stage broke `mngr create <agent>@.modal -b file=libs/mngr/imbue/mngr/resources/Dockerfile` for end users.
- Bumps offload `0.9.2` -> `0.9.5` to pick up `post_patch_cmd` and uses it to DRY all source-dependent setup between the Dockerfile RUN and offload's thin-diff path.
- Re-introduces the `current.tar.gz` keyframe workflow (transparent to existing producers) by routing extraction through the new shared script. The typical user flow via `.mngr/settings.toml`'s pre_command_scripts hook now works again.

See [specs/restore-modal-dockerfile/plan.md](specs/restore-modal-dockerfile/plan.md) for the full plan and rationale.

### Commits

1. Bump offload pin from 0.9.2 to 0.9.4
2. Collapse Dockerfile to single FROM stage
3. Move source-dependent setup into `scripts/post-source-setup.sh`
4. Wire `scripts/post-source-setup.sh` as offload's `post_patch_cmd`
5. Add plan, changelog, and `.sculptor/docs.md`

## Test plan

- [x] CI: `test-offload` (unit + integration) green
- [x] CI: `test-offload-acceptance` green, including `test_mngr_create_with_dockerfile_on_modal`
- [ ] CI: release tests green, including `test_mngr_create_with_default_dockerfile_on_modal` (the canonical reproducer)
- [ ] Local: `uv run mngr create test-modal-dockerfile@.modal -b file=libs/mngr/imbue/mngr/resources/Dockerfile` reaches `Done.`

Co-authored-by: Sculptor <sculptor@imbue.com>

PR #1600, opened 2026-05-12
support multistage dockerfiles in modal

evgunter:
`mngr create asdlkfj -t modal` was failing. i'm not sure if this is the right fix (it's more of an investigation)

---

## Summary
- `mngr create -t modal` (and `mngr clone ... -t modal`) hit `AssertionError: Multistage Dockerfiles are not supported yet` against the mngr default Dockerfile, which became multistage in ad60b5f70 when offload moved to a `rust:1-bookworm AS offload-builder` builder stage.
- The Modal provider asserted `not dfp.is_multistage` and chained per-instruction `image.dockerfile_commands`. Chaining produces one image layered on the next with no way to honor `COPY --from=<stage>`, so multistage genuinely cannot ride that path.
- Route multistage Dockerfiles through `modal.Image.from_dockerfile` (single-shot, no per-layer caching but supports multistage) while keeping the existing per-instruction caching path for single-stage Dockerfiles. Exposed via a new `image_from_dockerfile` method on `ModalInterface`.

## Root cause
`libs/mngr_modal/imbue/mngr_modal/instance.py:3289` asserted `not dfp.is_multistage`. Triggered from `_create_modal_image_for_host` -> `_get_modal_image_definition` -> `_build_image_from_dockerfile_contents`. Stack from a real `mngr clone transfer transfer1 -t modal` run is in the description below.

## Notes
- `modal.Image.from_dockerfile` reads the file lazily inside `.build()`, so the helper's `TemporaryDirectory` had to outlive the function call. For the multistage path, the temp dir handle is pinned in a module-level list for the process lifetime; OS-managed `/tmp` reclamation handles cleanup if Python exits uncleanly.
- Per-layer caching is intentionally preserved for single-stage Dockerfiles; only multistage takes the single-shot path. There is no straightforward way to per-layer-cache a multistage build through Modal's incremental image API.

## Test plan
- [x] `uv run pytest libs/mngr_modal/imbue/mngr_modal/testing_provider_test.py` (180 passed)
- [x] `uv run pytest libs/modal_proxy/` (77 passed)
- [x] New unit test: `test_build_image_from_multistage_dockerfile` exercises the new multistage branch through the testing provider.
- [ ] The existing release test `test_mngr_create_with_default_dockerfile_on_modal` should now actually exercise multistage end-to-end on Modal (it had been masked by the assert).


libs/modal_proxy/imbue/modal_proxy/interface.py
```diff
@@ -261,6 +261,27 @@ def image_from_id(self, image_id: str) -> ImageInterface:
...
+
+        Supports multistage Dockerfiles. Builds the whole Dockerfile as a single
+        Modal image without per-layer caching, in contrast to chained
+        ``dockerfile_commands`` which caches per instruction but does not
+        support multistage builds.
+
+        ``build_args`` override matching ``ARG`` defaults in the Dockerfile,
+        equivalent to ``docker build --build-arg KEY=VALUE``.
+        """
+        ...
```
evgunter, line 283: hm

joshalbrecht:
@qi-imbue and @danverbraganza have some context here!

I don't think this is the right way to fix it, but I do think we want modal creation to work again

qi-imbue:
I remember @danverbraganza has another pending PR that makes the Dockerfile single-stage again

danverbraganza:
@evgunter That's correct, the version that restore our Modal image to a single-stage is here: https://github.com/imbue-ai/mngr/pull/1536

evgunter:
closing in favor of https://github.com/imbue-ai/mngr/pull/1536

PR #1601, opened 2026-05-12
make `mngr clone` handle claude session files properly

evgunter:
## Summary

Makes a cloned claude agent actually **resume** the source's conversation end-to-end on the destination — the cloned claude sees the source's history and can answer recall prompts about it. Builds on the now-merged #1598, which fixed the cross-host plugin/ rsync but stopped short of making claude itself pick up the transferred session.

## Five things, all required

After #1598's rsync lands the source's `plugin/` on the destination, claude on the destination still won't pick up the source's session unless all five of these are also true:

1. **Project subdir matches the *resolved* destination work_dir.** Session JSONLs are filed under `plugin/claude/anthropic/projects/<encoded-work-dir>/<sid>.jsonl`. Claude on the destination encodes `realpath(cwd)`. On Modal `/mngr/projects/agent-X` is a symlink onto `/__modal/volumes/<vol-id>/projects/agent-X`; encoding the unresolved path puts the JSONL where claude won't find it.
2. **`claude_session_id` on the destination is the actual JSONL filename's stem.** The startup command resolves `MAIN_CLAUDE_SESSION_ID` from this file. The source's own `claude_session_id` is unreliable — when the source ran `claude -p`, claude generated its own id rather than honoring `--session-id`, so the source's file (defaulted to the agent UUID by the SessionStart hook) disagrees with the JSONL on disk. The filename is the ground truth.
3. **The rsynced `sessions-index.json` is removed.** Claude consults this per-project index for session lookup; the source's index has stale entries pointing at source paths, so claude reports `No conversation found with session ID: <sid>` and falls through to a fresh session even with the JSONL physically at the right path. (Same trick `--adopt-session` uses; this PR consolidates the logic.)
4. **`claude_session_id_history` is carried forward** verbatim for traceability of the adopted run.
5. **The combo with `--adopt-session` is rejected at the CLI boundary.** Both are session-adoption directives; without an explicit check, one would silently win and the other would be dropped.

## Fix layout

In `libs/mngr_claude/imbue/mngr_claude/plugin.py`:

- `_resolve_work_dir_on_host()` runs `readlink -f self.work_dir` over the destination's `OnlineHostInterface` (warns on fallback).
- `_transfer_source_plugin_data` (in `provision()`, before identity rewrites) is just the rsync — single-purpose now.
- `on_after_provisioning` is the single dispatch point for session adoption: independent `if`s for `--adopt-session` (`_adopt_explicit_sessions`) and clone (`_adopt_cloned_session`), with a defensive top-level assert that they're not both set (the CLI rejects the combo upstream).
- `_adopt_cloned_session(host, source_location: HostLocation)`:
  - Reads the source's active project subdir name + session id from the source host via a single `ls -t .../projects/*/*.jsonl | head -n1`.
  - Carries over `claude_session_id_history`.
  - Renames the source-encoded project subdir on the destination with a single `mv`. **Refuses to clobber** a pre-existing target (logs a warning and bails — collision could only happen if the source had a multi-cwd setup whose encoded name coincidentally matched ours; silent clobber would risk losing data we don't realize is there).
- `_finalize_adopted_session(host, project_dir, session_id)` is shared between both adoption paths — drops the stale `sessions-index.json` and writes the session id to `claude_session_id`.
- `on_before_create` rejects `--adopt-session` combined with `--from <agent>` with `UserInputError`.

`source_agent_state_location.host` (threaded through `CreateAgentOptions` by #1598) scopes the source-side `ls -t` correctly across hosts.

### Unit (`plugin_test.py`)
- `test_clone_adoption_copies_plugin_dir` — non-project plugin/ data preserved as-is, agent state-dir's own `data.json` untouched.
- `test_clone_adoption_skips_when_no_plugin_dir` — graceful no-op when source has no `plugin/`.
- `test_clone_adoption_rekeys_project_dir` — project-dir rename + `claude_session_id` set to JSONL stem.
- `test_clone_adoption_uses_jsonl_filename_not_source_session_id_file` — pins the rationale that the JSONL filename is the ground truth, not the source's `claude_session_id` file (which holds the stale agent UUID for `claude -p` runs).
- `test_clone_adoption_refuses_to_clobber_existing_target` — covers the data-safety guard. Seeds two source project subdirs (one whose name matches the destination's encoded work_dir, one holding the more-recently-active JSONL) with deterministic mtimes via `os.utime`, then asserts both subdirs survive intact and `claude_session_id` is not written.
- `test_on_before_create_rejects_adopt_session_with_clone_source` — covers the new combo rejection.

### Release (`test_claude_agent_modal.py`)
- `test_clone_local_claude_agent_to_modal_resumes_session` — full end-to-end resume verification against real Modal. Plants a unique two-token secret in the source's prompt, clones to a fresh Modal sandbox, drives the cloned agent via `mngr message --message <recall>` to combine the two tokens, polls `mngr capture --full` until the combined form appears in the tmux pane, then destroys the target and asserts the combined form survives the preserve-sessions round trip.
- The combined form does not appear in the source's portion of the JSONL, so finding it in the cloned pane proves both (a) claude on the clone resumed the source's session (and therefore saw both halves) and (b) the model produced new content informed by it. Same pattern as `test_adopt_session.py`, just routed through `mngr capture` since the agent's tmux session lives on the Modal sandbox.

**Verified locally end-to-end against real Modal**: 1 passed in 132s.

## Review

- **`/verify-architecture`**: passed across two iterations. First iteration applied 2 of 3 cleanups (read session id source-side instead of via post-rsync `ls -t`; single `mv` instead of `cp -a && rm -rf`). Second iteration consolidated session-id finalization between `--adopt-session` and clone via `_finalize_adopted_session`, then moved clone adoption into `on_after_provisioning` to share the lifecycle hook with `--adopt-session`.
- **`/autofix`**: 5 runs across the branch's life. Multiple commits accepted (logging improvements, changelog corrections, robustness fixes, dead-code drop, two unit-test additions, warn-on-fallback, DRY refactor). Two were reverted per user feedback: a PR-number reference in a docstring ("why is a docstring talking about a PR number at all"), and an `rm -rf` clobber on rename collision ("can we error if it's already there?" — replaced with refuse-and-warn).
- **`/verify-conversation`** (haiku): clean across all incremental reviews.


joshalbrecht:
lgtm

PR #1602, opened 2026-05-13
capture session costs in mngr usage

evgunter:
kind of sketchy and ad-hoc but should work in simple cases

---

## Summary

Adds per-session cost tracking to `mngr usage`, **with subscription cost (imputed) and API-key cost (real billable spend) aggregated separately** so the two never get conflated. Cost is process-aware: cumulative in Claude Code's emissions, the reader undoes that encoding so each session's reading reflects its own contribution.

**Writer (`mngr_claude_usage`):** Captures `session_id`, `cost.*`, and `rate_limits` from Claude Code's statusline payload on every render. Cost tracking works for direct-`ANTHROPIC_API_KEY` users too — they never receive `rate_limits` (Pro/Max only), but `cost` is always present, so the writer emits whenever `rate_limits` OR `cost` is in the payload.

**Reader (`mngr_usage`):** Scans every line of each agent's events file, **partitions each agent's events into Claude Code processes via cost-drop detection** (cost is process-cumulative; `/clear` does not reset it; only a quit+relaunch does), and within each process builds a `SessionCostRecord` per session whose `cost` is its delta from the prior session's cumulative reading. Sessions are then filtered to a recency window (`--since`, default 24h).

Each process is classified by **auth mode**: `SUBSCRIPTION` if any event in it carried a non-empty `rate_limits` payload (rate_limits is emitted only by Claude.ai Pro/Max auth, after the first API response), else `API_KEY`. Every session inherits its process's mode. Aggregates on `UsageSnapshot` are split accordingly into `subscription_cost` (sum of imputed contributions) and `api_cost` (sum of real-spend contributions). There is intentionally **no combined `cost` field** — predicates and templates must say which mode they mean.

Rate-limit windows still reduce freshest-wins across agents (they're an account-level counter, identical across all agents on the same Anthropic account). Every event must carry a non-empty `session_id`; events lacking one are dropped with a `logger.warning` so data loss is visible.

**Renames:**
- Events path: `events/<source>/rate_limits/events.jsonl` → `events/<source>/usage/events.jsonl`
- Writer script: `claude_rate_limits_writer.sh` → `claude_usage_writer.sh`
- Env var (test override): `MNGR_RATE_LIMITS_EVENTS_PATH` → `MNGR_USAGE_EVENTS_PATH`
- Event `type`: `rate_limit_snapshot` → `cost_snapshot`

**`mngr usage` surfaces:**

- **Human (default — summary):** Under each `[source]` header, one cost line per mode that contributed in the window, followed by window lines:

  ```
  [claude]
  subscription cost (imputed): $0.42 (2m ago)
  api cost: $1.25 (just now)
  5h: 73% used, resets in 2h 30m
  ```

  Usually only one of the two lines renders (most users are on one auth mode at a time). Subscription line first when both appear; subscription always carries an `(imputed)` callout to mark that it's informational, not billable.

- **Human with `--detail`:** Adds indented per-session lines (newest-first) between the cost lines and the window lines, **each tagged with the session's mode**:

  ```
  [claude]
  api cost: $1.30 across 2 sessions in last 24h
    [api] currents: $0.30 (just now)
    [api] oldersee: $1.00 (2h ago)
  ```

  Suppressed when there's exactly one session (the cost line already names that session's reading).

- **JSON (default — summary):** Per source — `subscription_cost`, `api_cost`, `subscription_session_count`, `api_session_count`, `session_count` (total), `since_seconds`, plus window keys. **No combined `cost` field.** `sessions[]` is omitted to keep the default payload small.

- **JSON with `--detail`:** Adds `sessions[]` (newest-first, each carrying a `cost_mode` of `"SUBSCRIPTION"` or `"API_KEY"`).

- **CEL / `mngr usage wait`:** `subscription_cost.*` / `api_cost.*` are the per-mode aggregates; `sessions[]` always present regardless of `--detail`; each session carries `cost_mode`. Predicates must pick the mode they mean.

- **Format template:** `{subscription_cost.total_cost_usd}` / `{api_cost.total_cost_usd}` plus `{subscription_session_count}` / `{api_session_count}` / `{session_count}`. The format-template surface intentionally doesn't expose list-indexed paths; per-session callers should use `--format json --detail`.

```
mngr usage                                                       # summary, per-mode lines
mngr usage --detail                                              # per-session breakdown, mode-tagged
mngr usage --since 7d                                            # aggregate over 7 days
mngr usage wait --until 'api_cost.total_cost_usd > 20.0'         # real billable spend > $20
mngr usage wait --until 'subscription_cost.total_cost_usd > 50'  # imputed >$50 of value
mngr usage wait --until 'sessions[0].cost.total_cost_usd > 5'    # most recent session only
```

`--detail` is deliberately separate from `--verbose` (which controls log level): conflating "log more" with "show more" would be a mild but real UX trap.

## Notes / accepted tradeoffs

- **No combined `cost.*` field by design.** Subscription cost is imputed by Claude Code (estimated value of usage that's actually billed flat-rate) and api cost is real metered spend. Summing the two would be misleading. Callers pick which they want.
- **Mode is classified per Claude Code process.** A user who swaps auth via quit+relaunch (subscription → API-key or vice versa) produces two separate process records with different modes — the cost-drop boundary detection already partitions the agent's stream.
- **Path rename (`rate_limits` → `usage`)** is a breaking change for already-provisioned agents: they keep writing to the old path until re-provisioned. Accepted because the original work hadn't shipped yet.
- **Process-cumulative cost handling.** Cost grows across `/clear` boundaries within one Claude Code process; only a relaunch resets it. The reader uses cost-drop signals (cumulative reading strictly lower than the prior event) to detect process restarts within each agent's stream and applies fresh delta baselines per process. Verified empirically by tmux-driven smoke test.
- The reader scans the whole events file on every poll. Fine today (<1MB after thousands of renders) but linear in file size; a tail-only fast path is plausible future work.

## Test plan

- [x] Unit + integration tests pass locally (`libs/mngr_usage`: 127; `libs/mngr_claude_usage`: 79; total 206 after the cost-mode split + autofix work).
- [x] Ratchet tests pass (incl. `PREVENT_LITERAL_MULTIPLE_OPTIONS` — `CostMode` is an `UpperCaseStrEnum`).
- [x] Autofix clean at branch tip (4 iterations on the cost-mode split alone, all 7 fixes kept).
- [x] Conversation review clean.
- [x] CI: Unit+Integration, test-offload, test-docker, cleanup-modal all PASS on the latest tip.
- [ ] CI test-offload-acceptance pending.
- [x] **Manual smoke:** Provisioned Claude agent (Pro/Max), confirmed `mngr usage` shows `subscription cost (imputed): $...` plus the rate-limit window lines; `--detail` shows `[sub]` per-session lines.
- [ ] **Manual smoke:** Provision a Claude agent with `ANTHROPIC_API_KEY` set, run a prompt, confirm `mngr usage` shows `api cost: $...` (no `(imputed)` callout, no rate-limit window lines).


joshalbrecht:
lgtm, excited for this!

PR #1606, opened 2026-05-13
destroy: split --force into --force and --yes

joshalbrecht:
[AGENT] Adds a dedicated `-y` / `--yes` flag to `mngr destroy` that just skips the
interactive confirmation prompt. `-f` / `--force` is unchanged in semantics
(destroy running agents, swallow `AgentNotFoundError`) and now implies
`--yes`, so existing scripts that pass `--force` keep working without
modification.

## Motivation

Today, scripted callers that simply want to skip the confirmation prompt
on a known non-running agent have to use `--force` because that's the
only escape hatch from the prompt -- but `--force` is more permissive
than the script needs (it also bypasses the running-agent safety check
and swallows missing-agent errors). The conventional CLI split is now
restored:

- `-f` / `--force` -- bypass safety checks (destroy running agents, ignore
  missing-agent errors). Implies `--yes`.
- `-y` / `--yes` -- skip the confirmation prompt only. Running-agent
  safety check still applies.

## Changes

- `libs/mngr/imbue/mngr/cli/destroy.py`: add `yes` field on
  `DestroyCliOptions`, register the `-y` / `--yes` click flag, and gate
  the confirmation prompt on `force or yes` (via a small
  `_should_skip_confirmation` helper). The running-agent check stays
  gated on `force` only (via `_bypasses_running_check`). Updates the
  command's help synopsis and description.
- `libs/mngr/imbue/mngr/cli/destroy_test.py`: covers the four-point
  matrix (stopped/running x yes/force), the no-flags default, the new
  `--yes` flag appearing in `--help`, and that `--yes` does NOT swallow
  `AgentNotFoundError` (unlike `--force`).
- `changelog/mngr-cvu0.md`: changelog snippet describing the
  user-facing change.

## Verification

- `mngr destroy --help` shows `-f, --force` and `-y, --yes` as separate
  options with distinct one-line descriptions.
- `mngr destroy <nonexistent> --yes` exits 1 (not bypassed).
- `mngr destroy <nonexistent> --force` exits 0 (bypassed -- unchanged).
- Full mngr test suite (`-m "not tmux and not modal and not docker and
  not docker_sdk and not acceptance and not release"`): 3856 passed,
  1 skipped, 2 failed. The 2 failing tests
  (`issue_reporting_test::test_handle_not_implemented_error_interactive_*`)
  are pre-existing on `main` and unrelated to this change (verified by
  `git stash`-ing and re-running).


joshalbrecht:
[AGENT] Driver review

**(A) Likely to approve:** yes — the split between `--force` (bypass safety checks) and `--yes` (skip prompt only) is the conventional CLI shape, `--force` still implies `--yes` so existing scripts keep working, the diff is small and surgical, and the changelog entry captures the user-facing change clearly.

**(B) Potential issues:**
- Two thin one-line helpers (`_should_skip_confirmation`, `_bypasses_running_check`) were extracted purely to make the predicate logic unit-testable without mocking `AgentInterface`. Reasonable given the style guide forbids `unittest.mock` and the real-agent tests are tmux-marked (skipped from the standard quick suite), but the worker itself flagged this as a small abstraction added for test access — worth deciding whether the abstraction earns its keep or whether you'd prefer the gate inlined.
- The four-point flag matrix (stopped/running × yes/force) was collapsed to three tests on the grounds that `(force=T, yes=T)` is predicate-identical to `(force=T, yes=F)`. Logically sound, but the literal task wording asked for the four explicit scenarios — worker would revert commit `b16812d00` if you'd prefer the documented four-test version.
- `AgentNotFoundError` swallow at `_destroy_single_online_agent` still gates on `opts.force` directly rather than going through `_bypasses_running_check`. Worker kept the helper's scope narrow to "running check" rather than renaming it to cover both behaviors. Slight smell that "bypass safety" is split across one helper + one raw flag check.
- `test_destroy_help_documents_yes_and_force_separately` only asserts both flags appear in `--help`; it doesn't pin their distinct-semantics wording. A future help-text refactor could silently weaken the docs without failing this test.
- CI still in progress at review time (`test-offload`, `test-offload-acceptance` running; others green). Full test-suite run locally was 3856 passed / 2 pre-existing failures on main (verified via stash + rerun) / 1 skipped.

- Are the two predicate helpers worth their weight, or should the gate logic just be inlined?
- Three predicate tests vs. four explicit scenario tests — preference?
- Should `--yes` (not just `--force`) also tolerate `AgentNotFoundError`? Worker assumed no, since the task said to preserve `--force` semantics and `--yes` is intentionally more conservative.

**Recommendation:** merge (after CI lands green)

PR #1607, opened 2026-05-13
mngr list: degrade gracefully on per-provider failures

joshalbrecht:
[AGENT] Fixes the bug where `mngr list` exits 1 if any single provider
fails (Modal unauthed, Docker daemon down, etc.), even when listings
from the other providers succeeded. New behavior: print the agents that
were found, log a WARNING per unavailable provider, exit 0. `mngr
destroy --force`'s GC pass has done this for a while; this aligns
`mngr list` with that shape.

## Repro (before)

From `/code/.external_worktrees/mngr` on `main`, with Modal unauthed
and Docker daemon unavailable:

```
$ uv run mngr list
... local-provider agents printed ...
Error: Modal is not authorized. ...
$ echo $?
1
```

The same environment with `mngr destroy --force <agent>` exits 0 with
a WARNING (`This does not affect the destroy operation, which
completed successfully`). That divergence is what this PR closes.

## Design and why I did not follow the in-flight PRs as-is

Five active PRs touch this area:

- **#1556** (`wz/mngr-list-fatal-as-error`, OPEN, big): typed
  `ProviderUnavailableError` / `ProviderNotAuthorizedError` /
  `ProviderDiscoveryError` hierarchy + two-tier severity. The "After"
  matrix explicitly keeps `mngr list` at **exit 1** for
  `ProviderNotAuthorizedError` (Modal auth fail). That is the opposite
  of what this ticket asks for — Modal auth fail must be a warning, not
  a fatal error, to match destroy.
- **#1538** (`wz/mngr-list-warnings-explicit`, OPEN): adds a
  `WarningInfo(source, type, message)` and an `on_warning` callback on
  `list_agents`. Providers explicitly construct `WarningInfo`. This is
  a clean structured-warnings channel but is a much bigger refactor
  than the ticket warrants.
- **#1537** (`wz/mngr-list-abort-partial-json`, OPEN, stacked on
  #1556): valid partial JSON under `--on-error abort`. Orthogonal.
- **#1534** (`wz/discovery-events-file-on-errors`, CLOSED): writes
  partial `DISCOVERY_FULL` event. Orthogonal.
- **#1532** (`wz/mngr-list-warnings`, OPEN): the loguru-sink variant of
  #1538. Same architectural scope.

This PR takes the **destroy-mirror** path the ticket asked for, which
is the simplest design that meets the success criteria and is the
smallest delta against `main`. Concretely: instead of introducing a
typed exception hierarchy or a separate `warnings` channel, it
demotes per-provider construction/discovery failures to warnings at
the same spot destroy does — the `_discover_hosts_for_gc` analogue in
`api/list.py`. The failure is still recorded as
`ProviderErrorInfo` on `result.errors` (no API loss for programmatic
consumers); a new `is_provider_unavailable=True` flag distinguishes
"a provider couldn't load" from "a real data error" so the CLI knows
which entries to skip when deciding the exit code. The two
WZ-authored PRs (#1538 and #1556) are not blocked by this — they can
land later and add their structured warning channel on top, treating
`is_provider_unavailable` as a routing hint or replacing it entirely.

### `libs/mngr/imbue/mngr/api/list.py`

- New `_record_provider_unavailable` helper: logs `WARNING`, emits the
  existing discovery error event, appends a `ProviderErrorInfo` (with
  `is_provider_unavailable=True`) to `result.errors`, fires
  `on_error`. Does **not** raise.
- `_construct_and_discover_for_provider` (batch) and
  `_construct_discover_and_emit_for_provider` (streaming) both
  delegate to it on provider-construction/discovery failure. They no
  longer honor `error_behavior` for that catch — matching how
  `_discover_hosts_for_gc` in `api/gc.py` behaves.
- `ProviderErrorInfo` gets `is_provider_unavailable: bool = False`.
  `build_for_provider` accepts the flag. Default `False`, so the
  existing batch-mode mismatched-provider path (which raises
  `ProviderInstanceNotFoundError`) keeps producing fatal errors.
- Host- and agent-level error handling
  (`_process_host_with_error_handling`, `_handle_listing_error`)
  unchanged: `--on-error` still controls them.

### `libs/mngr/imbue/mngr/cli/list.py`

- New `_has_fatal_errors(errors)` helper: returns True iff `errors`
  contains anything other than provider-unavailable entries.
- All four exit-1 sites (`_list_jsonl`, `_list_streaming_human`,
  `_list_streaming_template`, `_run_list_iteration`) replace
  `if result.errors: ctx.exit(1)` with the helper. Warnings are still
  logged.

### Tests

- `api/list_test.py`: two ABORT-mode-raises tests
  (`test_list_agents_abort_mode_propagates_top_level_mngr_error`,
  `test_construct_discover_and_emit_for_provider_abort_mode_propagates_error`)
  rewritten to assert the new soft behavior — same scenarios now
  populate `result.errors` with `is_provider_unavailable=True`
  instead of raising. The three CONTINUE-mode tests are
  strengthened to also assert the new flag. `allow_warnings` regex
  updated for the new log message.
- `api/test_list.py`: extends `ProviderErrorInfo` builder test, adds
  one new test for the flag.
- `cli/list_test.py`: new direct tests for `_has_fatal_errors`
  (provider-unavailable / fatal-provider / plain-error / empty),
  plus three new end-to-end CLI tests using the existing
  `FakeApiListAgents` (now accepting `errors=`): a parametrized
  human/json/jsonl exit-0 test, a "fatal error still exits 1" sanity
  test, and a streaming-human exit-0 test. No new
  `monkeypatch.setattr` site (extended the existing helper instead so
  the `PREVENT_MONKEYPATCH_SETATTR` ratchet count doesn't move).

### `changelog/mngr-tkt-mngr-list-exits-1-when-one-pro-mlf5.md`

One-paragraph user-facing summary.

## Verification

Ran (per CLAUDE.md):

```
cd libs/mngr && PYTEST_MAX_DURATION_SECONDS=540 \
  uv run pytest -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release' -q
```

Result: **3859 passed, 1 skipped, 400 deselected**. The 2 remaining
failures
(`issue_reporting_test.py::test_handle_not_implemented_error_interactive_opens_*`)
reproduce on a clean `main` checkout (verified with `git stash`) so they
are pre-existing and unrelated. Coverage: **81.70%** (≥80% gate
passes). Ratchets (`utils/test_ratchets.py`) green.

End-to-end smoke: `uv run mngr list` in this repo still exits 0 with
the existing WARNINGs for Docker / Vultr.

## Open question for Josh

If you prefer the warning live on a dedicated `result.warnings` field
(closer to #1538's structured-warnings channel) rather than a flag on
`ProviderErrorInfo`, I'm happy to swap that — see "Why this shape"
above. I picked the flag because it ships the fix in the smallest
diff and is forward-compatible with either #1538 or #1556 landing
later.


joshalbrecht:
[AGENT] Driver review

**(A) Likely to approve:** uncertain — the fix is correctly aimed at the ticket (mirror destroy's degrade-on-provider-failure shape), the diff is small and focused, but it sits squarely in territory where you have **four active in-flight PRs** (#1556, #1538, #1537, #1532) reworking the same exit-code / warning surface. The worker explicitly chose the smallest delta over the WZ-authored designs; you may want the typed-hierarchy or structured-warnings channel to land first instead.

**(B) Potential issues:**
- **ABORT-mode semantics changed silently**: per-provider construction/discovery failures used to propagate as `MngrError` under `error_behavior=ABORT`; they now degrade to warnings regardless of `error_behavior`. Two existing tests were rewritten from "raises MngrError" to "records ProviderErrorInfo without raising". If any internal caller (or downstream client) was relying on ABORT actually aborting for provider auth failures, this is a quiet behavior break. Consider whether ABORT should keep its old semantics here and only the CLI (which is always CONTINUE in practice) should change.
- **Scope of the new "provider unavailable" bucket is broad**: catches Modal auth, Docker daemon, Vultr API key, bad-backend-in-config, network — anything thrown out of `get_provider_instance` or `discover_hosts_and_agents`. PR #1556 distinguishes `NotAuthorized` vs `Discovery` vs `Unavailable`. The broad catch may mask a genuine bug (e.g. a real exception from inside the provider's discovery code that isn't "unavailable") as a soft warning.
- **Open question in the PR body**: worker asks whether you'd prefer the warning live on a dedicated `result.warnings` field (closer to #1538) instead of a flag on `ProviderErrorInfo`. Worth answering before merge.
- 2 pre-existing failures in `cli/issue_reporting_test.py` (`test_handle_not_implemented_error_interactive_opens_*`) — worker verified via `git stash` that they reproduce on clean `main`. Unrelated to this PR but worth noting they're sitting on `main`.

- Whether the destroy-mirror is the design you want here, or whether you'd rather wait on (or rebase onto) #1556 / #1538.
- The ABORT-mode behavior change in particular — read `test_list_agents_abort_mode_records_provider_unavailable_without_raising` and `test_construct_discover_and_emit_for_provider_abort_mode_records_provider_unavailable` and confirm you're OK with that.
- The breadth of `is_provider_unavailable=True` — verify a real provider bug (not an availability failure) doesn't slip into the warning path.

**Recommendation:** merge (with eyes on the ABORT-mode change and the design question)

PR: https://github.com/imbue-ai/mngr/pull/1607

weishi-imbue:
> mngr list exits 1 if any single provider fails 

I had an impression that this is the desired behavior. Am I wrong?

joshalbrecht:
you are correct, this was an auto created (wrong) PR from a random agent I was testing :)

PR #1612, opened 2026-05-13
docs: fix stale CLI references in README and customization.md

joshalbrecht:
[AGENT] Two small, defensible doc fixes uncovered by reading the user-facing markdown against the current CLI.

## Summary

- **`README.md`** (and its auto-generated mirror `libs/mngr/README.md`): the snapshot-and-fork example used `mngr snapshot create ... --format "{id}"`, but `snapshot create --format` only knows the fields `snapshot_id, host_id, provider, agent_names` (see `_emit_create_result` in `libs/mngr/imbue/mngr/cli/snapshot.py` and the `--help` text). `{id}` is a field on `snapshot list`, not `create` — so the example would have raised a `KeyError` when copy-pasted. Fixed to `{snapshot_id}`. `libs/mngr/README.md` was regenerated via `scripts/make_cli_docs.py`.

- **`libs/mngr/docs/customization.md`** ("Work Directory Extra Paths"): referenced `--worktree` and `--copy-source` as the flags users would pass to `mngr create`. Those flags no longer exist — they were consolidated into the single `--transfer` flag in commits `bde5c3646` and `36436099d`, with values `none|rsync|git-mirror|git-worktree`. Rewrote the lead-in to describe what actually happens today (a new work_dir built via the default `--transfer=git-worktree` for git or `--transfer=rsync` for non-git projects), without inventing flag names that don't exist.

Each commit body explains what the doc said before, what was wrong, and what it now says.

## Test plan

- [x] `uv run mngr create --help` confirms `--transfer` is the current flag and `--worktree`/`--copy-source` are not.
- [x] `uv run mngr snapshot create --help` confirms `--format` fields are `snapshot_id, host_id, provider, agent_names`.
- [x] `scripts/make_cli_docs.py` regenerates `libs/mngr/README.md` deterministically (committed regenerated copy).
- [x] No code touched; pure docs change.


joshalbrecht:
[AGENT] Driver review

**(A) Likely to approve:** yes — two small, surgical doc fixes, each with a citable source-of-truth, body explains what was wrong before and after.

**(B) Potential issues:**
- The customization.md rewrite says "the default `--transfer=git-worktree` for git repos, or `--transfer=rsync` for non-git projects." The `git-worktree` half is verbatim what `mngr create --help` says (default when source and target are on the same host). The `rsync for non-git` half is slightly looser than the code: in `libs/mngr/imbue/mngr/cli/create.py:1381` `TransferMode.NONE` is the first branch when transfer is disabled, and rsync is the implied fallback for non-git. The rewrite is still strictly more accurate than the old `--worktree` / `--copy-source` text, but you may want to tighten this clause to "or `--transfer=rsync` for projects without a git repo" or similar.
- Worker did not run the test suite. These are pure docs changes (no `.py` edits), so risk is bounded to a markdown lint; the worker also regenerated `libs/mngr/README.md` via the documented `scripts/make_cli_docs.py` rather than hand-editing, which is the supported workflow.

- The customization.md phrasing for the `--transfer` defaults (see above) — confirm the wording matches how you want users to reason about transfer modes.
- The two commits are split clean per fix; commit messages cite the source-of-truth files. No bundling.

**Recommendation:** merge

PR #1616, opened 2026-05-13
Key Latchkey state per-host instead of per-agent

hynek-urban:
Pass the Latchkey-related env vars (LATCHKEY_GATEWAY, _PASSWORD, _PERMISSIONS_OVERRIDE, _DISABLE_COUNTING) to 'mngr create' via --host-env instead of --env so every agent that ever runs on the new host shares the same gateway URL, password, JWT, and permissions file.

On-disk metadata moves accordingly: minds now stores the per-agent latchkey_permissions.json under <latchkey-dir>/mngr_latchkey/hosts/<host_id>/ instead of /agents/<agent_id>/. After 'mngr create' returns, minds reads the canonical host_id from the JSONL 'created' event and points the opaque permissions handle (referenced by the JWT minted at create time) at the new host-keyed path.

Renames in imbue-mngr-latchkey:

- agent_setup.finalize_agent_permissions -> finalize_host_permissions (takes a HostId)
- store.permissions_path_for_agent -> permissions_path_for_host
- store.link_opaque_permissions_to_agent -> link_opaque_permissions_to_host
- 'mngr latchkey link-permissions' takes --host-id instead of --agent-id

The minds UI's grant flow resolves the request event's agent_id to its host_id via the backend resolver before writing the grant; if the resolver hasn't seen the agent yet, the grant POST returns 503 so the UI can retry instead of silently writing the grant to the wrong file.


hynek-urban:
This PR replaces https://github.com/imbue-ai/mngr/pull/1576 which was already approved. (I basically addressed the feedback by creating a new PR from scratch.)

joshalbrecht:
lgtm, thanks hynek!

PR #1618, opened 2026-05-13
mngr-latchkey + minds: delegate permission management to Latchkey gateway extensions

hynek-urban:
(Accompanies https://github.com/imbue-ai/forever-claude-template/pull/80.)

- Add Latchkey gateway extensions implementing permission management via the /permissions and /permission-requests endpoints.
- Add the mngr latchkey admin-jwt command to get a "root" token that is allowed any operations on the gateway (including modifying permissions) - used by the minds app or by CLI consumers.
- Add the mngr latchkey gateway-info command to allow CLI consumers to discover running Latchkey gateway.
- Update the minds app to consume and update permission requests and permissions via the mngr_latchkey plugin (using the gateway extensions under the hood.)
- Restart the mngr latchkey forward process every time the minds app starts for easier upgrades.
- Give the agents the permission to create permission requests by default.


apps/minds/imbue/minds/cli/run.py
```diff
@@ -154,11 +169,35 @@ def run(
...
+    # _wait_for_gateway_port returns only after the supervisor has
+    # stamped a non-None gateway_port; re-assert here so the type
+    # checker can narrow.
+    assert forward_info.gateway_port is not None
+    gateway_client = LatchkeyGatewayClient(
+        base_url=f"http://{latchkey.listen_host}:{forward_info.gateway_port}",
+        password=gateway_password,
+        admin_jwt=admin_jwt,
+    )
+
```
joshalbrecht, line 194: hmm--I might want us to adjust this in a follow up PR

we want the app to start really quickly, and thus this probably should be done in parallel with other stuff (like mngr forward, below)

additionally, if it fails, permissions are broken, but the whole rest of the app can work. It *does* seem preferrable to be in a degraded state, rather than completely broken.

we have a number of different degraded states, and probably want to think about how to represent them, how to allow users to recover from them, etc

for now this is fine I guess, just flagging that it might be something we will want to come back to

joshalbrecht:
other than my one comment, lgtm!

(and we can do that comment in a follow up someday)

hynek-urban:
@joshalbrecht Makes sense, I'll note it down as another piece of work. Thanks!

PR #1619, opened 2026-05-13
tmr: redesign agent output transfer + clean up orchestration

qi-imbue:
See human summary below: https://github.com/imbue-ai/mngr/pull/1619#pullrequestreview-4284123404

## Summary

Two threads of change to ``mngr tmr``: a new output-transfer scheme that fixes brittleness around finalizing agents, and a decoupling of orchestration from result-data shape so the reporter and the polling loop can evolve independently.

### Output transfer

Testing agents now publish a single ``outputs.tar.gz`` into their state directory (``\$MNGR_AGENT_STATE_DIR/plugin/test-map-reduce/``), containing the renamed ``test_output/`` directory and an optional incremental ``branch.bundle``. The orchestrator polls for that archive via the per-agent volume API (works even when the host is offline), downloads it, extracts, and reconstructs the agent's branch from the bundle -- replacing the previous rsync + git-pull steps.

Reintegrate mode uses the same path. SSH provider, which does not expose a volume, is no longer supported for testing-agent outputs. The integrator agent's transfer mechanism is unchanged (still rsync), but its outcome is read from disk by the reporter rather than handed across in memory.

### Polling

Polling is now driven entirely by the existence of ``outputs.tar.gz`` on the agent's volume. ``mngr list`` is no longer called from the main loop. This fixes a race where an agent reaching a terminal lifecycle state before its archive was visible on the volume would be marked failed forever. ``--poll-interval`` default raised to 60s; ``--result-check-interval`` is gone (the volume check is the only signal now).

### Orchestration / reporter decoupling

The outcome JSON schema is a contract between agents and ``report.py`` only. Orchestration treats extracted contents as opaque blobs; it hands a list of ``AgentMetadata`` (kind, agent_name, test_node_id, branch_name, error_summary) to the reporter, which reads each agent's outcome from disk on demand and caches per-agent.

This collapses several pieces of orchestration code (``_collect_agent_results``, ``gather_results``, ``_build_current_results``, the ``cached_results`` dict) and makes the reporter the only place that needs to know about ``TestResult`` / ``IntegratorResult`` shape. Future pluggable reporting can swap ``generate_html_report`` without touching the polling loop.

### CLI flag changes

- ``--max-parallel`` → ``--max-parallel-launch`` (default 4 → 10).
- ``--max-agents`` → ``--max-parallel-agents``.
- ``--output-html PATH`` → ``--output-dir DIR``; the report is always written to ``\<dir\>/index.html``.
- ``--result-check-interval`` removed.
- ``--poll-interval`` default 10s → 60s.

### Other user-visible changes

- ``mngr list --format json`` no longer emits the redundant ``address`` field on agent/host records. Same value is still reachable on the parsed Python objects; removing it from the wire format lets the output round-trip through ``model_validate_json``.
- ``Volume`` gains a ``path_exists(path)`` method, implemented for local, Docker, Modal, and the ``ScopedVolume`` wrapper.
- New CI step in ``.github/workflows/tmr.yml`` pre-trusts the repo checkout for Claude (so the integrator agent's local-host trust-dialog flow doesn't hang in CI). CI now also passes ``--use-snapshot`` to the orchestrator.

### Internals worth mentioning

- ``api.py`` renamed to ``orchestration.py``; the rest of the package now imports from ``mngr_cli``, ``pulling``, ``utils``, ``launching``, ``orchestration`` directly instead of through a re-export hub.
- Module-private helpers prefixed with ``_`` across the package.
- ``AgentDetails.address`` and ``HostDetails.address`` are now plain ``@cached_property`` (not ``@computed_field``) so JSON round-trips cleanly.
- ``base_commit`` threaded through ``TmrLaunchConfig`` into ``AgentGitOptions.base_branch`` so ``\$MNGR_GIT_BASE_BRANCH`` is populated for agents (was empty, which broke the bundle command).

## Test plan

- [ ] CI: ``just test-offload`` -- unit + integration suites under the new shapes.
- [ ] CI: the dispatch ``TMR`` workflow with ``--use-snapshot`` and the pre-trust step -- should produce a valid HTML report and integrator branch end to end.
- [ ] Local: ``mngr tmr`` against the ``local`` provider on a small test set. Verify:
  - Per-agent ``output_dir/\<agent\>/test_output/testing_agent_outcome.json`` lands on disk.
  - Per-agent ``output_dir/\<agent\>/branch.bundle`` is created when the agent committed.
  - Each agent's branch ends up in the local repo via ``git fetch \<bundle\>``.
- [ ] Local: ``mngr tmr`` against ``--provider modal`` with at least two agents. Verify:
  - ``output_dir/index.html`` is regenerated during polling, not just at the end.
  - Once each agent publishes the archive, it shows up in the report (no permanent "still running" rows).
- [ ] Local: ``mngr tmr --reintegrate \<run-name\>`` against a previous run that produced fix branches. Verify:
  - Agent outputs are re-extracted from the volume.
  - Integrator runs and produces a branch.
  - Report renders with the integrator's squashed/impl/failed columns populated.
- [ ] Manual: a deliberately-killed agent (e.g., stop a testing agent before it writes the archive). Should time out per ``--timeout`` and render as "Agent was stopped because the timeout was reached." in the report.

### DIFF
--- a/.github/workflows/tmr.yml
+++ b/.github/workflows/tmr.yml
@@ -64,6 +64,32 @@ jobs:
           git config user.email "dev@imbue.com"
           git config user.name "imbue-dev"
 
+      - name: Pre-trust checkout for Claude
+        # The integrator agent runs on the local host, where mngr_claude's
+        # provisioning path goes through Claude Code's trust dialog (see
+        # mngr_claude/plugin.py:interactively_dismiss_claude_dialogs). In CI
+        # there is no human to accept the dialog and no prior ~/.claude.json
+        # entry to inherit, so without seeding trust the integrator either
+        # raises ClaudeDirectoryNotTrustedError or hangs waiting on the
+        # dialog. Pre-seed the global ~/.claude.json with trust for the
+        # checkout plus the other dialogs the plugin gates on; the per-agent
+        # integrator config then inherits the trust via the plugin's
+        # copy_project_config_from path.
+        run: |
+          uv run --project libs/mngr_claude python - <<'EOF'
+          from pathlib import Path
+          from imbue.mngr_claude.claude_config import (
+              add_claude_trust_for_path,
+              complete_onboarding,
+              dismiss_effort_callout,
+              find_user_claude_config,
+          )
+          config_path = find_user_claude_config()
+          add_claude_trust_for_path(config_path, Path.cwd().resolve())
+          dismiss_effort_callout(config_path)
+          complete_onboarding(config_path)
+          EOF
+
       - name: Run TMR orchestrator
         env:
           ADDITIONAL_AUTHORIZED_HOSTS: ${{ inputs.additional_authorized_hosts }}
@@ -89,7 +115,7 @@ jobs:
             --env "MODAL_TOKEN_SECRET=${MODAL_TOKEN_SECRET}" \
             --env "ANTHROPIC_API_KEY=${ANTHROPIC_API_KEY}" \
             "${authorized_host_args[@]}" \
-            --output-html tmr-report/index.html \
+            --output-dir tmr-report \
             --format jsonl \
             ${{ inputs.test_paths }} \
             -- ${{ inputs.pytest_args }} \
--- a/.github/workflows/tmr.yml
+++ b/.github/workflows/tmr.yml
@@ -111,6 +111,7 @@ jobs:
             -S 'agent_types.yolo.cli_args=["--dangerously-skip-permissions"]' \
             --agent-type "${{ inputs.agent_type }}" \
             --provider modal \
+            --use-snapshot \
             --env "MODAL_TOKEN_ID=${MODAL_TOKEN_ID}" \
             --env "MODAL_TOKEN_SECRET=${MODAL_TOKEN_SECRET}" \
             --env "ANTHROPIC_API_KEY=${ANTHROPIC_API_KEY}" \
--- a/libs/mngr/imbue/mngr/interfaces/data_types.py
+++ b/libs/mngr/imbue/mngr/interfaces/data_types.py
@@ -470,7 +470,6 @@ class HostDetails(FrozenModel):
         description="Reason for failure if the host failed during creation",
     )
 
-    @computed_field
     @cached_property
     def address(self) -> HostAddress:
         return HostAddress(host=HostName(self.name), provider=self.provider_name)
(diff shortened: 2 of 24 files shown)


.github/workflows/tmr.yml
```diff
@@ -64,6 +64,32 @@ jobs:
           git config user.email "dev@imbue.com"
           git config user.name "imbue-dev"
 
+      - name: Pre-trust checkout for Claude
```
qi-imbue, line 67: Disgusting but necessary. Not sure how to do this better

libs/mngr/imbue/mngr/interfaces/data_types.py
```diff
@@ -470,7 +470,6 @@ class HostDetails(FrozenModel):
         description="Reason for failure if the host failed during creation",
     )
 
-    @computed_field
```
qi-imbue, line 473: This is strictly no longer necessary for this PR, but I left this in as it's useful.

The `@computed_field` decorator was making the `address` field show up in the serialization, but deserialization doesn't like this. Removing this makes this model round-trip successfully again, which is a good property to have.

libs/mngr/imbue/mngr/interfaces/data_types.py
```diff
@@ -522,7 +521,6 @@ class AgentDetails(FrozenModel):
 
     plugin: dict[str, Any] = Field(default_factory=dict, description="Plugin-specific fields")
 
-    @computed_field
```
qi-imbue, line 525: Same

libs/mngr/imbue/mngr/interfaces/volume.py
```diff
@@ -30,6 +30,11 @@ def listdir(self, path: str) -> list[VolumeFile]:
         """List file entries in the given directory path on the volume."""
         ...
 
+    @abstractmethod
```
qi-imbue, line 33: Added this because TMR needs this, and it seems useful and easy to implement in general.

qi-imbue (commented):
Sorry for this large and messy change.

I've left comments on all the changes outside `mngr_tmr`.

I've babysat most of the changes in `mngr_tmr`, and am happy about it. The changes there mix two large changes with a bunch of small refactors:

- Simplify how testing agents deliver their output: the ultimate outcome is that each testing agent produces a single `.tar.gz` containing everything, and it puts it on a host volume that's accessible when the host is offline.
- Decouple the reporter from the rest of the TMR. Among [things that TMR hardcodes about mngr](https://imbue-ai.slack.com/archives/C0AA6LFV5S5/p1778588135247479) this is the most tightly coupled one. I wanted to do this in a followup PR, but the output delivery change broke the reporter, so I did this in this PR. We should now be ready to fully decouple TMR from the mngr codebase and make it truly generic.

joshalbrecht:
lgtm!

PR #1622, opened 2026-05-13
mngr_lima: use portable `tail -F` for the serial-log tailer

weishi-imbue:
### GNU tail
source: https://www.gnu.org/software/coreutils/manual/html_node/tail-invocation.html
image

### MacOS tail
source: man tail on mac. 
image

## What problem is tail solving here?
When Lima starts a new VM (the call that mngr_lima makes on a fresh agent), the only window into "what's happening inside the booting VM" before SSH is reachable is the serial console log: 
Path looks like: `~/.lima/<vm-name>/serial.log`

we need to:
1. follow the file as it grows
2. survive the file being recreated. 

-F gives us this behavior on both platforms.


joshalbrecht:
lgtm (assuming tests pass)

PR #1626, opened 2026-05-14
mngr_uncapped_claude: drop-in claude -p replacement plugin

joshalbrecht:
## Summary

Adds `imbue-mngr-uncapped-claude`, a new mngr plugin exposing the `mngr uncapped-claude` top-level command. It acts as a drop-in replacement for `claude -p` whose execution is routed through a fresh, ephemeral mngr claude agent instead of running `claude` directly.

- Every `claude` flag is forwarded verbatim to the spawned agent except the wrapper-consumed ones (`-p/--print`, `--input-format`, `--output-format`, `--replay-user-messages`).
- `--fallback-model`, `--max-budget-usd`, `--no-session-persistence`, `--include-hook-events`, `--include-partial-messages`, `-c/--continue`, `-r/--resume`, and `--session-id` are rejected with exit code 2 in v1.
- Input formats: `text` (positional argv or piped stdin) and `stream-json` (NDJSON `{"type":"user","message":{"role":"user","content":"..."}}` lines).
- Output formats: `text` (final assistant text), `json` (synthesized claude-native `result` envelope), and `stream-json` (live transcript replay).
- The spawned agent runs in-place at the current cwd with `auto_dismiss_dialogs=True` / `auto_allow_permissions=True` so dialogs and permission prompts never block. SIGINT/SIGTERM destroy the agent before the signal is re-raised.
- Exit codes: `0` success, `1` claude/api error, `2` mngr-side failure.

The full design lives in `specs/uncapped-claude/spec.md`.

## Test plan
- [x] Unit tests for arg partitioning, input modes, output mode envelopes, and orchestrator helpers (111 tests passing locally via `just test-quick libs/mngr_uncapped_claude`).
- [x] Ratchet tests pass for the new package (after a documented misfire bump for the wire-required `"num_turns"` JSON key).
- [x] `uv run mngr uncapped-claude --help` renders correctly; rejected-flag and missing-prompt paths exit with code 2.
- [ ] CI exercise of a real `claude -p` round-trip (gated on ANTHROPIC credentials in the offload sandboxes).


joshalbrecht:
lgtm

will need a different name eventually, but it works and fixes some other tricky bugs with the status line stuff, so, worth merging

PR #1628, opened 2026-05-14
fix `tmux send-keys -l` bug when the message starts with `-`

evgunter:
## Summary
- `tmux send-keys -l` still treats a leading `-` token as a flag, so any agent command starting with `-` (e.g. `--model gemma`) or message starting with `-` (e.g. `--help`) errored out with `invalid flag --`.
- Insert the `--` end-of-options separator at the three `send-keys -l` call sites (agent command, additional-window commands, agent message send).
- Added regression tests in `host_test.py` and `base_agent_test.py`.

## Test plan
- [x] `just test-quick libs/mngr/imbue/mngr/hosts/host_test.py libs/mngr/imbue/mngr/agents/base_agent_test.py` (246 passed)
- [ ] CI offload (`just test-offload`) green
- [x] Verified manually: `tmux send-keys -l -- '--model gemma'` types the literal text into the pane; without `--` tmux errors with `invalid flag --`.


libs/mngr/imbue/mngr/hosts/host.py
```diff
@@ -2211,7 +2211,7 @@ def rename_agent(
             new_session_name = f"{self.mngr_ctx.config.prefix}{new_name}"
             result = self.execute_idempotent_command(
                 f"tmux has-session -t {shlex.quote('=' + old_session_name)} 2>/dev/null && "
-                f"tmux rename-session -t {shlex.quote('=' + old_session_name)} {shlex.quote(new_session_name)} || true"
+                f"tmux rename-session -t {shlex.quote('=' + old_session_name)} -- {shlex.quote(new_session_name)} || true"
```
joshalbrecht, line 2214: is this correct? should it really be there for a rename? that seems  a little suspicious

joshalbrecht:
lgtm except for my one comment

libs/mngr/imbue/mngr/hosts/host.py
evgunter: experimentally confirmed it is correct

PR #1629, opened 2026-05-14
reject unknown agent type instead of falling back to BaseAgent

evgunter:
i kinda thought this behavior had been removed when i added the `command` agent type, but apparently not

also does some cleanup for how we check the agent registry

---

## Summary

- Fix silent fallback: `mngr create --type X -- ...` no longer treats an unregistered `X` as a generic `BaseAgent` (effectively turning `--type X --` into a hidden alias for `--type command --`). `resolve_agent_type` now requires `X` to be known via a registered class, registered config, or a `[agent_types.X]` block in user config; unknown names raise `UnknownAgentTypeError` (with the existing plugin install hint).
- Remove `set_default_agent_class` / `_default_agent_class`. The only caller that depended on the fallback was the built-in `command` plugin, which now registers `BaseAgent` explicitly. `check_agent_type_known` and `is_streaming_headless_agent_type` were also relying on the fallback (or on `is_agent_class_registered` directly) and got buggy answers for TOML custom types and the `command` type; they now go through the new `is_known_agent_type` / `resolve_agent_type`.
- Add canonical `is_known_agent_type(name, config)` so the gate, lifecycle-state check, and any future caller share one predicate.

## Test plan

- [x] `just test-quick libs/mngr` — 3856 passed locally
- [x] `just test-quick libs/mngr_claude libs/mngr_imbue_cloud libs/mngr_forward libs/mngr_opencode libs/mngr_pi_coding libs/mngr_claude_subagent_proxy` — 1047 passed locally
- [ ] CI offload (unit + integration + acceptance)


libs/mngr/imbue/mngr/config/agent_class_registry.py
```diff
@@ -21,28 +20,19 @@ def register_agent_class(
     _agent_class_registry[AgentTypeName(agent_type)] = agent_class
 
 
-def set_default_agent_class(agent_class: type) -> None:
-    """Set the default agent class returned when a type is not registered."""
-    global _default_agent_class
-    _default_agent_class = agent_class
-
-
```
evgunter, line 29: huh...

libs/mngr/imbue/mngr/config/agent_class_registry.py
```diff
@@ -21,28 +26,19 @@ def register_agent_class(
     _agent_class_registry[AgentTypeName(agent_type)] = agent_class
 
 
-def set_default_agent_class(agent_class: type) -> None:
```
evgunter, line 24: renamed to set_orphan_agent_class

joshalbrecht:
lgtm, thanks for noticing and fixing!

PR #1639, opened 2026-05-14
support transcript for gemini 

evgunter:
## Summary

- **`mngr transcript <gemini-agent>` now works just like it does for Claude.** A background converter (`libs/mngr_gemini/imbue/mngr_gemini/resources/common_transcript.sh`) reads the raw gemini transcript stream and emits the agent-agnostic format at `events/gemini/common_transcript/events.jsonl`.
- **New raw-transcript layer for gemini.** `libs/mngr_gemini/imbue/mngr_gemini/resources/stream_transcript.sh` tails gemini's session JSONL files (filtered by `.project_root == $MNGR_AGENT_WORK_DIR`) into `$MNGR_AGENT_STATE_DIR/logs/gemini_transcript/events.jsonl` **verbatim**. This preserves every field gemini emits (model metadata, internal blocks, etc.) and lives inside the agent state dir, so the bytes survive cleanup of gemini's `~/.gemini/tmp/` working directories. The common-transcript converter then reads from the raw stream rather than scanning `~/.gemini` directly, so the converter no longer depends on gemini's on-disk layout.
- **Two-tier mixin in `libs/mngr/imbue/mngr/interfaces/agent.py`:**
  - **`HasTranscriptMixin`** (raw layer, always provisioned) — requires `get_raw_transcript_scripts() -> Mapping[str, str]`. Raw capture is not user-gated because the bytes are the source of truth.
  - **`HasCommonTranscriptMixin`** (extends `HasTranscriptMixin`) — adds `is_common_transcript_enabled: bool` and `get_common_transcript_scripts()` for the (user-gated) common-schema converter.

  Provisioning helpers in `libs/mngr/imbue/mngr/agents/common_transcript.py`:
  - `provision_raw_transcript_scripts(agent, host, state_dir, cg)` — unconditional upload.
  - `maybe_provision_common_transcript_scripts(agent, host, state_dir, cg)` — checks `is_common_transcript_enabled` first.

  Both agents use the same helpers, so the raw/common split is consistent across Claude and gemini.
- **`mngr transcript <agent>` errors with a clear message** when the agent type's class doesn't implement the mixin (e.g. `Agent 'foo' has type 'codex', which does not produce a common transcript.`), rather than falling through to the misleading "no transcript yet" path. The precheck uses the public `find_one_agent` helper (the previously-public `filter_one_agent` became private on main during the merge).
- **`ClaudeAgent` refactored to satisfy both mixins.** `get_raw_transcript_scripts()` returns `stream_transcript.sh`; `get_common_transcript_scripts()` returns just the one gated converter. Claude's always-on script list keeps the background orchestrator and unrelated helpers (`claude_background_tasks.sh`, `wait_for_stop_hook.sh`, `sync_keychain_credentials.py`) but `stream_transcript.sh` itself now flows through the new mixin path.
- **`GeminiAgent` satisfies both mixins.** `provision()` always installs the raw streamer and conditionally installs the common converter; `assemble_command` always prepends the raw streamer subshell and prepends the common watcher subshell only when enabled.
- **`MNGR_EMIT_COMMON_TRANSCRIPT` env var removed.** Pre-refactor, the env var was the gate the Claude wrapper script consumed at runtime: `[ "${MNGR_EMIT_COMMON_TRANSCRIPT:-}" = "1" ] && [ -x "$COMMON_TRANSCRIPT_SCRIPT" ]`. With the new provisioning-time gate, the converter script is simply not on disk when emit is off, so the `-x` half of that compound check does all the work on its own. Removing the env var leaves a single mechanism (on-disk presence, controlled by `maybe_provision_common_transcript_scripts`), shared by both agents.
- `stub_mngr_log_sh` test fixture moved from `mngr_claude/conftest.py` to `mngr/conftest.py` so both packages can reuse it.
- New `GeminiAgentConfig.emit_common_transcript: bool = True` (set to `false` to opt out of the common converter); the matching field on `ClaudeAgentConfig` was already there. Raw capture is always on for gemini and Claude.
- `mngr_gemini` coverage gate bumped from 90 → 95 (currently at 98%).

## Behavior changes worth noting

- For Claude, when `emit_common_transcript=False`, `common_transcript.sh` is no longer written to `commands/` at all (previously it was written and then ignored at runtime via the env var). Toggling False → True after provisioning now requires re-provisioning to materialize the file.
- For gemini, raw capture runs even when `emit_common_transcript=False`. The raw bytes go to `$MNGR_AGENT_STATE_DIR/logs/gemini_transcript/events.jsonl` regardless of the flag, because raw capture is the source of truth and not user-gated.
- The `mngr transcript` command rejects agent types without the mixin up front instead of running through to a misleading "no transcript yet" error.

## Test plan

- [x] `just test-quick libs/mngr_gemini` — 151 tests pass (includes new `stream_transcript_test.py` for the raw streamer: discovery, `.project_root` filtering, incremental emission, id-based offset reconciliation).
- [x] `just test-quick libs/mngr_claude` — 483 tests pass; no regressions from the refactor.
- [x] `just test-quick libs/mngr` — 3900 tests pass.
- [x] `just test-quick libs/mngr/imbue/mngr/cli/transcript_test.py` — precheck assertions for the no-mixin error path.
- [x] Ratchets pass for mngr, mngr_claude, mngr_gemini.
- [x] mngr_gemini coverage at 98% (gate 95%).
- [ ] CI: full `just test-offload` and acceptance suites.


libs/mngr/imbue/mngr/cli/transcript.py
```diff
@@ -36,6 +43,40 @@ class TranscriptCliOptions(CommonCliOptions):
...
+        if issubclass(get_agent_class(type_name), HasCommonTranscriptMixin):
+            supporting.append(type_name)
+    return supporting
+
+
+def _assert_agent_type_supports_transcripts(address: AgentOrHostAddress, mngr_ctx: MngrContext) -> None:
+    """Raise UserInputError if the targeted agent's type does not implement HasCommonTranscriptMixin.
+
+    No-op for HostAddress targets; only agent-scoped lookups are validated here.
+    """
```
evgunter, line 48: it is kind of strange the way this is non-uniform between hosts and agents, even though in both individual cases it does the sensible thing

joshalbrecht:
I could have missed it, but I don't see a raw transcript for gemini?

That's honestly almost more important than the common one...

evgunter:
lol i thought that would happen automatically. added, see https://github.com/imbue-ai/mngr/pull/1639/changes/175ef1e30fb349584abffb7ca67211b9b1584fb5..c74bb0aec45dc09a167ff69c2975fef77e77f71b

joshalbrecht:
lgtm

PR #1640, opened 2026-05-14
ensure changelog runs before release

evgunter:
if there are unconsolidated changelog entries, stop and let the user kick off the changelog job first

this relates to a broader question: should we keep the changelog running every night, or only have it run at release?
- if it runs every night, they'll duplicate work if the previous changelog wasn't merged (e.g. https://github.com/imbue-ai/mngr/pull/1662), though i can fix this by making them branch off an existing open changelog PR instead of main
- if it runs for every release then the entries might be worse since there might be a lot of things to consolidate at once

---

## Summary

- Gate `scripts/release.py` on unconsolidated `changelog/*.md` entries. When the gate fires, prompt the user to invoke the existing `changelog-consolidation` Modal schedule on demand; the agent opens a PR the user merges before re-running `release.py`.
- New `scripts/trigger_changelog_consolidation.py` owns the gate, the `mngr schedule run` invocation, and the `--disable-plugin` computation; `setup_changelog_agent.sh` now sources the latter from there so deploy and on-demand trigger stay in sync.
- Tests at `scripts/trigger_changelog_consolidation_test.py` cover entry detection and gate branching (no monkeypatching — fns are injected).

## Test plan

- [x] `just test-quick scripts/trigger_changelog_consolidation_test.py` (10/10 pass)
- [x] `uv run python scripts/trigger_changelog_consolidation.py --print-disable-plugin-args` returns paired flags
- [ ] CI green (offload runs the full suite on PR)


joshalbrecht:
lgtm, but would like to discuss exactly how we want to do changelogs, moving discussion to slack

PR #1646, opened 2026-05-14
gemini misc settings / config / hooks stuff

evgunter:
## Summary

Foundational settings/hook plumbing for `mngr_gemini`, plus the readiness-hook integration the rest of the plugin needs to detect agent state without polling the TUI. After this PR, an mngr-managed Gemini agent:

- starts without consuming a keystroke on the trust dialog (env-var trust instead of `--skip-trust`),
- exposes a sentinel file so `mngr` knows when the TUI is ready,
- optionally auto-approves every tool call without dropping into yolo mode,
- leaves the user's workspace and `~/.gemini/` completely untouched.

The user's project never gains a `.gemini/` directory; everything mngr writes lives in the per-agent state dir.

### `libs/mngr_gemini/imbue/mngr_gemini/gemini_config.py` (new)
Settings helpers: read/write for the three Gemini settings tiers (`/etc/gemini-cli/`, `<project>/.gemini/`, `~/.gemini/`), atomic write with `.bak` backup, malformed-JSON-tolerant reads, env-var interpolation matching Gemini's `$VAR` / `${VAR}` / `${VAR:-default}` syntax. Two hook-config builders (`SessionStart` readiness sentinel, `BeforeTool` wildcard auto-allow) and `merge_hooks_config` / `hook_already_exists` for safe dedup-preserving composition. `GeminiDirectoryNotTrustedError` reserved for future interactive trust flows.

### `libs/mngr_gemini/imbue/mngr_gemini/plugin.py`
Extends `GeminiAgent`:
- `provision()` writes `$MNGR_AGENT_STATE_DIR/plugin/gemini/system_settings.json` (mngr-owned, rewritten each provision) with the `SessionStart` readiness hook. Mirrors `mngr_claude`'s `plugin/claude/anthropic/` namespacing.
- `modify_env_vars()` sets `GEMINI_CLI_SYSTEM_SETTINGS_PATH` so Gemini reads our settings as the system tier (Gemini's documented [enterprise override](https://geminicli.com/docs/cli/enterprise/)), plus `GEMINI_CLI_TRUST_WORKSPACE=true` for headless trust.
- New `auto_allow_permissions: bool = False` config flag (mirrors `mngr_claude`). When `True`, also installs a wildcard `BeforeTool` hook that emits `{"decision":"allow"}`.
- `--skip-trust` dropped from default `cli_args` — smoke-testing confirmed it strips workspace-tier hook registration, which would have silently disabled the readiness hook.

### `specs/mngr-gemini-feature-parity/concise.md` (new)
Multi-PR plan for bringing `mngr_gemini` closer to `mngr_claude`'s feature surface. Marks this PR's contents as LANDED; subsequent PRs (lifecycle hookimpls, session adoption, deploy/host-destroy, headless variant, skill-provisioned subtypes) are scoped with file paths, dependencies, and acceptance criteria so they can be picked up in parallel.

## Why these design choices

- **System-tier settings, not workspace**: writing `<project>/.gemini/settings.json` would pollute the user's repo. The `GEMINI_CLI_SYSTEM_SETTINGS_PATH` env-var override lets us keep mngr-owned state inside the per-agent state dir. Smoke-tested against Gemini CLI 0.42.0.
- **Env-var trust, not `--skip-trust`**: `--skip-trust` only trusts the workspace for tool execution, not hook registration — workspace-declared hooks silently disappear from the registry under it. `GEMINI_CLI_TRUST_WORKSPACE=true` is Gemini's [documented headless-automation path](https://geminicli.com/docs/cli/trusted-folders/) and registers hooks correctly. Confirmed end-to-end against the live CLI.
- **`BeforeTool` hook, not `--approval-mode yolo`**: the hook survives admin policies that disable yolo mode (`security.disableYoloMode`) and is observable in `--debug` output. Mirrors `mngr_claude`'s `PermissionRequest` approach.
- **`merge_hooks_config` for composition, not inline `dict.update`**: see the comment block on `_install_system_settings` for the rationale — short version, the helper appends matcher groups safely while `dict.update` would silently drop them under a shared event key.

## Test plan

- [x] `just test-quick libs/mngr_gemini` — 137 passed, 0 failed.
- [x] End-to-end smoke against real Gemini CLI 0.42.0: settings file generated by the plugin's builders is loaded as the system tier; `Hook registry initialized with N hook entries` matches the configured count; `$MNGR_AGENT_STATE_DIR/session_started` is created during the session; when `auto_allow_permissions=True`, tool calls execute without approval prompts and `BeforeTool` hook execution is observable in `--debug`.
- [x] User's workspace verified untouched after provision (no `.gemini/` directory created in the work_dir).
- [x] JSON shape produced by the builders validated against Gemini's published `settings.schema.json`.
- [ ] CI offload run.


joshalbrecht:
lgtm after tests pass!

evgunter:
(test failure is fixed in https://github.com/imbue-ai/mngr/pull/1652)

PR #1647, opened 2026-05-14
fix config merging behavior that was causing default claude permissions mode to be dropped

evgunter:
hm, apparently this is different from how claude code itself does things (i.e. it also would drop default permissions mode when setting other fields on permissions)

related to https://github.com/imbue-ai/mngr/pull/1650 but doesn't touch the same machinery; could possibly be updated so it does use that scheme (though that doesn't yet handle recursive `__extend` merges cleanly either), but that would mean it no longer matches claude code's native structure

---

## Summary
- Fix: per-agent `settings.json` was wiping `permissions.defaultMode` (e.g. `"auto"`) when an agent type's `settings_overrides` re-specified any key under `permissions` (such as `allow`).
- Root cause: `_build_settings_json` used a shallow `dict.update`, so a re-specified `permissions` dict replaced the base wholesale, losing sibling keys configured in the user's `~/.claude/settings.json`.
- Switch to a recursive deep merge for `settings_overrides` so nested sibling keys not present in the override are preserved.

## Test plan
- [x] New regression test `test_build_settings_json_preserves_nested_sibling_keys` (covers `defaultMode` surviving a `permissions.allow`-only override).
- [x] All 8 `_build_settings_json` tests pass locally.
- [x] All 199 `plugin_test.py` tests pass locally.
- [ ] CI green.


evgunter:
this should be fixed properly by josh's planned settings update-vs-replace systematization

PR #1648, opened 2026-05-15
some install.sh tests

evgunter:
these tests are kind of low-quality but hopefully kind of helpful

---

## Summary

Closes the test-coverage gap raised on `mngr/installer-sets-default-type` ("has this actually been tested on a clean machine?") by adding real-subprocess tests at two layers: the `scripts/install.sh` shell flow, and the `_install_via_brew` / `_install_via_apt` Python install functions. Writing those tests surfaced a real production bug, which is fixed in the same PR.

### Tests added

- **`libs/mngr/imbue/mngr/test_install_script.py`** — 5 integration tests that run `bash scripts/install.sh` against mock `uv` and `mngr` binaries on a synthetic PATH. Covers:
  - `uv tool upgrade` branch when `imbue-mngr` is already installed
  - `uv tool install` branch when it is not
  - PATH-not-set error after a successful install
  - `mngr dependencies -i` failure does not stop `mngr extras -i` (the `|| warn` pattern)
  - `mngr extras -i` failure does not stop the final `Get started` line
  - Warning assertions pinned to the exact `|| warn` message text for each step (so a regression that fires the wrong warning is caught at the right test).
- **`libs/mngr/imbue/mngr/utils/test_deps_install.py`** — 4 integration tests plus 1 release test:
  - `_install_via_brew(["tmux", "jq"])` actually invokes `brew install tmux jq`
  - `_install_via_brew` returns False (not raise) when brew exits non-zero
  - `_install_via_apt` runs `update` then `install` in order, with the right flags
  - `_install_via_apt` stops at first failure (doesn't proceed to install if update fails)
  - `@pytest.mark.release` end-to-end against real brew on macOS using an already-installed formula (no system mutation)

### Production bug fixed

`_install_via_brew` and `_install_via_apt` previously caught `ProcessError` only, but `ConcurrencyGroup.__exit__` wraps that in a `ConcurrencyExceptionGroup` before re-raising. The functions were therefore raising uncaught exception groups out of what was meant to be a True/False-returning helper, and ultimately out of `mngr dependencies -i`. The same bug existed latently in `check_bash_version`. All three call sites now catch via a shared `_SUBPROCESS_ERRORS` tuple.

### Reorganization

- `test_install.py` and `test_upgrade.py` moved from `libs/mngr/imbue/mngr/e2e/` to the package root — neither used the heavy e2e fixture; both used only `minimal_install_env`.
- `MinimalInstallEnv` and the `minimal_install_env` fixture moved from `e2e/conftest.py` to `mngr/conftest.py` to live next to their dependencies.
- New shared `write_executable_script` helper in `mngr/utils/testing.py` for the bash-mock pattern, used by both new test files.

### Branch base correction

This branch was accidentally based on `ev/main` (a personal stacking branch). Cherry-picked onto a fresh base off `main` and force-pushed; PR base retargeted to `main`.

## Known gaps (not in scope)

- The `uv` bootstrap branch of `install.sh` (the `curl https://astral.sh/uv/install.sh | sh` path) is not tested — mocking that pipeline pulls in network-shape concerns this PR isn't trying to take on.
- `_install_via_script` (the curl-pipe-bash code path) does not have an equivalent integration/release test. Worth a follow-up.
- The full chain `mngr dependencies -i` → interactive prompt → `_install_via_brew` is not exercised end-to-end (the CliRunner test environment has no tty, so the prompt short-circuits). Each segment is tested separately, but a regression at the seam could slip through.
- Two substring assertions in `test_install_script.py` are still loose (`"PATH" in stderr`, `"Get started" in stdout`) — worth tightening to specific phrases like the `WARNING`-side assertions already were. Follow-up.

## Test plan

- [x] `just test-quick libs/mngr/imbue/mngr/test_install_script.py` — 5 passed in ~3s
- [x] `just test-quick libs/mngr/imbue/mngr/utils/test_deps_install.py` — 4 passed in ~3s
- [x] `just test-quick "libs/mngr/imbue/mngr/utils/test_deps_install.py -m release"` — release test passes against real brew (~43s)
- [x] `just test-quick libs/mngr` (default marker exclusions) — 3871 passed, 1 skipped (preexisting)
- [ ] CI: full offload suite


libs/mngr/imbue/mngr/utils/test_deps_install.py
```diff
@@ -0,0 +1,159 @@
+"""Integration and release tests for system dependency installation.
+
+The integration tests in this file place mock ``brew`` / ``apt-get`` / ``sudo``
+binaries on PATH and call ``_install_via_brew`` / ``_install_via_apt``
+directly, verifying that they spawn the real subprocess with the exact
+command line we expect. Unit tests in ``deps_test.py`` cover only the
+*planning* layer (``describe_install_commands``, OS-branch selection,
```
evgunter: don't know if these tests are worth keeping

joshalbrecht: these look pretty stupid, I vote we remove

joshalbrecht:
lgtm except above comment + conflicts

PR #1650, opened 2026-05-15
Unified MNGR__* env-var overrides and assign-vs-extend settings semantics

joshalbrecht:
## Summary

- Any `MngrConfig` field can now be overridden from a single shell mechanism: `MNGR__X__Y__Z=value` env vars (double underscores) map 1:1 to the dotted config path `x.y.z`. The old, narrow `MNGR_COMMANDS_<CMD>_<PARAM>` scheme is removed, and plugin / CLI command names can now contain multiple words.
- `__extend` operator suffix on the leaf key opts into additive behavior (concat for lists/tuples, shallow key-merge for dicts, union for sets); the bare key is always assignment. Uniform across TOML, `--setting`, `MNGR__*` env vars, and `mngr config set / extend`.
- **Breaking change to merge semantics:** cross-layer TOML merging and `AgentTypeConfig` parent-type inheritance flip to assign-by-default for every aggregate field. Top-level container dicts (`agent_types`, `providers`, `plugins`, `commands`, `create_templates`) keep their per-key additive merge.
- `mngr config` gains `extend` and `schema` subcommands plus a `list --all` flag for discoverability. `config set KEY__extend VALUE` is accepted as an alias for `config extend`.
- Three legacy env vars are promoted to first-class config fields (`logging.enable_paramiko_logging`, `agent_ready_timeout`, `completion_cache_dir`); `MNGR_RETAIN_LOCK_*` is renamed to `MNGR_DEBUG_RETAIN_LOCK_*` to signal debug-only; `MNGR_ROOT_NAME` / `MNGR_HOST_DIR` / `MNGR_PREFIX` / `MNGR_HEADLESS` remain as documented aliases with conflict detection against their canonical `MNGR__*` form.

Spec: [specs/env-settings-overrides/concise.md](specs/env-settings-overrides/concise.md).
Changelog: [changelog/mngr-env-settings-overrides.md](changelog/mngr-env-settings-overrides.md).

## Test plan

- [x] `just test-quick libs/mngr` (3851 unit tests pass)
- [x] `just test-quick libs/mngr_claude libs/mngr_modal libs/mngr_pair libs/mngr_opencode libs/mngr_gemini libs/mngr_schedule libs/mngr_lima libs/mngr_claude_subagent_proxy libs/mngr_imbue_cloud libs/mngr_forward libs/mngr_tmr` (all plugins green)
- [x] `just test-quick apps` (all minds app tests pass)
- [ ] `just test-offload` — let CI run the full acceptance/release suite
- [ ] Manually exercise: `MNGR__COMMANDS__CREATE__CONNECT=false mngr create`, `mngr config extend agent_types.my.cli_args '["--foo"]'`, `mngr config schema`


libs/mngr/docs/commands/secondary/config.md
```diff
@@ -347,6 +405,53 @@ $ mngr config edit --scope user
 $ mngr config edit --scope local
 ```
 
+## mngr config schema
```
evgunter: this is cool! it would be neat if it supported cel filtering (though its output was less overwhelmingly long than i feared); i might have a claude add that when this lands (assuming you don't want to add it here)

libs/mngr/docs/concepts/environment_variables.md
```diff
@@ -2,79 +2,108 @@
...
-Additionally, the following variable controls where project-level config files are loaded from:
+Each `__`-separated segment after the `MNGR__` prefix maps 1:1 to a step in the dotted config path. Segments are uppercase-only and lowercased before lookup. Values are JSON-parsed first (so `true`, `false`, numbers, arrays, and objects all work) and fall back to the raw string.
 
-- `MNGR_PROJECT_CONFIG_DIR` - Directory containing project-level config files (`settings.toml` and `settings.local.toml`). When set, overrides the default `.{root_name}/` directory at the git root. This only affects where project settings are loaded from; it does not affect `MNGR_HOST_DIR`. Unlike the variables above, this can be changed freely at any time.
+```bash
+# These three are equivalent:
+MNGR__COMMANDS__CREATE__CONNECT=false   mngr create my-agent
+mngr --setting commands.create.connect=false create my-agent
+mngr config set commands.create.connect false
+mngr create my-agent
```
evgunter, line 31: this is kind of pedantic but i'd argue that these shouldn't be called "equivalent" rather than idk saying "these three commands create the same agent" because the last one leaves the config variable set. (or could include an unset line at the end)

libs/mngr/docs/concepts/environment_variables.md
```diff
@@ -2,79 +2,108 @@
...
+Changing `MNGR_ROOT_NAME`, `MNGR_PREFIX`, or `MNGR_HOST_DIR` after a host has been created is not supported.
+
+If both an alias and its canonical `MNGR__*` form are set with **different** values, mngr raises `ConfigParseError` at startup. Same value on both forms is fine.
+
+Additionally, the following non-config env vars are recognized:
 
-**Important**: Command names must be single words (no spaces, hyphens, or underscores). This is required for unambiguous parsing of the environment variable names. Any future plugins that register custom commands must follow this convention.
+- `MNGR_PROJECT_CONFIG_DIR` — directory containing project-level config files. When set, overrides the default `.{root_name}/` directory at the git root. Affects only where project settings are loaded from.
+- `MNGR_ALLOW_UNKNOWN_CONFIG` — when truthy, unknown TOML / env-var / `--setting` keys produce warnings instead of errors.
+- `MNGR_ALLOW_PYTEST` — explicit opt-in for end-to-end tests that intentionally run mngr inside a pytest subprocess against a config with `is_allowed_in_pytest = false`.
```
evgunter, line 92: this is apparently used in only one test and seems not worth the danger of it existing

libs/mngr/imbue/mngr/agents/base_agent.py
```diff
@@ -267,7 +266,10 @@ def get_resume_message(self) -> str | None:
 
     def get_ready_timeout_seconds(self) -> float:
         data = self._read_data()
-        return data.get("ready_timeout_seconds", get_agent_ready_timeout())
+        stored = data.get("ready_timeout_seconds")
+        if stored is None:
+            return self.mngr_ctx.config.agent_ready_timeout
+        return stored
```
evgunter, line 272: huh

libs/mngr/docs/commands/secondary/config.md
evgunter: huh, actually, why is this a separate command from `config list --all` + some flag to indicate that it should also include the type and description? (i know that it operates on the schema instead of data internally, but from a user's perspective they seem very similar.) i guess this one would work even if you've misconfigured things such that `config list` errors?

libs/mngr/imbue/mngr/cli/config.py
```diff
@@ -320,6 +446,37 @@ def _emit_config_value(key: str, value: Any, output_opts: OutputOptions) -> None
             assert_never(unreachable)
 
 
+def _format_extend_sentinel(value: Any) -> str:
+    """Render an extend value with a ``...`` sentinel indicating "extends base".
+
+    Lists become ``[..., item1, item2]`` and dicts become ``{..., k1: v1}``.
+    Other shapes fall back to the bare display form (the resolver would have
+    rejected them, but be defensive).
```
evgunter, line 481: this is nice

libs/mngr/imbue/mngr/config/agent_config_registry.py
```diff
@@ -83,30 +81,23 @@ def _apply_custom_overrides_to_parent_config(
...
         return parent_config
 
-    parent_values = parent_config.model_dump()
     custom_values = custom_config.model_dump()
-    updates: list[tuple[str, Any]] = []
-
-    for field_name in explicitly_set_fields:
-        if field_name in _METADATA_FIELDS:
-            continue
-        elif field_name in AGENT_TYPE_CONCAT_TUPLE_FIELDS:
```
evgunter, line 98: yayyy

libs/mngr/imbue/mngr/config/completion_cache.py
```diff
@@ -15,17 +15,27 @@
...
     """Return the directory used for completion cache files.
 
-    Uses MNGR_COMPLETION_CACHE_DIR if set, otherwise the mngr host directory
-    (MNGR_HOST_DIR or ~/.mngr). The directory is created if it does not exist.
+    Resolution order:
+    1. ``config_value`` (i.e. ``MngrConfig.completion_cache_dir``) when supplied
+       and non-None. Honoring the parsed config field is what makes
+       ``completion_cache_dir = "/foo"`` in ``settings.toml`` take effect.
+    2. ``MNGR__COMPLETION_CACHE_DIR`` env var (lightweight pre-reader path used
+       by tab-completion callers that run before the full config is loaded).
```
evgunter: this surprises me + makes me think that tabcomplete may not work correctly with `MngrConfig.completion_cache_dir`

libs/mngr/imbue/mngr/config/data_types.py
```diff
@@ -73,32 +72,35 @@ def split_cli_args_string(cli_args: str) -> tuple[str, ...]:
...
 
-@pure
-def merge_dict_fields(base: dict[K, V], override: dict[K, V] | None) -> dict[K, V]:
-    """Merge dict fields, with override keys taking precedence."""
-    if override is not None:
-        return {**base, **override}
-    return base
+    Keys present in only one side are preserved. Keys present in both are
+    merged via the value's ``merge_with`` method. This is the explicit
+    carveout from the otherwise assign-by-default rule.
```
evgunter, line 93: can we force the keys to use the __extend convention as well?

libs/mngr/imbue/mngr/config/data_types.py
```diff
@@ -221,15 +211,14 @@ def _normalize_cli_args(cls, value: str | list[str] | tuple[str, ...]) -> tuple[
...
-        Important note: despite the type signatures, any of these fields may be None in the override--this means that they were NOT set in the toml (and thus should be ignored)
-
         Uses model_fields_set to determine which fields were explicitly set in
         the override config, so that subclass-specific fields (e.g., ClaudeAgentConfig's
         auto_dismiss_dialogs) are correctly preserved during merges.
 
-        Scalar fields: override wins if explicitly set
-        Tuple fields (see AGENT_TYPE_CONCAT_TUPLE_FIELDS): concatenate
-        Lists (permissions): concatenate if explicitly set
+        All aggregate fields flip to assign-by-default: a list/tuple/dict/set
```
evgunter, line 218: "flip to"

libs/mngr/imbue/mngr/config/data_types.py
```diff
@@ -564,186 +542,95 @@ class MngrConfig(FrozenModel):
...
-            pager=merged_pager,
-            unset_vars=merged_unset_vars,
-            work_dir_extra_paths=merged_work_dir_extra_paths,
-            enabled_backends=merged_enabled_backends,
+            prefix=_assign_scalar(self.prefix, override.prefix),
+            default_host_dir=_assign_scalar(self.default_host_dir, override.default_host_dir),
+            pager=_assign_scalar(self.pager, override.pager),
+            unset_vars=override.unset_vars if override.unset_vars is not None else self.unset_vars,
+            work_dir_extra_paths=override.work_dir_extra_paths
+            if override.work_dir_extra_paths is not None
```
evgunter, line 583: could be good to have a unit test which ensures that this function doesn't miss any fields, e.g. one which creates a `MngrConfig` object which has non-default values of all keys (and asserts this is the case), then ensures that that object survives a round trip through this with an empty update

libs/mngr/imbue/mngr/config/data_types.py
```diff
@@ -564,186 +542,95 @@ class MngrConfig(FrozenModel):
...
-            disabled_plugins=merged_disabled_plugins,
+            # disabled_plugins is a deliberate carveout from the assign-by-default
+            # rule: it is populated by CLI ``--disable-plugin`` flags via
+            # ``_apply_plugin_overrides`` *after* layers have been merged, and
+            # individual TOML layers do not normally write to it. An override
+            # with an empty frozenset is indistinguishable from "not set", so we
+            # preserve the base value instead of wiping it. Users wanting to
+            # disable a specific plugin per-scope should use
+            # ``[plugins.<name>] enabled = false`` rather than this field.
+            disabled_plugins=override.disabled_plugins if override.disabled_plugins else self.disabled_plugins,
```
evgunter, line 599: sad

libs/mngr/imbue/mngr/config/loader.py
```diff
@@ -795,90 +846,108 @@ def parse_config(
...
+def _collect_env_overrides(environ: Mapping[str, str]) -> dict[str, Any]:
+    """Combine ``MNGR__*`` overrides with preserved-alias env vars into a single
+    raw config dict.
 
-    return result
+    Each preserved alias uses its historic value parser (see
+    ``_PRESERVED_ALIASES``) so backwards-compatible spellings like
+    ``MNGR_HEADLESS=yes`` keep their old meaning. Raises ``ConfigParseError``
+    when a preserved alias and its canonical ``MNGR__*`` form are both set
+    with different parsed values.
```
evgunter, line 900: the tyrant in me wants to make cases like this a warning so new users won't do it

evgunter:
looks great!

libs/mngr/docs/commands/secondary/config.md
joshalbrecht: you're right, this could probably just be a new flag for the format of "config list", I might just have it do that instead

libs/mngr/docs/concepts/environment_variables.md
joshalbrecht: you are correct, these are certainly not equivalent, will change the phrasing, thanks!

libs/mngr/docs/concepts/environment_variables.md
joshalbrecht: it's not about being used in tests--it's about preventing tests from using our .mngr/settings.toml (ie, tests that fail to set the home dir or mngr prefix or cwd)

leaving this, it's saved us a lot of pain for that reason

libs/mngr/imbue/mngr/config/completion_cache.py
joshalbrecht: you're right, I think I'll revert this one and make it special

libs/mngr/imbue/mngr/agents/base_agent.py
joshalbrecht: some fields *became* config vars (that were previously only settable as environment vars), ie, this is just how it was supposed to work in the first place

libs/mngr/imbue/mngr/config/data_types.py
joshalbrecht: unfortunately these felt just too awkward (adding a new agent type or provider and accidentally nuking all others isn't never what you expected)

such keys are better thought of as "registries" than config values, really

I wouldn't be opposed to a follow up PR that tried that out to see how it felt, but my guess is that it will be just too awkward

libs/mngr/imbue/mngr/config/data_types.py
joshalbrecht: thanks, will fix

libs/mngr/imbue/mngr/config/data_types.py
joshalbrecht: yeah, I'm not sure of a better way right now though

libs/mngr/imbue/mngr/config/loader.py
joshalbrecht: the preserved aliases are *very* special, they cannot really be used as config keys anyway

it's maybe better thought about the other way--we accept the "__" version just in case someone expects all fields to be consistently named this way, but logically, these fields cannot always be part of configs

libs/mngr/docs/concepts/environment_variables.md
evgunter: no, i'm talking about `MNGR_ALLOW_PYTEST`

libs/mngr/imbue/mngr/config/data_types.py
evgunter: ah, that makes sense

libs/mngr/imbue/mngr/config/data_types.py
evgunter: me neither

libs/mngr/imbue/mngr/config/key_resolver.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+def _apply_extend(
+    current_value: Any,
+    extend_value: Any,
+    field_path: str,
+) -> Any:
+    """Apply ``extend_value`` onto ``current_value`` and return the result.
+
+    Operates only at the leaf field, no recursion into nested aggregates.
```
evgunter, line 122: was this a deliberate design choice? i was hoping to build on this branch to fix a related issue where the permissions mode was getting wiped because an aggregate value of a key was replaced (https://github.com/imbue-ai/mngr/pull/1647)

evgunter: well it turns out claude code itself doesn't have the semantics i wanted to fix in that branch, but the question is still relevant in the abstract

PR #1651, opened 2026-05-15
add aws provider

evgunter:
vultr and ovh release tests now pass! (as well as aws, of course)

not sure if this should be a warning or error:
```
WARNING: Could not attach the mngr-aws self-stop IAM instance profile to the new instance (AWS InvalidParameterValue: Value (mngr-aws) for parameter iamInstanceProfile.name is invalid. Invalid IAM Instance Profile name). This usually means the launching credentials lack iam:PassRole for the mngr-aws role, or `mngr aws prepare` has not been run. Launching WITHOUT the profile, so the agent will not self-stop on idle (you can still `mngr stop --stop-host`). To enable self-stop, run `mngr aws prepare` and grant iam:PassRole, or set attach_self_stop_role = false to silence this warning.
```

---

## Summary

- **New `aws` provider** (`libs/mngr_aws/`) running agents in Docker containers on EC2. Uses `boto3` (+ `boto3-stubs[ec2]` for typed dicts), no hand-rolled SigV4. Marked `[experimental]` in its README.
- **Factored the parallel-SSH host-record discovery loop** (previously duplicated in `VultrProvider`) into `VpsDockerProvider`. Subclasses now implement two small hooks: `_list_provider_vps_hostnames()` and `_fetch_provider_instances()`.
- **Dropped `os_id` from the shared `VpsClientInterface`**: `VultrVpsClient` carries `os_id` locally, AWS uses AMI strings, OVH uses image names. `VpsHostConfig` / `ParsedVpsBuildOptions` / `VpsDockerProviderConfig` all lose the field.
- **Per-provider build-args prefix**: `--vps-region=` / `--vps-plan=` are gone. Each provider uses its native prefix:
  - AWS: `--aws-region=`, `--aws-instance-type=`, `--aws-ami=`, `--aws-spot`
  - Vultr: `--vultr-region=`, `--vultr-plan=`
  - OVH: `--ovh-datacenter=` (alias `--ovh-region=`), `--ovh-plan=`
  - Shared: `--git-depth=` (about the local mngr build context).

  Old `--vps-*` args raise a migration error pointing at the new name. `default_plan` is dropped from `VpsDockerProviderConfig`; each provider's config carries its own native field (Vultr/OVH `default_plan`, AWS `default_instance_type`). `vps_boot_timeout` → `instance_boot_timeout`.
- **Composable parser helpers + abstract hook**: the shared parser is rebuilt on `extract_single_value_arg`, `extract_git_depth`, `extract_presence_flag`, `raise_if_vps_migration_arg`, `raise_if_unknown_provider_arg`. `parse_vps_build_args` stays as a convenience for the region+plan+git-depth shape. `VpsDockerProvider._parse_build_args` is `@abstractmethod`. Per-provider parsers compose the low-level helpers; AWS uses the composables directly to add `--aws-ami=` and `--aws-spot` without growing the convenience helper's signature.
- **Per-provider `_create_vps_instance` hook + per-provider parsed-options subclass**: keeps `VpsClientInterface` minimal forever. `AwsProvider._create_vps_instance` overrides the default (which calls `self.vps_client.create_instance(...)`) to call `self.aws_client.create_instance(...)` with AWS-only kwargs `ami_id_override` and `spot`. `ParsedAwsBuildOptions(ParsedVpsBuildOptions)` carries the AWS-specific fields. Future AWS knobs (`--aws-spot-max-price=`, etc.) follow the same three-step recipe — no shared-interface changes needed.
- **`--aws-ami=AMI-ID`** per-host AMI override. Falls back to `default_ami_id` / `default_ami_by_region` when omitted.
- **`--aws-spot`** (presence-only) opts the host into EC2 spot capacity. AWS may reclaim with ~2 min notice; cloud-init auto-shutdown safety net still fires. Documented opt-in.
- **New `mngr aws prepare` CLI command**: idempotent one-time setup that creates the `mngr-aws` security group and authorizes SSH ingress. Lifts the privileged SG creation out of the `mngr create` hot path. The new lookup-only `AwsVpsClient.resolve_security_group_id()` (used by `create_instance`) needs only `ec2:DescribeSecurityGroups`, so developers can run `mngr create --provider aws` with restricted IAM (no `CreateSecurityGroup` / `AuthorizeSecurityGroupIngress`). Missing-SG errors point users at `mngr aws prepare`. A `[future]`-tagged `mngr aws ami` stub is also registered.
- **New `mngr aws cleanup` CLI command**: the safe inverse of `prepare`. Deletes the `mngr-aws` security group so a region returns to its pre-prepare state. Refuses (deletes nothing) while any mngr-managed instance still exists in the region, so it can't strand a running agent; idempotent when the SG is already gone. Needs `ec2:DescribeInstances` + `ec2:DescribeSecurityGroups` + `ec2:DeleteSecurityGroup`. Does not delete per-host keypairs (those follow the create/destroy lifecycle). Backed by new `AwsVpsClient.delete_security_group()` and `list_mngr_managed_instances()`.
- **`allowed_ssh_cidrs` default flipped to `("0.0.0.0/0",)`** to match the de-facto Vultr / OVH norm in this monorepo (neither ships a managed firewall; instances are reachable from anywhere on tcp/22 at the network layer; key-only SSH is the actual protection). A warning is logged at provision time when the effective range includes `0.0.0.0/0` or is empty, so production users get a visible nudge to tighten.
- **New `MinimalVpsDockerProvider`** in `mngr_vps_docker`: the public, instantiable `VpsDockerProvider` for the "externally-managed VPS lifecycle, host-setup only" role. Pairs with `ExternallyManagedVpsClient`. Used by `mngr_imbue_cloud`'s slow path; available for any other caller that needs the same shape.
- **New `auto_shutdown_minutes` field** on `VpsDockerProviderConfig` — cloud-init schedules `shutdown -P +N` when set. On AWS, paired with `InstanceInitiatedShutdownBehavior=terminate`, the EC2 instance auto-terminates from the inside even if the orchestrating process is killed.
- **`is_for_host_creation` removed from `ProviderBackendInterface`**: replaced with default-no-op `bootstrap_for_host_creation(name, config, mngr_ctx)` that Modal overrides; `mngr create` calls it before `build_provider_instance`.
- **New `resolve_backend_and_config(provider_name, mngr_ctx)` helper** on `mngr/providers/registry.py`. Both `get_provider_instance` and the `mngr create` bootstrap path now use it.
- **`wait_for_instance_active` lifted onto `VpsClientInterface`** as a default method with a `slow_provisioning_warning_threshold_seconds` field for per-provider tuning. AWS / Vultr no longer duplicate the polling implementation.
- **Modal-style three-layer leak defense** for AWS release tests: per-test `mngr destroy --force`, `pytest_sessionfinish` orphan scan that filters by an `mngr-pytest-launched=true` tag (which `AwsVpsClient.create_instance` adds when `PYTEST_CURRENT_TEST` is set), and cloud-init `shutdown -P +60` with `InstanceInitiatedShutdownBehavior=terminate`.
- **Security-tightened defaults** on AWS instances: IMDSv2 enforced (`HttpTokens: required`, `HttpPutResponseHopLimit: 1`); EBS root volume `Encrypted: True` always; `ensure_security_group` errors when >1 SG matches the configured name (avoids guessing across VPCs).
- **`security_group` config is a tagged union** (`ExistingSecurityGroup(id=...) | AutoCreateSecurityGroup(name=...)`) so invalid combinations are unrepresentable.
- **Construction-time validation** for AWS: missing credentials raise `ProviderUnavailableError` (state-unknown; the framework logs at error level on read paths so misconfigured providers stay visible in `mngr list` / `connect` / `gc` instead of silently vanishing). Missing AMI is a create-only concern — it surfaces as `MngrError` from `bootstrap_for_host_creation` so `mngr create` fails fast, while `mngr list` is unaffected (a misconfigured-AMI provider's running instances still appear). Credential and AMI resolution live in two module-level helpers (`_resolve_session_or_unavailable`, `_resolve_ami_or_raise`); neither logs.
- **`test_default_amis_describe_successfully` suggests a fix**: when a region's pinned AMI is stale, the release test queries the Debian publisher (owner `136693071363`) for the current `debian-12-amd64-*` AMI in that region and emits a copy-pasteable replacement dict literal in the failure message.

## What's new for AWS

- **One-time admin step**: `mngr aws prepare --region us-east-1` creates the `mngr-aws` security group. Run once with full creds; from then on, devs can `mngr create` with restricted creds.
- **Teardown**: `mngr aws cleanup --region us-east-1` deletes the `mngr-aws` security group once no mngr-managed instances remain in the region (refuses otherwise, so it can't strand a running agent).
- **Auth**: boto3 default credential chain only. No mngr-side credential config.
- **Networking**: the `mngr-aws` security group opens tcp/22 and the container SSH port to every CIDR in `allowed_ssh_cidrs` (default: `("0.0.0.0/0",)`). Tighten for production (e.g., `("203.0.113.4/32",)`). A warning is logged at provision time when the effective range is `0.0.0.0/0` or empty.
- **Per-host build-arg knobs**: `--aws-region=`, `--aws-instance-type=`, `--aws-ami=`, `--aws-spot`, `--git-depth=`. **Note**: `--aws-region=` is not a free per-host override — `AwsVpsClient` is bound to one region at construction and `create_instance` rejects cross-region requests with `VpsApiError "Cross-region create not supported"`. The arg exists so user-supplied region matches the provider's configured one; to target multiple regions, define one `[providers.aws-<region>]` block per region (see "Multi-region" below).
- **SSH key**: per-host EC2 KeyPair via `ImportKeyPair`, deleted on `destroy_host`.
- **Tags**: every instance gets `mngr-provider`, `mngr-host-id`, `mngr-created-at`. Discovery filters `DescribeInstances` by `tag:mngr-provider`. Under pytest, instances also get `mngr-pytest-launched=true` for the leak scanner.
- **Multi-region**: each provider instance is bound to a single region; configure one `[providers.aws-<region>]` block per region.

## Testing performed

- **Unit tests**: 873 passing across the touched packages (mngr_vps_docker 261, mngr_aws 114, mngr_ovh 202, mngr_vultr 88, mngr_imbue_cloud 176, plus mngr core via CI).
- **Type checking**: zero type errors across mngr, mngr_vps_docker, mngr_aws, mngr_vultr, mngr_ovh, mngr_imbue_cloud.
- **AWS release tests against real AWS** (`test_release_aws.py`):
  - All 6 lifecycle / API release tests passing (~340s end to end). Earlier "lifecycle test times out" issue was a layered cloud-init bug (sshd restart killing in-flight SSH, slow `curl get.docker.com | sh`, root authorized_keys not getting the key). All three root causes are fixed; the test is reliable.
  - `test_default_amis_describe_successfully` surfaced 6 of 8 stale AMI IDs on a previous run; refreshed `DEFAULT_AMI_BY_REGION` with current Debian 12 amd64 IDs. The improved test also emits a copy-pasteable replacement dict literal when entries go stale.
- **Security audit**: subagent-driven review of `mngr_aws` surface — no real vulnerabilities; defaults flagged (and now tightened); 3 follow-up nits resolved (unused IAM permissions in README trimmed; security-group name-collision guard added; AMI staleness test landed).
- **autofix sweeps caught three real cross-file breakages** introduced by the per-provider rename and folded them into the same PR: `mngr_imbue_cloud admin pool create` was still passing `--vps-datacenter=` to the inner OVH `mngr create`; `apps/minds`'s desktop_client CLOUD-launch path was still passing `--vps-region=` to Vultr; `mngr_imbue_cloud`'s delegated slow-path was instantiating bare `VpsDockerProvider` (whose now-abstract `_parse_build_args` would have raised at every slow-path rebuild). All three fixed.

## Limitations / future work

Tagged `[future]` in `libs/mngr_aws/README.md`:

- `mngr aws ami` (build a Debian + Docker + deps AMI). Stub command is registered today.
- mngr-published public AMI.
- GPU AMI automation. Pairs with the `mngr aws ami` work above.
- Optional EIP allocation for stable public addressing across stops/starts. Needs destroy-path lifecycle awareness (otherwise leaks ~$3.60/month per abandoned EIP); the parser-extension pattern handles the `--aws-eip` build arg trivially, but the cleanup work is a separate change.
- Auto SSM Parameter Store lookup for current Debian AMIs per region.
- Multi-container per EC2 instance packing.
- *Automatic* cleanup of the `mngr-aws` security group on the final `destroy` of a region. Manual teardown landed in this PR (`mngr aws cleanup`, which deletes the SG once no mngr-managed instances remain in the region); the remaining future work is doing it automatically on the last destroy.


joshalbrecht:
Ok, finally got a chance to digest this. Thanks for putting it together!

Let's chat about it live--there are a few architectural things we'll want to fix, like:

1. security group configuration should be a one-off command (like `mngr aws prepare` or something)
2. we'll want to figure out the right names (so stuff isn't `--vps-` prefixed)
3. I have questions about allowed_ssh_cidrs
4. AMI's really feel like they are about building images... but I'm not quite sure how we want to interact with that
5. we'll want to fix some of those limitations

libs/mngr/docs/commands/secondary/aws.md
```diff
@@ -0,0 +1,39 @@
...
+
+# mngr aws
+**Usage:**
+
+```text
+mngr aws [OPTIONS] COMMAND [ARGS]...
+```
+**Options:**
+
+
```
evgunter, line 12: i guess there are no options yet, but looks strange

libs/mngr_vps_docker/imbue/mngr_vps_docker/instance.py
```diff
@@ -126,50 +126,67 @@ class ParsedVpsBuildOptions(FrozenModel):
...
-                    f"{arg.split('=', 1)[0]} is no longer supported. OS image selection lives on the provider "
-                    "config: set ``default_os_id`` ([providers.<name>] for Vultr), ``default_image_name`` "
-                    "(OVH), or ``default_ami_id`` / ``default_ami_by_region`` (AWS) in your settings.toml. "
-                    "Per-host image overrides require a separate provider instance with its own config block."
+                    f"{arg.split('=', 1)[0]} is no longer supported. Build args are now "
+                    f"per-provider: use --{provider_prefix}-region=, --{provider_prefix}-{plan_arg_name}= "
+                    "(or the equivalent on a different provider) instead. The old --vps-os= / "
+                    "--vps-image= / --vps-ami= image-selection args are also removed; image selection "
+                    "lives on the provider config (default_os_id for Vultr, default_image_name for OVH, "
+                    "default_ami_id / default_ami_by_region for AWS)."
```
evgunter: is this how we want to do this?

joshalbrecht:
re:
> WARNING: Could not attach the mngr-aws self-stop IAM instance profile to the new instance (AWS InvalidParameterValue: Value (mngr-aws) for parameter iamInstanceProfile.name is invalid. Invalid IAM Instance Profile name). This usually means the launching credentials lack iam:PassRole for the mngr-aws role, or `mngr aws prepare` has not been run. Launching WITHOUT the profile, so the agent will not self-stop on idle (you can still `mngr stop --stop-host`). To enable self-stop, run `mngr aws prepare` and grant iam:PassRole, or set attach_self_stop_role = false to silence this warning.

that's a slightly strange way of doing it (because it requires the instance profile). I mean, sure, I guess, but also, you should be able to get away with a cloud-init script that does 99.99% of the same thing

THEN that can just be a warning

joshalbrecht:
> allowed_ssh_cidrs default flipped to ("0.0.0.0/0",) to match the de-facto Vultr / OVH norm in this monorepo (neither ships a managed firewall; instances are reachable from anywhere on tcp/22 at the network layer; key-only SSH is the actual protection). A warning is logged at provision time when the effective range includes 0.0.0.0/0 or is empty, so production users get a visible nudge to tighten.

I still don't fully understand what is being said here (is this incoming? outgoing? both?)

also worth us thinking a little bit harder about what our defaults ought to be... (and documenting those)

joshalbrecht:
> New auto_shutdown_minutes field on VpsDockerProviderConfig

Minor -- this should be seconds (like the rest of our units), and ideally the same flag name / value as what modal uses for similar effect (or at least closer to that one)

joshalbrecht:
Lmk if there are entries here that you don't think we should do in follow ups!

```
mngr aws ami (build a Debian + Docker + deps AMI). Stub command is registered today.
mngr-published public AMI.
GPU AMI automation. Pairs with the mngr aws ami work above.
Optional EIP allocation for stable public addressing across stops/starts. Needs destroy-path lifecycle awareness (otherwise leaks ~$3.60/month per abandoned EIP); the parser-extension pattern handles the --aws-eip build arg trivially, but the cleanup work is a separate change.
Auto SSM Parameter Store lookup for current Debian AMIs per region.
Multi-container per EC2 instance packing.
Automatic cleanup of the mngr-aws security group on the final destroy of a region. Manual teardown landed in this PR (mngr aws cleanup, which deletes the SG once no mngr-managed instances remain in the region); the remaining future work is doing it automatically on the last destroy.
```

libs/mngr/docs/commands/secondary/aws.md
joshalbrecht: how does this play with multiple accounts?

like, the AWS CLI allows you to specify "profile" IIRC

and what we've done with modal and imbue_cloud is a little bit different for each (w.r.t having multiple accounts for a given provider). We should probably figure out exactly how we want that to work...

libs/mngr/imbue/mngr/hosts/outer_host.py
```diff
@@ -404,6 +414,13 @@ def _run_shell_command(
         with self._notify_on_connection_error():
             try:
                 return self._run_shell_command_with_transient_retry(command, pyinfra_kwargs)
+            except TimeoutError as e:
+                # ``TimeoutError`` is a subclass of ``OSError``, so this
+                # must precede the OSError branch below. Reached when the
+                # retry decorator has exhausted its attempts on transient
+                # SSH read timeouts; surface as a structured
+                # HostConnectionError so callers don't see a raw timeout.
+                raise HostConnectionError("SSH command timed out reading output") from e
```
joshalbrecht: hmmm.... this comment and logic is copy-pasted in quite a few places... (within this file).

that makes me quite suspicious that this should probably be part of, for example, _notify_on_connection_error or something else

libs/mngr_aws/imbue/mngr_aws/backend.py
```diff
@@ -0,0 +1,316 @@
...
+        spot, args = extract_presence_flag(args, "--aws-spot")
+        git_depth, args = extract_git_depth(args)
+        valid_args = (
+            "--aws-region=",
+            "--aws-instance-type=",
+            "--aws-ami=",
+            "--aws-spot",
+            "--git-depth=",
+        )
+        docker_build_args: list[str] = []
```
joshalbrecht, line 145: it really feels like there are a lot more args than that for AWS...

libs/mngr_aws/imbue/mngr_aws/client.py
```diff
@@ -0,0 +1,624 @@
...
+    subnet_id: str | None = Field(default=None, description="Subnet ID, or None to let EC2 pick a default")
+    vpc_id: str | None = Field(default=None, description="VPC ID, used only to scope SG lookup")
+    allowed_ssh_cidrs: tuple[str, ...] = Field(
+        default=("0.0.0.0/0",),
+        description=(
+            "CIDR blocks allowed inbound on tcp/22 and tcp/container_ssh_port of the auto-created "
+            "security group. Default ('0.0.0.0/0',) matches Vultr/OVH default reachability in "
+            "this monorepo (neither provider ships a managed firewall). Tighten for production "
+            "(e.g. ('203.0.113.4/32',) for a single IP). Empty tuple means 'add no ingress' -- "
+            "the SG ends up unreachable from outside its VPC; logged as a warning."
```
joshalbrecht: I've read this a few times now, and it's just really confusing

libs/mngr_aws/imbue/mngr_aws/conftest.py
```diff
@@ -0,0 +1,189 @@
...
+_TEST_LEAK_TTL: Final[timedelta] = timedelta(minutes=AWS_TEST_INSTANCE_AUTO_SHUTDOWN_MINUTES)
+
+
+def _force_terminate_instances(ec2: Any, instance_ids: list[str]) -> None:
+    if not instance_ids:
+        return
+    try:
+        ec2.terminate_instances(InstanceIds=instance_ids)
+    except (BotoCoreError, ClientError) as e:
+        logger.warning("Failed to terminate leaked EC2 instances {}: {}", instance_ids, e)
```
joshalbrecht: so skeptical that it is reasonable to just warn when we fail like this...

same basic comment applies below as well

libs/mngr_vps_docker/imbue/mngr_vps_docker/vps_client.py
```diff
@@ -67,14 +76,49 @@ def get_instance_ip(self, instance_id: VpsInstanceId) -> str:
...
+                    if elapsed > threshold:
+                        logger.warning(
+                            "VPS {} provisioning took {:.1f}s (threshold: {:.0f}s)",
+                            instance_id,
+                            elapsed,
+                            threshold,
+                        )
+                    return ip
+                except VpsProvisioningError:
+                    pass
```
joshalbrecht: should at least log

joshalbrecht:
lgtm after the above are addressed!

we'll want to coordinate (I'll need to change some minds stuff for this to work out), but lmk when the above are addressed and I'm happy to do that testing

libs/mngr_aws/imbue/mngr_aws/conftest.py
evgunter: this is only used on the path which will raise an error and fail the test anyway (but it should probably be inlined so it doesn't get used anywhere else)

evgunter:
vultr and ovh release tests pass! but have flakes. will fix flakes in https://github.com/imbue-ai/mngr/pull/2110

libs/mngr/imbue/mngr/hosts/outer_host.py
evgunter: fixed

libs/mngr_aws/imbue/mngr_aws/backend.py
evgunter: i put a FIXME

libs/mngr_vps_docker/imbue/mngr_vps_docker/vps_client.py
evgunter: done

evgunter:
> that's a slightly strange way of doing it (because it requires the instance profile). I mean, sure, I guess, but also, you should be able to get away with a cloud-init script that does 99.99% of the same thing

i'm already reworking stuff related to this in https://github.com/imbue-ai/mngr/pull/2097, so i'll fold that in there

libs/mngr_aws/imbue/mngr_aws/client.py
evgunter: better now?

PR #1652, opened 2026-05-15
fixture-scoped resource guards

evgunter:
## Summary
Adds an opt-in `@fixture_uses_resources(...)` decorator to `imbue.resource_guards`. Fixtures that declare which resources they use run their setup and teardown under an independent guard scope: resource calls inside the fixture are authorized against the fixture's own declaration rather than whichever consuming test happens to trigger setup. Untagged fixtures keep their existing behavior (per-test attribution).

Applied to `deployed_snapshot_function` in `libs/mngr_modal/imbue/mngr_modal/routes/test_snapshot_and_shutdown.py` to fix `test_snapshot_and_shutdown_missing_host_id` and `test_snapshot_and_shutdown_missing_sandbox_id`, which were failing because the shared module-scoped fixture's modal calls were being attributed to a single consuming test and triggering the "marked with @pytest.mark.modal but never invoked modal" guard on siblings.

## Mark semantics on consumers of tagged fixtures

`@pytest.mark.<resource>` on a test is satisfied by *either*:
- the test body directly invoking the resource (the original meaning), OR
- the test consuming a `@fixture_uses_resources(<resource>)` fixture in its closure (the fixture's setup is independently verified to invoke the resource, so the mark stays meaningful).

The mark is **required** on every consumer of a tagged fixture, even consumers that only hit the fixture's output (e.g. HTTPing the URL it yields). This keeps `pytest -m <resource>` as the canonical "select every test that transitively needs the resource" selector — there's no escape hatch where a consuming test silently inherits the dependency without declaring it.

The block check (direct invocation without the mark) is unchanged.

## What's in the diff
- `libs/resource_guards/imbue/resource_guards/resource_guards.py`: new `fixture_uses_resources` decorator + `pytest_fixture_setup` hookwrapper wired into `_ResourceGuardPlugin`. New `_collect_fixture_covered_resources` helper that walks the test's fixture closure; raises `ResourceGuardViolation` if it finds an override of a tagged fixture (ambiguous case deliberately unsupported). `_check_guard_violations` gained a third check for undeclared fixture coverage. Backward-compatible for untagged fixtures.
- `libs/resource_guards/imbue/resource_guards/resource_guards_test.py`: pytester end-to-end tests for allow / block / never-invoked / non-leak-to-consumer / teardown-authorization / required-mark-on-consumer; direct unit tests for the new helper and the new check.
- `libs/resource_guards/imbue/resource_guards/test_ratchets.py`: bumped `inline_imports` (16 -> 33, from new pytester test strings) and `inline_functions` (3 -> 4, closures in the fixture wrapper helper).
- `libs/resource_guards/README.md`: documents the new decorator and the consumer-mark requirement.
- `libs/mngr_modal/imbue/mngr_modal/routes/test_snapshot_and_shutdown.py`: dropped module-level `pytestmark`; decorated the fixture with `@fixture_uses_resources("modal")`; all five tests carry `@pytest.mark.modal` per the new "consumers must declare" rule.
- `changelog/mngr-fix-test.md`.

## Test plan
- [x] All 131 resource_guards tests pass locally with 100% coverage (`just test-quick libs/resource_guards`)
- [ ] CI passes `test_snapshot_and_shutdown_missing_host_id` and `_missing_sandbox_id` on offload
- [ ] CI passes the three remaining `test_snapshot_and_shutdown_*` modal tests (unchanged behavior)
- [ ] No regressions in other modal-marked acceptance tests


libs/resource_guards/imbue/resource_guards/resource_guards.py
```diff
@@ -515,6 +519,133 @@ def _build_per_test_guard_env(marks: set[str], tracking_dir: str) -> dict[str, s
...
+# Such fixtures run setup/teardown under their own guard scope rather than
+# attributing resource calls to whichever test triggered the setup -- which
+# matters for module/session-scoped fixtures shared across tests. Untagged
+# fixtures keep today's per-test attribution behavior.
+
+
+# Maps a fixture function to the resources it has declared via
+# @fixture_uses_resources. WeakKeyDictionary so registering a decorator on
+# a transient function does not leak the function past its lifetime.
+_fixture_resource_marks: weakref.WeakKeyDictionary[Callable[..., Any], set[str]] = weakref.WeakKeyDictionary()
```
evgunter: huh haven't seen this before

joshalbrecht:
I think it's gotta be (1), not (2), if I'm understanding this correctly?  (unless you have some context I am missing)

other than that lgtm (and happy to chat if you had other reasons for going with (2) )

libs/resource_guards/imbue/resource_guards/resource_guards.py
joshalbrecht: scary...

evgunter:
(1) struggles to handle libs/mngr_modal/imbue/mngr_modal/routes/test_snapshot_and_shutdown.py, where there are tests which only use modal via `deployed_snapshot_function`; since `deployed_snapshot_function` is module-scoped,  the test its modal invocation gets attributed to isn't deterministic. we could work around that particular case, but in principle handling this via (1) would mean we'd need to introduce another kind of mark for "can, but need not, invoke the resource" (claude summarized it wrong; it wouldn't be `pytest.mark.modal` for the case where we need `fixture_uses_resources`)

libs/resource_guards/imbue/resource_guards/resource_guards.py
evgunter: should i do anything about this?

evgunter: ok leaking the function is not actually a real problem, so simplifying

evgunter:
switched to requiring every test which uses a fixture bearing the `fixture_uses_resources(resource_x)` being required to have `pytest.mark.resource_x` + added tests for a variety of edge cases

libs/resource_guards/imbue/resource_guards/resource_guards.py
```diff
@@ -515,6 +529,231 @@ def _build_per_test_guard_env(marks: set[str], tracking_dir: str) -> dict[str, s
...
+                f"RESOURCE GUARD: Fixture '{fixture_id}' invoked '{violation.resource}' but did not declare it via "
+                f"@fixture_uses_resources({violation.resource!r}). Add the declaration or remove the {violation.resource} usage."
+            ) from setup_exception
+        case _GuardViolationKind.NEVER_INVOKED:
+            raise ResourceGuardViolation(
+                f"RESOURCE GUARD: Fixture '{fixture_id}' declared @fixture_uses_resources({violation.resource!r}) "
+                f"but did not invoke {violation.resource} during setup. Remove the declaration "
+                f"or ensure the fixture exercises {violation.resource}."
+            ) from setup_exception
+        case _:  # pragma: no cover
```
evgunter, line 783: this package has a 100% coverage threshold i didn't want to weaken

libs/resource_guards/imbue/resource_guards/test_ratchets.py
```diff
@@ -69,8 +69,15 @@ def test_prevent_silent_decode_error_catches() -> None:
...
+# The count here is dominated by string-literal misfires, not real inline
+# imports in module bodies: resource_guards_test.py uses pytester to compile
+# small test files at runtime, and the Python source for those files lives
+# inside triple-quoted strings with indented `import` lines that the regex
+# matches the same way it would match a real inline import. Tightening the
+# rule to skip string literals lives in imbue_common, not this package, so
+# the snapshot bump is intentional.
 def test_prevent_inline_imports() -> None:
-    rc.check_inline_imports(_DIR, snapshot(16))
+    rc.check_inline_imports(_DIR, snapshot(85))
```
joshalbrecht, line 80: sure, though how bad would it be to fix our regex / approach to finding inline imports so that this is more robust anyway?

joshalbrecht:
lgtm assuming tests pass--my comment is only really meant as inspiration for a followup PR

PR #1655, opened 2026-05-15
mngr: refactor agent address resolution into find + resolve phases

qi-imbue:
## My summary

When trying to understand how various subcommands that operate on one agent resolve their agent addresses, I go really confused by how the old `find_one_agent`'s `is_start_desired` and `skip_agent_state_check` arguments work and interact. As it turned out, Claude was confused by it too, and it ended up passing the wrong values in a bunch of subcommands, causing them to start the agent unnecessarily when they don't actually require the agent to be started. Claude's description of the solution below is pretty decent: it's slightly more verbose but much more explicit and less confusing.

There are a couple refactor/cleanup changes related to address parsing and resolution in this PR; again, Claude's description is pretty decent this time.

Another two changes that I did not include as part of the PR:

- The `rename` subcommand today requires the host to be online, but it doesn't actually have to be. Today it operates on an online host because it also wants to rename the tmux session, but we can simply skip this when the host is offline, and manipulate the agent record through the host volume. https://github.com/imbue-ai/mngr/pull/1659
- I introduced `HostedLocation` to represent things like the argument of `mngr create --from`. I didn't realize at that time that there is already a `HostLocation` which is the runtime representation of it - this means that the former is in fact better named `HostLocationAddress` (to mirror Agent:AgentAddress, Host:HostAddress). https://github.com/imbue-ai/mngr/pull/1658

---------------

## Summary

Refactors how mngr's single-agent subcommands turn an `AgentAddress` into the live interfaces they operate on. The old flow welded three concerns into one `find_one_agent` call — discovery, host materialization, and agent lifecycle — controlled by two implicit-default flags (`is_start_desired`, `skip_agent_state_check`) that silently conflated host- and agent-state requirements. The new flow is two explicit phases:

```python
host_ref, agent_ref = find_one_agent(address, mngr_ctx)                # metadata
agent, host = resolve_to_started_host_and_[running_]agent(host_ref, agent_ref,
                                                          allow_auto_start=opts.start,
                                                          mngr_ctx=mngr_ctx)
```

Each subcommand picks the resolver that matches its actual requirement:

| Command | Resolver |
| --- | --- |
| connect, capture | `resolve_to_started_host_and_running_agent` |
| push, pull, provision, rename | `resolve_to_started_host_and_agent` (no agent-state requirement) |
| exec (single, API only) | `resolve_to_started_host_and_running_agent` |
| mngr_file | bespoke (intentionally supports offline-via-volume) |

Multi-agent commands (`start`, `stop`, `destroy`, `exec` multi, `message`) are deliberately out of scope here.

## User-visible behavior changes

- **push, pull, and provision no longer require the agent to be running.** Previously they failed against a stopped agent on an online host, even though their operations (rsync, file copy, provisioning) only need the host. Now they succeed.
- **push, pull, provision, and rename gain a `--start/--no-start` flag** (default `--start`). Previously they had no way to control host auto-start.
- **`mngr connect` no longer auto-picks the most recently created agent non-interactively.** Running `mngr connect` with no agent in a non-interactive shell used to silently pick the most recent agent; it now errors out like every other single-agent command. Use an interactive shell or pass an explicit agent.
- **`mngr file` now uses standard `AGENT_OR_HOST_ADDRESS` parsing.** A `SafeName`-shaped host name (no `host-` prefix, no dots) must be written as `@hostname` to be parsed as a host. Bare `hostname` parses as an agent. This matches every other mngr subcommand.
- **Cancelling the interactive agent selector now exits cleanly via `click.Abort`** instead of returning None silently.
- **`--start` help text** on `connect`, `capture`, `exec`, and the new flags on push/pull/provision/rename now accurately describes what each command starts.

## Notable internal changes

- New `api/find.py` helpers: `find_one_agent` (metadata-only), `find_one_agent_and_agents_by_host` (sibling returning full discovery), `resolve_to_started_host_and_agent`, `resolve_to_started_host_and_running_agent`.
- `api/exec.py:exec_command_on_agent` and `mngr_file/resolve_file_target` (agent half) now reuse the standard `find_one_agent` flow — duplicate inline resolution gone.
- The agent-selector TUI moved out of `cli/connect.py` into a new `cli/agent_selector.py` to break the import cycle the new structure exposed.
- `filter_one_agent`, `filter_all_agents`, and `materialize_agent` removed from `api/find.py`'s public surface; the unused `filter_agents_by_host` deleted.

## Test plan

- [ ] `mngr push <stopped-agent> .` — succeeds (previously failed).
- [ ] `mngr pull <stopped-agent>` — succeeds.
- [ ] `mngr provision <stopped-agent>` — works without `is_start_desired` flag, gets the host online.
- [ ] `mngr push --no-start <agent-on-offline-host>` — fails with a clear "host is offline" error.
- [ ] `mngr push <agent-on-offline-host>` (default `--start`) — auto-starts the host.
- [ ] `mngr connect` (no agent, non-interactive shell, e.g. piped output) — errors out cleanly, no longer silently picks most-recent agent.
- [ ] `mngr connect` (no agent, interactive terminal) — shows selector; Ctrl-C twice exits cleanly without stack trace.
- [ ] `mngr connect <stopped-agent>` (default `--start`) — starts agent and connects.
- [ ] `mngr connect --no-start <stopped-agent>` — fails with "agent is stopped" error.
- [ ] `mngr rename <stopped-agent> new-name` — renames the stopped agent (auto-starts host if needed).
- [ ] `mngr file get @localhost /etc/hostname` — works with explicit `@` for host.
- [ ] `mngr file get localhost /etc/hostname` — now errors (parses as agent name); previously fell back to host lookup.
- [ ] `mngr file get <stopped-agent-on-offline-host> path --relative-to state` — still works via volume backend (offline-host access preserved).
- [ ] Acceptance + release tests via CI.

### DIFF
--- a/libs/mngr/imbue/mngr/api/find.py
+++ b/libs/mngr/imbue/mngr/api/find.py
@@ -113,7 +113,7 @@ def filter_one_host(
 
 
 @pure
-def filter_all_agents(
+def _filter_all_agents(
     agent: AgentNameOrId,
     agents_by_host: Mapping[DiscoveredHost, Sequence[DiscoveredAgent]],
     resolved_host: DiscoveredHost | None = None,
@@ -147,7 +147,7 @@ def _filter_one_agent(
     The multi-match error lists each matching agent in ``NAME@HOST.PROVIDER``
     form so the user can disambiguate.
     """
-    matches = filter_all_agents(agent, agents_by_host, resolved_host)
+    matches = _filter_all_agents(agent, agents_by_host, resolved_host)
     if len(matches) == 0:
         if isinstance(agent, AgentId):
             raise AgentNotFoundError(str(agent))
--- a/libs/mngr/imbue/mngr/api/find.py
+++ b/libs/mngr/imbue/mngr/api/find.py
@@ -552,8 +552,8 @@ def find_one_agent_and_agents_by_host(
 
     Returns only metadata. Callers that need a live ``AgentInterface`` or
     ``OnlineHostInterface`` should compose with
-    :func:`imbue.mngr.cli.agent_utils.ensure_host_started_and_resolve_agent`
-    or :func:`imbue.mngr.cli.agent_utils.ensure_host_and_agent_started`.
+    :func:`resolve_to_started_host_and_agent` or
+    :func:`resolve_to_started_host_and_running_agent`.
 
     Raises :class:`UserInputError` if the host constraint matches no hosts.
     Raises :class:`AgentNotFoundError` / :class:`UserInputError` if the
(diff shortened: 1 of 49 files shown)


libs/mngr/imbue/mngr/api/find.py
```diff
@@ -571,31 +537,101 @@ def _post_filter_matches_by_addresses(
     return filtered
 
 
-def find_one_agent(
+def find_one_agent_and_agents_by_host(
```
qi-imbue, line 540: This is a little bit yucky, but it's needed by `rename`. Everywhere else uses `find_one_agent`.

libs/mngr/imbue/mngr/cli/agent_selector.py
qi-imbue, line 1: This is moved (hopefully verbatim) from `cli/connect.py`. Previously `agent_utils.py` imported `connect.py`, which was a bad idea and resulted in circular import when standardizing on the same TUI agent selector in `connect.py`.

libs/mngr/imbue/mngr/cli/agent_utils.py
```diff
@@ -19,100 +16,61 @@
...
-            mngr_ctx,
-            is_start_desired=is_start_desired,
-            skip_agent_state_check=skip_agent_state_check,
-        )
+        return find_one_agent(address, mngr_ctx)
 
     if not mngr_ctx.is_interactive:
         raise UserInputError("No agent specified and not running in interactive mode (specify an agent name or ID)")
 
-    return select_agent_interactively_with_host(
```
qi-imbue, line 109: The following delta is inlining `select_agent_interactively_with_host`, because it's only ever used here.

libs/mngr/imbue/mngr/cli/agent_utils.py
```diff
@@ -19,100 +16,61 @@
...
-    return find_one_agent(
-        AgentAddress(agent=selected.id),
-        mngr_ctx,
-        is_start_desired=is_start_desired,
-        skip_agent_state_check=skip_agent_state_check,
-    )
-
-
-def find_agent_for_command(
+def find_agent_by_address_or_interactively(
```
qi-imbue, line 19: Renamed to better reflect what it actually does.

libs/mngr/imbue/mngr/cli/connect.py
qi-imbue, line 1: This no longer has a bespoke agent-finding implementation and just uses the one in `agent_utils.py` like the other subcommands.

libs/mngr_file/imbue/mngr_file/cli/get.py
qi-imbue, line 1: This was missed in the previous address parsing refactor PR, now fixed.

libs/mngr/imbue/mngr/api/find.py
joshalbrecht: yup, makes sense

joshalbrecht:
lgtm

PR #1659, opened 2026-05-15
mngr: rename agents on offline hosts; drop --start from rename

qi-imbue:
## Summary
- `mngr rename` now works against offline hosts: the rename (and any `-l KEY=VALUE` labels) are written directly to the provider's persisted agent data, no host start required.
- Lifted `rename_agent` from `OnlineHostInterface` up to `HostInterface` with an `OfflineHost` implementation; both share `apply_rename_to_agent_data` and `BaseHost._check_rename_conflict`.
- Dropped `--start/--no-start` from `mngr rename` since the command no longer needs to start a host.

## Test plan
- [x] `just test-quick libs/mngr/imbue/mngr/cli/test_rename.py` (with and without `-m tmux`)
- [x] `just test-quick libs/mngr/imbue/mngr/cli/rename_test.py`
- [x] `just test-quick libs/mngr/imbue/mngr/utils/test_ratchets.py`
- [x] `just test-quick "libs/mngr -m 'not modal and not docker and not docker_sdk and not acceptance and not release and not lima and not rsync'"` (4020 passed)
- [ ] CI acceptance/release tests via offload


joshalbrecht:
erm, I believe I need to be able to start hosts from the rename command (for performance reasons) in minds right now, so please leave that flag on there :)

(it's fine if it doesn't actually start the host, you'll note a gross `-l` flag on there for simultaneously changing labels as well)

other than that, lgtm

qi-imbue:
Added back `--start`, but now defaulting to false.

joshalbrecht:
lgtm, thanks!

PR #1661, opened 2026-05-15
Bundle Lima into minds desktop app and default to lima

weishi-imbue:
## Summary

- `apps/minds/scripts/build.js` now downloads the official Lima 2.1.1 release tarball (`bin/` + `libexec/` + `share/`) into `resources/lima/`, alongside `uv` and `git`.
- `apps/minds/electron/backend.js` adds `resources/lima/bin` to the `PATH` of the `mngr` subprocess. `mngr_lima`'s `shutil.which("limactl")` now resolves the bundled binary with no plugin changes -- `limactl` discovers `share/lima/` (guest agents, templates) relative to its own binary location, so the bundle is fully relocatable.
- The default compute provider flips from `LOCAL` to `LIMA` in the create form (no-account fallback), the form/API/creating-page parse fallbacks in `app.py`, and `AgentCreator.start_creation`'s API default. A fresh user on a pristine macOS install can accept the defaults and click Create without installing anything.
- On Apple Silicon, Lima uses the built-in `vz` backend (Virtualization.framework), so we do not bundle QEMU.

## Test plan

- [x] `pnpm build` produces `resources/lima/bin/limactl` reporting version 2.1.1.
- [x] `shutil.which("limactl")` resolves to the bundled binary under the same `PATH` we inject into the `mngr` subprocess.
- [x] `just test-quick apps/minds` -- 544 tests pass.
- [ ] Full CI run (`just test-offload` / acceptance).
- [ ] Smoke-test on a clean macOS install: create an agent with all defaults and confirm it succeeds without prerequisites.

## Out of scope

`LaunchMode.LOCAL` (maps to `--template docker`) still requires Docker, which is not bundled. This PR does not address that gap.


weishi-imbue:
close in favor of #1731

PR #1663, opened 2026-05-16
Fix mngr config help text: --user should be --scope user

weishi-imbue:
## Summary
- `mngr config --help` showed an example using `--user`, which is not a real option. The actual flag is `--scope user`.
- Fix the example in both the CLI help (libs/mngr/imbue/mngr/cli/config.py) and the generated docs page (libs/mngr/docs/commands/secondary/config.md) so copy/paste works.

## Test plan
- [ ] CI passes
- [ ] `uv run mngr config --help` shows the corrected example


joshalbrecht:
lgtm, thanks!

PR #1664, opened 2026-05-16
mngr_lima: disable guest -> host port forwarding

weishi-imbue:
## Summary

- The previous `portForwards: []` was a no-op: Lima always appends a fallback rule that forwards any guest TCP/UDP socket to host loopback. Two coexisting VMs both binding `0.0.0.0:8082` collided on host `localhost:8082` (first binder won).
- Replace it with two ignore rules -- one matching `guestIP: 0.0.0.0` (with `guestIPMustBeZero: true`) and one matching `guestIP: 127.0.0.1`. Empirical Lima 2.1.1 testing shows user-supplied rules match the bind address literally, so a single-rule version leaves the opposite bind class still leaking. Lock `portForwards` in `merge_lima_yaml` via `_LOCKED_KEYS` so a user `--file` YAML cannot re-enable forwarding. SSH is unaffected -- Lima manages it through a separate top-level `ssh:` block, outside `portForwards`.

## Empirical verification (Lima 2.1.1 on macOS)

Throwaway VMs created with three candidate `portForwards` configs; inside each, `python3 -m http.server` was bound on `0.0.0.0:8082` and `127.0.0.1:9090`; from the host: `curl http://127.0.0.1:8082/` and `:9090/`.

| Config | `0.0.0.0:8082` | `127.0.0.1:9090` | SSH (`limactl shell`) |
|---|---|---|---|
| `portForwards: []` (pre-PR baseline) | HTTP 200 (leak) | HTTP 200 (leak) | ok |
| One rule, `guestIP: 0.0.0.0` | refused | HTTP 200 (leak) | ok |
| One rule, `guestIP: 127.0.0.1` | HTTP 200 (leak) | refused | ok |
| **Two rules (this PR)** | **refused** | **refused** | **ok** |

The two-rule form matched no `Forwarding TCP` log lines in `ha.stderr.log` after the binds (vs. an explicit `Forwarding TCP from 127.0.0.1:9090 to 127.0.0.1:9090` in the single-rule-0.0.0.0 case). SSH continued to respond with `lima-mngr-pftest-fixed3` as the guest hostname, confirming the SSH channel is unaffected.

## Test plan

- [x] Unit tests: `just test-quick libs/mngr_lima` -- 109 passed
- [x] Empirical: throwaway Lima VMs (above table)
- [x] CI: green on the prior single-rule commit; this commit re-triggers it
- [ ] Operational note: existing `assistant-host` / `new-assistant-host` VMs still have the old (no-op) `portForwards: []` in their Lima state and must be recreated for the fix to apply


joshalbrecht:
lgtm, thanks weishi!

PR #1673, opened 2026-05-18
Granular permission requests

hynek-urban:
When reviewing, please read the accompanying [changes to the forever-claude-template skill](https://github.com/imbue-ai/forever-claude-template/pull/86) first because they will make this more obvious.

Change the permission request data model and let agents reason about permissions.

- Add new endpoints to the `permissions` extension:
    - `/permissions/self` (for agents to see their own permissions)
    - `/permissions/available/<service-name>` (to retrieve all the known available named permissions for a given service)
    - `/permissions/available/` (to return the whole catalog of known permissions, including display names - used by the Minds app)
- Adjust the `permission-requests` creation endpoint
    - Make it so that agents explicitly request specific new permissions that they need.
- Improve the separation of concerns between `mngr_latchkey` and `apps/minds` (Minds now do not need to know about the list of supported services or about the supported permissions.)
- In the permission dialog:
  - Do not pre-check `any` by default any more.
  - Pre-check those permissions that are either newly requested or that were already approved.
- Make some other semi-related fixes (e.g. fix a `latchkey forward` startup race condition).


joshalbrecht:
lgtm!

PR #1682, opened 2026-05-19
mngr_modal: fix sharding-dependent flake in test_snapshot_and_shutdown

weishi-imbue:
## Summary
- Switch `deployed_snapshot_function` in `libs/mngr_modal/imbue/mngr_modal/routes/test_snapshot_and_shutdown.py` from `scope=\"module\"` to function scope so each test in the file invokes the Modal SDK during its own per-test phase.
- Removes a latent flake where `Acceptance Tests` / `test-offload-acceptance` could fail with `Test marked with @pytest.mark.modal but never invoked modal` whenever offload's partition map placed a fixture-reuser test first in a sandbox.
- Trade-off: ~2-3 min extra on the affected file on CI (5 extra Modal deploys), in exchange for eliminating the flake.

## Background
Full root-cause writeup, evidence, and rejected alternatives in `specs/mngr-modal-snapshot-shutdown-flake/concise.md` on the originating branch (`lima-resil-v2`). Short version: `pytestmark = [pytest.mark.modal]` at module scope + a module-scoped fixture that calls `deploy_function(...)` + `resource_guards` per-test enforcement = only the first test in the file's offload sandbox actually credits the modal call; the others trip the guard.

## Test plan
- [x] `just test-quick \"libs/mngr_modal -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'\"` → 374 passed.
- [ ] `test-offload-acceptance` on this PR passes.
- [ ] Optional stress test on a side branch: add ~5-10 trivial unit tests elsewhere to perturb offload's collection count and confirm `test-offload-acceptance` still passes on this fix (would have ~50% chance of failing on the snapshot tests without it).
- [ ] Measure runtime delta on `test-offload-acceptance` vs baseline. Expected: +2-3 min on the affected file.


weishi-imbue:
Closing as redundant. Main already landed a better fix for this exact flake in 6ef5deca1 (May 17): a thin function-scoped wrapper around the module-scoped deploy that calls `modal.App.lookup(create_if_missing=False)` per test to credit the resource guard. That preserves the one-deploy-per-module cost while still satisfying the per-test `@pytest.mark.modal` enforcement check. This PR's Option-A (function-scope the deploy itself) would have multiplied deploy/cold-start cost ~5x with no additional correctness benefit.

PR #1683, opened 2026-05-19
Make latchkey encryption-key writes atomic to prevent partial reads

hynek-urban:
A concurrent caller could observe the on-disk key file between `O_EXCL` open and `os.write`, reading an empty string. Replace the write-in-place scheme with a write-temp + fsync + os.link publish, so the final path only ever appears with complete contents.


joshalbrecht:
lgtm

PR #1685, opened 2026-05-19
settings: extend CLI flags, fix --setting __extend through CommandDefaults, add narrowing guard

joshalbrecht:
## Summary

- CLI tuple/list flags (e.g. `--env`) now extend the merged settings value in **config-then-CLI** order. Was previously CLI-then-config, which silently put CLI values before the merged settings entries.
- `--setting commands.<cmd>.<param>__extend=...` now correctly extends the merged value stored inside `CommandDefaults.defaults` instead of falling through to `None` and acting as a plain assign.
- Added a new top-level `allow_settings_key_assignment_narrowing` setting (default `False`). When `False`, a higher-precedence settings layer that would silently drop entries from a non-empty list/tuple/dict/set raises `ConfigParseError`. The error tells users how to opt in (set the field to `True`) or keep the additive behavior for the specific key (use `__extend`), and warns that the default will flip to `True` in a future version.

## How the safety net behaves

- Empty overrides (`env = []`, `--setting commands.create.env=[]`) → treated as explicit clears, no error.
- Supersets (`__extend` results, additive assigns that include every prior entry) → no error.
- Non-empty assigns that drop at least one base entry → error, with the deepest dotted path of the loss.

Checked at every settings-file/env-var/`--setting` merge in `load_config` and `apply_settings_to_config`. Violations are collected across all layers and reported in a single combined error after the merge completes. CLI flag merging is unaffected (it always extends).

## Test plan

- [x] Added unit tests for the new ordering of `apply_config_defaults` (config-then-CLI).
- [x] Added tests for `--setting env=[] --env X=6` (clear + CLI) and `--setting env__extend=[...] --env X=6` (extend through CommandDefaults + CLI extend).
- [x] Added unit tests for `detect_settings_narrowing` (list/superset/empty-override/CommandDefaults recursion).
- [x] Added load_config integration tests for the narrowing guard (raises by default, allowed with opt-in, `__extend` workaround preserves entries).
- [x] Existing `test_apply_settings_to_config_replaces_existing_command_defaults` updated to use the opt-in; added a sibling test for the default-False raise path.
- [x] `just test-quick libs/mngr` (3903 passed, no failures) covering unit + integration tests.
- [ ] CI to exercise acceptance + release suites.


libs/mngr/docs/concepts/environment_variables.md
```diff
@@ -52,6 +52,23 @@ The same `__extend` suffix is recognised in TOML, in `--setting`, and in `mngr c
 
 `__extend` on a scalar field raises `ConfigParseError`. A shape mismatch (e.g. a string value used to extend a list field) also raises.
 
+### Assignment narrowing safety net
+
+Assign-by-default makes it easy to accidentally drop earlier entries when a higher-precedence layer writes the same key. For example, with `commands.create.env = ["X=4"]` in `.mngr/settings.toml` and `commands.create.env = ["X=5"]` in `.mngr/settings.local.toml`, the merged value silently becomes `["X=5"]` — the `X=4` from the project layer is lost.
```
evgunter, line 57: this seems like a bad example. why not `commands.create.env = ["X=4"]` and `commands.create.env = ["Y=5"]`? right now it actually looks like what you want is to override X

libs/mngr/imbue/mngr/cli/common_opts.py
```diff
@@ -418,9 +419,38 @@ def apply_settings_to_config(
...
+    return ConfigParseError(
+        "Settings narrowing detected: a --setting override would assign over a non-empty "
+        "list/tuple/dict/set value from the merged config, silently dropping the earlier "
+        "entries.\n" + "\n".join(detail_lines) + "\n"
+        "To opt into this assign-by-default behavior (and silence this error), set "
+        "`allow_settings_key_assignment_narrowing = true` in your settings.toml.\n"
+        "To keep the additive behavior for a specific key, switch to the `__extend` suffix on "
+        "the --setting key (e.g. `--setting commands.create.env__extend='[\"X=5\"]'`).\n"
+        "NOTE: the default for `allow_settings_key_assignment_narrowing` will change to True "
+        "in a future version, and support for False may be removed entirely."
```
evgunter, line 467: "support for false" is sounds like supporting the old behavior rather than supporting this error message

evgunter: (this same phrasing is used a few places)

evgunter:
some slightly unclear things in the docs but the change itself looks good!

joshalbrecht:
merged via https://github.com/imbue-ai/mngr/pull/1756

PR #1689, opened 2026-05-19
Minds VM test harness (Tart-based, macOS)

weishi-imbue:
## Summary

End-to-end test infra for the packaged `minds.app` Electron bundle, running inside fresh macOS VMs via Tart. Teammates run `apps/minds/scripts/vm-testing/run-test.sh <build-url-or-local-dmg-or-zip> minds-fresh` on their own Macs and get a junit.xml + summary.json verifying install → launch → first message end-to-end.

## What's in here

- **`apps/minds/scripts/vm-testing/build-persona.sh`** + `personas/minds-fresh.sh`: clones the cirruslabs vanilla macOS image, runs per-persona provisioning, saves it as a local Tart image.
- **`apps/minds/scripts/vm-testing/run-test.sh`**: host orchestrator. Accepts a `.dmg`/`.zip` URL, a local `.dmg`/`.zip`, or an unzipped `.app`. Boots a throwaway VM cloned from the persona, ships the harness via shared dir, scp's results back, tears the VM down. Supports `--keep-vm` for debugging, `TEMPLATE_GIT_REF` to pin the FCT clone to a specific commit, `ANTHROPIC_API_KEY` → `ai_provider=API_KEY` forwarding.
- **`apps/minds/scripts/vm-testing/harness/run-harness.py`**: in-VM Python harness. Wipes `~/.minds/`, untars the bundle into `/Applications/`, exec's the binary directly with `SKIP_AUTH=1`, polls the backend, creates an agent against `forever-claude-template`, sends a deterministic message via the bundled `mngr message`, and verifies the agent reply in `minds-events.jsonl`. Writes per-step `junit.xml` + `summary.json` + raw logs.
- **README** with one-time setup, usage, credentials story, future personas, and known gotchas.

## Design notes / non-obvious bits

- The bundle is shipped to the VM as a tarball, not laid out raw, because Tart's virtiofs misreports Electron framework symlinks as cyclic ("Too many levels of symbolic links"), breaking any in-VM copy.
- The VM uses uv (bootstrapped from the bundle tarball) to get a managed Python — the cirruslabs vanilla image has no Xcode CLT, so `/usr/bin/python3` would trigger a GUI install prompt.
- SSH options force password auth (`PubkeyAuthentication=no`, `PreferredAuthentications=password`) so sshpass actually gets a chance against `sshd`'s `MaxAuthTries` — without this, the host's loaded identities get tried first and blow the quota.
- Latchkey gets a deterministic encryption key because the fresh-VM keychain isn't unlocked.

## Test plan

- [x] `build-persona.sh minds-fresh` clones, provisions, and registers a Tart image.
- [x] `run-test.sh` accepts `.dmg`/`.zip` URLs and local `.dmg`s; mounts/extracts cleanly.
- [x] Harness wipes state, installs, launches, and reaches the backend (steps 1-4 pass).
- [ ] Full happy path (create_agent + send_message) against a known-compatible bundle/template pair — in progress; the harness correctly surfaces version-skew (today's FCT main is incompatible with the 0.2.28 bundle) as a structured create_agent failure.
- [ ] CI integration (out of scope for v1).
- [ ] Additional personas: `minds-vanilla-brew`, `minds-stale-path` (out of scope for v1).


weishi-imbue:
cannot test nested virtualization. use self hosted mac runner for testing

PR #1691, opened 2026-05-19
bump timeout on two flaky tests

evgunter:
## Summary
Add `@pytest.mark.timeout(60)` to two flaky multi-agent destroy tests in `libs/mngr/imbue/mngr/cli/test_destroy.py`:

- `test_destroy_via_stdin`
- `test_destroy_multiple_agents`

Both tests spin up two `sleep` agents, wait for tmux sessions, run a parallel destroy of both, and wait for cleanup — a workload that lands at or over the global 10s pytest-timeout under modal-offload contention. The first exhausted all 5 `@pytest.mark.flaky` retries on PR #1652 (commit `1c72e84d`, check run `76815140017`) where the timeout fired inside `destroy.__exit__` joining destroy strands. The multi-agent shape is the actual coverage value of both tests (piping multiple names through stdin is the main use case for `destroy -`), so the workload cannot be shrunk. Matches the established in-file precedent on `test_destroy_transfer_none_keeps_shared_worktree`.

## What I tried and discarded along the way
- **Shrinking `test_destroy_via_stdin` to a single agent.** Local benchmark (N=15 interleaved pairs) showed paired Δ +0.42s, 15/15 pairs faster. But the multi-agent shape is the use case being exercised, so this would have removed real coverage. Reverted.
- **`poll_interval=1.0` on `wait_for(tmux_session_exists, …)` calls.** Local benchmark (N=15) showed paired Δ +0.00s, 8/15 pairs faster — pure noise, because the polled condition is true on the first iteration so interval doesn't matter. Reverted.

## Test plan
- [ ] CI green on `test_destroy_via_stdin` and `test_destroy_multiple_agents` over a few runs (the 39 failing acceptance tests on this branch's CI are an unrelated Modal infra issue — `User joshalbrecht does not have write access` — affecting every PR currently).

## Follow-up (out of scope here)
The architecture review noted a deeper fix: cut tmux-subprocess churn in `wait_for(tmux_session_exists, …)` (or replace polling with a `tmux wait-for` style call). That would benefit the whole family of tmux integration tests, not just these two.


joshalbrecht:
lgtm

PR #1692, opened 2026-05-19
Add tmr key

weishi-imbue:
Adds my SSH public key to `.github/tmr-authorized-keys` so I can debug TMR CI's modal agents (per the instructions in that file). Mirrors #1688.


joshalbrecht:
lgtm

PR #1694, opened 2026-05-19
mngr create: scope --reuse agent matching to the address's host

weishi-imbue:
## Behavior change

`mngr create <agent>@<host>.<provider> --reuse` now scopes the reuse lookup to the host named in the address.

**Previously**, when the target host was being newly provisioned (`--new-host`, or the `NAME@.PROVIDER` form), the reuse lookup dropped the host scope and matched the agent name across *every* host on the provider. With a `system-services` agent already on `myhost1`, running

    mngr create system-services@myhost2.lima --reuse --new-host

would find and adopt `myhost1`'s agent — `myhost2` was never created, and with `--update` the new work tree was pushed onto `myhost1`.

**Now** the lookup is scoped to `myhost2`: no host named `myhost2` exists, so nothing matches and `myhost2` is provisioned fresh with its own `system-services`; `myhost1` is untouched. If a host named `myhost2` already existed, its `system-services` would be reused.

Bare `--reuse` with no host in the address is unchanged — it still matches any host.

## Empirical verification

Run on lima (a real multi-host provider). `worker` is a `command`-type agent; the source is a plain directory (rsync transfer).

| # | command | host state | result |
|---|---|---|---|
| 1 | `worker@h1.lima --new-host` | `h1` fresh | host `h1` created with `worker` |
| 2 | `worker --reuse` | `h1` has `worker` | reused `worker@h1` — bare name matches any host |
| 3 | `worker@h1.lima --reuse` | `h1` has `worker` | reused `worker@h1` |
| 4 | `worker@h1.lima --reuse --new-host` | `h1` has `worker` | reused `worker@h1` — host exists, so `--new-host` is moot |
| 5 | `worker@h2.lima --reuse --update --new-host` | `h2` fresh | no reuse match → host `h2` created with its own `worker`; `h1` untouched |
| 6 | `worker@h1.lima --reuse --update --new-host` | `h1` has `worker` | reused `worker@h1`, updated in place |
| 7 | `worker --update` | — | rejected: `--update requires --reuse` |

Rows 5 and 6 use identical flags (`--reuse --update --new-host`) and differ only by whether the addressed host already exists — the host-scoping fix is what makes them diverge correctly: a fresh host name provisions a new host, an existing one reuses the agent there.

Note: row 6 was run with a plain-directory source (rsync). With a git-repo source, `--reuse --update`'s re-transfer currently fails on a pre-existing `git-mirror` / `git push --prune` issue (`hosts/host.py`, not touched by this PR).


libs/mngr/imbue/mngr/cli/create.py
```diff
@@ -1154,15 +1189,22 @@ def _parse_project_name(
...
     agent_name: AgentName,
     provider_name: ProviderInstanceName | None,
-    target_host_ref: DiscoveredHost | None,
+    address_host: HostName | HostId | None,
     mngr_ctx: MngrContext,
     agent_and_host_loader: Callable[[], dict[DiscoveredHost, list[DiscoveredAgent]]],
 ) -> tuple[AgentInterface, OnlineHostInterface] | None:
     """Try to find and start an existing agent with the given name.
 
-    Searches for an agent matching the name, scoped by provider and host if specified.
```
weishi-imbue, line 1169: When creating new, target_host_ref will be None, and this would search every host. 

Bug walk through
```
- system-services already exists on host mindtest-4 (lima).
- User runs mngr create system-services@mindtest-5.lima --reuse --update --new-host.
- --new-host → _parse_target_host returns NewHostOptions (mindtest-5 doesn't exist yet).
- Call site: target_host is NewHostOptions, not DiscoveredHost → target_host_ref = None.
- In _try_reuse_existing_agent: host filter disabled; provider filter keeps lima hosts.
- It finds system-services on mindtest-4 and reuses it.
- --update then stops that agent and pushes the new worktree onto mindtest-4 → refusing to update checked out branch.
```

qi-imbue:
TIL `--reuse` existed...

IMO we should simply disallow `--reuse --new-host`, it makes no sense.

- If you want to reuse an existing agent, you are not creating a new host.
- If you are creating an agent on a new host, then the agent is by definition new, and you're not going to use any existing agents.

We should also disallow `--reuse` when the agent name is missing because that's another nonsensical combination.

weishi-imbue:
closing as similar solution #2085 got in.

PR #1644, opened 2026-05-14
minds: track creation phases on the creating page

gnguralnick:
## Summary

The "Creating your project" page no longer pins "Cloning repository..." for the whole flow. Phase state is now carried on the existing `AgentCreationStatus` enum as the single source of truth -- the SSE stream resolves the spinner caption from the current enum value, which it polls on each loop iteration. The caption advances as setup progresses ("Starting...", "Cloning repository...", "Checking out branch '<x>'...", "Provisioning AI access...", "Creating workspace...", "Waiting for workspace to be ready...").

For IMBUE_CLOUD the first/last captions read "Connecting to host..." / "Setting up agent..." to match the existing server-rendered initial copy. The `/api/create-agent/{id}/status` JSON API now returns the new enum values (`INITIALIZING`, `CLONING_REPO`, `CHECKING_OUT_BRANCH`, `PROVISIONING_AI`, `CREATING_WORKSPACE`, `WAITING_FOR_READY`, `DONE`, `FAILED`) instead of the previous `CLONING` / `CREATING`.

## Test plan
- [x] `just test-quick apps/minds -m 'not docker and not docker_sdk and not modal and not tmux and not acceptance and not release'`
- [x] `uv run ruff check`
- [ ] Manual: kick off a real `minds run` agent creation and watch the spinner caption advance through the phases.

PR #1697, opened 2026-05-20
minds-start: scrub ambient ANTHROPIC_API_KEY / ANTHROPIC_BASE_URL

gnguralnick:
## What

`just minds-start` now runs `unset ANTHROPIC_API_KEY ANTHROPIC_BASE_URL` after sourcing `.env` and before launching the desktop client.

## Why

In dev, whoever runs `just minds-start` typically has `ANTHROPIC_API_KEY` (and sometimes `ANTHROPIC_BASE_URL`) exported by their shell rc. The desktop client inherits them and forwards them into every agent it creates, silently overriding the auth mode picked in the create form.

A real packaged macOS app launched from Finder never sources shell rc files, so this leak is purely a dev-environment artifact. The `minds-start` recipe is the right boundary to fix it -- no app-level changes needed.

This replaces the heavier create-form UI approach that was originally on the `gabriel/creation-status` branch (PR #1644); that env-anthropic UI has been dropped there.

### DIFF
--- a/changelog/gabriel-minds-start-unset-anthropic.md
+++ b/changelog/gabriel-minds-start-unset-anthropic.md
@@ -1,0 +1,1 @@
+- `just minds-start` now unsets `ANTHROPIC_API_KEY` and `ANTHROPIC_BASE_URL` before launching the desktop client, so credentials exported in the developer's shell no longer leak into agents created by the dev app.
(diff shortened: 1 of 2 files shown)


joshalbrecht:
lgtm, thanks for fixing!

PR #1698, opened 2026-05-20
File sharing

hynek-urban:
Support file and directory sharing between a Mind agent and the user's local computer.

(Accompanied by https://github.com/imbue-ai/forever-claude-template/pull/106.)

- Add the `minds_api_proxy.mjs` Latchkey Gateway extension which proxies requests to the Minds API.
- Update the `permission_requests.mjs` extension
  - Support different permission request types (`predefined` for the predefined Detent service permissions, `file-sharing` for the new file-sharing permission requests).
  - When saving a permission requests, save also the `target` (pointer to the permissions.json file to be updated upon approval` and the `effect` (actual changes that should be done to the permissions.json file).
  - Implement the `/permission-requests/approve/<request_id>` endpoint which internally uses the above.
- Add the new `FileSharingPermissionRequestHandler` and related machinery in the Minds app (mostly to show the file sharing permission dialog).
- Add a new endpoint in the Minds API (`/files`), implemented by the `wsgidav` library.
- Fixed a bug where `$MINDS_API_KEY` wasn't propagated to agents (only to the "default" one which isn't a real agent).
    - For simplicity, done just through --host-env, even though it's not entirely correct. We'll remove `$MINDS_API_KEY` altogether soon, though.


joshalbrecht:
lgtm

PR #1703, opened 2026-05-20
mngr create: reject --reuse + --new-host as incompatible

weishi-imbue:
## Summary

`mngr create` currently accepts `--reuse` together with `--new-host`, but the two contradict each other by construction:

- `--new-host` always provisions a brand-new host.
- `--reuse` looks up an *existing* agent on an *existing* host and reuses it.

A freshly provisioned host has no agents, so "reuse" there is meaningless. Today the combination is silently accepted. This PR makes it an explicit user error.

## Changes

- **`libs/mngr/imbue/mngr/cli/create.py`**: Add a sibling validator next to the existing `--update requires --reuse` check that raises `UserInputError` when both `--reuse` and `--new-host` are set. The `--update` validator already runs first, so `--reuse --update --new-host` is also caught here. Updated `--help` text on both flags to call out the incompatibility (CLI markdown docs regenerated via pre-commit hook).
- **`apps/minds/imbue/minds/desktop_client/agent_creator.py`**: Remove `--reuse` and `--update` from `_build_mngr_create_command` for all launch modes. Every minds launch mode (LOCAL, LIMA, CLOUD, IMBUE_CLOUD) passes `--new-host`, so this would otherwise break every minds workspace creation. This is safe because:
  - LOCAL/LIMA/CLOUD always provision a fresh host; there is never an existing agent to reuse.
  - IMBUE_CLOUD: the baked `system-services` agent name collision is already handled by mngr's duplicate-name check exempting `host.pre_baked_agent_id` in `libs/mngr/imbue/mngr/api/create.py`, so `--reuse` plays no role.
- **Tests**: Add CLI tests asserting `--reuse --new-host` and `--reuse --update --new-host` are rejected. Update the minds tests that previously asserted `--reuse`/`--update` were present in the built command.
- **Changelog**: `changelog/mngr-incompat-options.md`.

## Related

PR #1694 (branch `mngr/reuse-host-scoped`) independently scopes `--reuse`'s agent lookup to the host named in the address. That is a different approach to the same underlying bug and is not depended on here; the two touch `cli/create.py` in different functions, so any merge conflict will be trivial.


weishi-imbue:
imbue_cloud provider may still need these two for now. --reuse -> reuse existing physical host. --new-host -> means the host is new to the user.

PR #1705, opened 2026-05-20
add leak check to more provider release tests

evgunter:
Deprecated in favour of #2551 

there aren't actually aws and gcp credentials available to the ci job, so those are dead for now

---

## Summary

- Adds a tag-based safety net in `libs/mngr_vultr/imbue/mngr_vultr/conftest.py` that detects and destroys any Vultr VPSes leaked by tests at session end, modeled on the Modal leak detector in `libs/mngr_modal/imbue/mngr_modal/conftest.py`.
- `pytest_configure` injects `mngr-vultr-test-session=<uuid>` into `MNGR_VPS_EXTRA_TAGS`. `build_vps_tags` in `mngr_vps_docker` already reads this, so the tag rides through to every VPS the `mngr` CLI subprocess creates.
- `pytest_sessionfinish` lists Vultr instances, filters to survivors bearing the session tag, destroys them, and sets `session.exitstatus = TESTS_FAILED` so CI surfaces the bug rather than silently absorbing the cleanup.
- Skipped silently when `VULTR_API_KEY` is unset (unit-only runs).

## Motivation

Vultr release tests create real VPSes (~\$0.03/hr each). Each test wraps create in `try`/`finally`, but there are real leak windows:
1. Assertion fails between `result.returncode == 0` and the `try:` block.
2. Pytest timeout kills the test before `finally` runs.
3. The destroy call itself fails silently.

A subagent's audit flagged this gap (vs. the Modal-style session-end scanner). This PR closes it.

## Out of scope

- An opt-in `MNGR_VULTR_RELEASE_TESTS=1` flag — `VULTR_API_KEY` + `pytest.mark.release` already gate these tests, matching the OVH/Modal pattern.
- A create-time guard analogous to Modal's `_create_environment` `PYTEST_CURRENT_TEST` check — would require updating existing `client_test.py` unit tests that use `label="test"`. Easy follow-up if desired.

## Test plan

- [x] `just test-quick "libs/mngr_vultr -m 'not release and not acceptance'"` — 88 passed
- [x] `just test-quick libs/mngr_vultr/imbue/mngr_vultr/test_ratchets.py` — 56 passed (incl. type + ruff)
- [ ] Real Vultr release run via `just test-offload-release` to verify the hook doesn't fire spuriously


joshalbrecht:
I'm skeptical here: what is preventing this from nuking all vultr instances in an account?

I don't see any filtering at all...

evgunter:
the filtering is with `_SESSION_TAG`

joshalbrecht:
this seems like an entirely different way of checking for and destroying leased hosts though?

like, we already know how to do this correctly from modal (where it uses date based tags)

without date based tags, we cannot know which hosts were leftover from very old, killed tests, and thus cannot clean them up, and that seems bad?

(this same kind of cleanup logic applies to all new cloud providers as well)

DanverImbue:
Closing in favour of #2551

PR #1707, opened 2026-05-20
Tiered workspace restart and recovery flow for minds

gnguralnick:
## Summary

Reworks the minds workspace-recovery flow into a tiered restart built on clean `mngr` operations instead of poking tmux, and makes the whole flow resilient to a workspace whose container is stopped or whose sshd has died.

> **Stacked on #1775** (`mngr/tolerate-discovery-ssh-failures`). That PR makes provider discovery tolerate per-host SSH failures and preserve an SSH-dead container's agents via the offline fallback; this branch builds on it and merges it in, so its three source files (`interfaces/provider_instance.py` + tests, `mock_provider_test.py`) show up in this diff until #1775 lands. Review/merge #1775 first.

### Recovery flow (`apps/minds`)

- When a workspace's system interface stops responding, minds shows a recovery page. While it checks host health or a restart is in flight, it shows a single "Loading workspace" state and refreshes itself until the workspace is back. The loader is visually consistent with the forwarding plugin's own "Loading workspace" page, so the two pages a user may see during recovery look like one.
- The page picks its **dispatch tier** from the workspace host's state and recovers with no clicks where it safely can. Each tier names the *condition* (what is wrong with the workspace); the page action is a consequence of the condition:
  - **`INTERFACE_UNRESPONSIVE`** — the container is running and reachable, but the system interface is stuck. Gets an in-place `system-services` restart that does not interrupt the user's agents (auto-dispatched).
  - **`HOST_OFFLINE`** — the container is fully stopped, so it is cold-booted immediately (nothing is running, so nothing is interrupted). The auto-dispatched host restart skips the redundant `mngr stop --stop-host` and boots straight away.
  - **`HOST_UNRESPONSIVE`** — the container claims to be running but is unreachable (the common case is SSH transport down to a RUNNING host; this is also the fallback for any other ambiguous host state). Falls back to a confirmed "Restart workspace" button, since bouncing a live container requires explicit consent — no auto-dispatch.
  - **`WORKSPACE_MISCONFIGURED`** — `services.toml` is missing `[services.system_interface]` (the one case no restart can fix). The page explains that a restart will not help and offers only a "Try restart anyway" affordance.
- **Restart mechanism:** the in-place ("surgical") restart cleanly `mngr stop` + `mngr start`s the `system-services` agent (replacing the old `tmux kill-window` + `touch services.toml`); the host restart bounces the whole container via `mngr stop --stop-host` + `mngr start`. Recovery no longer depends on the bootstrap watch loop already running, and the user's claude agent is left untouched.
- **Restart-failed handling:** the pre-restart prompt and the post-failure state are one identical "Workspace unresponsive" page (heading, body, "Restart workspace" button, plus a collapsed, wrapping error detail shown only when a restart actually failed). The failed state shows the failure reason *and* the current diagnostics together, polls in the background, and returns the user to the workspace automatically the moment the background health probe recovers it — no more dead-end.
- **Tier-aware wait budget:** an in-place restart waits 15s; a host restart (full cold boot) waits 30s before declaring failure, so a still-booting container is no longer prematurely bounced to the unresponsive page.
- **Recovery diagnostics:** the host-health endpoint runs a batched in-container probe over `mngr exec` (5s hard ceiling, `--no-start`/`--quiet` so it can never start a stopped host) covering `tmux ls`, the `services.toml` declaration parse, and `ss`/`curl` on the system-interface inner port, plus a scoped `mngr list` and a plugin resolver-snapshot read. Results are returned as a flat `probes: [{question, command, output, answer}]` list plus the derived `dispatch_tier`, and rendered as check/x/? rows with expanders showing the exact command and raw output. The `mngr list` is scoped to this workspace's agents (CEL `id == ...`) and runs `--on-error continue`, so an unrelated wedged host on the same provider does not blank out the diagnostics. A page-level "Copy diagnostics" button captures everything. Probes only run on recovery-page load (RESTARTING refreshes skip them); healthy operation generates no new probe traffic.
- **Stuck detection is fully probe-driven:** a workspace is only treated as stuck after a sustained run of failed HTTP probes, so a single transient hiccup (e.g. a recycled SSE stream) starts active probing instead of triggering recovery. The page no longer flashes for a healthy workspace.
- **No more eviction to the landing page** on transient discovery loss: Electron navigates to landing only when a workspace was *explicitly* destroyed. The chrome SSE payload now carries `destroying_agent_ids`, and the client remembers which agent ids it has seen destroying; a workspace that merely disappears from the live list is handled by the recovery flow with no nav.
- **SSH-dead restart addressing:** minds caches the workspace-agent → system-services-agent mapping to `system_services_agent_cache.json` and falls back to it when live discovery has lost the pair, so a restart can still address the system-services agent when the provider can't currently enumerate agents.
- **Loopback + Host header:** the readiness/health probes and the refresh-service broadcast now dial the plugin over loopback and carry the agent's `agent-<hex>.localhost` vhost in the `Host` header instead of in the URL, making them independent of `*.localhost` name resolution (not available on every host).
- **Manual affordances:** a "Restart workspace…" entry (with confirmation) in the Electron sidebar context menu, and a per-row restart button on the home page. Opening a stopped-container workspace now routes to the recovery page instead of flashing a raw error.
- **Health tracker:** a new `RESTART_FAILED` state (carrying the reason), an on-recovery callback fired on every non-HEALTHY → HEALTHY transition (wired to a compact loguru INFO line), and a fix for a restart race where the background probe loop could prematurely declare recovery by probing the still-alive pre-restart interface — the loop now skips RESTARTING agents.
- **Backend-failure policy now lives in minds, not the plugin** (see the forwarding-plugin section below). The plugin reports *every* non-2xx response; minds' new `should_enroll_suspect_for_backend_failure` decides which ones enroll a probe suspect — connection-level failures and infrastructure 5xx (502/503/504) only. Application errors (app 500s, ordinary 4xx) are ignored on the failure-envelope path and left for the background probe to adjudicate. This preserves the prior end-to-end behavior while moving the decision out of the (now app-agnostic) plugin.

### `mngr stop --stop-host` (`libs/mngr`)

- New flag that stops an agent's whole host (every agent on it) instead of just the named agent, via the provider `stop_host` API (no snapshot). For container-backed providers it stops the container while the machine keeps running. Validates `supports_shutdown_hosts` up front and rejects combination with `--archive`.
- Idempotent: an already-offline host reports success instead of raising, so restarting an already-stopped workspace works.
- Resolves the target host **without SSH** (from the discovery event stream, cross-checked against an SSH-free host scan), so it works in exactly the situation it exists for — sshd crashed or PID-exhausted while the container is still running.
- Also fixes `mngr list --on-error continue --format json` crashing with `TypeError: ... ProviderErrorInfo is not JSON serializable` on the empty-agents path.

### Forwarding plugin (`libs/mngr_forward`)

This branch also makes the forwarding plugin a **dumber, application-agnostic utility**: it forwards bytes and reports what it observes, and leaves all interpretation/policy to whatever application consumes its envelope stream.

- A stopped/unreachable workspace backend is now reported as a backend failure (`CONNECT_ERROR` `system_interface_backend_failure` envelope + the styled "Loading workspace" loader) instead of a raw error, so a consumer can drive its own recovery UI. Covers both SSH-tunnel setup failure and a refused host-loopback dial.
- **Simplified HTTP error handling.** Instead of deciding which HTTP statuses matter, the plugin now forwards every response unchanged and emits a single `ERROR_RESPONSE` reason (carrying `status_code`) for *any* non-2xx response. This removes the prior special-casing — `FIVEXX_RESPONSE` (a 502/503/504 allowlist) and `NOT_FOUND_RESPONSE` (a 404-on-`GET` "is the real system interface answering?" SPA heuristic) — along with the `_INFRASTRUCTURE_5XX_STATUSES` set and the branching in the forward path. The status-interpretation policy moves entirely to the consumer (minds; see the recovery-flow section). This intentionally drops the GET-404 "wrong process on the port" hint: the failure envelope carries no HTTP method, and adding one would turn a health signal into a partial access log — a genuinely-wrong backend is still caught by minds' background probe.
- New `resolver_snapshot` envelope: the plugin emits the full per-agent service map on every mutation — `update_services` and the destruction paths (`remove_known_agent` / `update_known_agents` when they drop an agent that had services) — so consumer mirrors don't retain stale entries. Minds mirrors the latest snapshot and uses it in recovery diagnostics. Version-skew safe in both directions.
- **Decoupling from minds.** The plugin should not know about minds in any meaningful way, so this branch removes the minds-specific framing earlier work had baked into its docstrings and comments, replacing "minds-side health tracker / minds chrome / recovery page" with neutral "downstream consumer" language. This reverts/alters wording introduced by **#1589** ("Workspace-server restart and health-recovery UX on mngr_forward" — the original restart that bounced only the system-interface/workspace-server process, plus its health-check), and generalizes a few residual references from the original plugin-creation commit. No behavior changes from this decoupling; only the framing (and the now-relocated status policy) differ. Factual "Adapted from `minds.desktop_client.X`" provenance attributions are kept.

## Plan / design docs

- `blueprint/tiered-restart-v2/plan-tiered-restart-v2.md` — the two-tier restart design.
- `blueprint/recovery-diagnostics/plan-recovery-diagnostics.md` — the diagnostics-probe design.

## Test plan

New/updated tests land alongside each area: `recovery_probe_test.py`, `backend_resolver_test.py`, `system_interface_health_test.py`, `templates_test.py`, `test_desktop_client.py` (minds); `cli/stop_test.py`, `api/discovery_events_test.py` (mngr); `data_types_test.py`, `envelope_test.py`, `resolver_test.py`, `server_test.py` (mngr_forward).

- [x] `test-offload` (full unit + integration suite) — green (re-runs in CI for the latest commits)
- [x] `test-offload-acceptance` — green (re-runs in CI for the latest commits)
- [x] `test_no_type_errors` for the touched projects — green
- [ ] `test-docker-electron` — `test_create_local_docker_workspace_via_electron` is currently timing out waiting for a freshly-created workspace's dockview to render (a heavy, historically-flaky e2e on the healthy-creation path, unrelated to this PR's recovery changes). Tracking separately; not a regression from this branch.

The most recent commits (the `mngr_forward` error-handling simplification, the minds-side status filter, and the minds-decoupling docstring/comment edits) were validated locally with the focused `mngr_forward` and minds `desktop_client` test suites (mngr_forward 229, minds health/forward-cli 66), `ruff`, `pyright`, and the repo's code-review + conversation-review gates; the full `test-offload` / acceptance suites re-run in CI on this PR.

Interactive recovery-page behavior was verified manually in the running Electron app (in-place / host / unresponsive / misconfigured tiers, restart-failed auto-recovery); per repo convention those checks are not crystallized into pytest.


apps/minds/imbue/minds/desktop_client/agent_creator.py
```diff
@@ -69,6 +70,17 @@
...
+# Path the workspace-readiness / health probes hit through the plugin. We probe
+# a real system-interface API route rather than ``/`` so a 200 is an *identity*
+# check, not merely "something is answering on the port". The SI serves its SPA
+# index for every unmatched GET (a ``/{path:path}`` catch-all), so ``/`` returns
+# 200 even from an unrelated process (e.g. a static file server) that happens to
+# hold the inner port. ``/api/agents`` is a core SI endpoint that returns 200
+# JSON from the real system interface but 404 from anything that does not
+# implement it, so probing it lets the health loop distinguish the real backend
+# from a port squatter / wrong process.
+_WORKSPACE_PROBE_PATH: Final[str] = "/api/agents"
```
joshalbrecht: I'm a little skeptical of this--ultimately, we don't know anything about what is running inside the workspace, and we shouldn't!

Why not just check "/"?

apps/minds/imbue/minds/desktop_client/app.py
```diff
@@ -1711,129 +1777,426 @@ def _sanitize_recovery_return_to(raw: str) -> str:
...
+            is_checked_after=False,
+            env=env,
+        )
+    except (OSError, RuntimeError, subprocess.TimeoutExpired, ConcurrencyGroupError) as exc:
+        # OSError covers fork/exec failures, RuntimeError the executor itself,
+        # ConcurrencyGroupError the strand-level failures the group raises, and
+        # TimeoutExpired any internal wait that surfaces it.
+        if log_failures:
+            logger.warning("mngr command {} failed: {}", argv, exc)
+        return None, str(exc)
```
joshalbrecht: highly suspicious

these seem likely to be real errors and failures that we should be alerted about, and seem to be getting suppressed as just warnings, which isn't good

in general we need to audit the other places that claude does this, but yeah, we definitely should not be making a helper that just masks real errors!

apps/minds/imbue/minds/desktop_client/app.py
```diff
@@ -1711,129 +1777,426 @@ def _sanitize_recovery_return_to(raw: str) -> str:
...
+        if log_failures:
+            logger.warning("mngr command {} exited {}: {}", argv, finished.returncode, finished.stderr)
+        return finished, f"exited {finished.returncode}: {finished.stderr.strip()}"
+    return finished, None
+
+
+def _run_mngr_command(concurrency_group: ConcurrencyGroup, argv: list[str], env: dict[str, str]) -> str | None:
+    """Run an ``mngr`` subprocess to completion; return an error message, or None on success."""
+    _finished, failure_reason = _run_mngr_subprocess(concurrency_group, argv, env)
+    return failure_reason
```
joshalbrecht: this function looks dumb too

in fact, a lot of these mngr interface functions look bad...

apps/minds/imbue/minds/desktop_client/backend_resolver.py
```diff
@@ -258,6 +280,76 @@ def parse_service_log_records(text: str) -> list[ServiceLogRecord | ServiceDereg
...
+        return {}
+    except OSError as exc:
+        logger.warning("Could not read services-agent-id cache from {}: {}", path, exc)
+        return {}
+    try:
+        payload = json.loads(raw)
+    except json.JSONDecodeError as exc:
+        logger.warning("Services-agent-id cache at {} is not valid JSON: {}", path, exc)
+        return {}
+    if not isinstance(payload, dict):
```
joshalbrecht: I want a ratchet against logger.warning--it's really using it to suppress a bunch of stuff that should not be suppressed

Who the heck is going to see such a warning? No one! This needs to go via a different error mechanism (which we have already IIRC)

apps/minds/imbue/minds/desktop_client/backend_resolver.py
```diff
@@ -258,6 +280,76 @@ def parse_service_log_records(text: str) -> list[ServiceLogRecord | ServiceDereg
...
+    the discovery thread over a transient I/O error would be worse than
+    losing one update.
+    """
+    try:
+        path.parent.mkdir(parents=True, exist_ok=True)
+        tmp = path.with_suffix(path.suffix + ".tmp")
+        tmp.write_text(json.dumps(cache, indent=2, sort_keys=True), encoding="utf-8")
+        tmp.replace(path)
+    except OSError as exc:
+        logger.warning("Could not write services-agent-id cache to {}: {}", path, exc)
```
joshalbrecht: I'm skeptical about this whole agent id cache thing anyway... can we get rid of this?

Also another logger.warning--I'll stop commenting on them, but we should address them all

libs/mngr/imbue/mngr/api/discovery_events.py
```diff
@@ -592,20 +598,71 @@ def find_latest_full_snapshot_offset(events_path: Path) -> int:
...
+    name_by_agent_id: dict[str, str] = Field(default_factory=dict)
+    # agent_id -> host_id
+    host_id_by_agent_id: dict[str, str] = Field(default_factory=dict)
+    # host_id -> host_name
+    host_name_by_host_id: dict[str, str] = Field(default_factory=dict)
+    # agent ids known to be destroyed
+    destroyed_agent_ids: set[str] = Field(default_factory=set)
+    # host ids known to be destroyed
+    destroyed_host_ids: set[str] = Field(default_factory=set)
+
```
joshalbrecht, line 623: I don't really understand what this code is / why it was necessary

libs/mngr/imbue/mngr/cli/list.py
```diff
@@ -712,7 +712,13 @@ def _run_list_iteration(params: _ListIterationParams, ctx: click.Context) -> Non
...
             write_human_line("No agents found")
         elif params.output_opts.output_format == OutputFormat.JSON:
-            emit_final_json({"agents": [], "errors": result.errors})
+            # Route through `_emit_json_output` so errors get pydantic-dumped to
+            # JSON-friendly dicts; passing raw ErrorInfo objects to
+            # `json.dumps` crashes with "Object of type ProviderErrorInfo is
+            # not JSON serializable", which is what made `mngr list --on-error
+            # continue --format json` hard-fail whenever discovery hit a
+            # broken host.
+            _emit_json_output([], result.errors)
```
joshalbrecht, line 724: wait, this seems really weird--why did this change?

libs/mngr/imbue/mngr/cli/stop.py
```diff
@@ -45,12 +52,75 @@ class StopCliOptions(CommonCliOptions):
...
+                # filesystem anyway. No discovery-event emission: the host is
+                # offline afterwards, so an online-host sweep would fail.
+                provider.stop_host(online_host, create_snapshot=False)
+                _output(f"Stopped host: {resolved.host_name}", output_opts)
+            case HostInterface():
+                # The host is already offline -- the desired end state is
+                # already reached, so this is an idempotent no-op.
+                _output(f"Host '{resolved.host_name}' is already stopped", output_opts)
+            case _ as unreachable:
+                assert_never(unreachable)
```
joshalbrecht: should almost certainly be done in parallel / in a concurrency group

apps/minds/imbue/minds/desktop_client/agent_creator.py
gnguralnick: (confirmed in person) - we want to remove this coupling. i switched to this so that we could recover from something taking over the system interface's declared port and causing issues, but we decided that that's not a scenario we need the app to auto-recover from. will update to just /

apps/minds/imbue/minds/desktop_client/app.py
gnguralnick: agreed. i think claude may have overabstracted. i'll audit all the usages of this and switch to more precise error handling rather than the general catch-all. depending on the mngr command you're running you may expect (and want to handle) particular types of errors; doesn't make sense to operate generally.

apps/minds/imbue/minds/desktop_client/backend_resolver.py
gnguralnick: agree about the logger.warning. the point of the cache here was that it's necessary for the app to know the ID of the system-services agent (as opposed to other agents running in the container), but if the container is in a failure state or actively restarting then the methods to query for that agent by name might not work. so the goal was to just keep track of this ID in memory so it could be referenced later.

libs/mngr/imbue/mngr/api/discovery_events.py
gnguralnick: here's the process that this code was fixing:

1. i have the minds app open to a workspace
2. in a separate terminal that's connected to the container, i do something that totally breaks the connection and would require a restart to fix (e.g. `pkill sshd`)
3. the minds app redirects to the home landing page for a moment before redirecting to the recovery/loading page

the reason is that the app suddenly thinks (for a moment) that the workspace was destroyed; the logic in this scenario is to redirect to the home page. so the fix here is to keep track of which workspaces we actually have destroyed so that in this failure case we know that there's something else going on and we should just go to loading/recovery pages directly.

but i can see if there's a cleaner way to sidestep that destroy check logic.

libs/mngr/imbue/mngr/cli/stop.py
gnguralnick: sounds good. i can add

libs/mngr_forward/imbue/mngr_forward/server.py
```diff
@@ -409,6 +410,27 @@ async def _stream() -> AsyncGenerator[bytes, None]:
...
+            agent_id,
+            SystemInterfaceBackendFailureReason.NOT_FOUND_RESPONSE,
+            backend_response.status_code,
+        )
+    else:
+        # Every other response (normal 2xx/3xx, an application-layer 4xx other
+        # than a GET 404, a non-infrastructure 5xx like a 500 stack trace) is a
+        # legitimate backend reply, not a health signal: forward it as-is and
+        # emit no failure envelope.
+        pass
```
gnguralnick: we should just log all responses, not try to differentiate between good and bad. maybe config option?

gnguralnick: this should allow us to make mngr forward dumber - less coupling and parsing needs to happen here

libs/mngr_forward/imbue/mngr_forward/data_types.py
```diff
@@ -21,19 +21,26 @@ class SystemInterfaceBackendFailureReason(UpperCaseStrEnum):
...
       bytes had already been delivered. Despite the name (motivated by
       the SSE forwarding path that originally surfaced this), it also
       covers non-SSE mid-response read failures.
     - ``FIVEXX_RESPONSE``: the backend returned a 502/503/504. Other 5xx
       codes (e.g. application-layer 500s) are *not* tagged as failures.
+    - ``NOT_FOUND_RESPONSE``: the backend answered a ``GET`` with a 404. The
+      real system interface serves its SPA index for every unmatched ``GET``
+      (a catch-all route), so it never 404s a page/route load -- a 404 reaching
+      the proxy means whatever is answering on the port is not behaving like
+      the system interface (e.g. a different process has bound the port).
```
gnguralnick: mentioned in later comment - mngr_forward shouldn't have any particular HTTP failure handling besides CONNECT_ERROR - tthese response values are currently just used to decide whether to emit; we should just always emit instead (or do much simpler filtering at emit time based on status code)

libs/mngr_forward/changelog/gabriel-workspace-restart.md
```diff
@@ -0,0 +1,38 @@
...
+  startup. No periodic flushes, no debouncing, no initial empty
+  emission; the first envelope is sent on the first real services
+  event.
+- The plugin now also treats a 404 on a proxied `GET` as a backend
+  failure (`NOT_FOUND_RESPONSE`). The system interface serves its SPA
+  index for every unmatched `GET`, so it never 404s a page/route load --
+  a 404 reaching the proxy means whatever holds the inner port is not
+  the system interface (e.g. a different process has bound the port).
+  This enrolls the agent as a probe suspect so the minds health loop can
+  confirm and recover, instead of letting a wrong-process responder look
```
gnguralnick: in general we should be trying to make mngr forward less knoweledgeable about minds - functionality here should be justified in a purer sense rather than constantly in reference to minds

libs/mngr_forward/imbue/mngr_forward/data_types.py
gnguralnick: minds app will just decide what to do with the events

libs/mngr/imbue/mngr/cli/list.py
gnguralnick: (discussed in person) there's some weirdness around these functions. the emit_final_json function is just a pointless public wrapper on _write_json_line - we should just make _write_json_line public instead and remove emit_final_json.

libs/mngr/imbue/mngr/api/discovery_events.py
gnguralnick: (discussed in person) there should be a better existing knowledge source to identify whether we've manually destroyed this host or not. regardless, logic relating to that should be in the minds app rather than in mngr, since the issue being fixed was purely within the minds app (the unnecessary homepage redirect). the minds app should just be smarter in knowing whether it has issued a destroy request for the agent who is stuck so that it can avoid redirecting.

apps/minds/imbue/minds/desktop_client/backend_resolver.py
gnguralnick: (discussed in person) there should be a better existing cache within the minds app that contains this needed mapping - we shouldn't have to store this state anywhere new. need to just wire things up better

apps/minds/imbue/minds/desktop_client/backend_resolver.py
gnguralnick: (discussed in person) there should be an error mechanism that surfaces up to the minds app in the notifications menu. might be janky rn but we should be using that

apps/minds/imbue/minds/desktop_client/app.py
```diff
@@ -1711,129 +1776,409 @@ def _sanitize_recovery_return_to(raw: str) -> str:
...
+    """
+    try:
+        finished = concurrency_group.run_process_to_completion(
+            argv,
+            timeout=_RESTART_COMMAND_TIMEOUT_SECONDS,
+            is_checked_after=False,
+            env=env,
+        )
+    except (OSError, ConcurrencyGroupError) as exc:
+        return MngrRun(failure_reason=str(exc))
```
gnguralnick: error scenarios should raise, not return. we do still want error messages to propagate properly so need to confirm that exception propagation will take the right route. but returning violates the python semantics.

this means that MngrRun potentially doesn't need a failure_reason - error would use an exception rather than this model

libs/mngr/imbue/mngr/api/discovery_events.py
gnguralnick: update: i was actually wrong about the reason for these changes. these changes are actually supporting `mngr stop --stop-host` - this needs to work regardless of the state of the host. so these changes let us resolve `agent identifier -> host` without using SSH by replaying the discovery event stream and cross-checking against `provider.discover_hosts`. but i'm checking whether this is a larger change than is actually necessary.

gnguralnick: update on the update: i decreased the extent of this change somewhat. the bulk of these lines changed is just extracting the mappings `name_by_agent_id`, `host_id_by_agent_id`, and `destroyed_agent_ids` into a MutableModel for clearer direct access to the fields (rather than just using a tuple of all three). the actual substantive change is `resolve_hosts_for_identifiers` and its usage, which just lets us identify hosts without relying on them working correctly. if you think this is not necessary i can remove it.

libs/mngr/imbue/mngr/cli/list.py
gnguralnick: (starting a separate branch to handle this)

gnguralnick: #1836

joshalbrecht: that other PR lgtm!

joshalbrecht:
lgtm!  thanks for all of the fixes

PR #1708, opened 2026-05-20
minds: let an agent spawn peer minds via latchkey curl

ysshah:
## Summary

Enables an agent inside an mngr/minds workspace to spawn peer minds via a latchkey-gated `POST 127.0.0.1:<minds-port>/api/create-agent`.

- Persists a per-installation bearer token at `<auth-dir>/api_token` and registers `minds` as a latchkey service, so the gateway injects `Authorization: Bearer ...` on the agent's behalf — the agent never sees the credential.
- `/api/create-agent`, its status, and its logs endpoints accept the bearer token as an alternative to the browser session cookie.
- Materializes a `minds` detent scope with three named permissions (`minds-create`, `minds-status`, `minds-logs`) gated to `domain=127.0.0.1` AND path prefix `/api/create-agent`, so a user clicking `any` can't escape into unrelated localhost endpoints. First spawn pops the standard latchkey "Spawn a peer mind" dialog; subsequent spawns by the same agent are silent.
- Latchkey UNKNOWN credentialStatus is now treated as proceed-and-trust, so the pre-installed rawCurl bearer token no longer pops "Manual credential setup required" on Approve.
- Idempotent startup migration backfills the schema into existing per-host permissions files before the gateway starts.

Original implementation in the first three commits; the remaining 18 commits are autofix cleanups (robustness, single-statement try/except, error-class hierarchy, dedup, docstrings).

## Test plan

- [ ] CI passes (full offload suite + acceptance)
- [ ] Manually verify: fresh install → spawn a peer mind from an agent shell, observe the latchkey "Spawn a peer mind" dialog the first time and silent success thereafter
- [ ] Manually verify: status and logs endpoints reachable from within the agent using the bearer-token path
- [ ] Manually verify: existing installs get the schema migration backfilled on next `minds run`
- [ ] Manually verify: graceful degradation when latchkey binary is missing/unexecutable (no crash, warning logged, rest of desktop client still works)


apps/minds/imbue/minds/cli/run.py
```diff
@@ -170,6 +171,34 @@ def run(
...
+    # gateway is up so there is no race with the ``permissions.mjs``
+    # extension.
+    #
+    # Wiring is best-effort: a failure in any one step (disk error
+    # during the schema migration, an unreadable api_token file, a
+    # latchkey CLI hiccup) leaves the spawn-peer capability disabled
+    # but lets the rest of the desktop client come up normally.
+    try:
+        _wire_latchkey_minds_service(latchkey=latchkey, auth_store=auth_store, port=port)
+    except (LatchkeyError, ApiTokenError, OSError, ConcurrencyExceptionGroup) as exc:
```
joshalbrecht, line 194: probably worth thinking through some of the edge cases here! what errors actually happened? are there ways for us to recover?

ysshah: Good point - I didn't experience any errors here when I performed my own local testing, but it could be worth failing more loudly (or even refusing to start up) depending on the specific error that's encountered

gnguralnick:
Superseded by #2074, recreated as a direct branch in imbue-ai/mngr (was a fork-based PR). Closing in favor of the in-repo PR so the stacked PR #2075 can target its branch.

PR #1710, opened 2026-05-20
Pin minds desktop toolchain (Node/pnpm/Python/Electron/uv) and add a 7-day dependency cooldown

gdenisov:
## Summary
Make the `apps/minds` desktop app build from an exactly-pinned toolchain, ship a runtime whose Node matches what we develop against, and harden dependency resolution with a 7-day release-age cooldown.

**JS toolchain pins**
- Pin `engines` in `apps/minds/package.json`: `pnpm` -> `10.33.4`, `node` -> `24.15.0`.
- Add `apps/minds/.npmrc` (`engine-strict=true`) so installs fail fast on engine mismatch instead of breaking confusingly downstream.
- Add `apps/minds/.nvmrc` (`24.15.0`) so nvm/fnm users pick up the pinned Node automatically.
- Bump the `test-docker-electron` CI job to match (`node-version: '24.15.0'`, `npm install --global pnpm@10.33.4`).
- Drop the now-redundant "v11 is not yet supported" note from the docs (engine-strict + exact pin make the error self-explanatory).

**Electron -> bundled-Node parity**
- Upgrade Electron `35.7.5` -> `40.10.1`. Electron 35 shipped a Node 22.x runtime to end users; 40.10.1 bundles **Node 24.15.0** (Chromium 144.0.7559.236), matching the pinned dev/CI Node. Electron 40 is the *lowest* major on the Node 24 line, so this is the smallest jump that achieves parity; staying on 40 (vs the newest 42.x) avoids the behavior changes in majors 41 (cookie `changed`-event cause values) and 42 (macOS notification code-signing), neither of which our Electron code depends on. Reviewed all 36->42 breaking changes against our actual API surface: none apply on 40.

**Python pin + lockfile**
- Pin the end-user-packaged interpreter to exact `==3.12.13` in `apps/minds/electron/pyproject/pyproject.toml` (the pyproject uv reads at first launch), and regenerate `apps/minds/electron/pyproject/uv.lock` to match (revision 3).
- Bump the bundled `UV_VERSION` in `apps/minds/scripts/build.js` `0.7.12` -> `0.11.15`: the relative `exclude-newer` form (below) needs uv >= 0.10.0, and 0.11.15 is the uv that generates the committed lockfile. (uv 0.7.12 reads revision-3 lockfiles fine; the revision is not the issue.)

**7-day dependency cooldown (supply-chain hardening)**
- Refuse to resolve any distribution -- including transitive deps -- published less than 7 days ago, so a freshly-compromised release cannot be pulled in before it is noticed/yanked.
- pnpm: `minimumReleaseAge: 10080` (minutes) in `apps/minds/pnpm-workspace.yaml` (pnpm >= 10.16.0; we pin 10.33.4).
- uv: `exclude-newer = "7 days"` under `[tool.uv]` in the packaged `apps/minds/electron/pyproject/pyproject.toml`.
- Only bites during resolution (`pnpm install` without `--frozen-lockfile`, `pnpm add`/`update`, `uv lock`/`add` or a re-resolve); frozen-lockfile installs replay the committed lockfile unaffected. Scoped to the `apps/minds` desktop app; the monorepo root `pyproject.toml` is unchanged.

**Docs / spec**
- Document the pinned-toolchain install recipes (nvm/fnm for Node, `npm install --global pnpm@10.33.4`, Homebrew-drift caveat) and the dependency cooldown in `apps/minds/docs/desktop-app.md`.
- Refresh the example `pyproject.toml` in `specs/electron-desktop-app/spec.md` to match the real packaged file and correct its standalone-pyproject path (`electron/pyproject.toml` -> `electron/pyproject/pyproject.toml`).

An earlier iteration tried a dual-form `pnpm-workspace.yaml` (pnpm 10 `onlyBuiltDependencies` + pnpm 11 `allowBuilds`); pinning matches team philosophy and avoids ongoing deny-list maintenance.

## Test plan
- [x] **pnpm engine check**: with the exact pin + `engine-strict=true`, `pnpm install` in `apps/minds/` fails fast with `ERR_PNPM_UNSUPPORTED_ENGINE` on any non-10.33.4 version.
- [x] **Real install under pinned toolchain**: `pnpm install --frozen-lockfile` under Node 24.15.0 / pnpm 10.33.4 installs Electron 40.10.1 cleanly; the cooldown does not block the frozen install.
- [x] **Python pin + uv cooldown**: `uv sync --frozen --python 3.12.13` installs cleanly; `uv lock --check` confirms uv parses `exclude-newer` and computes the 7-day cutoff; `uv lock --check` is clean against the regenerated lockfile.
- [x] **pnpm cooldown**: `pnpm config get minimumReleaseAge` returns `10080` under the pinned pnpm.
- [x] **Build script**: uv 0.11.15 release assets resolve for all 3 build targets; real download+extract yields a working `uv 0.11.15`; `copyPyproject`'s strip preserves `[tool.uv] exclude-newer` and removes `[tool.uv.sources]`; `node --check` + existing `build_test.py` pass.
- [x] **CI**: `test-docker-electron` passes **with Electron 40.10.1**; `test-offload` + `test-offload-acceptance` pass. (`test-docker` is unrelated to this branch and has been hitting a hung-runner infra issue.)
- [x] **Changelog entries** under each touched project: `apps/minds/changelog/`, `dev/changelog/`.

## Notes
- The cooldown is intentionally scoped to `apps/minds` only; pinning the monorepo-wide uv (`required-version`) and any root-level cooldown are deliberately left out of this PR.
- The packaged Python pin (`==3.12.13`) and the monorepo `.python-version` (`3.12`) are intentionally independent; CI's `setup-python: '3.12'` is intentionally not patch-pinned.


apps/minds/docs/desktop-app.md
```diff
@@ -150,7 +150,7 @@ URL -- env selection happens via `MINDS_CLIENT_CONFIG_PATH` /
 ### Prerequisites
 
 - Node.js >= 20
-- pnpm >= 10
+- pnpm 10.x (v11 is not yet supported)
 - Python >= 3.11, uv, git (for the Python backend)
```
joshalbrecht: can we not pin all of these to more specific versions?

I don't love dynamically pulling in basically anything--leads to weird bugs

gdenisov: For pnpm — we could, but I am pretty sure every dev has a slightly different versions of 10 right now — is it worth forcing the local build to fail, given that as far as we know there are no issues with any of the 10s at the moment?

PR #1713, opened 2026-05-20
use one shared env for each offload run

evgunter:
## Summary
- Opt-in via `MNGR_TEST_SHARED_MODAL_ENV_NAME`: when set, every modal test fixture and the session-scoped subprocess-env fixtures land in the same pre-created Modal environment (threaded through `MngrConfig.prefix` + `ModalProviderConfig.user_id`) and skip env creation / deletion / leak-tracking at the fixture layer. Apps and volumes are still created and deleted per-test.
- `just test-offload-acceptance` and `just test-offload-release` now wrap the offload invocation with `modal environment create` + `trap`-delete around a `mngr_test-YYYY-MM-DD-HH-MM-SS-shared-<uuid>` name and forward it via `--env`. Local pytest behavior (no env var set) is unchanged.
- Mitigates the 1500-envs-per-workspace Modal cap that was driving intermittent env-create failures.

## Test plan
- [x] `just test-quick libs/mngr_modal` (387 passed)
- [x] `just test-quick libs/mngr` (3928 passed, 1 skipped)
- [x] Full local non-acceptance/release run (10492 passed)
- [x] Ruff format + lint clean
- [x] CI `test-offload-acceptance` validates shared-env path end-to-end (passed in 4m58s)
- [x] Manual against real Modal: with `MNGR_TEST_SHARED_MODAL_ENV_NAME` set, ran `test_create_host_creates_sandbox_with_ssh` -- env list went 85 → 86 → 85 (only the wrapper env was minted; no per-test env). Test passed in 12.65s. Wrapper env deleted explicitly post-run.


joshalbrecht:
lgtm, merging because things are so borked without this that it's hard to tell what is going on

PR #1714, opened 2026-05-21
fix: retry Modal app context entry through async permission propagation

joshalbrecht:
## Summary

- Modal recently migrated their permission system. After `modal environment create` returns success, the per-user permission entry for the new env is propagated **asynchronously** (~3-7s). Any operation during that window raises `modal.exception.PermissionDeniedError` instead of the previous `NotFoundError`.
- Our retry layer in `mngr_modal.backend` only caught `ModalProxyNotFoundError`, so every Modal acceptance test crashed at `real_modal_provider` fixture construction time before any test body ran. `_translate_modal_error` also didn't have a typed `ModalProxyPermissionDeniedError`, so it fell through to a bare `ModalProxyError`.
- This change adds the typed error at the modal_proxy boundary and includes it in the existing tenacity retry policies (5 attempts, 1s→10s exponential backoff) in both the persistent and ephemeral app-entry paths. It also mirrors the retry in the test cleanup helper so fast teardowns don't spuriously leak Modal envs/volumes through the same window.

Repro before the fix (drives only the Modal SDK, no mngr code):
```
[1] app.run() against nonexistent env:
   NotFoundError: Environment '...' not found        <- our retry layer catches this
[2] modal environment create: success
[3] app.run() immediately after create:
   PermissionDeniedError: User josh-staging does not have write access ...
   PermissionDeniedError: ...
   success after ~3s                                  <- transient window
```

After the fix, the originally failing test passes (29s, 4 transparent retries through the ~7s window logged inline).

## Test plan
- [x] modal_proxy unit/integration tests pass (`just test-quick libs/modal_proxy ...`)
- [x] mngr_modal unit/integration tests pass (`just test-quick libs/mngr_modal ...`)
- [x] Ratchet tests pass
- [x] Originally failing acceptance test passes: `just test libs/mngr_modal/imbue/mngr_modal/test_modal_instance.py::test_get_host_by_id`
- [ ] Full Modal acceptance suite in CI (offload)

### DIFF
--- a/changelog/mngr-fix-ci-with-modal.md
+++ b/changelog/mngr-fix-ci-with-modal.md
@@ -1,0 +1,35 @@
+## Modal provider retries through async permission propagation
+
+Modal recently migrated their permission system so that the per-user
+permission entry for a just-created environment is propagated asynchronously
+(typically ~3-7 seconds after `modal environment create` returns success).
+During that window, any operation on the new environment — including by the
+creating user — raises `modal.exception.PermissionDeniedError` instead of
+the previous `NotFoundError`. The Modal team confirmed this is intended
+short-term behavior on their side and that they plan to add a server-side
+buffer.
+
+This was breaking every Modal acceptance test at fixture construction time:
+`real_modal_provider` would create a fresh test env, immediately try to enter
+its `app.run()` context, and crash on the permission error before any test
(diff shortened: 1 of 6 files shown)


joshalbrecht:
lgtm

PR #1715, opened 2026-05-21
Add minds electron acceptance test for local docker workspace

joshalbrecht:
## Summary
- Adds an acceptance test (`@pytest.mark.acceptance`) for the minds desktop client that creates a local Docker workspace from forever-claude-template and verifies the workspace's `system_interface` is reachable through the desktop client proxy
- Spec and implementation in progress (currently in `/architect` Q&A phase); placeholder spec is at `specs/minds-electron-acceptance-test/spec.md`

## Test plan
- [ ] To be filled in once spec is finalized and implementation lands


joshalbrecht:
lgtm

I don't LOVE this test, but I'm going to continue from here over the weekend in order to make better testing that uses docker in modal, caches a bunch of this, etc

and it's better to have something that checks that it kinda works rather than nothing at all

PR #1716, opened 2026-05-21
Adding mngr preferred tmux config: status-left-length

weishi-imbue:
## Problem
image

Tmux's default `status-left-length` is 10.  Mngr prepend `mngr-` to the agent name. 
So in the format of `[mngr-<agent_name>]`, user rarely sees the full agent name, since it's just 4 chars left. 

mngr should have an opinion that `status-left-length` should be longer, say 20. So it's more likely to show full agent name. 

User's own tmux config can still override it.


joshalbrecht:
lgtm, thanks!

PR #1719, opened 2026-05-21
replace mngr_gemini with mngr_antigravity

evgunter:
https://github.com/user-attachments/assets/af1ea8e7-e119-447e-806e-dece03b339c3

(slow because antigravity re-auths every time and my network is slow because i'm on a plane)

---

## Summary

- Renames `libs/mngr_gemini` → `libs/mngr_antigravity`; agent type `gemini` is replaced by `antigravity`. Google announced on 2026-05-19 that Gemini CLI is being superseded by Antigravity CLI (`agy`); the legacy request path turns off for paid-tier accounts on 2026-06-18. `mngr_gemini` was never released, so this is a destructive rename with no shim.
- Despite the Gemini lineage, agy is architecturally closer to Claude Code than to Gemini -- hook event names, `--dangerously-skip-permissions` flag spelling, settings file shape, and the conceptual "trust this folder" gate all match Claude's surface. The plugin's design choices mirror `mngr_claude`'s wherever the surface aligns.

## What ships

- **Agent type**: `antigravity`. Process name `agy`. `TUI_READY_INDICATOR = "? for shortcuts"` -- a footer-hint substring that only appears once agy's input row is fully drawn. Pinned with comment explaining why the obvious-but-wrong choice (the splash banner substring "Antigravity CLI") fails: agy renders "Welcome to the Antigravity CLI. You are currently not signed in." during the OAuth flow, which contains "Antigravity CLI" but is not yet ready to receive input; pasting then drops the keystrokes and the `mngr create --message` flow times out.
- **`auto_allow_permissions=True`** appends `--dangerously-skip-permissions` to the launch command. Picked over a permission hook because live testing showed agy *loads* `hooks.json` but hook *execution* is gated behind the `json-hooks-enabled` experiment Google enables per-account; once it ships GA the sentinel + permission hook can come back.
- **Workspace symlink workaround**: agy refuses to add any path with a dot-prefixed segment as a workspace (logs `Failed to add workspace folder ... is hidden: ignore uri`) and silently falls back to `$HOME` as the project root. Since mngr's work_dir lives under `~/.mngr/worktrees/...`, this means workspace-scoped tooling would operate against the wrong tree. `agy --help` has no flag for it; [Google forum thread](https://discuss.ai.google.dev/t/add-workspace-rule-failed/114582) documents the bug and lists "avoid hidden paths" as the only workaround. So `assemble_command` prepends `mkdir -p ... && ln -sfn <work_dir> /tmp/mngr_antigravity_workspaces/<agent_id> && cd <symlink>` before launching agy. agy then treats the symlink path as its workspace identity (verified via the `project: using project "/tmp/..."` log line). Recreated via `ln -sfn` on every launch so `/tmp` wipes are self-repairing. Symlinks under `/tmp` are safe to leave behind because cleanup operates on the symlink inode, not the target.
- **`auto_dismiss_dialogs` first-launch trust dismissal** (default `False`, mirroring `mngr_claude.auto_dismiss_dialogs`). agy uses exact-match against `trustedWorkspaces`, so the trust write targets the **symlink path** (what agy actually checks against), not work_dir. Branches:
  - Symlink path already trusted: idempotent no-op.
  - Source repo root already trusted (e.g. a prior agent for the same repo was approved): silently append the new symlink path. Subsequent agents for the same repo don't re-prompt.
  - `mngr_ctx.is_auto_approve` (`--yes`) or `auto_dismiss_dialogs=True`: silently append both source repo root and symlink path. Future agents in the same repo then take the silent-extend branch.
  - Interactive shell: prompt via `click.confirm` referencing the source repo root (same wording as Claude). On accept, write both source + symlink path.
  - Non-interactive without opt-in, or user declines: log a clear `logger.error` and `raise SystemExit(1)`. `SystemExit` is a `BaseException`, which bypasses `provision_agent`'s `ConcurrencyExceptionGroup` wrapping (see `imbue.concurrency_group.concurrency_group.ConcurrencyGroup._exit` lines 173-179 vs. 190-191) so the operator sees a clean exit instead of a noisy auto-diagnostics traceback. (Mirrors the user-visible posture of `mngr_claude.ClaudeDirectoryNotTrustedError`, which goes through a different path that doesn't hit the wrapping.)
- **`emit_common_transcript=True`** (default) wires `mngr transcript` support. agy keeps a per-conversation JSONL transcript at `~/.gemini/antigravity-cli/brain/<conv_id>/.system_generated/logs/transcript.jsonl`. The streamer scopes itself to this agent's conversations by grepping agy's own `--log-file` for `(Created|Resumed) conversation <uuid>` lines, then tails each matching transcript, augmenting each line with `_mngr_conv_id`. The converter maps `USER_EXPLICIT/USER_INPUT → user_message`, `MODEL/PLANNER_RESPONSE → assistant_message` (with synthetic `tool_call_id`s since agy doesn't carry ids), `MODEL/CODE_ACTION → tool_result` (paired with the last in-conversation tool call).
- **Hard-error on malformed/non-list settings**: if `~/.gemini/antigravity-cli/settings.json` is malformed JSON, has a non-object top-level value, or has a non-list `trustedWorkspaces` value, provisioning raises `UserInputError` rather than overwriting. Aligns with the `check_silent_decode_error_catches` ratchet's user-authored-config rule and protects unknown future agy schemas from being silently coerced.
- **Loud-error on schema breaks** inside the JSONL converter: if a `USER_INPUT` event arrives without the `<USER_REQUEST>...</USER_REQUEST>` envelope agy 1.0.0 always wraps it in, the converter writes a clear stderr message and drops the event rather than silently baking agy's bookkeeping metadata into the user-facing transcript.

## Deliberately out of scope (deferred)

- **Readiness sentinel**: live spike found that `hooks.json` is loaded but execution is gated behind the `json-hooks-enabled` experiment that Google enables per-account. Falls back to `InteractiveTuiAgent`'s banner poll. Re-introduce once the experiment ships GA.

## Antigravity-specific caveats documented in README

- **`agy` PATH shadowing**: the Antigravity 2.0 desktop app's bundled `agy` shim can shadow the standalone CLI in `PATH`. Workaround: remove the desktop app's `bin/agy` or override `command` in the agent type config.

## Test plan
- [x] `just test-quick libs/mngr_antigravity` -- 129 passed
- [x] `just test-quick "libs/mngr -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'"` -- 3906 passed, 1 skipped
- [x] `just test-quick "libs/mngr_claude -m 'not tmux and not docker and not docker_sdk and not acceptance and not release'"` -- 481 passed
- [x] `uv sync --all-packages` succeeds after rename
- [x] Live `mngr c <name> antigravity --message "..." --no-connect --yes`: completed cleanly; `mngr transcript` shows both user and assistant messages; agy's workspace correctly logged as the symlink path (no `$HOME` fallback)
- [x] Live silent-extend: second agent in same source repo with no `--yes` and no `auto_dismiss_dialogs` -- created without re-prompting (source repo trust grant carried over)
- [ ] CI: `just test-offload` + `just test-offload-acceptance` (deferred to CI)


libs/mngr_antigravity/imbue/mngr_antigravity/resources/common_transcript.sh
```diff
@@ -0,0 +1,288 @@
...
+#                                       recent PLANNER_RESPONSE tool_call
+#                                       in the same conversation)
+#   SYSTEM/CONVERSATION_HISTORY    -> dropped (bookkeeping)
+#   everything else                -> dropped (best-effort: forward-compat
+#                                       with future agy schema additions)
+#
+# Tool-call ids are synthetic: agy's transcript does not carry an id on
+# tool_calls (only `name` + `args`), so we mint
+# "<conv_id>-<step_index>-tc<idx>" using the conversation id smuggled in
+# from the streamer's `_mngr_conv_id` field. Pairing with CODE_ACTION
```
evgunter, line 32: huh

libs/mngr_antigravity/imbue/mngr_antigravity/plugin.py
```diff
@@ -86,6 +86,21 @@
...
+# /path/.mngr/...: is hidden: ignore uri`` and falls back to the user's
+# home directory as the project root, which means workspace-scoped tooling
+# (file search, project_id, .agents/) operates against the wrong tree.
+#
+# Verified via google-forum bug report (no flag override exists) and
+# confirmed live: launching agy with cwd set to a /tmp symlink that targets
+# the dotted ``work_dir`` produces ``project: using project "/tmp/..."``
+# (the symlink path, not the resolved target), and the workspace-add error
+# disappears. The symlink is recreated on every ``assemble_command`` call
+# via ``mkdir -p`` + ``ln -sfn`` so /tmp wipes self-repair on next launch.
```
evgunter, line 101: sigh

joshalbrecht: indeed--wtf kind of crazy behavior is that? Like, isn't that the default for tons of linux programs?

joshalbrecht:
lgtm!

PR #1720, opened 2026-05-21
mngr create: reject --reuse + --new-host as incompatible

weishi-imbue:
## Summary

`mngr create` accepts `--reuse` together with the explicit `--new-host` flag, but the two contradict each other:

- `--reuse` adopts an existing agent on an existing host.
- `--new-host` always provisions a fresh host, which by definition has no agent to adopt.

When both are passed, `--reuse`'s name-based lookup runs first and silently wins or fails depending on what already exists, so `--new-host` is effectively ignored. This PR rejects the combination with a clear `UserInputError`.

Note this is scoped to the explicit `--new-host` *flag*. The implied-new-host form (`NAME --provider X` with no host) stays compatible with `--reuse` -- that is the documented idempotent-create pattern (`mngr create --reuse --provider modal my-task`).

## Changes

- **`libs/mngr/imbue/mngr/cli/create.py`**: Reject `--reuse` + `--new-host` next to the existing `--update requires --reuse` check. Since `--update` requires `--reuse`, `--reuse --update --new-host` is caught by the same check. `--help` text on both flags updated; CLI markdown docs regenerated.
- **`apps/minds/imbue/minds/desktop_client/agent_creator.py`**: Drop `--reuse` / `--update` from `_build_mngr_create_command`. Every minds launch mode (LOCAL, LIMA, CLOUD, IMBUE_CLOUD) passes `--new-host`, so the flags would now be rejected. This is safe:
  - LOCAL/LIMA/CLOUD always provision a fresh host -- there is no agent to reuse.
  - Minds names every workspace agent the constant `system-services`, so `--reuse`'s provider-scoped name lookup would match *any* existing minds workspace (or raise "Multiple agents found"). It was never doing what the old comment claimed.
  - IMBUE_CLOUD: the baked `system-services` agent collision is handled by mngr's duplicate-name check exempting `host.pre_baked_agent_id` in `api/create.py`, so `--reuse` plays no role.
- **Tests**: New CLI tests for `--reuse --new-host` and `--reuse --update --new-host`; updated minds tests that previously asserted `--reuse`/`--update` were present.
- **Changelog**: `changelog/mngr-1703.md`.

## Note

Supersedes the content of PR #1703 (`mngr/incompat-options`), which had picked up a large unrelated merge. This branch is the same change re-derived cleanly off `main`.


qi-imbue (approved):

weishi-imbue:
This has evolved into a duplicate of #1703

The blocking error is fixed in #2085. 
Closing this now as imbue_cloud provider may still need these two for now. --reuse -> reuse existing physical host. --new-host -> means the host is new to the user.

PR #1721, opened 2026-05-21
TMR: daily scheduled run via wrapper workflow with stale-PR gate

qi-imbue:
## Summary

Run TMR on a daily cron, with a gate that prevents stacking up periodic PRs while leaving manual runs untouched.

- New `.github/workflows/tmr-scheduled.yml`: cron at 08:00 UTC (00:00 PST; GitHub Actions cron has no timezone support, so this shifts to 01:00 PDT in summer). A `gate` job inspects the oldest open PR with label `tmr-periodic`:
  - none open: proceed.
  - older than 4 days: post a closing comment, close the PR (with `--delete-branch`), proceed, and forward the closed PR's number so the new PR can post a "Superseded by #N" follow-up on it.
  - 4 days or younger: skip and post a comment on the open PR explaining the policy. A fresh comment is posted per skipped day; the recurring nudge is intentional.
  - A `run` job invokes `tmr.yml` via `workflow_call` only when the gate said proceed.
  - Concurrency group `tmr-scheduled` (cancel-in-progress: false) prevents a second firing from racing through the gate while a long prior run is still in flight (i.e. before it has had a chance to open its labeled PR).
- `.github/workflows/tmr.yml` becomes a reusable workflow (`workflow_call`) in addition to `workflow_dispatch`. Two new internal inputs (`is_scheduled`, `supersedes_pr`) are set only by the scheduled wrapper and aren't surfaced in the manual-dispatch UI. The PR-creation step:
  - adds `--label tmr-periodic` and `--assignee qi-imbue --assignee joshalbrecht` only when `is_scheduled` is true,
  - posts a "Superseded by ..." follow-up comment on the prior PR when `supersedes_pr` is non-empty.
- Default `test_paths` is now `libs/mngr/imbue/mngr/e2e/` (whole dir) instead of `test_basic.py`, for both scheduled and one-click runs.
- Manual `workflow_dispatch` of TMR is intentionally independent of all of the above: PRs are unlabeled, unassigned, and the gate is not consulted.

The cron only fires after this merges to `main` (standard GitHub Actions behavior). The gate workflow has a `workflow_dispatch` trigger so it can be smoke-tested on a branch before merging.

## Test plan
- [ ] On this branch, manually trigger `TMR (scheduled)` and confirm: the `tmr-periodic` label is created if missing, no open `tmr-periodic` PR -> `gate` outputs `proceed=true`, `run` is invoked.
- [ ] After merge, the first cron firing finds no open periodic PR -> proceeds; the auto-opened PR carries the `tmr-periodic` label and assignees `qi-imbue` + `joshalbrecht`.
- [ ] Confirm a manual `workflow_dispatch` of TMR does not label or assign its PR.


.github/workflows/tmr-scheduled.yml
```diff
@@ -0,0 +1,117 @@
+name: TMR (scheduled)
+
+# Daily scheduled wrapper around the main TMR workflow.
+#
+# At the top of each run, the `gate` job looks for an open PR from a prior
```
qi-imbue, line 5: This may be actually useful or over-engineered, we can get rid of the complexity if we don't find it useful later.

PR #1723, opened 2026-05-21
mngr: split minds env activate into use-vs-deploy modes

joshalbrecht:
## Summary
- Draft PR for the in-flight `minds env activate` split: default activation exports only use-side env vars; new `--deploy` flag adds `MODAL_PROFILE` (and any other deploy-side credentials).
- Goal: stop spurious Modal-discovery / Latchkey breakage for users who only want to *use* a deployed minds env but have no token for that tier's Modal workspace.
- Spec is being authored via /architect; implementation follows.

## Test plan
- [ ] Spec written and reviewed
- [ ] Implementation lands
- [ ] Unit + acceptance tests pass


joshalbrecht:
lgtm

PR #1724, opened 2026-05-21
Discovery: per-provider state in snapshots + minds providers panel

joshalbrecht:
## Summary

- Discovery snapshots (`FullDiscoverySnapshotEvent`) now carry per-provider state: `providers` (loaded successfully) and `error_by_provider_name` (discovery raised). Always emitted on every poll, even when zero providers succeeded.
- New `UNKNOWN` state on `AgentLifecycleState` / `HostState` for previously-tracked agents whose provider just failed discovery. `AgentObserver` emits these (sticky) and `mngr_notifications` recognizes the indirect `RUNNING → UNKNOWN → WAITING` transition.
- Minds: silent auto-disable-on-auth-error machinery deleted. Replaced with a visible providers panel on the landing page (status badges, last-error messages, Enable/Disable buttons, freshness counters) + a POST endpoint that toggles `is_enabled` and bounces observe.

Full spec: `specs/discovery-providers-and-errors/concise.md`

## Notes / things worth flagging

- **Cross-build breakage window**: older builds of `mngr_forward` / `mngr_latchkey` / `mngr_notifications` will hit `DiscoverySchemaChangedError` against snapshots from the new mngr observe. This is the intended loud-failure mode (no extra migration scaffolding per spec answer 3d). Users with stale plugin installs must rebuild.
- **Spec said `AgentState`, code has `AgentLifecycleState`**: the actual enum is named `AgentLifecycleState`. The UNKNOWN value was added there (and to `HostState`).
- **Manual UI verification not yet done**: tests pass; the providers panel was not exercised end-to-end in a running browser. Recommended before merge.
- **Snapshot heartbeat volume**: always-emit-on-poll roughly doubles the size of `events/mngr/discovery/events.jsonl` for steady-state systems with no provider errors. Existing rotation handles it (50 MB cap, one rotated file kept).

## Test plan

- [x] `just test-quick libs/mngr` (4102 passed)
- [x] `just test-quick apps/minds` (758 passed)
- [x] `just test-quick libs/mngr_notifications libs/mngr_forward libs/mngr_imbue_cloud libs/mngr_latchkey` (671 passed)
- [x] `just test-quick libs/mngr_modal libs/mngr_claude libs/mngr_lima libs/mngr_ovh libs/mngr_recursive` (1320 passed)
- [x] ratchet tests pass across all touched projects
- [ ] Manual UI verification: open the minds landing page, exercise Disable/Enable on a provider, confirm "Waiting…" → status update flow
- [ ] CI green


apps/minds/imbue/minds/desktop_client/app.py
```diff
@@ -1418,6 +1484,86 @@ def _on_health_change(agent_id: AgentId, status: AgentHealth) -> None:
     )
 
 
+# The `local` provider is always present and always healthy; it would just be
```
gnguralnick: is this necessarily true? couldn't docker or lima not be running on the machine?

gnguralnick: not sure whether this assumption is just a UI thing or if there's some functionality that's relying on the provider being healthy

joshalbrecht: sigh, this is my fault for naming things wrong :(

what is called "local" in the UI means "docker", but *this* version of "local" means "local in mngr", which means "process on your computer"

We should really just rename the docker one to docker...

PR #1726, opened 2026-05-21
mngr forward: bind port dynamically; minds learns it from the listening envelope

gnguralnick:
## Summary

`mngr forward` no longer crashes when its bind port is already in use, and `minds` no longer dictates that port — it learns the actual bound port from the plugin's `listening` envelope.

### `mngr forward` (`libs/mngr_forward`)
- `--port` is now optional. Omitted → tries the default (`8421`) and falls back to an OS-assigned port if that is taken. Supplied explicitly → binds exactly that port and fails fast with a clean error if it is unavailable (a caller that picked a port did so deliberately).
- The listen socket is bound up front and handed to uvicorn via `Server.run(sockets=[...])`, so there is no close-then-rebind race and the `listening` envelope always reports the port actually bound.

### `minds` (`apps/minds`)
- Removed `_DEFAULT_MNGR_FORWARD_PORT`, the `--mngr-forward-port` flag, the `MINDS_MNGR_FORWARD_PORT` env var, and `ForwardSubprocessConfig.port`.
- `EnvelopeStreamConsumer` now handles the `listening` envelope and exposes `wait_for_listening(timeout)`. `minds run` blocks on it after `consumer.start()`, then builds `AgentCreator` / the desktop app / the health probe with the reported port. If the plugin fails to report a port within 30s, startup aborts with a clear error.
- Dropped the now-unneeded per-test port plumbing from the Electron e2e test.

## Behavioral note
`minds run` startup now briefly blocks on the plugin's `listening` envelope (sub-second in practice; 30s timeout).

## Testing
- New unit tests: port fallback, explicit-port collision error, `listening`-envelope handling, `wait_for_listening` timeout.
- Local: `mngr_forward` 196 passed / 1 skipped; `minds desktop_client` 439 passed; `run_test` 8 passed; ratchets 112 passed; type checks + ruff clean.
- Full suite + acceptance tests run in CI.


apps/minds/imbue/minds/cli/run.py
```diff
@@ -81,9 +81,14 @@
...
 from imbue.mngr_latchkey.forward_supervisor import LatchkeyForwardSupervisor
 
-_DEFAULT_MNGR_FORWARD_PORT: Final[int] = 8421
 _AUTH_ERROR_TYPE: Final[str] = "ImbueCloudAuthError"
 
+# How long `minds run` waits for the spawned `mngr forward` plugin to report
+# its bound port via a `listening` envelope before treating startup as failed.
+# The plugin emits this from its FastAPI lifespan startup, so the only slow
+# part is the subprocess's own interpreter start -- 30s is generous headroom.
+_MNGR_FORWARD_LISTEN_TIMEOUT_SECONDS: Final[float] = 30.0
```
gnguralnick: i think 30 seconds is far longer than we should be waiting here

apps/minds/test_desktop_client_e2e.py
```diff
@@ -231,18 +231,14 @@ def _build_electron_env(workspace_git_url: Path, workspace_name: str, mngr_forwa
...
     don't silently leak into every workspace the test creates.
 
-    Passes ``MINDS_MNGR_FORWARD_PORT`` so the spawned ``mngr forward``
-    subprocess binds to a per-test free port -- ``backend.js`` never
-    passes ``--mngr-forward-port`` so ``minds run`` falls back to the
-    click default (``_DEFAULT_MNGR_FORWARD_PORT = 8421`` in
-    ``apps/minds/imbue/minds/cli/run.py``), which collides with any
-    concurrent ``just minds-start`` (or with a prior crashed test that
-    didn't fully tear down its forward supervisor).
+    The spawned ``mngr forward`` subprocess no longer needs a per-test
```
gnguralnick: there's a few places that refer to the old implementation - let's not do that - just explain what the current implementation is

apps/minds/imbue/minds/cli/run.py
gnguralnick: Dropped to 10s in 817df3b1d. Observed real-process startup of the spawned `mngr forward` is ~1-2s; 10s still covers a cold CI interpreter start with margin while surfacing a genuinely stuck startup much sooner.

apps/minds/test_desktop_client_e2e.py
gnguralnick: Fixed in 817df3b1d. Removed the orphaned port paragraph from the `_build_electron_env` docstring (the function no longer touches the port at all) and rewrote the `wait_for_listening` comment in run.py to describe current behavior rather than contrasting with the removed `--mngr-forward-port` flow.

apps/minds/imbue/minds/cli/run.py
gnguralnick: Updated to 5s in 848f3ad9b per your preference.

libs/mngr_forward/imbue/mngr_forward/cli.py
```diff
@@ -251,14 +312,13 @@ def _on_listening() -> None:
...
-            app,
-            host=opts.host,
-            port=opts.port,
-            timeout_graceful_shutdown=1,
-            log_level="warning",
-        )
+        # Hand uvicorn the socket we already bound rather than a host/port
+        # pair, so there is no window between resolving the port and the
+        # server claiming it (and no chance uvicorn binds a different one).
+        server = uvicorn.Server(uvicorn.Config(app, timeout_graceful_shutdown=1, log_level="warning"))
```
joshalbrecht, line 318: thank you for doing this correctly (whether it was "you, claude" or "you, gabe", I'm still thankful :-P )

LLMs fuck this up so often, it's the one thing I was coming here to check :)

joshalbrecht:
lgtm!

PR #1728, opened 2026-05-21
fix tmux prefix matching hopefully for the last time

evgunter:
## Summary

- **Root cause.** `BaseAgent.get_lifecycle_state()` queried `tmux list-panes -t '<session>:0'` without the `=` exact-match prefix. When the targeted session was gone but a sibling session whose name started with the same prefix was still alive, tmux's default prefix matching silently routed the query to the sibling. The lifecycle check read the sibling's pane, saw a live `claude`, and reported the stopped agent as `WAITING`. This explains why `gemini` showed `WAITING` while `gemini-to-antigravity` was the live one.
- **Prevention.** Adds `tmux_session_target()` and `tmux_window_target()` helpers in `imbue.mngr.hosts.tmux` that encode the parsing rule. Routes every tmux `-t` call site through them. Adds a `PREVENT_BARE_TMUX_TARGETS` ratchet in `imbue_common` (baseline 1, for the intentional `list-panes -s` after a `has-session` guard).
- **Tests.** Adds live prefix-collision tests in `hosts/tmux_test.py` that spin up two real tmux sessions with overlapping names, kill the shorter one, and assert each helper-built target refuses to misroute to the sibling on `list-panes` / `send-keys` / `capture-pane` / `has-session` / `kill-session`. Includes a sanity-check test confirming bare names still misroute, so we'll notice if tmux ever changes default behavior.

## Why the prior fix (PR #1145) walked it back

That PR added `=` everywhere, then partially reverted on the (incorrect) belief that `=` doesn't work for target-window/-pane commands. The reverter conflated `=name` (which does fail) with `=name:0` (which works). For `list-panes`, the lifecycle check had been `'={session}:0'` and was demoted to `'{session}:0'` -- exactly the line that caused this bug.

## The actual parsing rule

| Command class | Working form |
|---|---|
| target-session (`has-session`, `kill-session`, `rename-session`, `attach`, `list-windows`) | `=name` |
| target-window/-pane (`list-panes`, `send-keys`, `capture-pane`, `paste-buffer`, `set-option`, etc.) | `=name:window` (the `:window` component is required) |
| `list-panes -s` | tmux's `cmd-find.c` ignores `=` despite `-s` being a documented target-session form. Use a `has-session -t '=name'` guard first; the only intentional ratchet violation. |

## Behavior change worth flagging

`build_post_attach_resize_script` was refactored from `xargs` to a per-window `while` loop. Previously only the active window's panes got SIGWINCH; now every pane in every window does. This matches the behavior of the `resize-window` step (which already iterated all windows) and is what "post-attach resize" implies.

## Test plan

- [x] `libs/mngr/imbue/mngr/hosts/tmux_test.py` -- 11 tests passing (5 unit + 6 live prefix-collision)
- [x] `libs/mngr/imbue/mngr/api/test_connect.py::test_post_attach_resize_delivers_sigwinch_to_pane_process` -- passes
- [x] `libs/mngr/imbue/mngr/utils/test_ratchets.py` -- 57/57 pass, including new `test_prevent_bare_tmux_targets`
- [x] Broader sweep across `libs/mngr/imbue/mngr/{hosts,agents,api}` -- 1307 passed, 1 skipped
- [x] `uv run mngr list` -- `gemini` now correctly shows `STOPPED` (was `WAITING` before)
- [ ] CI


.reviewer/autofix/auto-accept.md
```diff
@@ -14,6 +14,7 @@ The following categories of fixes should always be accepted without manual revie
 - Removing module-level docstrings
 - Removing Args/Returns sections from docstrings
 - Removing code from `__init__.py` (moving it to the appropriate module)
+- Pure DRY cleanup which is a clear improvement and doesn't change any behavior
```
evgunter, line 17: flagging

joshalbrecht:
thanks for fixing!

lgtm

PR #1732, opened 2026-05-22
Disable Latchkey usage counting in all pytest sessions

hynek-urban:
Set LATCHKEY_DISABLE_COUNTING=1 in os.environ inside the shared register_conftest_hooks(). Any test that ends up spawning a latchkey gateway -- directly, through 'mngr latchkey forward', or transitively via Electron / minds in the e2e suite (e.g. the test-docker-electron CI job) -- inherits the opt-out through the environment. Test runs are not real usage and must not bump the public daily counter.


joshalbrecht:
lgtm

PR #1737, opened 2026-05-22
Fold mngr_kanpan GitHub fetches into one gh api graphql call

qi-imbue:
## Summary
Collapse the four separate `gh` calls per refresh into one `gh api graphql`:
- `gh pr list --state open` + `gh pr list --state all` per repo (the two calls that raced on gh's `SearchType` HTTP cache)
- `gh pr view --json mergeable` per open PR (conflicts)
- hand-rolled `gh api graphql` per open PR (unresolved threads)

The new query uses GitHub search's OR semantics over `repo:` and `head:` qualifiers to filter to exactly the (repo, branch) pairs the board needs, with `mergeable`, `statusCheckRollup { state }`, `reviewThreads`, and `comments` inlined on every returned `PullRequest`. One HTTP request, atomic snapshot, no `gh pr list` so no race condition by construction.

## What this removes
- `_GH_PR_LIST_LOCK` / `_run_gh_pr_list` (the lock from #1735)
- `_fetch_pr_metadata` (second-pass conflicts + unresolved fetcher)
- `_build_unresolved_query`, `_parse_conflicts`, `_build_gh_pr_list_cmd`, `_parse_gh_output`, `_PrLookup`, `_FetchPrsResult`, `fetch_all_prs`, `_fetch_repo_prs`, `_build_pr_branch_index`, `_lookup_pr`, `_pr_priority`
- `ThreadPoolExecutor` fan-out in `compute()` — no longer needed
- Net diff: **-433 lines** across `github.py` (-491) + `github_test.py` (+58 due to new fixtures, but ~75 tests collapsed to ~58)

Supersedes #1735. If #1735 lands first, this PR rebases and removes the lock as part of the refactor; if not, this PR's commits include both the lock fix history and its removal in a single net diff.

## Verification
- Local `just test-quick libs/mngr_kanpan` → 389 passed, 0 failed
- Ratchets pass (broad-exception count dropped, snapshot still valid)
- Manual smoke test against real GitHub: PR matching across two repos works, draft detection works, `mergeable=CONFLICTING` is detected on a real PR with conflicts, `ignore_user` unresolved filtering returns the expected booleans

## Behavior notes worth mentioning
- CI status now comes from `StatusCheckRollup.state` (GitHub's server-side aggregation) instead of iterating per-context conclusions client-side. Compared on 12 of my open PRs: 12/12 match. Switching to per-context if a divergence ever bites is a small change (re-add `contexts(first: 100) { nodes { ... } }` to the rollup field and revive the per-context parser).
- `search()` is hard-capped at `first: 100`. With per-branch filtering this is essentially never reached (matched set is bounded by agent count, typically <30). The parser surfaces a warning in `errors[]` if `hasNextPage` ever trips.
- Single-shot failure mode: if the combined query times out or 5xx's, the whole refresh fails (today, one slow per-PR call didn't block others). The existing cached-fallback path handles this.

## Test plan
- [x] `just test-quick libs/mngr_kanpan/imbue/mngr_kanpan/data_sources/github_test.py` (58 passed)
- [x] `just test-quick libs/mngr_kanpan` (389 passed)
- [x] Manual smoke test against live GitHub for one + multi-repo cases
- [ ] CI acceptance/release runs


libs/mngr_kanpan/imbue/mngr_kanpan/data_sources/github.py
```diff
@@ -163,145 +156,199 @@ def display(self) -> CellDisplay:
...
+    ``first: 100`` is GitHub's hard per-page cap. The board only renders
+    PRs whose branch matches an active agent, so even on busy monorepos
+    the matched set is bounded by the agent count -- which is typically
+    well under 100. ``hasNextPage`` is surfaced as an error if it ever
+    trips, so we'd notice before silently dropping data.
+    """
+    repos = sorted({rb[0] for rb in repo_branches})
+    branches = sorted({rb[1] for rb in repo_branches})
+    repo_clause = " OR ".join(f"repo:{r}" for r in repos)
+    branch_clause = " OR ".join(f"head:{b}" for b in branches)
```
qi-imbue, line 251: This can result in fetching too many things if two agents have the same branch name, but nah seems harmless

libs/mngr_kanpan/imbue/mngr_kanpan/data_sources/github.py
```diff
@@ -163,145 +156,199 @@ def display(self) -> CellDisplay:
...
+            # Defensive: should be impossible given our query, but skip rather than misattribute.
+            continue
+        by_key.setdefault(key, []).append(_parse_pr_node(node, unresolved_ignore_user))
+
+    # If multiple PRs share the same head branch (e.g. closed-then-reopened),
+    # prefer OPEN > MERGED > CLOSED to match what kanpan used to do via
+    # `_build_pr_branch_index`.
+    prs: dict[tuple[str, str], PrInfo] = {key: max(cands, key=_pr_state_rank) for key, cands in by_key.items()}
+
+    if (search_result.get("pageInfo") or {}).get("hasNextPage"):
```
qi-imbue, line 330: Claude was too lazy to actually address this, but I'm just hoping that we never have more than 100 agents with PRs

libs/mngr_kanpan/imbue/mngr_kanpan/data_sources/github.py
```diff
@@ -163,145 +156,199 @@ def display(self) -> CellDisplay:
...
+    unresolved_ignore_user: str | None = None,
+) -> FetchBoardResult:
+    """Fetch every PR the kanpan board needs in one ``gh api graphql`` call.
+
+    Sends a single embedded ``search()`` query whose query string ORs over
+    every requested ``repo:`` and ``head:`` qualifier. That gives an atomic
+    snapshot of every (repo, branch) pair we care about with one HTTP
+    request -- no per-repo fan-out, no second-pass metadata fetch, no
+    shared HTTP cache to race on. See `_build_board_graphql` for the
+    query shape and rationale.
```
evgunter: some unnecessary comments here and above

libs/mngr_kanpan/imbue/mngr_kanpan/data_sources/github.py
evgunter: seems fine for now

libs/mngr_kanpan/imbue/mngr_kanpan/data_sources/github.py
```diff
@@ -311,28 +358,66 @@ def _parse_pr_state(state_str: str) -> PrState:
...
+    we can resurrect the per-context iteration by re-adding
+    ``contexts(first: 100) {{ nodes {{ ... }} }}`` to the rollup field.
+    """
+    if state == "SUCCESS":
+        return CiStatus.PASSING
+    if state == "FAILURE":
+        return CiStatus.FAILING
+    if state == "PENDING":
+        return CiStatus.PENDING
+    return CiStatus.UNKNOWN
```
evgunter: i think we might have this in a const `Mapping` somewhere

libs/mngr_kanpan/imbue/mngr_kanpan/data_sources/github.py
```diff
@@ -311,28 +358,66 @@ def _parse_pr_state(state_str: str) -> PrState:
...
+
+    Reads directly off the PullRequest node returned by the combined search
+    query (no separate JSON to parse, no separate GraphQL request).
     """
-    if not rollup:
-        return CiStatus.UNKNOWN
+    threads = (node.get("reviewThreads") or {}).get("nodes") or []
+    for thread in threads:
+        if not isinstance(thread, dict) or thread.get("isResolved", True):
+            continue
```
evgunter: what is in here that's not a dict?

libs/mngr_kanpan/imbue/mngr_kanpan/data_sources/github.py
```diff
@@ -435,114 +522,91 @@ def compute(
...
-                    agent_fields.update(
-                        _compute_failed_fetch_fields(
-                            cached_fields, agent.name, branch, agent_repo, this_created, self.config
-                        )
+            else:
+                # Fetch errored. Use a cached PrField if one matches this branch,
+                # otherwise show a PrFetchFailedField sentinel.
+                agent_fields.update(
+                    _compute_failed_fetch_fields(
+                        cached_fields, agent.name, branch, agent_repo, this_created, self.config
```
evgunter: this is nice, could have a helper for the first branch as well since it's pretty long

libs/mngr_kanpan/imbue/mngr_kanpan/data_sources/github.py
```diff
@@ -435,114 +522,91 @@ def compute(
...
-
-    prs: tuple[_PrLookup, ...] = Field(description="Fetched PRs paired with their CI status")
-    error: str | None = Field(default=None, description="Error message if fetch failed")
-
-
 def _get_cached_repo_field(
     cached_fields: dict[AgentName, dict[str, FieldValue]], agent_name: AgentName
 ) -> RepoPathField | None:
-    """Get the cached RepoPathField (with its `created` timestamp) if available."""
+    """Get the cached RepoPathField (with its ``created`` timestamp) if available."""
```
evgunter: huh

libs/mngr_kanpan/imbue/mngr_kanpan/data_sources/github_test.py
```diff
@@ -106,284 +157,293 @@ def test_get_cached_repo_field_wrong_type() -> None:
...
-    assert _lookup_pr(index, "repo", "b") == lookup
 
+def test_parse_rollup_state_none() -> None:
+    assert _parse_rollup_state(None) == CiStatus.UNKNOWN
 
-def test_lookup_pr_not_found() -> None:
-    assert _lookup_pr({}, "repo", "branch") is None
 
+def test_parse_rollup_state_success() -> None:
+    assert _parse_rollup_state("SUCCESS") == CiStatus.PASSING
```
evgunter: sort of silly tests

libs/mngr_kanpan/imbue/mngr_kanpan/data_sources/github_test.py
```diff
@@ -106,284 +157,293 @@ def test_get_cached_repo_field_wrong_type() -> None:
...
+    assert _parse_rollup_state("PENDING") == CiStatus.PENDING
 
-def test_build_create_pr_url() -> None:
-    url = _build_create_pr_url("org/repo", "my-branch")
-    assert url == "https://github.com/org/repo/compare/my-branch?expand=1"
 
+def test_parse_rollup_state_unrecognized_is_unknown() -> None:
+    # Future-proofing: any new enum value we don't know about should
+    # render as UNKNOWN rather than crash.
+    assert _parse_rollup_state("EXPECTED") == CiStatus.UNKNOWN
```
evgunter: this one is good though

libs/mngr_kanpan/imbue/mngr_kanpan/data_sources/github_test.py
```diff
@@ -106,284 +157,293 @@ def test_get_cached_repo_field_wrong_type() -> None:
...
+def test_parse_pr_node_minimal() -> None:
+    node = _make_pr_node(number=42, head_branch="mngr/foo")
+    pr = _parse_pr_node(node, unresolved_ignore_user=None)
+    assert pr.number == 42
+    assert pr.head_branch == "mngr/foo"
+    assert pr.state == PrState.OPEN
+    assert pr.check_status == CiStatus.PASSING
+    assert pr.is_draft is False
+    assert pr.has_conflicts is False
+    assert pr.has_unresolved is False
```
evgunter, line 229: might be possible to make this a full round trip instead of asserting specific fields

libs/mngr_kanpan/imbue/mngr_kanpan/data_sources/github_test.py
```diff
@@ -426,64 +485,62 @@ def test_compute_mixed_agents_with_and_without_repo() -> None:
...
-    assert FIELD_PR in fields[agent_with_label.name]
-    assert FIELD_CI in fields[agent_with_label.name]
+    cg = _make_board_cg(response)
+    ctx = make_mngr_ctx_with_cg(cg)
+    agent = make_agent_details(name="a1", initial_branch="branch-1", labels={"remote": "git@github.com:org/repo.git"})
+    fields, _errors = ds.compute(agents=(agent,), cached_fields={}, mngr_ctx=ctx)
+    af = fields[agent.name]
+    assert FIELD_PR in af and FIELD_CI in af and FIELD_CONFLICTS in af and FIELD_UNRESOLVED in af
+    assert isinstance(af[FIELD_CONFLICTS], ConflictsField) and af[FIELD_CONFLICTS].has_conflicts is True  # type: ignore[union-attr]
+    assert isinstance(af[FIELD_UNRESOLVED], UnresolvedField) and af[FIELD_UNRESOLVED].has_unresolved is True  # type: ignore[union-attr]
```
evgunter, line 507: vaguely suggests that we could make this a typed object instead of a dict

evgunter:
looks good! i commented on some random stuff i saw but really the only one that needs to be addressed is https://github.com/imbue-ai/mngr/pull/1737#discussion_r3295220917 (might be silently skipping stuff that should be an error)

libs/mngr_kanpan/imbue/mngr_kanpan/data_sources/github.py
qi-imbue: I don't see that anywhere, but I did move this to be a static method on `CiStatus`.

I also renamed the enums to agree with the rollup status values, which turned out slightly more invasive than I wanted, but it feels worthwhile to get rid of this gratuitous inconsistency.

libs/mngr_kanpan/imbue/mngr_kanpan/data_sources/github.py
qi-imbue: Interesting, apparently it can be `None` because the schema (https://docs.github.com/en/graphql/reference/objects#pullrequestreviewthreadconnection) says so here. Changed to a check against `None`.

Claude said that it could be `None` because of some kind of backend error or race, which I'm not totally convinced by (it also says that in its testing this never actually happens, which I am convinced by), and the proper way to resolve this is probably to turn the unresolved status into a three-valued true|false|unknown, but that feels overkill here.

libs/mngr_kanpan/imbue/mngr_kanpan/data_sources/github.py
qi-imbue: A bizarre change here, but it seems that there's some prior art in this file.

When I asked Claude why, it tells me that this is RST syntax, and I'm convinced by that.

libs/mngr_kanpan/imbue/mngr_kanpan/data_sources/github_test.py
qi-imbue: i guess that'd entail making `_make_pr_node` take `PrInfo` instead? that's not straightforward since `PrInfo` contains strictly less information than what the graphql query returns. kept it as is for now.

libs/mngr_kanpan/imbue/mngr_kanpan/data_sources/github_test.py
qi-imbue: kanpan's api wants `dict[str, FieldValue]`, so making this a typed object will require conversion, which feels overkill.

If we make kanpan's api  accept`Mapping` instead, which is covariant on its value type, we can use a `TypedDict` with tighter constraint on the values - we can do that to all data types if we want, but it also feels slightly overkill

PR #1741, opened 2026-05-25
permission-requests: preserve symlinks at the approval target

hynek-urban:
renameSync(2) acts on the symlink itself, so writing through a symlinked target replaced it with a regular file. mngr latchkey link-permissions swings per-agent opaque paths into the canonical host permissions file via such symlinks, so approving a permission request silently desynced agents from the canonical file.

Resolve the target with lstatSync + realpathSync before computing the temp path and renaming, so the atomic swap lands on the underlying file and the symlink stays in place. Adds a regression test that creates a symlinked target and asserts both that the symlink survives approval and that the merged permissions end up in the canonical file underneath.


joshalbrecht:
lgtm

PR #1742, opened 2026-05-25
Access Minds API through the Minds API proxy.

hynek-urban:
(Accompanied by https://github.com/imbue-ai/forever-claude-template/pull/107/.)

This is how agents can now access (agent-scoped) Mind API endpoints:

```
latchkey curl -XPOST \\
  http://latchkey-self.invalid/minds-api-proxy/api/v1/agents/$MNGR_AGENT_ID/notifications \\
 -d '{"title": "test", "message": "test", "urgency": "normal"}'
```

- Per-agent `$MINDS_API_KEY` keys were replaced with a single Minds API key that Latchkey Gateway uses to authenticate against the Minds API.
- Agents can only access the Minds API through Latchkey Gateway now. The direct reverse tunnel has been removed.
- Those Minds API endpoints where the caller ID (agent ID) matters have been moved under `api/v1/agents/<agent_id>/...`. Latchkey Gateway is now responsible for rejecting unauthorized agents from accessing endpoints nested under someone else's ID.
  - (With the caveat that in terms of permissions, Latchkey currently doesn't really distinguish between different agents on a single host.)
- The `LatchkeyAutoRegister` class has been introduced: it observes agent creation events and calls `register_agent_for_host()` to update the permission.json file accordingly. (Non-Minds users can use the `mngr latchkey register-agent` command instead.)
  - `mngr latchkey register-agent  --host-id <host-id> --agent-id <agent-id>`
- Latchkey's reverse tunneling logic has been consolidated with mngr_forward to avoid unnecessary duplicate logic related to ssh tunneling (while keeping the `mngr forward --reverse` still working for backwards compatibility).

Note: I abandoned the original idea of encoding the agent ID in the JWT after I realized that we don't know the agent ID at the time of JWT minting (it actually happens before the agent creation so that we can easily supply the JWT to the agent upon creation as an env var).


apps/minds/imbue/minds/cli/run.py
```diff
@@ -172,6 +180,16 @@ def run(
     latchkey = _build_latchkey(data_directory=data_directory)
     latchkey.initialize()
 
+    # Mint a fresh central minds API key for this process. The same
+    # value is handed to the latchkey gateway's ``minds-api-proxy``
+    # extension (via the supervisor restart below, so it can inject
+    # ``Authorization: Bearer <key>`` on every forwarded request) and
+    # to the desktop client's own bearer-auth gates (so they accept
+    # the header the proxy just injected). Generated in memory rather
+    # than persisted because the supervisor is always restarted on
```
joshalbrecht, line 189: that's a little strange, no?

seems weird to constantly restart the latchkey gateway... I thought we wanted to work towards a slightly different design (eg, it permanently running, even when minds is not, it running on remote hosts, etc), that this seems to take us a little farther away from

it's not a huge deal, something just feels a little off (it feels like this is making minds more "special" rather than more like other APIs). I guess we haven't figured out how to deal with other APIs having multiple accounts anyway though...

hynek-urban: Latchkey Gateway was constantly restarted even before this PR. The original motivation was upgrades: it just makes it easier when a new Minds version drops (which also potentially ships an updated version of `mngr_latchkey`). This just happens to fit nicely with that model. We keep it running when Minds the app is down.

What exactly do you feel is off? Would you like me to make some adjustments here?

hynek-urban: I could just save the API key to disk with `0600` permissions - that was the version Claude came up with first :) I told him not to do it for now because we don't really need to persist the secret at the moment.

joshalbrecht:
lgtm

PR #1748, opened 2026-05-26
Make integrator less of a snowflake, and make orchestrator mostly agnostic of agent output content

qi-imbue:
## My summary

This PR prepares for the big TMR generalization change by making the integrator bit a bit more general. Previously:

- The integrator agent was a bit of a snowflake: it always ran locally, and the orchestrator had a different code path for feeding input and retrieving output.
- The orchestrator also looked inside the `outputs.tar.gz` to decide which branches to feed to the integrator.

This PR gets rid of these two behaviors:

- The integrator is now launched basically like the testing agent, just with a different prompt.
- The orchestrator now mostly treats `outputs.tar.gz` as a blackbox, and uploads all of them to the integrator.

Also bundled a few other cleanups into this PR; Claude's summary below is decent. The net delta of the PR is negative, which is also a great sign for a simplification change!

## What changed

`mngr tmr` now uses a single orchestrator code path for every provider. Notable behavior changes:

- **Integrator runs on `--provider`**, reusing any snapshot built for the testing agents. On `--provider modal` it spins up at snapshot speed instead of running locally.
- **Snapshot building is automatic** on providers that support it (modal today). The `--use-snapshot` flag is gone. `--snapshot <ID>` still works for reusing an existing snapshot.
- **`--integrator-provider`, `--integrator-type`, `--integrator-template` are removed.** The integrator follows `--provider`, `--agent-type`, `--agent-template`.
- **Result delivery unified**: the integrator publishes its result the same way testing agents do — `outputs.tar.gz` with `test_output/` + `branch.bundle` — and the orchestrator pulls it via the same volume-poll + extract + apply-bundle path that already serves testing agents. The old rsync / `pull_git` / remote-file-poll path is deleted.
- **Local provider now matches the remote path.** Testing agents on local provider now use `GIT_MIRROR` instead of `GIT_WORKTREE`, so their branches no longer auto-appear in the orchestrator's source repo — they arrive only via the published `branch.bundle`, exactly like on a remote provider. The integrator path is then identical: the orchestrator rsyncs the local `output_dir` into `<work_dir>/.tmr_inputs/` on the integrator host and sends the integrator prompt via `send_message`. The integrator's inline bash walks the inputs dir, applies the should-pull predicate, fetches qualifying bundles into local branches, and cherry-picks.

  Tradeoff: local runs lose the GIT_WORKTREE shortcut and pay a clone-and-mirror push per agent. The win is that `mngr tmr --provider local ...` now exercises the same orchestrator path as `--provider modal`, so a passing local run gives real signal about the remote one.
- **Orchestrator no longer hands the integrator a branch list.** `launch_integrator_agent`'s `fix_branches` parameter is gone — the integrator discovers its own work by walking the uploaded inputs dir. The remaining tarball-content coupling on the orchestrator side is just `_apply_branch_bundle` in `pulling.py`.

## Manual testing

- [ ] `mngr tmr --provider modal tests/<some path>` — confirm the full remote-integrator path end-to-end: snapshot auto-builds, integrator launches from it, rsync delivery → predicate filter → cherry-pick → outputs tarball pulled back → integrated branch fetched into the local repo.
- [ ] `mngr tmr --provider local tests/<some path>` — confirm the unified path works locally too (branches arrive only via bundles, integrator gets `.tmr_inputs/` rsynced, integrated branch ends up in the source repo).
- [ ] `mngr tmr --reintegrate --run-name <prior>` — confirm reintegrate re-pulls test-agent tarballs and re-runs the integrator under the new flow.
- [ ] `mngr tmr --snapshot <existing-modal-snapshot-id> tests/<some path>` — confirm `--snapshot` short-circuits the auto-build step and the supplied snapshot is used for both testing agents and the integrator.
- [ ] `mngr tmr --max-parallel-agents 4 tests/<some path>` — confirm batched mode still works (it does not snapshot, by design).
- [ ] Inspect the integrator agent's tmux to confirm the inline-bash predicate / `git fetch` / cherry-pick loop runs cleanly with at least one filtered-out agent (one whose outcome did not pass the should-pull predicate).

## Test plan

- [x] `just test-quick libs/mngr_tmr -m 'not modal and not docker and not docker_sdk and not acceptance and not release'` — 167 passed, ratchets clean, type-checker clean.
- [ ] CI: docker / modal integration + acceptance tests.


joshalbrecht:
lgtm!

qi-imbue:
Closing this to merge https://github.com/imbue-ai/mngr/pull/1764 (stacked on top of this one) directly into main

PR #1747, opened 2026-05-26
Changelog consolidation (run 2026-05-26)

evgunter:
Automated changelog consolidation run on 2026-05-26.

PR #1750, opened 2026-05-26
skip some non-notable entries in CHANGELOG.md

evgunter:
not sure if this is exactly what we want to do, but i figure we can err on the side of a concise changelog given that we also publish the unabridged ones

inspired by a non-notable change i saw in https://github.com/imbue-ai/mngr/pull/1747

---

## Policy change

Updates the nightly changelog consolidation prompt (`scripts/changelog_consolidation_prompt.md`) so the concise per-project `CHANGELOG.md` is a notable-only summary. Previously step 3 told the consolidation agent to give "internal refactors, doc edits, or test-only tweaks" a catch-all `Changed:` bullet. Now:

- Non-notable changes are omitted from `CHANGELOG.md` entirely instead of being forced into a bullet.
- The canonical example called out is a change that only affects tests rather than user-facing behavior.
- It is explicitly fine for a project's consolidation run to produce no `CHANGELOG.md` bullets at all if none of its entries are notable.

Entries are still preserved verbatim in each project's `UNABRIDGED_CHANGELOG.md` (the consolidator inserts those sections in step 2, independent of this prompt), so nothing is dropped from the full history.

## Retroactive cleanup

Applied the new policy to all 37 existing project `CHANGELOG.md` files (via 5 parallel subagents, each owning a disjoint batch). Pruned test-only, internal-refactor, and doc-only entries — and the boilerplate "adopted the new per-project changelog layout" notes — across both `[Unreleased]` and released `## [vX.Y.Z]` sections. Empty `### Category` subheadings were removed; `## [Unreleased]` headings were preserved (required by tooling); two now-empty legacy date sections in `dev/CHANGELOG.md` were dropped. Each touched project gets the required per-PR entry file.

The pruning erred aggressive by design (to be manually reviewed). `UNABRIDGED_CHANGELOG.md` files were untouched, so every pruned entry remains in full history.

Verified: `test_every_project_has_changelog_layout` and `test_pr_has_changelog_entry` pass.


libs/mngr_antigravity/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
-# Changelog - mngr_gemini
+# Changelog - mngr_antigravity
 
-A concise, human-friendly summary of changes for the `mngr_gemini` library. Entries are categorized using the [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) categories: Added, Changed, Deprecated, Removed, Fixed, Security.
+A concise, human-friendly summary of changes for the `mngr_antigravity` library. Entries are categorized using the [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) categories: Added, Changed, Deprecated, Removed, Fixed, Security.
```
evgunter, line 3: unrelated but seemed sensible enough to fix it here

joshalbrecht:
lgtm, thanks for fixing!

PR #1751, opened 2026-05-26
Lay groundwork for faster minds e2e tests (offload 0.9.6 + VM snapshots)

joshalbrecht:
## Summary

Two changes that together set up the tooling we need to make minds end-to-end
tests dramatically faster.

- **Bump offload to v0.9.6** in both the CI workflow and the in-image
  pin baked into `libs/mngr/imbue/mngr/resources/Dockerfile`.
  [v0.9.6](https://github.com/imbue-ai/offload/releases/tag/v0.9.6) adds
  `offload run --override-image-id <ID>` (Modal provider only), which
  lets a run point at a pre-built Modal image and skip the entire
  image-setup pipeline.
- **Add `scripts/snapshot_minds_e2e_state.py`**, a demonstration script
  that creates a Modal sandbox with
  `experimental_options={"vm_runtime": True}`, installs the
  Docker + Node + pnpm + xvfb stack the `test-docker-electron` CI job
  needs, runs the existing minds Electron e2e test
  (`apps/minds/test_desktop_client_e2e.py`), and then calls
  `sandbox.snapshot_filesystem()`. The resulting Modal image ID can be
  fed to offload via `--override-image-id` so subsequent test sandboxes
  boot from an already-warm workspace + Docker container in seconds
  instead of rebuilding from scratch.

`vm_runtime` is opted into only by this script. Modal has capacity issues
with that runtime, so the general `mngr_modal` provider is unchanged.

## Test plan

- [ ] CI is green on this PR (offload v0.9.6 installs and runs as expected
      in both the `test-offload` and `test-offload-acceptance` jobs, and
      the in-image offload binary still satisfies offload's
      `apply-diff` requirement)
- [ ] Manual: `uv run python scripts/snapshot_minds_e2e_state.py --skip-test`
      produces a snapshot image id (validates the image build + bare-VM
      snapshot path)
- [ ] Manual: `uv run python scripts/snapshot_minds_e2e_state.py` runs
      the e2e test inside the sandbox to completion and prints a
      snapshot image id with the workspace's Docker container baked in


joshalbrecht:
lgtm

PR #1755, opened 2026-05-26
fix key generation race in minds

evgunter:
## Problem

The `test-docker-electron` CI job (`test_create_local_docker_workspace_via_electron`) is flaky — it failed **11 of 53** completed runs (~21%) over the last ~60 CI runs, across many branches including `main`. 10 of 11 failures are the same symptom: `page.wait_for_selector("#create-form")` times out after 10s.

## Root cause

`/create` is a server-rendered page; its handler returns `403 "Not authenticated"` (no `#create-form` in the body) whenever the session cookie fails to verify. So every `#create-form` timeout means `GET /create` came back 403.

`FileAuthStore.get_signing_key` (`auth.py`) generated the cookie signing key lazily on first access with no synchronization. FastAPI dispatches sync route handlers on a threadpool, so on the e2e test's fresh data directory the desktop client's startup burst — the `/authenticate` request plus the `/`, `/_chrome`, and `/welcome` handlers, each of which calls `get_signing_key` — all reached generation concurrently. Two interleavings (both observed in CI logs) broke auth:

- **500 path** (5/11): a thread saw the just-created key file as momentarily empty (the old non-atomic `write_text`) and raised `SigningKeyError` → `/authenticate` returned 500, no cookie set.
- **200 path** (5/11): two threads generated *different* keys and raced to write; last-writer-wins silently invalidated the cookie just signed with the earlier key, so the next request's `verify_session_cookie` failed.

(The 11th failure is a separate, rarer `wait_for_url` post-submit timeout — not addressed here.)

## Fix

`get_signing_key` now:
- reads the existing key on a lock-free fast path;
- serializes first-time generation behind a per-store `threading.Lock` with a double-checked re-read, so concurrent callers converge on one key;
- writes the key via `atomic_write` so a concurrent reader never observes a partial/empty file.

This makes `get_signing_key` correct regardless of call site (so eager startup init is unnecessary; atomic write alone would not fix the divergent-key case).

## Tests

Added `test_get_signing_key_is_consistent_under_concurrent_first_access`, which hammers `get_signing_key` from 32 threads released simultaneously and asserts they all converge on the single persisted key. Verified it reliably **fails** against the old implementation and passes against the fix. All 14 `auth_test.py` tests and all 56 minds ratchets pass locally.


joshalbrecht:
lgtm, thanks for the fix!

PR #1757, opened 2026-05-26
upgrade deps, fix resulting type errors, add test that no deps are younger than 2 weeks (unless explicitly allowed)

evgunter:
the test that no deps are younger than 2 weeks doesn't protect devs from supply chain attacks, but at least it prevents us from republishing anything younger than that. could set a fixed `exclude-newer` date and remember to bump it sometimes instead

---

## Summary

Update the type checker and dependencies (under a supply-chain cooldown), fixing every resulting type error.

- **ty 0.0.24 → 0.0.39.** ty 0.0.39 stopped honoring the bracketed PEP-484 form `# type: ignore[<mypy-code>]`; all such comments were converted to ty's own `# ty: ignore[<ty-rule>]` form.
- **`uv lock --upgrade` under a two-week cooldown** — bumped floating dependencies to the latest versions that have been public for ≥2 weeks. Notable bumps: paramiko 3.5.1→4.0.0, coolname 3.0.0→5.0.0, starlette 0.50→1.0, urwid 3.0→4.0, pydantic 2.12→2.13, cryptography 46→48, typer 0.21→0.25, uvicorn 0.40→0.46.

## Supply-chain cooldown
- `test_no_dependencies_younger_than_two_weeks` (in `test_meta_ratchets.py`) fails if any locked dependency was published within the last two weeks, except deliberately-trusted exemptions.
- uv's static `[tool.uv] exclude-newer` only accepts a **fixed** date (relative durations work only on the `uv lock --exclude-newer` CLI), so the **relative** two-week cutoff is enforced by this (time-relative) test instead. Regenerate a compliant lock with `uv lock --upgrade --exclude-newer "2 weeks"`.
- **Caveat:** this does *not* protect us from a compromise that stays undetected past two weeks, nor shield the first project to lock a fresh release. Its only value is the detection delay before we adopt (and, for runtime deps shipped in our wheels, re-propagate) a release.
- Exemptions: `ty` (dev-only checker, never in a published wheel; pinned to latest 0.0.39) and `modal` (explicitly pinned `==1.4.3`).
- A side effect of the cooldown: `wsgidav` stays 4.3.3 (4.3.4 is <2 weeks and adds a `bcrypt<5` cap + `passlib` dep), so **bcrypt stays at 5.0**.

## Dependency-resolution notes
- paramiko caps at 4.0.0: the latest `pyinfra` (3.8.0) requires `paramiko<5`, so 5.0.0 is blocked upstream regardless of the cooldown.
- `supertokens-python` floor raised to `>=0.31.3` so the resolver keeps the latest auth library instead of backtracking to 0.30.3 (which would cap `aiosmtplib<4`). Keeps `aiosmtplib` at 5.x.
- `pyinfra` 3.8.0 adds **`types-paramiko`** transitively, whose stubs make ty type-check paramiko usage for the first time (errors fixed below).

## Type fixes (real fixes preferred over suppression)
- `concurrency_group`: accumulate a typed `list[Exception]` on the exit path (behavior unchanged).
- `minds`: `_resolve_ws_name_and_account` returns `list[AccountSession]` (`list` is invariant, so `list[object]` was rejected).
- `mngr` name_generator: annotate coolname configs with `CoolnameConfigT` (coolname 5.0 widened the config type).
- `mngr` outer_host: type the paramiko `Transport`/`SFTPClient` helpers; narrow `_put_file*` to `str | IO[bytes]`.
- `mngr` / `resource_guards`: annotate pytest hookwrapper generators' send type as `pluggy.Result[...]` so `outcome.get_result()`/`.excinfo` resolve.
- `resource_guards`: bind the `fixture_uses_resources` TypeVar to a `_NamedCallable` protocol so `func.__name__` type-checks.
- `mngr_kanpan`: `_submit_batch_item` dispatches on command type with a `match` (+ `case _: assert_never`) instead of an isinstance chain, narrowing to `CustomCommand` for the `.command` access.

Suppressions only where intentional: paramiko-private monkeypatch/access, a duck-typed SSH test fake's `get_transport` override, and an urwid `Widget`→`Text` downcast that is always a `Text` by construction.

## Ratchets
- Tightened ratchet violation counts repo-wide to their exact current values via `--inline-snapshot=trim` (test-config only).
- Documented `--inline-snapshot=trim` for tightening counts in `CLAUDE.md`'s Ratchets section.

## Verification
- `uv run ty check`: returncode 0 (only the pre-existing `cel_utils` `possibly-missing-submodule` warning, unchanged from 0.0.24).
- The ty/paramiko/coolname baseline passed full CI green; this revision re-runs full CI for the broader dep set.


CLAUDE.md
```diff
@@ -84,6 +84,8 @@ Only after doing all of the above should you begin writing code.
 
 Each project has a `test_ratchets.py` file containing automated code quality checks ("ratchets"). Each ratchet tracks a count of violations for a specific anti-pattern (e.g. raising built-in exceptions, using monkeypatch.setattr). The count can only stay the same or decrease -- increasing it fails the test. To add or modify ratchets, use `/writing-ratchet-tests`.
 
+When your change *reduces* a ratchet's violations, lock in the improvement by tightening its recorded count: run `uv run pytest --inline-snapshot=trim <path/to/test_ratchets.py>` (scoped to the affected file(s), and without xdist so inline-snapshot stays enabled).
```
evgunter, line 87: flagging

joshalbrecht:
lgtm, though I would prefer a fixed date for exclude-newer in the pyproject.toml

PR #1758, opened 2026-05-26
default is_allowed_in_pytest to False; remove MNGR_ALLOW_PYTEST

evgunter:
Hardens the pytest config-loading guard around `is_allowed_in_pytest`. (Started as the small `is_error_reporting_enabled` description fix below, then grew per review discussion.)

## `is_allowed_in_pytest` hardening (main change)
**Default flipped `True` → `False`**, and the `MNGR_ALLOW_PYTEST=1` env override is **removed** (it had a single user and wasn't worth the risk of being reached for as a quick bypass instead of fixing a leaky test env).

The guard is now **secure by default** with three properties:
- **Only fires when a config file was actually loaded.** No config picked up ⇒ nothing to protect against ⇒ mngr runs fine. (A test that loads no config needs no opt-in.)
- **Per-file.** Every loaded layer (user/project/local) must set `is_allowed_in_pytest = true` on its own — the merged value isn't enough, so a leaked real config can't ride in under a test config that opts in.
- **Covers the pre-reader path too.** The guard lives in one helper (`enforce_pytest_config_opt_in`) invoked from `_resolve_config_files` — the shared chokepoint behind `read_disabled_plugins` / `read_default_command` — so config reads at CLI startup / group dispatch are guarded, not just `load_config`.

Also made `context_dir` explicit (no default) across the pre-reader chain and threaded `load_config`'s `context_dir` into `read_disabled_plugins`, fixing a latent inconsistency where the disabled-plugin set was read from cwd while the rest of the config came from `context_dir`.

### Test infra
Tests that write a config and load it now opt in within that config (every layer); multi-layer writers (e2e `settings.local.toml`, `disable_remote_providers` fixture) opt in per layer; pre-reader/dispatch/startup tests pass the explicit `context_dir`. Added coverage for the new default and the no-config-loaded path. The repo's `.mngr/settings.toml` keeps its explicit `is_allowed_in_pytest = false`.

## `is_error_reporting_enabled` description fix (small aside)
The original task: the field description was stale (claimed it controls prompting to file GitHub issues). It actually gates whether, on an unexpected interactive error, mngr suggests launching a diagnostic agent. Description now matches behavior.

## Testing
Local (non-slow markers), all green: `libs/mngr` (4001), `mngr_usage`, `mngr_claude`, `mngr_schedule` (292), `mngr_modal` (388). Modal/Docker/acceptance/e2e tiers run in CI. (`origin/main` merged in; its narrowing/extends feature in `loader.py` reconciled with this guard.)


libs/mngr/imbue/mngr/config/data_types.py
```diff
@@ -535,15 +535,15 @@ class MngrConfig(FrozenModel):
...
+        default=False,
         description=(
-            "Set this to False to prevent loading this config in pytest runs. "
-            "Tests that intentionally need to load a config with this set to False "
-            "(e.g. end-to-end tests of real mngr subprocesses) must set "
-            "MNGR_ALLOW_PYTEST=1 in the subprocess env as an explicit opt-in."
+            "Whether this config may be loaded during a pytest run. Defaults to False so a "
+            "poorly-scoped test cannot pick up a real config (e.g. ~/.mngr) and perform real "
+            "operations; configs written for tests set this to True to opt in."
         ),
```
evgunter, line 736: not sure if this should actually be in config at all (it isn't really treated like a normal config option and it's only relevant for tests; we could have a comment sentinel or something instead)

joshalbrecht:
lgtm after conflicts are fixed, thanks for the follow up!

this is quite a bit of code, but I guess is probably the right way

PR #1759, opened 2026-05-27
Fix "Deny" functionality for invalid requests.

hynek-urban:
This fixes an issue observed by Kanjun where clicking "Deny" wouldn't actually delete the requests in some cases (specifically, if a permission request's scope was invalid).


joshalbrecht:
lgtm

PR #1760, opened 2026-05-27
Validate predefined permission requests against services catalog

hynek-urban:
Validate `POST /permission-requests` payloads to prevent issues similar to what Kanjun observed here: https://imbue-ai.slack.com/archives/C0B1JACQ9HT/p1779835487077399.

(When agents get confused and ask for non-existent scopes or permissions, their requests are now politely refused right away.)


joshalbrecht:
lgtm

PR #1762, opened 2026-05-27
Bump the Latchkey version to 2.12.1.

hynek-urban:
The newest Latchkey version is capable of reusing Google Projects which is important because the default limit on Google Project count is low.


joshalbrecht:
lgtm

PR #1763, opened 2026-05-27
Fix two settings-narrowing false positives

joshalbrecht:
## Summary

Two bug fixes for the settings-narrowing safety net introduced in the recent unified-settings work:

- **Templates lost their `__extend` intent for new templates.** In `[create_templates.<name>]`, an `<opt>__extend = [...]` whose template name didn't exist in any earlier layer used to be eagerly resolved against a `None` base and stored as a bare assign. When `mngr create --template <name>` later applied that template, the bare assign tripped the narrowing guard against the create command's runtime params. The user's intent (extend the runtime value) was silently lost. Fix: preserve the `__extend` suffix inside template options when the base has nothing to extend, so `apply_create_template` can apply it as an extend against the runtime params.

- **String-form `cli_args` tripped tuple-narrowing.** `agent_types.<name>.cli_args = "..."` is shell-tokenized into a tuple before the narrowing check runs, so two layers each supplying a different string would trip the guard against each other's individual tokens. A user-typed string is a coherent single value (scalar replacement intent), not an aggregate of tokens. Fix: tuples that were normalized from string-form values are now exempt from the per-entry narrowing check (via a `StringDerivedTuple` marker subclass tracked through `model_construct`).

## Test plan

- [x] New unit tests cover both fixes end-to-end through `load_config` (including the template-application path for the first bug).
- [x] All 3994 fast `mngr` tests pass locally.
- [x] All `mngr_claude` (482), `mngr_uncapped_claude`/`mngr_opencode`/`mngr_pair`/`mngr_antigravity`/`mngr_modal` (824) tests pass locally.
- [x] Manually verified against the user's real `~/project/mngr/.mngr/settings.local.toml`: the previously-commented `env__extend` and string-form `cli_args` lines now load successfully and the template options retain the `__extend` suffix.


joshalbrecht:
lgtm, even if a bit gross

PR #1766, opened 2026-05-27
Consolidate docker_vps state container + host volume into one volume

joshalbrecht:
## Summary

- For the `docker_vps` provider, fold the per-user Alpine state container's volume and the per-host data volume into a single per-host Docker volume (`mngr-host-vol-<host_id_hex>`).
- mngr now reads/writes `host_state.json` and `agents/<agent_id>.json` directly on the VPS filesystem via the volume's docker mountpoint (resolved with `docker volume inspect`), with `host_dir/` as a sibling subdirectory of the same volume. The Alpine state container and `<prefix>docker-state-<user_id>` volume are no longer created or read.
- Discovery is unchanged in shape (one SSH round-trip per VPS), but derives the unified volume name from the VPS's single mngr-labeled container's `host-id` label rather than probing a state container.
- Spec: [`specs/vps-docker-unified-volume/concise.md`](specs/vps-docker-unified-volume/concise.md).

## Motivation

A future backup story (`docker run --rm -v <vol>:/data ... tar ...`) only works cleanly if every byte for a host lives on one named volume. The previous two-volume split (metadata in a per-user state container, host_dir in a per-host volume) existed only for historical symmetry with the local `docker` provider, which doesn't have the same 1:1 host-per-VPS invariant.

## Breaking change

Existing `docker_vps` hosts created before this change cannot be discovered or managed after upgrade. They must be `mngr destroy`'d and recreated.

Snapshot semantics are unchanged: `docker commit` still does not capture named-volume contents.

## Test plan

- [x] `just test-quick libs/mngr_vps_docker` (168 passed)
- [x] `just test-quick libs/mngr -m '...'` (4000 passed)
- [x] `just test-quick libs/mngr_vultr` (89 passed)
- [x] `just test-quick libs/mngr_ovh` (202 passed)
- [x] `just test-quick libs/mngr_vps_docker/.../test_ratchets.py` (57 passed)
- [x] `just test-quick libs/mngr/imbue/mngr/utils/test_ratchets.py` (57 passed)
- [ ] CI acceptance tests on this PR
- [ ] Manual: create a fresh vultr host with this branch; confirm only one `mngr-host-vol-*` volume exists on the VPS, that `host_state.json` and `agents/` are at the volume root, and that mngr commands (`list`, `connect`, `stop`/`start`, `destroy`) all work.


joshalbrecht:
lgtm

PR #1767, opened 2026-05-27
clearer narrowing errors

evgunter:
```
ev@mac mngr % echo '[commands.create]\npass_env = ["TEST"]' > .mngr/settings.local.toml && mngr c sjdklfjkd
Error: Settings narrowing detected: a higher-precedence settings layer would assign over a non-empty list/tuple/dict/set value from a lower-precedence layer, silently dropping the earlier entries.
  commands.create.defaults
      assigned by local settings (~/mngr/.mngr/settings.local.toml) [edit with: mngr config set --scope local ...]
      would drop a value from user settings (~/.mngr/profiles/ev-main/settings.toml) [edit with: mngr config set --scope user ...]
To opt into this assign-by-default behavior (and silence this error), set `allow_settings_key_assignment_narrowing = true` in one of the settings files above (or MNGR__ALLOW_SETTINGS_KEY_ASSIGNMENT_NARROWING=true).
To keep the additive behavior for a specific key, use the `__extend` suffix on the key in the higher-precedence layer (e.g. `env__extend = ["X=5"]`).
NOTE: the default for `allow_settings_key_assignment_narrowing` will change to True in a future version, and support for False may be removed entirely. Migrate now so the eventual default flip is a no-op for your config.
```

or

```
ev@mac mngr % MNGR__COMMANDS__CREATE__PASS_ENV='["TEST"]' mngr c sdjklfjk
Error: Settings narrowing detected: a higher-precedence settings layer would assign over a non-empty list/tuple/dict/set value from a lower-precedence layer, silently dropping the earlier entries.
  commands.create.defaults
      assigned by local settings (~/mngr/.mngr/settings.local.toml) [edit with: mngr config set --scope local ...]
      would drop a value from user settings (~/.mngr/profiles/ev-main/settings.toml) [edit with: mngr config set --scope user ...]
To opt into this assign-by-default behavior (and silence this error), set `allow_settings_key_assignment_narrowing = true` in one of the settings files above (or MNGR__ALLOW_SETTINGS_KEY_ASSIGNMENT_NARROWING=true).
To keep the additive behavior for a specific key, use the `__extend` suffix on the key in the higher-precedence layer (e.g. `env__extend = ["X=5"]`).
NOTE: the default for `allow_settings_key_assignment_narrowing` will change to True in a future version, and support for False may be removed entirely. Migrate now so the eventual default flip is a no-op for your config.
```

or

```
ev@mac mngr % mngr c sjkdljfkldj --setting commands.create.env='["TEST"]'
Error: Settings narrowing detected: a --setting override would assign over a non-empty list/tuple/dict/set value from the merged config, silently dropping the earlier entries.
  --setting: commands.create.defaults
To opt into this assign-by-default behavior (and silence this error), set `allow_settings_key_assignment_narrowing = true` in your settings.toml.
To keep the additive behavior for a specific key, switch to the `__extend` suffix on the --setting key (e.g. `--setting commands.create.env__extend='["X=5"]'`).
NOTE: the default for `allow_settings_key_assignment_narrowing` will change to True in a future version, and support for False may be removed entirely.
```

---

## Problem

Reported confusion from a real `mngr c ... --setting allow_settings_key_assignment_narrowing=true` invocation that still errored:

1. The error didn't say which configuration files or scope levels were implicated — just an opaque layer label (`project local settings`) plus the dotted key, with no file path and no indication of which lower-precedence layer held the value being dropped.
2. The error advertised `--setting allow_settings_key_assignment_narrowing=true` as a remedy, but passing it had no effect on file/env-layer narrowing.

## What this PR does

- **Both-sides attribution in the error.** Each offending key now names the assigning layer *and* the lower-precedence layer whose value would be dropped, each with the resolved file path and the matching `mngr config set --scope <user|project|local>` flag. The `MNGR__*` env-var layer is named as such (no `config set` scope). Attribution reuses `detect_settings_narrowing` against the already-merged layers (no new field-walking — keeps it within the `PREVENT_GETATTR` ratchet's existing accounting).

  ```
  commands.create.defaults
      assigned by project local settings (/…/.mngr/settings.local.toml) [edit with: mngr config set --scope local ...]
      would drop a value from project settings (/…/.mngr/settings.toml) [edit with: mngr config set --scope project ...]
  ```

- **Corrected the misleading `--setting` suggestion.** The error (and `docs/concepts/environment_variables.md`) no longer claim `--setting` can silence the guard. It now explicitly says to opt in via a settings file or the `MNGR__ALLOW_SETTINGS_KEY_ASSIGNMENT_NARROWING=true` env var, and that `--setting` cannot toggle the guard.

## Why `--setting` can't opt into this guard (the chicken-and-egg)

I first made `--setting allow_settings_key_assignment_narrowing=true` actually work, but backed it out as not worth the seam it required. The reason it's awkward:

- Narrowing can only be *detected* during the layer-by-layer merge (you compare each incoming layer against the accumulated base; a single merged config has lost the per-layer info). That merge — and the guard — live inside `load_config`, over the settings files + `MNGR__*` env vars.
- `--setting` is applied **after** `load_config`, in `setup_command_context` (and `apply_settings_to_config` is also reused standalone by `mngr_uncapped_claude`). It's the highest-precedence layer, applied even after the `on_load_config` plugin hook.
- Crucially, `--setting` **can't** be fully parsed before `load_config`: its `__extend` keys resolve against the already-loaded config (`resolve_extends(config, raw)`). So the only way to feed a `--setting` opt-in to the in-loader guard was a bespoke pre-scan that re-parsed the raw `--setting` strings just to pluck out the one boolean — a real smell for little benefit.

So the opt-in is intentionally limited to the settings files / env var, and the message now says so. (The guard still exists for `--setting`-introduced narrowing, gated by the flag as loaded from files/env.)

## Verification

Manually reproduced the user's scenario (project + local `[commands.create]` conflict): confirmed the improved both-sides message (paths + scope flags), and that `--setting allow_...=true` now errors with a message that explains why it can't opt in here. Added/updated unit + integration tests; the getattr-ratchet CI failure is fixed (the rework removed the added `getattr`).


joshalbrecht:
lgtm, thanks for the qol fix!

PR #1768, opened 2026-05-27
add exclude-newer to pyproject.toml; automatically update it at release to 2 weeks ago

evgunter:
## What

Moves the two-week supply-chain cooldown from a time-relative meta test to uv's native resolver enforcement, and refreshes the cutoff at release time (the natural maintenance point, per discussion).

## Why

The previous approach (`test_no_dependencies_younger_than_two_weeks`) used a time-relative test specifically because "uv only accepts a fixed date in static config." But `scripts/release.py` already exists as a maintenance hook, so we can keep a static `exclude-newer` fresh by bumping it at each release — which makes the cooldown *enforced at lock time* (you cannot lock a too-new package) rather than merely detected after the fact.

## Changes

- **`pyproject.toml`**: add `[tool.uv] exclude-newer` (initial `2026-05-23T00:00:00Z`). uv refuses to consider any package uploaded after the cutoff when resolving. The initial value sits just past the newest currently-locked package (`ty`, 4 days old), so the lock is unchanged and no per-package exemptions are needed.
- **`scripts/release.py`**: `update_exclude_newer` advances the cutoff to `(Pacific release date − 2 weeks)` just before re-locking, **forward-only** via `max(current, release_date − 2 weeks)`. A release cut while the cutoff is still younger than the window leaves it untouched — so a deliberately-pinned fresh dep (e.g. `ty`) is never re-excluded, which would otherwise break resolution. The root `pyproject.toml` is committed with the release.
- **`test_meta_ratchets.py`**: remove the now-redundant `test_no_dependencies_younger_than_two_weeks` and its `ty`/`modal` exemptions; uv enforces the cooldown at lock time now.
- **`scripts/release_test.py`**: unit tests for the advance, the within-window no-op, and the boundary case.

## Verification

- `just test-quick "test_meta_ratchets.py scripts/release_test.py"`: 26 passed, 0 failed.
- ruff check + format and `ty check` pass on all touched files.
- Manually exercised `update_exclude_newer` against the real `pyproject.toml`: a release "tomorrow" keeps the recent cutoff (no-op), a release 3 weeks out advances it to `release − 2 weeks` while still admitting `ty`; tomlkit preserves the comment block.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,7 @@ dependencies = [
 # when resolving, so we only adopt releases public long enough for malware to
 # surface. scripts/release.py advances it to 2 weeks before each release, but
 # only ever forward -- pushing it back would re-exclude a deliberately-pinned
-# fresh dep (e.g. `ty`) and break resolution. This only buys a detection delay;
+# fresh dep and break resolution. This only buys a detection delay;
 # it can't catch a compromise that stays undetected past the window.
 exclude-newer = "2026-05-23T00:00:00Z"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,6 +10,19 @@ dependencies = [
     "click>=8.0.0",
 ]
 
+[tool.uv]
+# Supply-chain cooldown: uv won't consider any package version uploaded after
+# this timestamp when resolving/locking, so we only adopt registry releases
+# that have been public long enough for the community to flag malware before we
+# pull them. scripts/release.py advances this cutoff toward (release date - two
+# weeks) at each release, but only ever forward -- never pushing it back to
+# re-restrict a cutoff that is still younger than the cooldown window (doing so
+# would re-exclude a deliberately-pinned fresh dep like `ty` and break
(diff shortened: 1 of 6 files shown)


scripts/release_test.py
evgunter, line 14: sigh

pyproject.toml
```diff
@@ -10,6 +10,15 @@ dependencies = [
...
 ]
 
+[tool.uv]
+# Supply-chain cooldown: uv ignores any package uploaded after this timestamp
+# when resolving, so we only adopt releases public long enough for malware to
+# surface. scripts/release.py advances it to 2 weeks before each release, but
+# only ever forward -- pushing it back would re-exclude a deliberately-pinned
+# fresh dep (e.g. `ty`) and break resolution. This only buys a detection delay;
+# it can't catch a compromise that stays undetected past the window.
+exclude-newer = "2026-05-23T00:00:00Z"
```
evgunter, line 20: only 4 days ago because of the `ty` bump

joshalbrecht:
lgtm, thanks for the fix!

PR #1769, opened 2026-05-27
replace per-project unit tests for ruff and ty with single top-level ones + add ty to the pre-push hook

evgunter:
ty and ruff will no longer run in per-project pytest--each project's run was actually running on the whole repo. they will instead run on pre-push (both ty and ruff) and precommit (ruff only). this adds a delay of ~0.7s to pre-push, but speeds up the tests more, since it's not doing the same work 36 times (ty doesn't cache so all of the work is indeed redone). this is especially meaningful on modal where sometimes the per-project `test_no_type_errors` would hit their 10s timeout

---

## Summary

Each package's `test_no_type_errors` ran `uv run ty check`, which — because the root `pyproject.toml` declares `[tool.uv.workspace] members = ["libs/*", "apps/*"]` — resolves the **workspace root** and scans every member (1186 files) on *every* invocation, regardless of `cwd`. So all ~36 per-project copies re-ran the identical full-workspace scan (~0.7s each, with no cross-process cache benefit — ty has no on-disk cache, only in-memory Salsa incrementality for its server/watch modes). The per-project `test_no_ruff_errors` were likewise redundant with the repo-wide ruff check that already lived in `test_meta_ratchets.py`.

This consolidates both to run **once** repo-wide:

- Keep a single `test_no_type_errors` and `test_no_ruff_errors` in `test_meta_ratchets.py` (in its existing `# --- Repo-wide ratchets (run once, not per-project) ---` section). The ty test carries `@pytest.mark.timeout(60)` because `uv run ty check` can exceed the default 10s under cold-cache/loaded-runner conditions; it is intentionally **not** marked `flaky` (the check is deterministic, so a retry would only risk masking a real failure).
- Remove `test_no_type_errors` / `test_no_ruff_errors` (and their now-unused imports) from all 36 per-project `test_ratchets.py` files, and drop them from the meta-ratchet expected-test-name set (`_get_expected_ratchet_test_names`).
- Remove the now-dead `check_no_ruff_errors` helper from `imbue_common/ratchet_testing/ratchets.py` (its only callers were the deleted per-project tests; the repo-wide ruff test inlines its own `ruff check` + `ruff format --check`). `check_no_type_errors` is kept and reused by the repo-wide ty test.

### Local type-check gate (ty pre-push hook)

`ruff` was already in pre-commit, so dropping its per-project copies lost no local gate. But `ty` was **not**, so after consolidation a scoped run like `just test-quick libs/<project>` no longer type-checked at all (the repo-wide test in `test_meta_ratchets.py` isn't collected when you scope pytest to a single project). To close that gap, this adds a `ty` hook in `.pre-commit-config.yaml`. Because ty scans the whole workspace and can't scope to staged files (a fixed ~0.7s scan), it is staged at **`pre-push`** rather than per-commit — the commit loop stays fast and ty still gates before code leaves the machine. The single `test_no_type_errors` remains the CI backstop. Verified end-to-end: the hook is skipped at the commit stage, runs on `git push`, passes on a clean tree, and fails on a deliberately-introduced type error.

Type/lint coverage is unchanged — the single repo-wide checks are equivalent to (ty) or a strict superset of (ruff) the per-project ones. Under local xdist the ratchet group no longer serializes ~36 redundant scans on one worker; under offload they no longer fan out into ~36 sandboxes that each re-scan the whole workspace.

Docstrings and one changelog entry per touched project (36 projects + the synthetic `dev` bucket for the root-level files) are updated to match.

## Testing

- `just test-quick "test_meta_ratchets.py"` — 18 passed (incl. `test_no_type_errors`, `test_no_ruff_errors`, `test_all_test_ratchets_files_have_same_tests`).
- `just test-quick "scripts/sync_common_ratchets_test.py <sample per-project test_ratchets.py>"` — 117 passed.
- Manually verified the `ty` hook staging: skipped at `pre-commit`, runs at `pre-push` (and on a real `git push`), and fails on an introduced type error.
- Full unit/integration + acceptance suites run in CI via offload (`test-offload-acceptance` green, which exercises the per-project changelog-entry check).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,10 +15,8 @@ repos:
         language: system
         types: [python]
     -   id: ty
-#        ty resolves the uv workspace root and scans every member regardless of which files
-#        changed, so it runs over the whole workspace (pass_filenames: false) whenever any
-#        Python file is staged. This is the local type-check gate; the single
-#        test_no_type_errors in test_meta_ratchets.py is the CI backstop.
+#        ty scans the whole uv workspace regardless of which files changed, so it does
+#        not take staged filenames (pass_filenames: false).
         name: "Python type checker (ty)"
         entry: bash -c 'uv run ty check'
         language: system
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,6 +14,16 @@ repos:
         entry: bash -c 'uv run ruff check --fix --force-exclude --config pyproject.toml "$@" && uv run ruff format --force-exclude --config pyproject.toml "$@"' --
         language: system
         types: [python]
+    -   id: ty
+#        ty resolves the uv workspace root and scans every member regardless of which files
+#        changed, so it runs over the whole workspace (pass_filenames: false) whenever any
+#        Python file is staged. This is the local type-check gate; the single
+#        test_no_type_errors in test_meta_ratchets.py is the CI backstop.
+        name: "Python type checker (ty)"
+        entry: bash -c 'uv run ty check'
+        language: system
+        types: [python]
+        pass_filenames: false
     -   id: regenerate-cli-docs
         name: "Regenerate CLI markdown docs"
         entry: bash -c 'uv run python scripts/make_cli_docs.py && git diff --quiet libs/mngr/docs/commands/ libs/mngr/README.md'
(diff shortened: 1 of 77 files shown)


joshalbrecht:
lgtm, thanks for fixing!

I'm not totally sure on that pre-push hook, might disable that if it is meaningfully slow for me

PR #1771, opened 2026-05-27
minds: bundle real git binary instead of xcode-select shim

weishi-imbue:
## Summary

Packaged `minds.app` SIGKILLs on `git clone` on Macs without Xcode CLT at the expected path. Cause: `scripts/build.js` ran `which git`, which on macOS returns the 118 KB xcode-select shim. Bundling the shim into the sandboxed app and re-signing it makes runtime `git` calls SIGKILL.

## Fix

New `apps/minds/scripts/download-binaries.js` resolves git per-platform:

- **macOS**: `xcrun --find git` for the real binary, plus `libexec/git-core/` helpers and templates. Skips Apple's ~100 argv-shim symlinks (each dereferenced would add ~7.6 MB).
- **Linux**: `which git`.
- **Windows**: SHA256-verified MinGit download from git-for-windows releases.

`build.js` imports `downloadGit` from the new module. `todesktop.json` bumps `uploadSizeLimit` from 300 → 600 MB (the real binary plus libexec pushes the bundle over the prior limit).

## Why bundle git instead of using the user's git

Minds targets non-engineers. The alternatives are user-hostile:

| Platform | Without bundled git |
|---|---|
| macOS | `/usr/bin/git` is a shim; first invocation pops Apple's "Install command line developer tools" dialog → ~700 MB download → license accept → 8–30 min. Dialog says Xcode, not minds. |
| Linux | Distro-dependent; minimal images lack git. |
| Windows | No system git at all. |

Precedent: **GitHub Desktop** and **SourceTree** (non-eng audiences) bundle git via [dugite-native](https://github.com/desktop/dugite-native). VS Code / Cursor (engineer audiences) use system git.

A cleaner long-term version is to pull dugite-native tarballs for macOS/Linux instead of copying from the build host (eliminates the Xcode-CLT dependency on the build machine and makes builds bit-for-bit reproducible). Deferred.

## Provenance

First carve-out from #1317 (`wz/minds_onboard`). Stacked follow-up #1772 activates the `todesktop:beforeInstall` hook and restores the uv/pnpm bootstrap paths.

## Test plan

- [x] `node -e "require('./apps/minds/scripts/download-binaries.js').downloadGit('/tmp/git-test', { platform: 'darwin' })"` produces a 7.6 MB universal Mach-O at `/tmp/git-test/git/bin/git`; `git --version` → `2.50.1 (Apple Git-155)`.
- [x] `/tmp/git-test/git/libexec/git-core/` contains 31 real helper binaries.
- [x] `git init` against the bundled binary succeeds.
- [x] `just test-quick apps/minds/scripts/build_test.py` — 1 passed.
- [x] `just test-quick apps/minds/imbue/minds/test_ratchets.py` — 57 passed.
- [ ] Full `just test-offload` — deferred to CI.
- [ ] Packaged ToDesktop build — deferred; build-server path unchanged from `origin/main`.

### DIFF
--- a/apps/minds/changelog/mngr-extract-bundled-git.md
+++ b/apps/minds/changelog/mngr-extract-bundled-git.md
@@ -1,0 +1,12 @@
+- `apps/minds`: bundle the real macOS `git` binary plus its `libexec/git-core`
+  helpers instead of the xcode-select shim. The previous inline `downloadGit()`
+  in `scripts/build.js` ran `which git`, which on macOS returns the 118 KB
+  `/usr/bin/git` shim -- a launcher that re-invokes the real git from
+  `/Library/Developer/CommandLineTools/`. Bundling the shim into a sandboxed
+  packaged app meant runtime `git clone` SIGKILLs on any Mac without Xcode CLT
+  installed at the expected path. The new `scripts/download-binaries.js`
+  resolves the real binary via `xcrun --find git`, copies it plus its
+  `libexec/git-core` helpers and templates, and SHA256-verifies all
+  downloaded archives. Also bumps the ToDesktop `uploadSizeLimit` from 300 to
+  600 because the real binary plus its libexec push the bundle over the
+  previous limit.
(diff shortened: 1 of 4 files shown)


joshalbrecht:
lgtm!

this is kind of disgusting, but if it works, great :)

PR #1773, opened 2026-05-28
add examples for mngr usage with cron

evgunter:
## Summary

Adds a cookbook doc, `libs/mngr_usage/docs/cron_recipes.md` (linked from the README), for driving `mngr` from `cron` using **check mode** (`mngr usage --format json`) instead of the blocking `mngr usage wait` — aimed at *recurring* usage-driven automation. Includes a shared `spare-capacity.sh` helper.

Recipes:

- **Use up an about-to-expire 5h window** — one cron job owns a dedicated agent's whole lifecycle, branching on its state each tick: start it in the tail of an open 5h window when there's budget left and the week is on pace; stop it once the window rolls over or the week falls off pace (no scheduled `at` stop; running one tick into the fresh window warms it). The weekly guard is a tapering-margin pace check for START (`used% < elapsed% * (1 - 0.30*(100-elapsed%)/100)`) and a looser `used% = elapsed%` line for STOP — the gap is a hysteresis band so it can't thrash. Keeps age-stale readings; skips an already-reset window.
- **Warm a fresh 5h window early** — when the last recorded 5h window has elapsed (`resets_at < now`), nudge a *reused* warming agent (create once, then start/message/stop via `mngr wait … WAITING`, never destroy) to open the next window. Stopping rather than destroying keeps its usage reading, so the trigger self-clears (no marker).
- **Dispatch a queue of task files** — pop a task file, name the agent after it, launch it from the project repo, gated on `spare-capacity.sh`. Pool agents carry `queue=live`; finished (`WAITING`) ones are stopped and relabeled `queue=in-review` (slot freed, parked for `mngr list --label queue=in-review`); the cap counts only `RUNNING`.

## Verification

- Validated every `jq` predicate against sample snapshots: the use-extra START/STOP/KEEP classifier (hysteresis gap + the empty/no-data case), the warm elapsed check, `spare-capacity.sh`'s `jq -e` exit codes (0 spare / 1 no-spare / 4 no-data), and the weekly pace formula across the cycle.
- Confirmed `mngr` syntax against the CLI: `create --from :PATH --label`, `start`/`message`/`stop`, `mngr wait <agent> WAITING --timeout`, `mngr label --label` (set/overwrite; no remove), and `labels.queue == ...` / `state == ...` filters; `seven_day` carries `window_seconds: 604800` so `elapsed_percentage` is derived.
- Could not run the warm/dispatch recipes end-to-end (they'd spin up real agents); the agent-lifecycle flows are syntactically verified, not live-tested.

Docs-only change; no production code touched.


libs/mngr_usage/docs/cron_recipes.md
```diff
@@ -0,0 +1,222 @@
...
+
+## Scheduling
+
+`cron` runs with a bare `PATH`, so set one that finds `mngr`, `jq`, and `claude`:
+
+```cron
+PATH=/usr/local/bin:/usr/bin:/bin:/home/you/.local/bin
+
+*/5 * * * * /path/to/your/script.sh
+```
```
evgunter, line 222: could be cut since claude already knows this,  but seems harmless

joshalbrecht:
lgtm, thanks for the write-up!

PR #1774, opened 2026-05-28
add accuracy review to changelog job

evgunter:
## What

The nightly changelog consolidation agent now reviews the `CHANGELOG.md` bullets it just generated for factual accuracy against the code, before pushing and opening its PR. After committing the consolidation, it spawns one or more fresh-context `general-purpose` reviewer subagents (spec: `scripts/changelog_accuracy_reviewer.md`) and **partitions the projects that gained new bullets across them at its discretion** — a trivial change touching every package can go to a single reviewer; a large run can be split. Each verifies its assigned projects' newly-added bullets against the actual code, correcting/removing inaccurate ones and collapsing bullets that another bullet materially supersedes. Guards against stale or inaccurate changelog entries.

## Why not the real `/autofix`

We deliberately did **not** route this through the `/autofix` skill or a registered `subagent_type`:

- `/autofix` ships in the `imbue-code-guardian` Claude Code plugin, which is **not staged into the changelog sandbox** (the schedule deploys with `--exclude-user-settings`).
- A registered project `subagent_type` would need `project` in `--setting-sources`, which would also load the repo's `.claude/settings.json` hooks; and `verify-and-fix` only reviews a diff against fixed code-issue categories — not built to check changelog-vs-code accuracy.

So the reviewer is a committed, autofix-inspired spec that built-in `general-purpose` subagents read and follow. Zero deploy plumbing (a tracked `scripts/` file rides along in the repo checkout); relies only on the native `Task` tool + the built-in agent type (the schedule already disables `claude_subagent_proxy`, so `Task` runs natively).

## Behavior

- **Reviewers partitioned at the orchestrator's discretion, run in parallel.** Each project that gained bullets is assigned to exactly one subagent (disjoint partitions). One subagent can cover many trivially-touched packages, or a heavy project can get its own; a failed reviewer may be retried, and never fails the whole run.
- **Each reviewer edits + commits only its assigned `CHANGELOG.md` files.** The code is ground truth; reviewers never modify source. Each stages only its own files (never `git add -A`), so the parallel committers don't clobber each other. It may collapse a pre-existing `[Unreleased]` bullet that one of its new bullets materially supersedes (within the same project).
- **Unattended:** reviewers self-review (no `AskUserQuestion`) and commit their own corrections. They report their findings back to the orchestrator, which decides what to do with them (PR body content is at its discretion).
- **Outcome JSON is per-status:** success `{status, pr_url}`, no-entries `{status}`, failure `{status, notes}` (failing step + error detail).

## Files

- `scripts/changelog_accuracy_reviewer.md` — reviewer spec (handles a set of assigned projects)
- `scripts/changelog_consolidation_prompt.md` — new step 8 (spawn + partition reviewers), renumbered steps, per-status outcome schema
- `scripts/setup_changelog_agent.sh` — header/echo comments updated
- `dev/changelog/mngr-changelog-autofix.md` — changelog entry

## Validation

- Changelog meta-tests + changelog unit tests pass locally; no Python logic changed.
- `/verify-architecture` (positive verdict; findings addressed) and `/autofix` (clean; one stale-comment fix kept) run across iterations.
- True end-to-end behavior (a real consolidation run spawning the reviewers in the Modal sandbox) requires a deploy with `CHANGELOG_VERIFY=quick` or the next nightly — not triggered here.


scripts/changelog_consolidation_prompt.md
```diff
@@ -123,24 +124,57 @@ the right project's consolidated files.
...
+   change that touched many packages can be reviewed by one subagent
+   covering all of them; a large run, or a project with many substantial
+   bullets, is better split across several subagents so no one subagent is
+   overloaded. Assign each project to **exactly one** subagent (disjoint
+   partitions, so no two subagents touch the same file). Spawn them **in
+   parallel** (issue all the Task calls in a single batch). Give each
+   subagent exactly this prompt, with the project directory or directories
+   you assigned it substituted in: "Read
+   `scripts/changelog_accuracy_reviewer.md` and follow its instructions
+   exactly. You are assigned these project(s): `<project_dirs>` -- review
```
evgunter: fake skill was more convenient than a real skill

scripts/changelog_consolidation_prompt.md
```diff
@@ -5,21 +5,19 @@ ephemeral Modal sandbox. The schedule creates a fresh worktree at
...
 
 ```
-{
-  "status": "done" | "skipped-no-entries" | "failed",
-  "pr_url": "<url>" | null,
-  "notes": "<freeform human-readable string; multi-line ok>"
-}
+{"status": "done", "pr_url": "<url>"}
+{"status": "skipped-no-entries"}
+{"status": "failed", "notes": "<failing step number + error detail>"}
```
evgunter, line 16: we don't actually use the returned value for anything, it just seems good to have

joshalbrecht:
lgtm

PR #1775, opened 2026-05-28
Tolerate per-host SSH failures in discovery; preserve SSH-dead containers' agents via offline fallback

gnguralnick:
## Problem

In the minds desktop app, killing `sshd` inside a running Docker container (e.g. `pkill sshd` in `minds-staging-babaabab`) made the workspace **disappear** from the landing-page tile list and from the chrome SSE workspaces payload. A fully `docker stop`'d container, by contrast, keeps its workspace visible -- which is the behavior we want the SSH-dead case to match.

Relatedly, a single unreachable host (sshd hang, banner-reset, auth failure) on a Docker provider used to poison the *entire* provider's enumeration: `mngr_forward`'s resolver blanked its known-agents set, every workspace on that daemon began 503'ing ("Backend not yet available"), and minds' readiness probe redirected healthy workspaces to the recovery page.

## Root cause

The default `ProviderInstanceInterface.discover_hosts_and_agents` submits a `_discover_agents_on_host` future per host, then reads results back via `future.result()`.

1. When one host's future raised `HostConnectionError`, the comprehension re-raised; the error bubbled to `_construct_and_discover_for_provider` and was recorded as a per-provider failure. The resulting `DISCOVERY_FULL` event reported `agents=[]` / `hosts=[]` for the whole provider, so `mngr_forward` blanked its resolver and every workspace on the provider became unrouteable.

2. Even once the per-host error was tolerated (host kept, agents set to `[]`), a RUNNING-but-SSH-dead container still lost its agents: `get_host` returns an online `Host` (because the container is running), whose `discover_agents()` SSHes and raises `HostConnectionError`. A fully-stopped container instead gets an `OfflineHost`, which reads agent state from the persisted mngr-ctx records and succeeds -- which is why stopped containers stayed visible but SSH-dead ones did not.

## Fix

`discover_hosts_and_agents` now handles per-host `HostConnectionError` in three steps:

1. **Tolerate** the failure per host instead of aborting the whole provider (`HostAuthenticationError` / `HostOfflineError` are subclasses, so the base catch covers them). The catch stays narrow -- programming errors and non-connection `MngrError`s still abort.
2. **Invalidate caches** via `self.on_connection_error(host_id)`, matching `get_host_and_agent_details`, so providers that cache per-host state (docker's container cache, modal/lima/vps_docker host caches) drop the wedged entry instead of replaying a stale handle next cycle.
3. **Fall back to the offline view**: try `self.to_offline_host(host_id).discover_agents()` before degrading to `[]`. A RUNNING-but-SSH-dead Docker container thus preserves its agents in discovery -- the persisted records carry the same `workspace` / `is_primary` labels minds filters on, so the workspace stays visible, mirroring fully-stopped containers. Providers without an offline view (`NotImplementedError`) or without a persisted host record (`HostNotFoundError`) degrade to the previous empty-list behavior. On successful fallback the log is downgraded to debug; the warning only fires when both online and offline paths fail.

## Tests

In `libs/mngr/imbue/mngr/interfaces/provider_instance_test.py`:

- `test_discover_hosts_and_agents_tolerates_per_host_connection_error` -- two-host provider, one host's `discover_agents` raises `HostConnectionError` with no offline record; asserts the healthy host's agents come through, the broken host appears with `[]`, both hosts had `disconnect()` called, and `on_connection_error` fired for the broken host.
- `test_discover_hosts_and_agents_falls_back_to_offline_on_connection_error` -- online path raises, offline view returns one agent; asserts the agent surfaces **and its labels (`workspace`, `is_primary`) are preserved** through the fallback (so minds' filter still matches).
- `test_discover_hosts_and_agents_offline_fallback_handles_not_implemented` -- provider whose `to_offline_host` raises `NotImplementedError` degrades to `[]`.

All pass locally, along with the full `libs/mngr/imbue/mngr/interfaces` suite (91 tests) and the docker provider suite (`libs/mngr/imbue/mngr/providers/docker/instance_test.py`, 63 tests).

## Test plan

- [x] New unit tests pass
- [x] `libs/mngr` interface + docker provider suites pass locally
- [x] `/autofix` clean, `/verify-conversation` clean, `vet` clean
- [ ] CI green (currently blocked on a pre-existing main breakage fixed by #1789)


libs/mngr/imbue/mngr/interfaces/provider_instance.py
```diff
@@ -446,7 +447,56 @@ def discover_hosts_and_agents(
...
+                # back here keeps the host's workspaces visible in
+                # discovery, matching the behavior of a fully-stopped
+                # container. Providers without an offline view
+                # (NotImplementedError) or without a persisted record for
+                # this host (HostNotFoundError) degrade to the legacy
+                # empty-list behavior.
+                try:
+                    offline_agents = self.to_offline_host(host_ref.host_id).discover_agents()
+                except (NotImplementedError, HostNotFoundError):
+                    logger.warning(
```
weishi-imbue: Have we tried `--on-error continue` during mngr list from minds app? would that work? 

My understand is that each provider instance should throw errors as is, and then they're caught and handled in the discovery/listing layer.

gnguralnick: I think --on-error continue refers to per-provider error (e.g. if docker was entirely gone or modal is entirely not working) whereas this issue is relating to per-host errors (e.g. one host has ssh broken)

gnguralnick: currently if you have a bunch of docker minds and one of them has ssh broken (e.g. by running `pkill sshd` inside the container) it'll cause the app to show that there are 0 docker minds. and even with the basic fix of iterating one at a time we still don't see that host with ssh broken despite it having a state dir on disk

libs/mngr/imbue/mngr/interfaces/provider_instance.py
```diff
@@ -446,7 +447,56 @@ def discover_hosts_and_agents(
                     host_ref.host_id,
                 )
 
-        return {host_ref: future.result() for host_ref, future in future_by_host_ref.items()}
+        results: dict[DiscoveredHost, list[DiscoveredAgent]] = {}
+        for host_ref, future in future_by_host_ref.items():
+            try:
+                results[host_ref] = future.result()
+            except HostConnectionError as e:
```
joshalbrecht, line 462: FYI, if this is ever transient (it's not in this case, but just flagging), then it should be retried at the level of a tenacity wrapper on `_discover_agents_on_host` (try a few times, quickly, within a few seconds)

Otherwise if there is a transient network failure you'll suddenly think that one of the hosts is offline, which will be fairly confusing

Nothing to change right now, just flagging so you know how to address that failure if you ever see it!

joshalbrecht: Although now that I think about it, it's also potentially worth updating the comments here to include that info too, so the future claudes know, ie, maybe leave a FIXME

libs/mngr/imbue/mngr/interfaces/provider_instance.py
```diff
@@ -446,7 +447,56 @@ def discover_hosts_and_agents(
...
+                # the docker daemon (labels + on-host-volume data). Falling
+                # back here keeps the host's workspaces visible in
+                # discovery, matching the behavior of a fully-stopped
+                # container. Providers without an offline view
+                # (NotImplementedError) or without a persisted record for
+                # this host (HostNotFoundError) degrade to the legacy
+                # empty-list behavior.
+                try:
+                    offline_agents = self.to_offline_host(host_ref.host_id).discover_agents()
+                except (NotImplementedError, HostNotFoundError):
```
joshalbrecht: Two issues:
1. I struggle to imagine why we would ever catch NotImplementedError--how could that be right?
2. Catching anything here (and simply warning instead) isn't really allowed--discover *must* fail if we failed to properly even understand which hosts exist. What case prompted you to try to catch these errors?

libs/mngr/imbue/mngr/interfaces/provider_instance.py
gnguralnick: good point - i can add that

libs/mngr/imbue/mngr/interfaces/provider_instance.py
gnguralnick: updated to remove this - we assume that any host provider whose hosts can raise HostConnectionError also implements to_offline_host . currently the only exception to this is the SSH provider - added a fixme for that.

libs/mngr/imbue/mngr/interfaces/provider_instance.py
weishi-imbue: As per prev discussion with Josh, my understanding is that  `--on-error continue` should penetrate all layers (provider,host,agent) and all phases (discover, fetch details). 
However I acknowledge the current state isn't perfectly aligned. (That's what made this complicated. And I've struggled with it a lot as well. )

I think if you can figure out to cleanly guard this by `--on-error` that'll be ideal!

gnguralnick: i don't think the minds app is actually running `mngr list` here so i'm not sure it is possible to apply `--on-error continue`

gnguralnick: from some more investigation - the minds app uses `mngr observe --discovery-only`, not `mngr list` - this already runs with ErrorBehavior.CONTINUE . so actually this bug happens despite `--on-error continue` being effectively set. so maybe the issue here is that we're not fulfilling the contract that --on-error continue sets. will investigate more

joshalbrecht:
lgtm now that this is so minimal!

PR #1778, opened 2026-05-28
Automatic auth browser prepare

hynek-urban:
When logging in to a Google service for the first time in Minds, before this change, the login would fail on the missing `latchkey auth browser-prepare` step (which creates the Google Cloud project and OAuth client) and the user would have to ask the agent to run this manually. Now it should happen automatically.


joshalbrecht:
lgtm

PR #1770, opened 2026-05-27
Move VPS docker per-host volume onto a btrfs subvolume

joshalbrecht:
## Summary
- Per-host unified docker volume on Vultr/OVH VPSes is now backed by a btrfs subvolume on a loop-mounted btrfs filesystem (`/var/lib/mngr-btrfs.img` -> `/mngr-btrfs/<host_id_hex>`), so the agent data is eligible for consistent ``btrfs subvolume snapshot -r`` (in a follow-up PR).
- Docker keeps default `data-root=/var/lib/docker` and `storage-driver=overlay2` on the ext4 root; only this one volume's storage lives on btrfs. No btrfs-aware docker volume plugin.
- The named volume continues to exist, now created via ``docker volume create --driver=local --opt type=none --opt device=<path> --opt o=bind``; `host_store.py` reads the on-disk path from ``Options.device``.
- ``_prepare_btrfs_on_outer`` is idempotent step-by-step (installs btrfs-progs, fallocates the loop file, mkfs, mounts, appends `/etc/fstab` line, creates subvolume) so partially-failed creates can be retried cleanly. `destroy_host` best-effort `btrfs subvolume delete`s before removing the volume.
- New `VpsDockerProviderConfig` fields: `btrfs_mount_path`, `btrfs_loop_file_path`, `outer_disk_reserved_gb` (default 20GB held back from the loop file).

## Breaking change
Existing vultr / ovh hosts on the prior named-volume layout cannot be discovered or managed after upgrade. Destroy and recreate them. (Same shape as the earlier "two-volume consolidation" change.)

## Spec
``specs/vps-docker-btrfs/concise.md``

## Test plan
- [x] Unit + integration tests pass locally for `libs/mngr_vps_docker` (200), `libs/mngr_vultr` (89), `libs/mngr_ovh` (202)
- [x] Dependents pass: `libs/mngr` (4028), `apps/minds` (824)
- [x] Meta-ratchets pass (changelog entries, ruff lint, format)
- [ ] CI green
- [ ] Manual: `mngr create --provider vultr` on a fresh VPS lands `host_state.json` under `/mngr-btrfs/<hex>/`, `mngr destroy` cleans up the subvolume; same for `--provider ovh`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1782, opened 2026-05-28
Make imbue_cloud destroy_host wipe data then release the lease

joshalbrecht:
## Summary
- `mngr destroy <agent>` against an imbue_cloud-leased pool host is now *terminal*: it wipes the user's data on the leased VPS, then releases the lease back to the pool.
- Privacy-first ordering: data is gone before the connector flips the row to `released`, so the eventual VPS-destroy by `cleanup_released_hosts.py` is belt-and-suspenders rather than the only barrier.
- Wipe steps (on the VPS, best-effort, single SSH session): stop+`rm -fv` workspace container → `docker volume rm` the per-host named volume → `btrfs subvolume delete /mngr-btrfs/<host_hex>` → `docker system prune -a -f --volumes` → wipe `/root` (preserving `authorized_keys`) and `/tmp`.
- `mngr stop <agent>` unchanged: still the "I'll resume this workspace later on the same VPS" path that preserves the lease + on-disk volume.
- `mngr delete <agent>` (the GC path after the destroyed-host grace period) now runs the same wipe+release flow; safe to re-run on an already-released lease.
- The wipe script is a pure free function (`build_pool_host_wipe_script`) so the rendered shell is unit-testable without an SSH transport.

## Base
Stacked on `mngr/vps-docker-btrfs` (PR #1770) because the wipe needs to know about the per-host btrfs subvolume layout that PR introduces.

## Test plan
- [x] `just test-quick libs/mngr_imbue_cloud` -- 129 passed
- [x] `test_meta_ratchets` -- 18 passed
- [x] End-to-end: ran the new `destroy_host` against two live OVH pool leases in `dev-josh-1` (one online with the new btrfs storage, one offline from an older bake). Both rows flipped to `status=released` cleanly; the online host's `Wiped pool VPS data for host ...` debug line confirmed the wipe SSH ran end-to-end.
- [ ] CI

### DIFF
--- a/libs/mngr_imbue_cloud/README.md
+++ b/libs/mngr_imbue_cloud/README.md
@@ -58,13 +58,25 @@ The pool host is fully pre-provisioned, so mngr's create pipeline only
 writes the agent env file (and patches the claude config when an
 `ANTHROPIC_API_KEY` lands in env) before starting the tmux session.
 
-## Destroy vs delete
+## Destroy / delete / stop
 
-- `mngr destroy <agent>` stops the docker container on the leased VPS only;
-  the lease and on-disk data are preserved. `mngr start <agent>` brings it
-  back on the same VPS.
+- `mngr destroy <agent>` is *terminal* for imbue_cloud-leased hosts. It
+  (1) stops + removes the workspace container, drops the per-host docker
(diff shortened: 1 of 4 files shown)


joshalbrecht:
lgtm

PR #1784, opened 2026-05-28
Destroy hosts immediately when last agent is destroyed

joshalbrecht:
## Summary
- `mngr destroy` now actually destroys the host when the last agent on it is destroyed (the documented contract), regardless of how recently the host was created.
- Previously this fired through the post-destroy GC pass, which gates on `min_online_host_age_seconds` (default 10 min), so any host destroyed within minutes of creation leaked cloud-side resources until the 7-day destroyed-host grace period eventually triggered `provider.delete_host`. End result: leases / VPSes stuck around long after the user destroyed their workspace.

## Two changes in `destroy.py`
1. **Partition reconciles discover-vs-on-host disagreement.** Matched agents that exist in the provider's discover but not in `online_host.get_agents()` ("ghosts") are now routed to host-level destruction (`provider.destroy_host`) instead of silently dropped. Fixes the "No agents found" message when the same agent was destroyed twice on an imbue_cloud host (first destroy removed `/mngr/agents/<id>/` on the VPS; the connector's lease list still reported the agent).
2. **Post-loop sweep destroys hosts whose last agent was just destroyed.** For each online host that had an agent destroyed by this invocation, re-check `host.get_agents()` and call `provider.destroy_host` directly if empty. Bypasses GC's age filter; the GC pass that runs after is the safety net for transient failures.

## C: minds destroying.py comment
The stale "Lease release is intentionally NOT chained here" docstring is updated to describe the new contract: `mngr destroy` calls `provider.destroy_host` automatically when the last agent on a host is destroyed, which for imbue_cloud wipes data + releases the lease at destroy time.

## Base
Stacked on `mngr/imbue-cloud-destroy-wipes` (PR #1782), which makes `provider.destroy_host` for imbue_cloud actually do the wipe + release. Without that, the destroy-on-empty contract here would just stop the container without releasing the lease.

## Test plan
- [x] `just test-quick libs/mngr` — 4033 passed
- [x] `just test-quick libs/mngr_imbue_cloud libs/mngr_vps_docker libs/mngr_ovh libs/mngr_vultr` — 622 passed
- [x] `just test-quick apps/minds` — 824 passed
- [x] `test_meta_ratchets` — 18 passed
- [ ] CI green
- [ ] Manual end-to-end: bake a fresh pool host, lease via minds UI, destroy via UI, confirm `minds pool list` shows the row as `released` immediately (no 7-day wait)


joshalbrecht:
lgtm

PR #1786, opened 2026-05-28
run uv sync --all-packages pre-push; use uv run pre-commit install instead of manual hook installer

evgunter:
i'm not sure if i fully understand why `scripts/githooks/install.sh` and `scripts/githooks/pre-commit` were created in the first place, so the change that removed them may not be correct 

---

## What

Two related changes to how pre-commit hooks behave and how they're installed:

1. **New `uv-sync-pre-push` hook** in `.pre-commit-config.yaml` (registered for the `pre-push` stage, ordered as the **first** local hook) that runs `uv sync --all-packages` before a push **only when** that push's commit range changes `uv.lock` or any `pyproject.toml`.
2. **Retire the hand-written git-hook installer**: delete `scripts/githooks/install.sh` and `scripts/githooks/pre-commit`, and update `scripts/ruff-precommit-setup-guide.md` to install hooks with `uv run pre-commit install`.

## Why

**The hook.** The code-guardian stop hook merges `origin/main` into the working branch and then pushes the merge commit. When that merge adds a new workspace member (or otherwise changes dependencies), the local shared `.venv` is left stale. The other pre-push hooks (`ruff`, `ty`, `regenerate-cli-docs`, `compile-style-guide`) all shell out to `uv run`, which does **not** install all workspace members on its own — so they import a member missing from the env and fail. This was hit for real on PR #1773: a merge of main that added `libs/mngr_mapreduce` (imported by `mngr_tmr`) made `regenerate-cli-docs` fail with `ModuleNotFoundError: No module named 'imbue.mngr_mapreduce'`. Running `uv sync --all-packages` **first** (before those hooks) populates the env so they pass. The complementary removed-member case is already handled by the existing `clean-stale-workspace-dirs` post-checkout hook.

**The retire.** The hand-written `scripts/githooks/` shim + `install.sh` existed to avoid `pre-commit install` depending on the system Python. But running `pre-commit install` through uv (`uv run pre-commit install`) already pins the generated hooks to the uv-managed `.venv`, so the shim was redundant. The symlink installer was also incomplete — it only ever installed the `pre-commit` hook, never the `pre-push` or `post-checkout` hooks the config relies on — whereas `pre-commit install` installs every type listed in `default_install_hook_types` and writes self-contained shims that are robust across git worktrees.

## Notes

- The hook is scoped via `files: '(^|/)(uv\.lock|pyproject\.toml)$'`, so pushes that don't change dependency files skip it (negligible overhead). A newly added workspace member always adds its own `pyproject.toml`, so it's caught even without a root-lock change.
- Verified: ran the hook via `uv run pre-commit run --hook-stage pre-push` for `uv.lock` and a package `pyproject.toml` (runs) and a non-dependency file (skips); real branch pushes show it running ahead of `ruff`/`ty`/etc.


joshalbrecht:
lgtm

PR #1790, opened 2026-05-28
Derive offline agent lifecycle state from host state

evgunter:
## Summary

Implements `get_offline_agent_state` — the canonical, pure derivation of an agent's `AgentLifecycleState` when its host is offline/unreachable and the agent process cannot be observed directly. This mirrors the existing `derive_offline_host_state` for hosts and lives beside `determine_lifecycle_state` (the online derivation) in `imbue.mngr.hosts.common`.

Replaces the two hardcoded `AgentLifecycleState.STOPPED` fallbacks:
- `mngr_wait.api.poll_target_state` (both the already-offline-host branch and the `HostConnectionError` fallback)
- `build_agent_details_from_offline_ref` in `interfaces/provider_instance.py` (discovery's offline fallback)

## Behavior

No agent lifecycle state is ever persisted (it's computed live from tmux/process info), so the host state is the only signal available offline:

| Host state | Offline agent state |
|---|---|
| `BUILDING`, `STARTING`, `STOPPING`, `STOPPED`, `PAUSED`, `CRASHED`, `FAILED`, `DESTROYED` | `STOPPED` |
| `RUNNING` (unreachable), `UNAUTHENTICATED`, `UNKNOWN` | `UNKNOWN` |

The change from prior behavior: indeterminate host states now yield `UNKNOWN` instead of a misleading `STOPPED`. There is deliberately no agent equivalent of `DESTROYED` — agent destruction is modeled at the host/discovery layer (host `DESTROYED` state + the `include_destroyed` listing filter), so a destroyed host maps to a `STOPPED` agent.

The mapping is an exhaustive `match` with `assert_never`, and a unit test asserts coverage of every `HostState` member so a future addition forces a deliberate choice.

## Tests

- `hosts/common_test.py`: parametrized over every `HostState`, plus an exhaustiveness guard.
- `mngr_wait/api_test.py`: `poll_target_state` offline paths — down host → `STOPPED`, indeterminate host → `UNKNOWN`.
- `interfaces/provider_instance_test.py`: existing offline-fallback tests still pass.


evgunter:
agent misunderstood me

PR #1794, opened 2026-05-28
imbue_cloud auth oauth: don't let followup browser GETs clobber the captured callback

gdenisov:
## Summary
- Fixes a hang in `mngr imbue_cloud auth oauth`: after the browser successfully returned the OAuth `code`, the CLI sat polling until the 300s timeout instead of completing in seconds.
- Root cause: the localhost callback handler called `box.set(params)` unconditionally on every GET. Browsers fire secondary GETs at the listener after the success page renders (favicon.ico, prefetches, service-worker pings); those carry no query params, so they overwrote the captured `{code, state}` with `{}`. The polling loop only breaks on a truthy box, so it kept polling.
- Fix: gate the capture on both `parsed.path == "/oauth/callback"` and non-empty `params` before touching the box. Other paths still get the success page.

## Test plan
- [x] `just test-quick libs/mngr_imbue_cloud/imbue/mngr_imbue_cloud/cli/auth_test.py` — 4 new tests pass (captures real callback; ignores follow-up `/favicon.ico`; ignores paramless `/`; ignores query params on wrong path).
- [x] `just test-quick libs/mngr_imbue_cloud` — full project suite (122 tests) passes.
- [ ] Manually re-run `mngr imbue_cloud auth oauth google` in the minds app and confirm the CLI completes in seconds instead of hanging.


joshalbrecht:
thanks for finding and fixing this!

lgtm

PR #1795, opened 2026-05-28
Add R2 bucket support to mngr_imbue_cloud

joshalbrecht:
## Summary

Adds R2 bucket support to the `mngr_imbue_cloud` provider and the `remote_service_connector` service, per `specs/imbue-cloud-r2-buckets/spec.md`.

- Users can create an R2 bucket (paid accounts only) and mint multiple scoped S3 keys for it (read-only or read-write) to hand to different agents. Isolation is **per-bucket** (one per host), not per-prefix — R2 has no persistent prefix-scoped credentials.
- New CLI: `mngr imbue_cloud bucket create|list|info|destroy` plus `bucket keys create|list|destroy`. Credentials are emitted as JSON; the secret access key is shown once and never stored by the service.
- Connector routes `/buckets/*` and `/bucket-keys/*`, gated by the existing paid-account allowlist. Each key is an account-owned Cloudflare API token scoped to one bucket (Access Key ID = token id, Secret = sha256(token value)). Only key *metadata* is persisted, in a new `r2_keys` table; buckets are listed straight from the R2 API with an in-code owner-prefix re-check (mirrors the tunnel ownership check).
- `bucket destroy` refuses a non-empty bucket and otherwise cascades to revoke its keys.

## Operator migration (required before this works on a tier)

Each tier's `CLOUDFLARE_API_TOKEN` must be replaced with an **account-owned** (`cfat_`) token carrying `Workers R2 Storage: Edit` + `Account API Tokens: Edit` (plus the existing tunnel/DNS/Access/KV permissions), and **R2 must be enabled** on the Cloudflare account. See the connector README. **dev is already migrated + R2-enabled**; staging and production still need it.

## Notes / deviations

- R2 ops are folded into the existing `CloudflareOps` surface (shared authenticated client) rather than a separate `R2Ops` abstraction; only the key-metadata DB is a separate `KeyStore` abstraction. This avoids a second httpx client and keeps the self-contained `app.py` lean.
- Account-wide key listing/destroy uses `/bucket-keys` (not `/buckets/keys`) to avoid a path collision with `/buckets/{name}`.
- The `.minds/template/cloudflare.sh` comment update from the spec was dropped to avoid forcing a `dev/` changelog entry (per the resolved spec decision); the token guidance lives in the connector README instead.

## Test plan

- `apps/remote_service_connector`: 232+ tests incl. bucket lifecycle, cap, non-empty destroy, cascade, cross-user prefix re-check, paid gate, key ownership, sha256 derivation, migration drift guard. All against in-memory R2 + KeyStore fakes (no live R2 in CI).
- `libs/mngr_imbue_cloud`: client status→typed-error mapping, account-wide vs per-bucket key paths, primitives, CLI wiring.
- `ty` type checks pass on both projects. Full suite still to be confirmed green in CI.


joshalbrecht:
lgtm

PR #1797, opened 2026-05-29
fix bug where `mngr help` installed from pypi didn't show topics + let plugins register help topics + display .md properly on the cli

evgunter:
built on https://github.com/imbue-ai/mngr/pull/1773

the docs aren't in the wheel, so `mngr help` for mngr installed from pypi only showed address (since it was registered directly in code instead of from a .md)

also adds the usage recipes as a topic, but it's kind of a poor fit for that; i can remove it / edit it to make it a better fit

adds a couple fancy things like tabcomplete for help topics and fancy formatting (though the fancy formatting takes 50ms to import so it's done conditionally, so it'd be defensible to get rid of it and just display the markdown as plain text even though that's ugly)

https://github.com/user-attachments/assets/d105a33c-4571-4aac-961b-40b403e2d1b6

---

## Summary

Two headline changes (plus polish):

1. **Fix the PyPI/wheel bug**: `mngr help <topic>` showed no doc-based topics when mngr was installed from PyPI, because the topic docs were read from the top-level `docs/` tree at runtime and that tree isn't packaged in the wheel (only `address`, registered inline in code, survived). Verified against `imbue-mngr` 0.2.9 on PyPI.
2. **Let plugins contribute help topics** via a new `register_help_topics` hook, so an installed plugin's topics appear in `mngr help` / `mngr help <topic>` just like built-ins.

Along the way the topic system was reworked to be explicit/structured, terminal rendering was made nice, and `mngr help` arguments are now tab-completable.

## What changed

**Topic model — structured, explicit, no markdown-as-structure parsing**
- `TopicHelpPage` (`interfaces/help_topic.py`, in the interfaces layer so the plugin hookspec can reference it without importing the CLI) declares its metadata explicitly (key, one-line description, aliases, see-also). Its body comes from exactly one of: inline `content`, or a markdown `body_path` (read lazily, rendered as markdown). A `model_validator` enforces the XOR.
- Built-in topics are an **explicit registry** in `cli/builtin_help_topics.py` (each doc-backed topic points at its `docs/*.md` via `body_path`; `address` stays inline). This replaces the old directory scan that inferred the key from the filename and the description from the first heading.
- The runtime registry (`get_topic`/`get_all_topics`/`register_topic`) stays in `cli/help_topics.py`. Built-in topics register through the **same** `register_help_topics` hook as plugins (as a built-in plugin module), so there's one registration path; a topic whose key or alias collides with an already-registered one is skipped, so built-ins always win.

**`rich` markdown rendering in the terminal**
- `mngr help <topic>` now renders markdown nicely on an interactive terminal — headings, bold, inline code, links, and tables (the worst case, `idle_detection`'s mode table, renders as a proper bordered table) — wrapped to the terminal width. The same shared `render_markdown` helper renders command `--help` description/sections.
- Rendering is TTY-conditional: non-interactive output (pipes, scripts, tests, the doc generator) stays plain markdown. `rich` is **lazy-imported** (via a small `cli/markdown_render.py`) so it stays out of CLI startup; this is the one deliberate inline import (ratchet bumped 3→4, documented).

**Wheel packaging**
- The topic doc dirs are `force-include`d under the package (`imbue/mngr/docs`, and the plugin's under `imbue/mngr_usage/docs`), so `mngr help <topic>` works in a wheel install. Built a wheel and confirmed all 16 doc topics ship and render. Added `rich` as a dependency. Added a note in `CLAUDE.md` that the wheel must be self-contained (so this class of bug is caught in review).

**`mngr_usage` plugin**
- Registers its `cron_recipes.md` as the `usage_cron_recipes` topic ("mngr usage: Cron automation recipes"); key/description are namespaced so they're unambiguous in the global topic list.

**Tab completion**
- `mngr help <TAB>` now completes every command and topic (a `help_targets` entry in the completion cache, populated by the cache writer and resolved by the lightweight completer).

**Drive-by**
- Corrected the `mngr_schedule` PACKAGE-mode Modal deploy Dockerfile to install the real PyPI names `imbue-mngr` / `imbue-mngr-schedule` (were `mngr` / `mngr-schedule`, which don't resolve). Surfaced by an audit for other wheel-unsafe runtime file loads. (Note: `imbue-mngr-schedule` isn't published yet, so PACKAGE-mode schedule deploys still require publishing it.)

Docs: `concepts/plugins.md` + `concepts/api.md` updated; changelogs for `mngr`, `mngr_usage`, `mngr_schedule`.

## Testing

- New/updated unit + integration tests: the `register_help_topics` hook (registration, alias resolution, built-in precedence over plugin keys *and* aliases, doc-backed bodies), `format_topic_help` dual layout, `TopicHelpPage` body model, `mngr help` tab-completion targets, and the corrected deploy package names.
- Broad `just test-quick` sweeps over `libs/mngr/imbue/mngr/cli`, `plugins`, `interfaces`, `config`, `libs/mngr_usage`, ratchets, and the import-layer contract all pass.
- Built the wheel (`uv build libs/mngr --wheel`) and confirmed in a fresh install that all topics list and render (the bug is fixed).
- Full CI (offload) runs on this PR.


libs/mngr/imbue/mngr/cli/help_formatter.py
```diff
@@ -147,6 +147,23 @@ def get_terminal_width() -> int:
...
+    When ``use_ansi`` is True, render via rich (tables, bold, code, links,
+    wrapped to ``width``). When False, return the markdown unchanged -- this
+    preserves plain output for pipes, non-interactive runs, tests, and the doc
+    generator, so only interactive terminals get the rich rendering.
+    """
+    if not use_ansi:
+        return markdown
+    # Lazy import: keep rich (and its import cost) out of the CLI startup path;
+    # it is only needed here, when rendering help for an interactive terminal.
+    from imbue.mngr.cli.markdown_render import markdown_to_ansi
```
evgunter, line 162: flagging

libs/mngr_schedule/changelog/mngr-help-docs-plugins.md
```diff
@@ -0,0 +1 @@
+Fixed the PyPI package names in the PACKAGE-mode Modal deploy Dockerfile generator: it now installs `imbue-mngr` and `imbue-mngr-schedule` (the published distribution names) instead of `mngr` / `mngr-schedule`, which do not resolve on PyPI. (Note: `imbue-mngr-schedule` is not yet published, so PACKAGE-mode schedule deploys still require publishing it.)
```
evgunter, line 1: this is unrelated to the rest of the PR but came up when looking for other cases where there were pypi distribution issues

libs/mngr_usage/imbue/mngr_usage/plugin.py
```diff
@@ -10,18 +10,50 @@
...
 
+# Docs root resolution, mirroring mngr's built-in topics. In a wheel the docs are
+# force-included under the package at imbue/mngr_usage/docs; in a source/editable
+# checkout they live at libs/mngr_usage/docs (this file is
+# .../imbue/mngr_usage/plugin.py). Prefer the packaged copy, else the source tree.
+# _PACKAGED_DOCS_DIR is imbue/mngr_usage/docs (force-included in the wheel);
+# _SOURCE_DOCS_DIR is libs/mngr_usage/docs (the source/editable checkout).
+_PACKAGED_DOCS_DIR = Path(__file__).resolve().parent / "docs"
+_SOURCE_DOCS_DIR = Path(__file__).resolve().parents[2] / "docs"
+_DOCS_DIR = _PACKAGED_DOCS_DIR if _PACKAGED_DOCS_DIR.is_dir() else _SOURCE_DOCS_DIR
```
evgunter, line 32: not great

CLAUDE.md
```diff
@@ -59,6 +59,7 @@ Only after doing all of the above should you begin writing code.
 - If you see a flaky test, YOU MUST HIGHLIGHT THIS IN YOUR RESPONSE. Mark it with `@pytest.mark.flaky` so offload retries it automatically. Then try to fix the underlying flakiness in a separate commit.
 - Do not add TODO or FIXME unless explicitly asked to do so
 - Code must work on both macOS and Linux. It's ok if it doesn't work on Windows.
+- `mngr` is installed by end users from PyPI, so the built wheel must be self-contained: it only packages the `imbue` package (`packages = ["imbue"]`), so files outside it (e.g. the top-level `docs/` tree) are NOT shipped and must not be read at runtime by production code.
```
evgunter: flagging

libs/mngr/imbue/mngr/cli/help_formatter.py
joshalbrecht: acceptable here

libs/mngr_usage/imbue/mngr_usage/plugin.py
joshalbrecht: yeah... I'm ok with it for now though

joshalbrecht:
lgtm assuming tests pass!

(not really sure why it's still running 10 minutes later though...)

PR #1772, opened 2026-05-27
minds: activate ToDesktop beforeInstall hook; pin runner node+pnpm via todesktop.js

weishi-imbue:
## Summary

Three changes to `apps/minds`, all about what runs on ToDesktop's cloud build server:

1. **Activate the `todesktop:beforeInstall` hook** so the runner re-resolves `uv` and `git` for its target platform, instead of shipping whatever bytes the dev's local `pnpm build` produced.
2. **Pin `pnpm` and `node` on the runner** via ToDesktop's first-class `pnpmVersion` / `nodeVersion` fields. Replaces `todesktop.json` with `todesktop.js`, which `require()`s `package.json` and reads `engines.pnpm` / `engines.node` — so `package.json` is the single source of truth.
3. **Consolidate `downloadUv`** into one definition in `scripts/download-binaries.js`, imported by `scripts/build.js`. Mirrors how `downloadGit` and `download()` already work.

## Why the runner pins

- Without `pnpmVersion`, ToDesktop's runner does `npx pnpm@latest install` → pnpm 11.1.0, which requires Node ≥22.13 and crashes on ToDesktop's Node 20.20.0 Linux runner with `ERR_UNKNOWN_BUILTIN_MODULE: node:sqlite`.
- Without `nodeVersion`, the project's `engines.node: "24.15.0"` (#1710) is checked unconditionally by `pnpm install` regardless of `engine-strict` — the build silently relies on ToDesktop's default Node staying ≥ 24.15.0.

## Diff (11 files)

| File | Change |
|---|---|
| `apps/minds/todesktop.js` (new) | `require('./package.json')`, read `engines.{node,pnpm}` |
| `apps/minds/todesktop.json` (delete) | Replaced by `.js` form |
| `apps/minds/package.json` | +`"todesktop:beforeInstall": "./scripts/download-binaries.js"` |
| `apps/minds/scripts/download-binaries.js` | Adds `UV_VERSION = '0.11.15'`, uv SHA256s, `downloadUv`, `beforeInstall` default-export hook |
| `apps/minds/scripts/build.js` | Drops local `UV_VERSION` / `getUvDownloadUrl` / `downloadUv`; imports them from `download-binaries.js` |
| `apps/minds/scripts/build_test.py` | Reads `todesktop.js` via `node -e` instead of `json.loads` |
| `apps/minds/docs/desktop-app.md`, `specs/electron-desktop-app/{concise,spec}.md` | File-tree refs `todesktop.json` → `todesktop.js` |
[...]

PR #1800, opened 2026-05-29
Run Node-dependent Python tests in CI by installing Node in the shared mngr image

weishi-imbue:
Follow-up to #1772.

## Problem

Three Python tests shell out to a `node` runtime and were gated with `skipif(shutil.which("node") is None)`:

- `apps/minds/scripts/build_test.py::test_bundled_limactl_is_signed_with_virtualization_entitlement` — evaluates `apps/minds/todesktop.js` (the `.json`→`.js` rename in #1772 introduced this Node dependency).
- `libs/mngr_latchkey/.../minds_api_proxy_test.py` and `permission_requests_test.py` — spawn a Node child process to drive the latchkey gateway's `.mjs` extensions.

The shared mngr Modal image (which backs offload) has no Node, so these tests were collected by `test-offload` and then **silently skipped** — zero CI coverage. The latchkey ones had been skipping since they were written.

## Fix

1. **Install Node in the shared mngr image** (`libs/mngr/imbue/mngr/resources/Dockerfile`), pinned to `apps/minds/package.json`'s `engines.node` (24.15.0). Node is also a genuine runtime dependency of the latchkey gateway, so it belongs in the image regardless.
2. **Remove the `skipif`** from the three tests. The pre-existing `assert _NODE_BINARY is not None` becomes the hard-fail: with Node now in the image they run on offload's existing `all` group, and a missing Node fails loudly instead of silently skipping.

No new CI job, no pytest markers, no resource guard — Node joins the ~20 other tools already used unguarded in the image (git, curl, jq, ripgrep, …). (An earlier revision of this PR routed these via a dedicated `test-node` job + a `node` resource guard; per review discussion it was simplified to this minimal form. `ci.yml` / `offload-modal.toml` / `justfile` end up net-zero vs `main`.)

## Diff (7 files)

- `libs/mngr/imbue/mngr/resources/Dockerfile` — Node install
- `apps/minds/scripts/build_test.py` — drop `skipif` (+ now-unused `import pytest`)
- `libs/mngr_latchkey/.../minds_api_proxy_test.py`, `permission_requests_test.py` — drop `skipif`; refresh stale "skips cleanly" docstrings
- changelog entries for `apps/minds`, `libs/mngr`, `libs/mngr_latchkey`

## Verification

- The three test modules (32 tests) pass locally with Node present; `ruff` + `ty` clean.
- The "Node present on offload" half is validated by CI (the Modal image can't be built locally); the Dockerfile hunk was reviewed by inspection (arch mapping, extraction, pin, `test -x`).

## Follow-up (optional, not in this PR)

The Dockerfile `NODE_VERSION` pin has no drift-guard test against `package.json`'s `engines.node`, unlike `CLAUDE_CODE_VERSION` (`apps/minds/imbue/minds/test_claude_version_alignment.py`). A small alignment test would prevent drift.


joshalbrecht:
lgtm, thanks for the fix!

PR #1802, opened 2026-05-29
Exclude Latchkey from the minimum age check.

hynek-urban:
Exclude the Latchkey from the minimum age check so that we can quickly bump Latchkey versions while developing.


joshalbrecht:
lgtm

PR #1804, opened 2026-05-29
GitHub backups over Latchkey: prerequisites

hynek-urban:
(Accompanied by https://github.com/imbue-ai/forever-claude-template/pull/116.)

- Bump Latchkey version to 2.14.0 in order to get support for Smart HTTP git operations on GitHub.
  - The minimum version check has also been bumped! You'll probably need to run `pnpm install` from `apps/minds`.
- Add support for multiple scopes per service. (The GitHub service now covers two relatively distinct scopes: `github-rest-api` and `github-git`.)

This is how I've been able to successfully test pushes to GitHub:

```
Please create a new repository in /tmp/foo (on your disk inside the sandbox), add a dummy readme and push it as a new private repository under my account. For this to work, you will need to:
- Configure the remote as http://localhost:1989/gateway/https://github.com/... 
- Include the x-latchkey-gateway-password header with the value from the $LATCHKEY_GATEWAY_PASSWORD env var when running the git operations. (Probably via -c http.extraHeader.)
- Do also the same for x-latchkey-gateway-permissions-override and $LATCHKEY_GATEWAY_PERMISSIONS_OVERRIDE.
```


joshalbrecht:
lgtm

PR #1805, opened 2026-05-29
mngr_mapreduce: reuse snapshot's /code for the reducer too

qi-imbue:
## Summary

- The snapshot-based code-reuse optimization in `_create_agent` only fired when an `existing_host` was passed in — mappers got that via the pre-created host pool, but the reducer called `_create_agent` with no `existing_host` and so always re-uploaded source from the laptop.
- Pre-create the host inside `_create_agent` when `config.snapshot` is set (and the agent isn't the snapshotter itself) so the existing `GIT_WORKTREE` branch fires for the reducer too. As a bonus, mappers that fell through this path (e.g. if the pool was empty) now also benefit.

## Test plan

- [x] `just test-quick libs/mngr_mapreduce` — 98 passed, 0 failed
- [ ] CI (offload acceptance/release)


joshalbrecht:
lgtm

PR #1807, opened 2026-05-29
Make the permission dialog less daunting

hynek-urban:
During user testing, people apparently found the permission dialog screen complicated. This changes the appearance so that users are only faced with a single choice: "Approve" vs "Deny". (The original list of checkboxes can still be summoned when clicking on "Adjust".)

Before:

before

After:
after


joshalbrecht:
lgtm, thanks for fixing!

also, I laughed out loud at "people **apparently** found the permission dialog screen complicated" (emphasis mine) :)

PR #1809, opened 2026-05-29
mngr_mapreduce: re-strip trailing hyphens after sanitize-and-truncate

qi-imbue:
## Summary

- TMR fails to spawn a mapper for any e2e test whose sanitized-and-truncated slug lands on a hyphen at the 40th character. The most recent victim is \`test_create_modal_idle_mode_ssh_timeout_300\`, which sanitizes to \`test-create-modal-idle-mode-ssh-timeout-300\` (43 chars) and then truncates to \`test-create-modal-idle-mode-ssh-timeout-\`. \`AgentName\` rejects that as "alphanumeric with dashes/underscores allowed in the middle".
- \`sanitize_for_agent_name\` already strips leading and trailing hyphens *before* truncating; it just needs to re-strip after truncation. One-line fix plus a regression test using the exact slug from the failing TMR run.

## Test plan

- [x] \`just test-quick libs/mngr_mapreduce/imbue/mngr_mapreduce/utils_test.py\` — 15 passed (14 prior + 1 new regression test).
- [ ] CI


joshalbrecht:
lgtm

PR #1810, opened 2026-05-29
Run git/rsync with terminal stdio passthrough in mngr CLI

qi-imbue:
## My summary

Make these commands even thinner wrappers by running them directly in the terminal.

The only caveat is that `mngr rsync` no longer emits structured logs for the rsync stats. I think this is overall a win:
- it's hard to do that now when the outputs now go directly to the terminal - we'll need to tee the output
- the user may specify options that change the stats output
- if the user is interested it's not that hard for them to parse the output themselves anyway.
- the two stats we choose to parse and emit are arbitrary

## Summary

- `mngr git push`, `mngr git pull`, and `mngr rsync` now run the underlying `git` / `rsync` binary with the user's terminal stdio (no redirection), so progress, prompts, errors, and pager-style output flow through directly. mngr does its setup work first (URL/env construction, `safe.directory` + `receive.denyCurrentBranch` config, default refspec resolution, destination stash check, destination mkdir), then waits for the child to exit -- which preserves destination-side cleanup like `mngr rsync --uncommitted-changes=merge`'s stash pop. On non-zero exit, mngr raises its own typed error so the CLI still surfaces failure (with the child's exit status in the message).
- The terminating `{"event": "git_push_complete", ...}` / `{"event": "rsync_complete", ...}` JSONL events and the "Rsync complete: N files, M bytes transferred" human line are gone -- callers should rely on exit code instead of a synthetic terminator. `rsync --stats` is still on by default, so CLI users still see the transfer summary block in their terminal at the end of a run.
- API change: `imbue.mngr.api.git.git_push` / `git_pull` and `imbue.mngr.api.rsync.rsync` / `rsync_to_remote` / `rsync_from_remote` grow an optional `run_in_terminal: bool = False` parameter. Default behavior (captured stdout/stderr via `ConcurrencyGroup`) is unchanged, so in-process callers like `mngr_pair` and `mngr_mapreduce` are unaffected; only the `mngr` CLI passes `True`. The terminal-stdio branch calls the pre-existing `imbue.mngr.utils.interactive_subprocess.run_interactive_subprocess` helper directly.
- Cleanup: the rsync API functions now return `None` instead of `RsyncResult`. `RsyncResult`, `imbue.mngr.utils.rsync_utils.parse_rsync_output`, and the whole `rsync_utils` module are deleted. Existing callers (the CLI and `mngr_mapreduce.launching`) already discarded the return value, and the parsed counts were already unavailable in `run_in_terminal=True` mode (nothing to parse).

## Test plan

- [ ] `mngr git push my-agent` -- output (`Pushing to ...`, then `To <url>` / `<oldsha>..<newsha>  <branch> -> <branch>`) flows directly to the terminal; exit code is 0
- [ ] `mngr git push my-agent -- --force --all` -- arg passthrough works; rejected pushes show git's native error and mngr exits non-zero
- [ ] `mngr git pull my-agent -- --no-edit` -- merge / fast-forward output flows through; non-merge prompts (`--no-edit` defaulting) work because stdin is real
- [ ] `mngr rsync ./src/ my-agent` -- `rsync --stats` block streams to the terminal as it goes (not at the end), and mngr exits 0
- [ ] `mngr rsync ./src/ my-agent -- --dry-run` -- arg passthrough; mngr exits 0
- [ ] `mngr rsync my-agent ./dst` where `./dst` is a dirty git repo, with `--uncommitted-changes=merge` -- mngr stashes the dirty changes, runs rsync, then pops the stash on top (this is the regression that motivated dropping the earlier `os.exec` approach)
- [ ] `mngr rsync my-agent ./dst` with `--uncommitted-changes=fail` against a dirty destination -- mngr errors before invoking rsync
- [ ] `mngr git push ./bare-local-path` -- mngr rejects with "requires an agent or remote host" before invoking git
- [ ] Ctrl-C during a slow `mngr rsync` -- SIGINT reaches rsync directly (no mngr-level signal trap); mngr exits with rsync's signal-derived exit code
- [ ] `mngr_mapreduce` smoke run (which now calls a `None`-returning `rsync_to_remote`)
- [ ] Full acceptance / release suite via CI offload


joshalbrecht:
the code looks ok, but see slack thread

something feels a little wrong about this

joshalbrecht:
lgtm, thanks for fixing!

PR #1811, opened 2026-05-29
Clean up leaked docker build images and orphaned state containers

joshalbrecht:
## Summary

Fixes two Docker resource leaks.

**1. mngr docker provider — leaked build images.** `create_host` tags a per-host image `mngr-build-<host_id>` that was never removed, so built images piled up in `docker images`. Now untagged eagerly in `destroy_host` and again defensively in `delete_host`. Snapshot restore is unaffected (snapshots are independent commits that keep their own layers).

**2. minds — orphaned state containers.** Each minds env has its own `MNGR_PREFIX`, hence its own singleton Docker state container `<MNGR_PREFIX>docker-state-<user_id>` (with `restart_policy=unless-stopped`). It was abandoned (kept running forever) whenever an env was reset or destroyed. Now removed — the **one exact** container + its same-named volume — in both reset/destroy paths:
- `minds env destroy` / `destroy_env` (via an injected `cleanup_state_container` Providers callable; skipped under `--keep-agents`).
- The activate-time generation-mismatch auto-wipe, which now also destroys the env's agents first, then removes the state container, then wipes local state.

Supporting changes:
- New `apps/minds/imbue/minds/envs/docker_cleanup.py` (shells out to `docker`; best-effort when no daemon, raises on a real removal failure; targets the exact name only).
- Agent teardown refactored from per-agent to a single batched `mngr destroy -f <ids…>` call.
- Shared `read_active_profile_dir` helper extracted into `bootstrap.py`.
- Tests added (mngr docker lifecycle + minds docker_cleanup + auto-wipe smoke); `docker_cleanup.py` added to minds coverage omit; changelog entries for `libs/mngr`, `apps/minds`, `dev`.

Design doc: `specs/docker-cleanup-state-and-images/spec.md`.

## Testing

- Local: minds fast suite (826 passed; 5 pre-existing unrelated `webdav_test.py` failures verified against base), mngr docker non-docker tests (96 passed), all new/changed targeted tests, `ty` type checks, autofix clean.
- `just test-offload` cannot run in this environment (no `MODAL_TOKEN_ID`); the full suite including `docker`-marked tests runs in CI.

### DIFF
--- a/apps/minds/imbue/minds/cli/env.py
+++ b/apps/minds/imbue/minds/cli/env.py
@@ -34,6 +34,7 @@ from pydantic import AnyUrl
 from pydantic import SecretStr
 
 from imbue.concurrency_group.concurrency_group import ConcurrencyGroup
+from imbue.concurrency_group.errors import ConcurrencyGroupError
 from imbue.minds.bootstrap import DEFAULT_MINDS_ROOT_NAME
 from imbue.minds.bootstrap import MINDS_ROOT_NAME_ENV_VAR
 from imbue.minds.bootstrap import mngr_host_dir_for
@@ -888,7 +889,10 @@ def _destroy_agents_and_state_container_for_wipe(env_name: str) -> None:
             if destroyed_count:
                 logger.info("Destroyed {} mngr agent(s) during wipe of env {!r}", destroyed_count, env_name)
             cleanup_env_state_container(dev_env_name, parent_concurrency_group=cg)
-    except (MngrAgentCleanupError, DockerCleanupError, OSError) as exc:
+    except (MngrAgentCleanupError, DockerCleanupError, ConcurrencyGroupError, OSError) as exc:
+        # ConcurrencyGroupError covers a failure to launch the `mngr` / `docker`
+        # subprocess (e.g. the binary is missing) so a teardown hiccup can never
+        # wedge the eval-sourced activation.
         logger.warning(
             "Could not fully tear down agents / Docker state container for env {!r} during wipe "
             "({}); proceeding with the local-state wipe anyway.",
--- a/apps/minds/imbue/minds/cli/env.py
+++ b/apps/minds/imbue/minds/cli/env.py
@@ -51,13 +51,13 @@ from imbue.minds.config.loader import EnvConfigError
 from imbue.minds.config.loader import load_client_config
 from imbue.minds.config.loader import load_deploy_config
 from imbue.minds.config.loader import repo_tier_client_config_path
+from imbue.minds.envs.docker_cleanup import DockerCleanupError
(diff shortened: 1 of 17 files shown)


apps/minds/imbue/minds/cli/env.py
```diff
@@ -854,6 +864,43 @@ def _try_run_generation_check(*, env_name: str, client_config_path: Path, env_ro
...
+    except (MngrAgentCleanupError, DockerCleanupError, ConcurrencyGroupError, OSError) as exc:
+        # ConcurrencyGroupError covers a failure to launch the `mngr` / `docker`
+        # subprocess (e.g. the binary is missing) so a teardown hiccup can never
+        # wedge the eval-sourced activation.
+        logger.warning(
+            "Could not fully tear down agents / Docker state container for env {!r} during wipe "
+            "({}); proceeding with the local-state wipe anyway.",
+            env_name,
+            exc,
+        )
```
joshalbrecht: No. Don't just catch all of these errors--there's no reason why we should fail to clean things up!  And if something goes wrong, the user should see it so they can fix it

libs/mngr/imbue/mngr/providers/docker/instance.py
```diff
@@ -734,6 +734,28 @@ def _log_docker_creation_command_output(self, line: str, is_stdout: bool) -> Non
...
+        images that retain the underlying layers, so untagging here does
+        not break snapshot restore.
+        """
+        tag = self._build_image_tag(host_id)
+        try:
+            self._docker_client.images.remove(tag)
+        except docker.errors.ImageNotFound:
+            logger.trace("No build image to remove for host {}", host_id)
+        except docker.errors.DockerException as e:
+            logger.warning("Error removing build image {}: {}", tag, e)
```
joshalbrecht: No.

Don't make a method called "_remove_build_image" that does not, in fact, remove the build image :-P

Get rid of this catch

joshalbrecht: Fixed in 67610a6fd. `_remove_build_image` now checks `images.list(name=tag)` and removes the image with no surrounding catch, so a genuine removal failure propagates. The image-absent case (pulled-image host, or already removed) is a clean no-op.

apps/minds/imbue/minds/cli/env.py
joshalbrecht: Fixed in 67610a6fd. Removed the broad try/except in `_destroy_agents_and_state_container_for_wipe` -- agent-destroy and state-container removal failures now propagate so the operator sees them. (The sweep is still a no-op when there's simply no Docker daemon; only real failures raise.)

joshalbrecht:
lgtm

PR #1812, opened 2026-05-29
add mngr_ctx to on_before_create and on_before_host_create hooks

evgunter:
`on_before_create` was needed for a fix to `--adopt-session`, but also most hooks have this parameter and presumably these should not differ

---

## Summary

Adds the `MngrContext` as a parameter to two plugin lifecycle hooks:

- `on_before_create(args, mngr_ctx)`
- `on_before_host_create(name, provider_name, mngr_ctx)`

This brings them in line with the other lifecycle hooks (`on_host_created`, `on_before_provisioning`, `on_after_provisioning`, `on_before_host_destroy`, etc.), all of which already receive `mngr_ctx`. Plugins implementing these hooks can now reach `mngr_ctx.config`, `mngr_ctx.pm`, and `mngr_ctx.concurrency_group`.

## Changes

- `hookspecs.py`: added `mngr_ctx: MngrContext` to both hookspec signatures and docstring examples.
- `create.py`: pass `mngr_ctx` at both call sites (the chained `on_before_create` invocation and the `pm.hook.on_before_host_create` call).
- `mngr_claude/plugin.py`: updated the `on_before_create` implementation (the `--adopt-session` validator) to accept `mngr_ctx`.
- Updated docs (`plugins.md`) and all affected tests; added an assertion that `on_before_host_create` receives the correct context.

This is a breaking change for any out-of-tree plugins implementing these two hooks: they must add a `mngr_ctx` parameter.

## Testing

- `libs/mngr_claude/.../plugin_test.py` (unit): 197 passed
- `libs/mngr/api/test_create.py` + `agent_host_hooks_test.py` (unit): 10 passed
- `test_meta_ratchets.py::test_no_type_errors`: passed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1813, opened 2026-05-29
accept claude subtypes in --adopt-session

evgunter:
built on https://github.com/imbue-ai/mngr/pull/1812 and https://github.com/imbue-ai/mngr/pull/1817

---

## Problem

`mngr c <name> --adopt-session <id>` failed with:

```
Error: --adopt-session can only be used with the claude agent type, not 'write-plus'.
```

even though `write-plus` is a subtype of the claude agent (a config-defined template whose `parent_type` chain reaches `claude`).

## Cause

The `on_before_create` validation in the Claude plugin compared the agent type *name* literally against the string `"claude"`, so any subtype was rejected.

## Fix

Route the check through the centralized `resolve_agent_type` registry and accept any type that resolves to a `ClaudeAgent` subclass — the same notion of "is a claude agent" used by the `isinstance(agent, ClaudeAgent)` checks elsewhere (e.g. `mngr_claude_usage`). Valid non-claude types (e.g. `codex`) still get the clear `UserInputError`; an unknown type now surfaces `UnknownAgentTypeError` (which `create()` would raise regardless).

Added a regression test asserting a `parent_type=claude` subtype (`write-plus`) is accepted.

## Dependency

Based on #1812, which threads `MngrContext` into the `on_before_create` hook so the validation can resolve agent types. Retarget to `main` once that merges.


joshalbrecht:
lgtm

PR #1815, opened 2026-05-29
kinda support lifecycle states for antigravity

evgunter:
lifecycle states now work to detect running vs stopped, but not waiting for permissions:

https://github.com/user-attachments/assets/8748b4f8-882f-4049-af62-609ca0056143

---

## What

Re-verified against `agy` 1.0.3 that Antigravity CLI hooks **load and execute** (an earlier `mngr_antigravity` comment claimed execution was gated behind a per-account `json-hooks-enabled` experiment — that is obsolete). This PR puts hooks to use for lifecycle reporting.

mngr provisions a **per-agent** `hooks.json` into the agent state dir and points agy at it via `--add-dir`, recovering the per-agent isolation the old `mngr_gemini` got from the (removed) `GEMINI_CLI_SYSTEM_SETTINGS_PATH`, without touching the user's global `~/.gemini/config/`:

- **Lifecycle marker**: `PreInvocation` touches `$MNGR_AGENT_STATE_DIR/active`, `Stop` removes it. `BaseAgent.get_lifecycle_state` reads that marker (marker-only, no pane-activity fallback), so antigravity agents now report **RUNNING while working and WAITING when idle**. agy writes no `active` marker on its own, so before this they could never report RUNNING.

## Two agy rough edges worked around

- **`--add-dir` rejects hidden paths**: the hooks dir lives under `~/.mngr/` (dotted), which agy refuses (`is hidden: ignore uri`) — so a direct `--add-dir` silently loaded zero hooks. Routed through a per-agent `/tmp` symlink (`/tmp/mngr_antigravity_hooks/<id>` → `<state>/agy_hooks`), mirroring the existing workspace symlink.
- **The hook allow-decision doesn't gate `run_command`**: agy's documented `PreToolUse` `{"decision":"allow"}` output runs but the confirmation dialog still appears (verified live). So `auto_allow_permissions` uses the `--dangerously-skip-permissions` **flag**, not a hook.
- Also: the in-TUI `/hooks` command writes to `~/.gemini/antigravity-cli/hooks.json`, which the execution engine never runs ([antigravity-cli#49](https://github.com/google-antigravity/antigravity-cli/issues/49)); only `~/.gemini/config/hooks.json` and per-workspace `.agents/hooks.json` execute. mngr writes its own file and never relies on the TUI.

## Verification

End-to-end with `mngr create`: hooks load (0 hidden-path rejections), no permission dialog under auto-allow, `mngr` reports RUNNING during a turn, and Stop clears the marker back to WAITING. All 137 `mngr_antigravity` package unit tests pass.

## Deferred (agy limitations, documented in code/README)

- **No permission-specific WAITING reason**: agy exposes no permission/idle hook event and nothing fires while blocked at a dialog, so mngr can't surface *why* an agent waits (as `mngr_claude` does). With auto-allow there are no dialogs anyway.
- **No `SESSION_GUARD` equivalent**: `mngr_claude` guards its marker hooks (`MAIN_CLAUDE_SESSION_ID`) so resumed sub-agent sessions don't corrupt the parent's `active` marker. The agy equivalent can't be copied directly (no session-id env var is set, and copying it as-is would always early-exit and break the marker). Left as a follow-up pending live confirmation that agy subagents even corrupt the parent marker + a real session-id env var set by provisioning.
- Readiness still uses the `InteractiveTuiAgent` banner-poll (agy's hook events are execution-loop events with no "input prompt drawn" analog).


libs/mngr_antigravity/changelog/mngr-agy-hooks.md
```diff
@@ -0,0 +1,7 @@
+The `antigravity` agent type now uses agy hooks to report lifecycle state (verified working against agy 1.0.3).
+
+- mngr provisions a per-agent `hooks.json` and points agy at it with `--add-dir` (via a `/tmp` symlink, since agy rejects the dotted state-dir path), so the user's global `~/.gemini/config/` is untouched and each agent's state stays isolated.
+- A `PreInvocation`/`Stop` hook pair maintains an `active` marker so antigravity agents now report RUNNING while working and WAITING when idle (previously they had no `active` marker and could not report RUNNING).
+- `auto_allow_permissions = true` continues to use the `--dangerously-skip-permissions` CLI flag. agy's documented `PreToolUse` `{"decision": "allow"}` hook output does not actually gate the `run_command` confirmation dialog, so a hook can't replace the flag.
```
evgunter, line 5: thanks gogle

libs/mngr_antigravity/imbue/mngr_antigravity/antigravity_config.py
```diff
@@ -8,11 +8,31 @@
...
+per-agent ``hooks.json`` into the agent state dir and points agy at it with
+``--add-dir`` (see ``build_antigravity_hooks_config`` and the plugin's
+``assemble_command``), so the user's global config stays untouched and each
+agent's marker files land in its own state dir.
+
+The in-TUI ``/hooks`` command instead writes to
+``~/.gemini/antigravity-cli/hooks.json``, which the execution engine does not
+run -- that path is loaded only for the TUI's display, while hooks execute only
+from ``~/.gemini/config/hooks.json`` and per-workspace ``.agents/hooks.json``
+(google-antigravity/antigravity-cli#49). mngr therefore does not use the TUI.
```
evgunter, line 29: cool

joshalbrecht:
lgtm!

PR #1816, opened 2026-05-29
fix flaky test_host_volume_is_symlinked_and_persists_data

evgunter:
## What

`test_host_volume_is_symlinked_and_persists_data` (a `mngr_modal` acceptance test) flaked in [PR #1797's Acceptance Tests run](https://github.com/imbue-ai/mngr/runs/78595441450): it failed on attempt 1 with `assert None is not None` at the final `assert get_volume_for_host(host) is not None`, then passed on retry.

## Root cause

The assertion ran `get_volume_for_host(host)` exactly once, immediately after `create_host`. That method probes the volume with a fresh `volume_from_name(create_if_missing=False)` followed by `listdir("/")`. Just after sandbox creation, Modal's control plane can lag in making the newly-created volume resolvable by name, so the `listdir` probe transiently raises a volume-level `NotFoundError`. That is *not* retried by `_should_retry_volume_op` (which only retries environment-level not-found), so `get_volume_for_host` catches it and returns `None`.

The volume genuinely exists -- the sandbox mounted it and the test had already written through it in the preceding assertions -- so this is pure eventual-consistency lag, not a real bug.

## Fix

Poll the assertion with `wait_for`, mirroring the sibling test `test_host_volume_data_readable_via_volume_interface`, which already handles the same class of lag. No production code changed: real callers of `get_volume_for_host` (`on_before_host_destroy`, event handlers, mapreduce pulling) only run against long-lived hosts, never within a second of creation, so they were never exposed to this race.


joshalbrecht:
lgtm

PR #1817, opened 2026-05-29
make CreateAgentOptions.agent_type required and don't default to claude

evgunter:
not sure how this got missed when removing the default from the cli layer

---

## Summary

We recently removed the CLI's behavior of defaulting `mngr create`'s agent type to `claude`. The `or AgentTypeName("claude")` fallback still lived on in `api.create.create` (and a couple of mirrored spots). This makes `CreateAgentOptions.agent_type` a **required** field so the type system enforces a concrete agent type on every creation path, rather than silently falling back to `claude`.

## Changes

- `CreateAgentOptions.agent_type`: `AgentTypeName | None = None` -> required `AgentTypeName`.
- Removed the `agent_type or AgentTypeName("claude")` fallback in `api.create.create` and `Host.create_agent_state`.
- Dropped the now-dead `if options.agent_type is not None:` guard around agent-type provisioning merging in `Host.provision_agent`.
- Simplified the `--adopt-session` `on_before_create` validation hook in `mngr_claude` (no more unset special-case).
- Updated the 7 `_merge_agent_type_provisioning` unit tests (the only real construction sites that omitted `agent_type`) to pass a type.

## Verification

- `uv run ty check` clean on `libs/mngr` and `libs/mngr_claude` (only the pre-existing `cel_utils` warning).
- `host_test.py` (166 passed) and the `--adopt-session` plugin tests (5 passed) pass locally. Full suite via CI.


joshalbrecht:
lgtm, thanks for fixing!

PR #1818, opened 2026-05-29
make links in help topics clickable in the terminal

evgunter:
Stacked on #1797 (`mngr/help-docs-plugins`). Implements the deferred "links fix": relative and anchor markdown links in `mngr help <topic>` doc pages are now rewritten to absolute GitHub URLs at terminal render time, so the OSC-8 hyperlinks rich emits are actually clickable and open the right page/section.

## Problem

`mngr help <topic>` renders doc-backed topics via `rich`, which turns markdown links into clickable terminal hyperlinks. But topic docs use **relative** links (`provider_backends.md`, `../README.md#waiting-on-a-predicate`) and **anchor-only** links (`#user-input-tracking`), whose targets are meaningless to a terminal/browser. Those hyperlinks were dead.

## Approach

The PyPI README generation in `scripts/make_cli_docs.py` already rewrites relative links to `https://github.com/imbue-ai/mngr/blob/main/...`. This does the analogous thing for the terminal, but per-doc (each doc is at a different path, and plugins live in different repos), and also handles anchors.

- `DocFile` (in `interfaces/help_topic.py`) gains a `source_url: str | None` field: the doc's own canonical GitHub blob URL. `interfaces` stays a pure plugin-facing model layer.
- New `cli/doc_links.py` holds the application/CLI doc-link logic: `imbue_mngr_doc_url(repo_relative_path)` builds the blob URL for docs in the imbue-ai/mngr repo, **pinned to the installed release tag** (`v{version}` via `importlib.metadata`, falling back to `main`); and `rewrite_links_to_absolute(markdown, base_url)` resolves relative/anchor links against a doc's `source_url` with `urljoin`. Built-in topics and the `mngr_usage` plugin set `source_url` via `imbue_mngr_doc_url`.
- At render time (`help_formatter.render_markdown`, gated on `use_ansi`), each relative/anchor link is rewritten: `#x` -> `<doc-url>#x`, `sib.md` -> the sibling, `../README.md#x` -> the parent. Already-absolute (`https:`, `mailto:`) links pass through. Plain non-terminal output (pipes, scripts, doc generator) keeps the original relative links.

### Ref policy

Links pin to the installed release tag (`v0.2.9`), so a PyPI user's terminal links resolve to exactly the docs their installed version shipped (verified: the repo is public and `.../blob/v0.2.9/.../idle_detection.md` returns 200). Falls back to `main` if the dist version can't be read. Caveat: in a source checkout whose version predates a not-yet-released doc, that doc's link can 404 until the next release -- inherent to version-pinning and harmless for real installs.

## Testing

- `cli/doc_links_test.py`: link-rewrite helper (anchor, sibling, parent, absolute, mailto, no-link text) and `imbue_mngr_doc_url`.
- `help_formatter_test.py`: `render_markdown` (rewrite when ANSI, passthrough when not).
- `help_test.py` / `plugin_test.py`: integration through `format_topic_help` (terminal rewrites links; plain output leaves them) and that built-in + `mngr_usage` topics carry a GitHub `source_url`.
- Manually verified the real `idle_detection` topic renders OSC-8 hyperlinks to the correct absolute GitHub URLs.

### Known limitations

The rewriter uses the same simple `](...)` regex as `make_cli_docs.py`: it handles inline links only (not reference-style `[text][ref]` or `](url "title")`) and doesn't skip code spans/fences. The topic docs only use inline links without titles, so this is fine in practice.


joshalbrecht:
lgtm -- leaving this unmerged because it doesn't target main

(hopefully I didn't miss any other stacked ones like that!)

PR #1820, opened 2026-05-30
minds: animated wordmark loading screen

gdenisov:
## Summary

Replaces the placeholder logo + spinner on the Electron startup screen (`apps/minds/electron/shell.html`) with the Minds wordmark SVG, animated to convey "something is happening" while the Python backend boots. On startup failure the animation freezes with all letters visible so the brand stays present alongside the "Something went wrong" panel (left unchanged).

All in scope to one file: `apps/minds/electron/shell.html`.

## Visual behavior

- True-black background, white wordmark.
- 5.4s cycle, infinite:
  - **Fade in:** 0.5s linear, 0.1s stagger (M → I → N → D → S).
  - **Hold visible:** 1s.
  - **Fade out:** 1.8s `cubic-bezier(.05, .25, .08, .99)`, 0.3s stagger.
  - **Hold invisible:** 0.5s.
- Error state: `body.error .wordmark path { animation: none; opacity: 1 }` freezes the cycle and shows every letter. JS error handler adds `body.error` and reveals `#error-view`.

## Before / After
| Before | After |
|-|-|
| loading-screen-before | loading-screen-after |

## Commits

- `a0469405d2` minds: animated wordmark loading screen with frozen error state
- `54f3c744cf` Remove dead `#loading-view` CSS rule from shell.html
- `7616720c53` Give wordmark SVG an accessible label (`role="img"` + `aria-label="Minds"`)

## Test plan

**Fastest review:** open the file directly in a browser — `shell.html` is self-contained (inline CSS + JS, no Electron bridge required for the visual). The interactive buttons are no-ops without `window.minds`, but the animation and error toggle work fine.

```bash
open apps/minds/electron/shell.html
# or paste this into a browser address bar:
# file:///<your-checkout>/apps/minds/electron/shell.html
```

- [ ] **Loading state:** wordmark animates indefinitely with the cycle described above.
- [ ] **Error state:** in DevTools console:
  ```js
  document.body.classList.add('error');
  document.getElementById('error-view').style.display = 'flex';
  ```
  Animation freezes, all five letters visible at full opacity, error panel renders below.
- [ ] **Accessibility:** screen readers announce "Minds" via the SVG's `aria-label`.

**In Electron (optional, slower):** `just minds-start` should show the same loading screen during backend startup. To force the error path, kill the backend mid-startup or temporarily throw inside `runEnvSetup` in `main.js`.

## Known follow-ups (deferred)

- Five hand-computed `@keyframes` could be collapsed to one + per-letter `animation-delay`.
- `prefers-reduced-motion` handling not yet added.
- The `status-update` IPC channel in `main.js`/`preload.js` is now dead (no listener); the emitters can be removed when the dust settles.


joshalbrecht:
lgtm!

PR #1821, opened 2026-05-30
Open external links in the default browser (minds desktop app)

gnguralnick:
## What

In the Minds desktop app, external links clicked anywhere (agent content including embedded iframes, sidebar, request panels, title bar) now open in the user's default browser instead of taking over the in-app workspace view or spawning a bare chrome-less Electron window. Covers both ordinary link clicks / JS navigations (via `will-frame-navigate`, which also fires for clicks inside the iframes that embed agent content) and `target="_blank"` / `window.open` popups (via `setWindowOpenHandler`).

"External" = non-localhost http/https, plus `mailto:`/`tel:`. In-app navigation (the app's own pages and `agent-<id>.localhost` workspace subdomains) and non-web schemes (`file:`, `about:`, `blob:`, `data:`, `devtools:`) are unchanged. Wired globally via `app.on('web-contents-created')`, following Electron's documented security pattern. External URLs are routed through a shared `openInBrowser` helper that defers via `setImmediate` and swallows `shell.openExternal` rejections (avoids a main-process crash).

## Verification

- `node --check` passes; `isExternalUrl` logic checked against representative cases (incl. IPv6 loopback `[::1]`, mailto/tel, agent subdomains).
- Architecture and autofix review gates passed. Autofix caught two issues (IPv6 `[::1]` bracket classification; unhandled `openExternal` rejection), both fixed.


joshalbrecht:
lgtm, thanks for fixing!

PR #1731, opened 2026-05-22
minds: bundle Lima into the desktop app

weishi-imbue:
## Summary

Bundles the [Lima](https://lima-vm.io/) VM manager into the packaged
Minds desktop app so a user can run agents in Linux VMs out of the box,
with no `brew install lima` prerequisite.

**Extracted from PR #1317** (`wz/minds_onboard`), which is the *proven*
implementation: a fully signed, notarized macOS build was produced from
it and verified end-to-end (download → launch → create a Lima agent →
first message). It **supersedes #1681 and #1661**, which both bundle Lima
but miss the macOS code-signing pieces and would fail notarization.
Those two PRs should be closed in favor of this one (left for a human to
do).

## What this PR does

- `scripts/build.js` downloads the official Lima 2.1.1 release tarball
  for the build host's platform/arch and extracts it into
  `resources/lima/`. A new shared `downloadAndExtractTarball` helper now
  backs both uv and Lima.
- The unsigned `lima-guestagent.Darwin-*.gz` Mach-O payloads are
  stripped after extraction — Apple's `notarytool` unzips and rejects
  them, and they are unreachable since we run Linux VMs only.
- `entitlements.mac.plist` (new) carries
  `com.apple.security.virtualization` (required by `limactl`'s VZ
  driver) plus the standard Electron entitlements. `todesktop.json`
  wires it in via a `mac` block and lists `resources/lima/bin/limactl`
  in `additionalBinariesToSign` so ToDesktop deep-signs `limactl` with
  that plist. **This is the piece #1681 / #1661 miss.**
- `paths.js` exposes `getLimaPath()` / `getLimaBinDir()`; `backend.js`
  prepends the bundled lima bin dir to the packaged-mode `PATH`.
- `scripts/build_test.py` guards that the signing config grants
  `limactl` the virtualization entitlement.

## Out of scope

- Defaulting the create-form's compute provider to `LIMA`. This was
  originally commit 2 of this PR, then reverted (commit `20d347620`):
  the justification docstring was over-opinionated for a small
  ergonomic change, and the matching e2e workaround
  (`page.select_option("#launch_mode", "DOCKER")`) only existed to
  paper over the LIMA default on the docker-only CI runner. Flipping
  the default is a small follow-up PR that should stand on its own
  merits, independent of bundling Lima.
- Bundled-`git` work, `_bundled/` client-config staging, workspace-wheel
  build / pyproject rewrite, latchkey changes, auto-update changes, and
[...]

PR #1822, opened 2026-05-30
mngr_sbx: Docker Sandboxes provider on top of #1317 build machinery

weishi-imbue:
## Summary

The packaged `minds.app`'s `git clone` of the FCT template died with
SIGKILL (exit -9) at runtime on any Mac without Xcode Command Line Tools
installed at the expected path. Root cause: `scripts/build.js`'s inline
`downloadGit()` ran `which git`, which on macOS returns the 118 KB
`/usr/bin/git` xcode-select **shim** -- a launcher that re-invokes the
real git from `/Library/Developer/CommandLineTools/usr/bin/git`. The shim
itself isn't a runnable git binary; bundling it into a sandboxed app
means runtime `git` calls SIGKILL because the shim cannot reach the real
binary at the libexec layout it expects.

## Fix

Introduce `apps/minds/scripts/download-binaries.js` as the
platform-specific binary-download module:

- Resolves the real git binary via `xcrun --find git` on macOS (instead
  of trusting `which git`), then copies it plus its `libexec/git-core/`
  helpers and `share/git-core/templates/` into `resources/git/` so the
  bundled git works standalone with no dependence on the user's Xcode
  install layout.
- Skips Apple's ~100 argv-shim symlinks under `libexec/git-core/`
  (`git-add`, `git-commit`, etc. all point back at the main binary;
  dereferencing them would balloon the bundle by ~1GB).
- SHA256-verifies every downloaded archive against a pinned hash before
  extraction.

`build.js` now imports `downloadGit` from `download-binaries.js`.

Bumps `todesktop.json`'s `uploadSizeLimit` from 300 -> 600: the real git
binary plus its libexec helpers push the bundle over the previous limit
(empirically observed as "app larger than 300MB (409.5MB)" from
ToDesktop).

## Provenance

Extracted from #1317 (`wz/minds_onboard`). This is the second carve-out
from that branch, after the Lima-bundling extraction in #1731.

`download-binaries.js` also contains uv-download and pnpm-install
helpers for the ToDesktop `beforeInstall` hook -- they ship as part of
the same coherent module on `wz/minds_onboard`, so they come over whole
even though only `downloadGit` is wired into `build.js` here. Wiring the
ToDesktop hook (`package.json` -> `todesktop:beforeInstall`) is out of
scope for this PR; the git fix takes effect at the local-build path
that `pnpm build` already exercises.

## Should we be bundling git at all?

The PR fixes a bug caused by *how* we bundle. It's worth being explicit
about *why* we bundle, since the alternative (rely on the user's system
git) would let us drop ~30 MB and all of this code. Minds targets
non-engineer users, which is the load-bearing assumption.

| Dimension | Bundle git (this PR) | Rely on system git |
|---|---|---|
| macOS user without Xcode CLT | Works out of the box | First `git` call pops Apple's "command line developer tools" dialog -> ~700 MB download -> license accept -> 8-30 min wait; dialog mentions Xcode, not minds -- non-eng users may cancel or get stuck |
| Linux minimal / server image | Works out of the box | Distro-dependent; user must know their package manager (apt/dnf/pacman/...) |
| Windows | Works (Windows has no system git) | User must install Git for Windows separately (~50 MB installer) before minds runs |
| Installer size impact | +30 MB (current Apple-binary copy) or +57-60 MB (dugite-native) per platform | None |
| Build reproducibility | Today we copy from build host's Xcode -> different Xcode versions ship different gits. dugite-native (pinned versioned tarballs) would eliminate this | N/A -- runs whatever the user has |
| macOS code-signing | Have to re-sign with ToDesktop identity; re-signing Apple's xcode-select shim is exactly what produced this PR's SIGKILL bug | None |
| CVE / freshness tracking | Our burden -- stale bundled git stays stale until we ship an update | User's responsibility |
| First-launch failure surface | None -- git is there | High -- failure at the worst possible UX moment |
| Precedent (non-eng audience) | **GitHub Desktop** bundles git via [dugite-native](https://github.com/desktop/dugite-native) (~57-60 MB tarballs per platform); **SourceTree** ships embedded git with an option to switch to system; minds fits this pattern | None -- no non-eng git GUI relies on system git |
| Precedent (dev audience) | None of the major dev tools bundle | **VS Code** uses system git via PATH, prompts to install if missing; **Cursor** same model |

**Conclusion:** for a non-engineer audience, the same calculus that made
GitHub Desktop and SourceTree bundle applies to minds. Bundling stays.
The cleaner strategic upgrade (deferred to a future PR) is switching
macOS/Linux from "copy from build host's git" to pulling
[dugite-native](https://github.com/desktop/dugite-native)'s prebuilt
tarballs, mirroring what we already do for Windows MinGit. That removes
the build-host Xcode dependency, makes builds bit-for-bit reproducible
across developer machines, and pins git version + SHA256 like the
Windows path. Today's PR is the tactical fix that unbreaks shipping;
dugite-native adoption is the follow-up.

## Test plan

- [x] Locally: `node -e "const {downloadGit} = require('./apps/minds/scripts/download-binaries.js'); downloadGit('/tmp/git-test', { platform: 'darwin' })"` produces a 7.6 MB Mach-O universal binary at `/tmp/git-test/git/bin/git` (not the 118 KB shim), and `git --version` reports `git version 2.50.1 (Apple Git-155)` exit 0.
- [x] `ls /tmp/git-test/git/libexec/git-core/` contains 31 real helper binaries (post-shim-skip).
- [x] `git init` against the bundled binary creates `.git/{config,description,HEAD,hooks,info,...}` successfully.
- [x] `just test-quick apps/minds/scripts/build_test.py` -- 1 passed.
- [x] `just test-quick apps/minds/imbue/minds/test_ratchets.py` -- 57 passed.
- [ ] `just test-offload` (deferred to CI; ran subset locally; 5 unrelated `webdav_test.py` flakes also fail on `origin/main`).
- [ ] Full packaged-build verification on ToDesktop -- deferred; the build-server path is unchanged from #1317.


weishi-imbue:
The idea of mngr_sbx is to speed up local agents creation via docker sandbox. But it turns out to require user log into docker first. Closing it now as the gap is kind of small. 
Docker (docker desktop) -> requires user install & login.
Lima VM -> scriptable, no user input needed.

PR #1826, opened 2026-05-31
Stop nested claude -p from being adopted into the parent agent's transcript

gnguralnick:
## Problem

When a mngr-managed Claude agent runs a nested `claude -p` (or any nested `claude`) through its Bash tool, the child session was being adopted into the **parent** agent: its messages showed up in the parent's transcript, and the parent's lifecycle-state files / session tracking were corrupted.

Root cause: the nested process inherits `MAIN_CLAUDE_SESSION_ID` (and `CLAUDE_CONFIG_DIR`) from the parent session. mngr's readiness/transcript hooks are all guarded on `MAIN_CLAUDE_SESSION_ID` (the `SESSION_GUARD` in `claude_config.py`), so with that var non-empty the child's `SessionStart` hook appended its session id to `claude_session_id_history` — which `stream_transcript.sh` then streams into the parent's transcript — and `UserPromptSubmit`/`PermissionRequest`/`Stop` flipped the parent's state files.

The existing subagent-proxy already scrubs `MAIN_CLAUDE_SESSION_ID` for its Task-tool children, but a raw `claude -p` via Bash bypasses it entirely.

## Fix

Each Claude agent now provisions a small `claude` shim (`resources/claude_shim.sh`) into a dedicated dir (`$MNGR_AGENT_STATE_DIR/claude_shim/`) that `assemble_command` prepends to PATH. The shim unsets `MAIN_CLAUDE_SESSION_ID` before `exec`-ing the real binary, so the **existing** guard correctly no-ops for nested sessions across all hooks. `CLAUDE_CONFIG_DIR` is left intact, so the nested `claude -p` keeps the same credentials/config — it simply isn't streamed (the streamer is history-driven).

The **main** launch resolves the real binary via `command -v claude` *before* the shim is on PATH and runs it by absolute path, so the main session's environment — and therefore `/clear`, `/compact`, and resume — are untouched. No hook changes.

If `claude` is absent from PATH, the prelude now aborts with a clear message instead of falling through to a confusing empty-binary invocation.

## Test plan

- New `resources/claude_shim_test.py`: shim scrubs the var, forwards argv, and errors loudly when no real `claude` exists.
- New end-to-end test in `plugin_test.py` that executes the assembled command against a stub `claude` and asserts the main launch keeps `MAIN_CLAUDE_SESSION_ID` while a nested call (routed through the shim) has it scrubbed.
- New test asserting the clean `exit 127` + diagnostic when `claude` is absent.
- Updated the 5 existing `assemble_command` string-comparison tests.
- Headless agent is unaffected (it launches with `MAIN_CLAUDE_SESSION_ID` unset by design).

Locally: `plugin_test.py` + `claude_shim_test.py` (202), ratchets (55), subagent-proxy (166) all pass; ruff clean. Full offload suite runs in CI.

### DIFF
--- a/libs/mngr_claude/changelog/gabriel-claude-p-transcript.md
+++ b/libs/mngr_claude/changelog/gabriel-claude-p-transcript.md
@@ -1,0 +1,12 @@
+Fixed a bug where a nested `claude -p` (or any nested `claude`) run from inside
+a mngr-managed Claude agent's session would be adopted into the parent agent:
+the child session inherited `MAIN_CLAUDE_SESSION_ID`, so mngr's readiness and
+transcript hooks (all guarded on that variable) fired for it -- leaking the
+child's messages into the parent's transcript and flapping the parent's
+lifecycle-state files and session tracking.
+
+Each Claude agent now provisions a small `claude` shim onto its session PATH
+that strips `MAIN_CLAUDE_SESSION_ID` from nested invocations before exec-ing the
+real binary. The main agent launch resolves and runs the real binary directly,
+so its environment -- and therefore `/clear`, `/compact`, and session resume --
+are unaffected.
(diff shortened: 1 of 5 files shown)


joshalbrecht:
this is a bit too much complexity to me for now (esp given the other changes I want to make to claude -p)

I like your suggestion in slack though!

PR #1827, opened 2026-06-01
Permission descriptions

hynek-urban:
- Source permission descriptions from detent (via the `generate_services_json.py` script).
- Display permission descriptions in the permission dialog.
- Make permission descriptions part of `/permissions/available` so that agents can better reason about the permissions to ask for.
- Validate `POST /permission-requests` to disallow asking for non-existent permissions.

permission_dialog


joshalbrecht:
lgtm!

though I will note that from a UI perspective something feels wrong about having both information displayed to the user--my intuition is that they should *either* see `slack-auth-read` *or* they should see `Check authentication status and list teams`, but not both

PR #1829, opened 2026-06-01
fix misc warnings

evgunter:
Folds together several warning/cleanup fixes surfaced by `ty`, `uv audit`, and CI logs.

## Changes

**1. `ty` `possibly-missing-submodule` (libs/mngr)**
`cel_utils.py` subclassed `celpy.celtypes.MapType` via attribute access on the
top-level `celpy` package. Now imports `MapType` directly from `celpy.celtypes`.
No behavior change.

**2. GitHub Actions Node.js-20 deprecation (dev)**
GitHub forces Node-20 JS actions to Node 24 starting 2026-06-16 and removes
Node 20 on 2026-09-16. Bumped all flagged actions to their latest Node-24
majors: `actions/cache` v4->v5, `actions/upload-artifact` v4->v7,
`actions/setup-node` v4->v6, `actions/checkout` v4->v6 (vet.yml),
`extractions/setup-just` v2->v4, `mikepenz/action-junit-report` v5->v6,
`astral-sh/setup-uv` v6->v7. (setup-uv stays on v7 rather than v8 because v8
drops floating major tags, which would break the repo's `@vN` convention.)
Verified breaking-change notes; no usage relies on removed inputs, and
setup-node v5+ auto-caching does not trigger (no `packageManager` field).

**3. pytest collection warnings (libs/mngr_tmr, libs/modal_proxy)**
Added `__test__ = False` to `TestResult`, `TestMapReduceResult`, and
`TestingModalInterface` so pytest stops trying to collect these domain/testing
classes (names start with "Test"). Renaming was rejected as churn since the
names are semantically correct.

**4. `uv audit` vulnerable dependencies (dev / uv.lock)**
Upgraded `idna` 3.14->3.16 and `starlette` 1.0.0->1.0.1 (clean, isolated bumps).
Could not address: `litellm`/`fastapi-sso` (reaching fixed litellm forces a
workspace-wide downgrade cascade of cryptography/fastapi/starlette/boto3),
`paramiko` (no fix published), `rsa` (archived, no replacement pinned).

**5. Stale skill reference (libs/mngr_imbue_cloud)**
Updated a `minds-dev-iterate` -> `minds-dev-workflow` reference in the
unabridged changelog (kept the former name for historical clarity).

## Verification
- `uv run ty check` (whole repo): All checks passed
- `uv run ruff check` (whole repo): All checks passed
- `uv audit`: idna + starlette advisories cleared
- Collection warnings: 0 remaining on the affected test files
[...]

PR #1833, opened 2026-06-01
Add audit-ci skill, plus the CI fixes it surfaced

evgunter:
Adds a new \`audit-ci\` Claude skill, plus the concrete CI fixes the first audit surfaced.

## New skill: \`audit-ci\`
Documents the non-obvious parts of this repo's CI so an agent can find subtle degradations a human misses on green runs:
- **Synthesized check-runs.** \`test-offload\`/\`test-offload-acceptance\` don't contain the test results; a composite action POSTs separate \`Unit + Integration Tests\` / \`Acceptance Tests\` check-runs at the end (shown as "in 0s", \`/runs/<id>\` URLs). The retry+flaky table lives there; \`neutral\` = passed-after-retry is the flaky signal.
- The flaky-table format, the image-cache (git-note pointer to a Modal image, ~48h TTL), a repeated-log-noise/misconfig category, a catch-all, and a Calibration section for common false positives.
- **Subagent calibration**: fan-out subagents over-claim, so paste the caveats into their prompts or have them return raw evidence.

## Test flakiness: timeout bumps instead of \`@flaky\`
A census across recent runs found that every \`@flaky\`-marked test that actually flaked did so via \`pytest-timeout\` (10s default), all in the \`libs/mngr\` agent create/clone/destroy/list/cleanup families that do real work over tmux + subprocess -- never transient errors. So for those ~14 tests, \`@flaky\` (which just reruns a slow-but-correct test, and hides it) is replaced with a modest \`@pytest.mark.timeout(30)\`. This deliberately drops them from offload's retry group so a genuinely-broken one surfaces instead of being silently retried.

## Electron e2e log noise
\`test_create_local_docker_workspace_via_electron\` was logging \`Modal is not authorized\` every ~10s because the job has no Modal token. Disabled the modal provider for the Electron child (\`MNGR__PROVIDERS__MODAL__IS_ENABLED=false\`, inherited via \`_build_electron_env\`).

## Scope the offload checkout to HEAD's history
\`test-offload\` and \`test-offload-acceptance\` used \`fetch-depth: 0\`, which fetches the full history of *every* branch (~717, including a heavy orphan screenshots branch offload never needs). They now do a default shallow checkout + \`git fetch --unshallow origin <head-sha>\`, deepening only the checked-out commit's ancestry -- all offload needs to find and thin-diff its checkpoint. Measured on this PR's CI: checkout+deepen dropped to **~11s** (fetching 1 ref instead of 717), vs ~19-30s for a normal \`fetch-depth: 0\` and 206s on a congested run; offload still gets a cache hit (full blobs of HEAD's ancestry are retained, so the base-image export still resolves -- unlike a blobless partial clone).

The broader audit findings (Node20 GHA deprecation handled by #1829, Modal acceptance-test flakiness that retries absorb, healthy docker caching) were reported separately, not committed.


joshalbrecht:
lgtm, thanks for fixing!

PR #1836, opened 2026-06-01
replace emit_final_json with public write_json_line

gnguralnick:
## What

`emit_final_json` (in `imbue.mngr.cli.output_helpers`) was a one-line pass-through to the private `_write_json_line`. Its name was also misleading: despite "final JSON" implying the single terminating object emitted in `--format=json` mode, it was called from streaming JSONL callbacks too (e.g. `mngr list`'s `_emit_jsonl_agent` / `_emit_jsonl_error`).

This makes the primitive public as `write_json_line` (sibling to the existing `write_human_line`), removes `emit_final_json`, and updates all ~74 call sites + imports across 6 packages (`mngr`, `mngr_wait`, `mngr_usage`, `mngr_schedule`, `mngr_latchkey`, `mngr_file`).

`write_json_line` honestly describes what it does -- write one JSON object as a line -- for both the JSON (single terminating object) and JSONL (streamed) paths. The thin `list.py` wrappers (`_emit_json_output`, `_emit_jsonl_agent`, `_emit_jsonl_error`) are intentionally left in place; only their inner call is renamed.

No user-visible behavior change -- the byte output is identical.

## Testing

- `just test-quick libs/mngr/imbue/mngr/cli/output_helpers_test.py` -- 49 passed, 0 failed (dropped the now-redundant `test_emit_final_json_outputs_json`).
- Smoke-imported all 26 changed modules; confirmed `write_json_line` present and `emit_final_json` gone.
- Full suite via CI.


joshalbrecht:
lgtm!

Issue #1825, opened 2026-05-31
Modal: SSH tunnel connection reset by peer during provisioning (consistent, all accounts)


DaRubberDuckieee:
## Summary
`mngr create` on Modal consistently fails with SSH `Connection reset by peer` (errno 54) during file upload provisioning. The `wait_for_sshd` check passes (SSH banner received via raw socket), but paramiko's subsequent SSH session is immediately reset by Modal's tunnel.

## Repro
```bash
mngr create test@.modal claude --no-connect --idle-timeout 60s --message "hello"
```

## Environment
- **mngr**: 0.2.9 (also reproduced on 0.2.8)
- **modal**: 1.4.3
- **paramiko**: 4.0.0 (also 3.5.1 on 0.2.8)
- **OS**: macOS (Apple Silicon)
- **Modal profiles tested**: 2 different workspaces, same result
- **Attempts**: 5+ across both profiles

## Error
```
HostConnectionError: Failed to connect to host: SSH error (Error reading SSH protocol banner)
```

## Debug log detail
```
[paramiko] Exception (client): Error reading SSH protocol banner[Errno 54] Connection reset by peer
```

## Analysis
1. Modal sandbox creates successfully (host name generated)
2. Image builds and packages install
3. sshd starts, `wait_for_sshd()` passes — raw socket receives SSH-2.0 banner
4. paramiko opens a new connection for file upload → **connection reset by peer**

The tunnel appears to work for the initial TCP/banner check but resets subsequent SSH sessions. This happens on every attempt, not intermittently.

## Modal container list
`modal container list` works fine — Modal API connectivity is not the issue. Only the SSH tunnel is broken.

PR #1837, opened 2026-06-01
use rsync instead of individual file uploads more + add helpers and ratchet test

evgunter:
## Summary

Fixes github issue #1825: `mngr create @.modal` (and other remote hosts) consistently failed during provisioning with `Error reading SSH protocol banner` / `Connection reset by peer`.

## Root cause

Deploy/provision files were uploaded **one SFTP channel per file** (open→putfo→close inside a loop). Each channel open is a full round-trip over the SSH tunnel — ~0.7s/file over Modal. The deploy set is dominated by the user's `~/.claude/plugins` tree (hundreds of files), so the upload took minutes: it blew past the 300s timeout (`TimeoutError`), and the sustained churn intermittently reset the transport (banner error). The SSH tunnel/sshd were healthy throughout.

## What's in this PR

1. **Shared bulk-upload helper** `imbue.mngr.hosts.file_upload.upload_files_in_bulk`: stages files into a local temp dir mirroring their remote paths and transfers them with a **single rsync** (remote) or direct writes (local). Measured ~900 files in seconds vs. minutes.

2. **New `copy_local_directory` host primitive**: rsync from the local machine to a host. The source is implicitly the local machine (no source-host object), so `provision_agent`'s signature is unchanged — this avoids the host-layer→`api.providers` import cycle (via `base_provider`) without threading a `local_host` parameter through the call stack.

3. **Routed every per-file-over-SSH loop through the helper**, unifying staging that was previously duplicated across `Host.provision_agent` (user uploads + agent file transfers), `mngr_recursive` (deploy files), `mngr_claude` (generated config + home-dir rsync), and `mngr_pi_coding` (resource dirs). The shared shell-lib writes stay per-file because they need `mode=0755`, which rsync staging doesn't preserve.

4. **`get_local_host` hoisted** into `imbue.mngr.api.providers`, deduping its copies.

5. **New ratchet `check_per_file_host_upload`** (AST): flags `write_file`/`write_text_file`/`put_file` inside loops; rolled to all projects via `sync_common_ratchets.py`. Baselines after the fixes: mngr=2 (the two `0755` shell-lib writes), mngr_claude=1, modal_proxy=1, rest 0.

6. **`created_host` test context manager** (`imbue.mngr.api.testing`), adopted across the Modal provider tests.

## A deterministic Modal regression found and fixed along the way

While unifying the staging, the rsync target was briefly `/`, which deterministically broke `mngr create` on Modal with `AgentNotFoundOnHostError`. Root cause (fully understood, not worked around):

- On Modal, `host_dir` (`/mngr`) is a **symlink** into the mounted volume, so `agents/<id>/data.json` lives on the volume *through* that symlink.
- `rsync` **without `--keep-dirlinks`** deletes a receiver-side symlink-to-directory and replaces it with a real directory on the ephemeral filesystem when the sender has a real directory at that path. Rsyncing a staged tree with a top-level `mngr/` into `/` therefore clobbered the `/mngr` symlink into a fresh ephemeral dir, stranding `data.json` on the volume — config appeared to "upload fine" (it landed in the ephemeral dir) but the agent became undiscoverable.

Fix: `copy_local_directory` now passes `--keep-dirlinks` so rsync writes *through* such symlinks to the underlying storage, and `upload_files_in_bulk` rsyncs into the tightest common-ancestor directory of its destinations (hygiene — never stamps perms/mtimes on directories it isn't deliberately writing into). Verified with GNU rsync 3.2.7 (clobber without `-K`; correct write-through with `-K`, even staging from `/`).

## Verification

- Reproduced both failure variants on `main`.
- Real `mngr create @.modal claude` completes end-to-end (host created, config rsynced, agent started, message sent, snapshot created, discoverable in `mngr list`); a real **local** create also verified.
- Modal acceptance test uploads a large (600-file) deploy set to a real host: asserts an upload-time budget, full delivery, **and** that a pre-existing (non-uploaded) file survives (rsync is additive, no `--delete`).
- New `file_upload_test.py` covers resolve/stage/local/remote/skip-missing/no-delete/same-file. Unit + ratchet + type + lint green; offload suite: 0 failures.


libs/mngr/imbue/mngr/hosts/file_upload.py
```diff
@@ -54,27 +56,39 @@ def _write_source(source: FileSource, dest: Path) -> None:
...
-    ``copy_directory`` (rsync). For a local target the files are written directly to
-    the destination as given (preserving ``write_file``'s relative-to-cwd semantics);
-    ``remote_home`` is unused and rsync is unnecessary since writes are local.
+    temp dir, and transferred with one ``copy_local_directory`` (rsync) into the
+    *tightest common-ancestor directory* of the destinations. Targeting the common
+    ancestor (rather than "/") keeps rsync from stamping perms/mtimes on intermediate
+    directories we are not writing into (the volume root, the agent-state dir, ...).
+    Note that what actually protects against clobbering a symlinked directory (e.g.
+    Modal's host_dir symlink into the mounted volume) is ``--keep-dirlinks`` inside
+    ``copy_local_directory`` -- the common-ancestor scoping is hygiene on top of that.
```
evgunter, line 80: not sure if the common ancestor thing is actually necessary

joshalbrecht:
I wanted to fix this for a while, thanks ev!

lgtm

PR #1839, opened 2026-06-01
[review-only] workspace-restart: changes since review of #1707

gnguralnick:
## Review-only companion to #1707

This PR exists **only to make re-review easy** — it is **not for merge**. The real PR is #1707; that's where these changes land.

Review comments on #1707 were all against commit `8cc31a4`. This PR is based on exactly that commit and contains **only my own changes since then** — the main-branch merges and other contributors' commits that also landed on the feature branch in the meantime have been excluded, so the diff is just my response to the review.

### Commits (oldest → newest)
- Simplify discovery offline fallback to assume offline-host support
- Add FIXME for transient-failure retry in agent discovery
- minds: minimize the mngr-subprocess recovery interface; narrow its exception catch
- minds: decouple workspace readiness/health probes from the SI app
- minds: let mngr launch failures propagate instead of laundering them into a return value
- minds: unify system-services fallback on a last-good agent topology
- mngr: drop redundant destroyed-host tracking from stop-host resolution
- mngr: resolve --stop-host via provider get_host, not a fleet-wide scan
- mngr_forward: collapse HTTP failure reasons to ERROR_RESPONSE; de-minds-ify framing
- mngr_forward: de-minds-ify the coupling PR #1589 added to the plugin
- mngr_forward: de-minds-ify remaining behavioral references in plugin-creation framing

Base branch `gabriel/workspace-restart-josh-reviewed-base` is pinned at `8cc31a4` (the snapshot reviewed). Do not merge this PR.


apps/minds/imbue/minds/desktop_client/app.py
```diff
@@ -2152,9 +2141,40 @@ def _run_host_health_probe(
...
     list_argv = _build_mngr_host_state_argv(mngr_binary, agent_id, services_agent_id)
     list_command = shlex.join(list_argv)
-    finished, _mngr_list_error = _run_mngr_subprocess(concurrency_group, list_argv, env)
-    list_json: str | None = finished.stdout if finished is not None else None
-    in_container_stdout = _run_batched_probe(concurrency_group, mngr_binary, services_agent_id, env)
+    try:
+        list_run = _run_mngr(concurrency_group, list_argv, env)
+        list_stdout = list_run.stdout
+        list_error = list_run.failure_reason
+    except (OSError, ConcurrencyGroupError) as exc:
```
joshalbrecht, line 2148: this feels like a weird thing to have to catch... why always both of these errors?

generally our concurrency_group.run_process_to_completion should be returning a fairly nice error that will even include the stdout, stderr, exit code, etc already on it?

(same comment in other place below as well)

joshalbrecht:
lgtm!  other than my one comment

apps/minds/imbue/minds/desktop_client/app.py
gnguralnick: turns out despite me asking several rounds claude still refused to follow the common practice in the repo for shelling out to mngr. fixing

PR #1840, opened 2026-06-01
require .gitkeep in changelog dir

evgunter:
## What

`test_meta_ratchets.py::test_every_project_has_changelog_layout` was failing because `libs/mngr_mapreduce/changelog/` no longer existed.

## Why it happened

Every project's `changelog/` directory is kept in git by a `.gitkeep` file, but `libs/mngr_mapreduce/changelog/` never had one. When the 2026-06-01 changelog consolidation run deleted its last per-PR entry file, the now-empty directory vanished from git (git does not track empty directories), tripping the meta-ratchet that requires every project to have a `changelog/` directory.

## Fix

1. Add an empty `.gitkeep` to `libs/mngr_mapreduce/changelog/` so the directory stays tracked even when it holds no pending entries, matching every other project.
2. Tighten `test_every_project_has_changelog_layout` to also require a `.gitkeep` inside each project's `changelog/` directory. Previously only the directory's existence was checked, so a newly added project lacking a `.gitkeep` would pass until a later consolidation run drained its entries and the empty dir silently vanished. Checking for the placeholder upfront catches the omission when a project is first added, rather than later when it happens to be consolidated.

Verified `test_every_project_has_changelog_layout` and `test_pr_has_changelog_entry` both pass locally.


joshalbrecht:
lgtm

PR #1842, opened 2026-06-02
Fix Minds App request refreshing

hynek-urban:
The request side panel used to be stale - changes in permission requests wouldn't propagate until the panel was closed & reopened. Auto-opening didn't work and the red dot notification didn't work, either.


joshalbrecht:
lgtm

PR #1843, opened 2026-06-02
Do not deduplicate permission requests

hynek-urban:
Permission requests used to be "deduplicated" by their content. That meant that when an agent sent the same request more than once, multiple requests would be represented by a single item on the frontend. When the user approved or denied the request, it looked broken because the processed request was immediately replaced by another one with the same content that was previously hidden due to "deduplication".

It's better to not attempt any magic and instead transparently show all the permission requests that were sent by agents, duplicates or not.


joshalbrecht:
we'll need to come back to this one, but lgtm for now

PR #1844, opened 2026-06-02
Consolidate error hierarchy: HostError inherits from MngrError

joshalbrecht:
## What

Makes `HostError` (and all of its subclasses) inherit from `MngrError` instead of `BaseMngrError`, consolidating the mngr error hierarchy so that all user-facing errors share a single parent class.

`MngrError` is a `click.ClickException`; `BaseMngrError` is not. So this change makes every host error a `ClickException`:

- An uncaught host error reaching the CLI now renders as a clean `Error: <message>` (plus any `user_help_text`) instead of a full Python traceback. Exit code stays 1.
- `except MngrError` handlers now treat host errors as the user-facing errors they are.

## Behavior notes

- All `HostError`-family raise sites pass exactly one string message, compatible with `ClickException.__init__`; the multi-arg subclasses (`CorruptedAgentDataError`, `HostDataSchemaError`) already funnel to a single `super().__init__(message)`. MRO for `InvalidActivityTypeError(HostError, ValueError)` is consistent (mirrors the existing `ParseSpecError(MngrError, ValueError)`).
- Audited all ~110 `except MngrError` sites: none required modification. The broadening is either a no-op (sites already catching `BaseMngrError`/`HostError`) or benign/an improvement (best-effort/log-and-continue paths now uniformly handle host errors; the `error_behavior`-driven API handlers now honor CONTINUE mode for host errors instead of aborting).
- One observable change: in `api/list.py` provider-discovery `ErrorBehavior.ABORT`, a host error is now re-raised as its original type instead of being flattened into a generic `MngrError` (an improvement -- preserves type and help text).

## Category C simplifications

Removed the now-redundant host-error entries from catch tuples that also caught `MngrError`:

- `mngr_vps_docker/instance.py`: `except (HostConnectionError, MngrError)` -> `except MngrError` (7 sites)
- `mngr_imbue_cloud/instance.py`: dropped `HostConnectionError`/`HostNotFoundError` from the tuple
- `mngr_mapreduce/launching.py` + `cli.py`: dropped `HostError` (kept `AgentError`, which is still a `BaseMngrError`, not a `MngrError`)
- `providers/docker/testing.py`: dropped redundant `HostNotFoundError`
- Updated two stale comments in `mngr_vps_docker/host_store.py` and `mngr_mapreduce/agent_stopper.py`.

## Tests

- Added `errors_test.py` coverage: host errors are `MngrError`/`ClickException` subclasses, and a host error renders with a single `Error: ` prefix via Click.
- Locally: `errors_test.py` + `api/list_test.py` (132 passed), changed plugin unit tests (445 passed), ratchet tests for affected projects (220 passed), `ruff` and `ty` clean. Full suite runs in CI.

### DIFF
--- a/libs/mngr/changelog/mngr-fix-error-hierarchy.md
+++ b/libs/mngr/changelog/mngr-fix-error-hierarchy.md
@@ -1,0 +1,6 @@
+`HostError` (and all of its subclasses, e.g. `HostConnectionError`, `HostOfflineError`,
+`HostAuthenticationError`, `CommandTimeoutError`, `HostDataSchemaError`) now inherit from
+`MngrError` instead of `BaseMngrError`, consolidating the error hierarchy under a single
+user-facing parent class. Host errors are now `ClickException` instances, so when one reaches
+the CLI it renders as a clean `Error: ...` message (plus any help text) instead of a Python
+traceback, and `except MngrError` handlers treat them as the user-facing errors they are.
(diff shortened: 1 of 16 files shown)


joshalbrecht:
lgtm

PR #1846, opened 2026-06-02
Permission requests in modal view

hynek-urban:
Render the permission request dialog in a modal window so that user doesn't lose the context of their work. (Requested by Kanjun and partially also by Gleb.)

modal_window


joshalbrecht:
lgtm!

PR #1847, opened 2026-06-02
Use find_all_agents in `mngr message`

qi-imbue:
## My summary

When trying to fix `mngr message` to pass the set of providers to discovery, I realized that it could just use `find_all_agents`, which doesn't have this bug. What's previously preventing it from using `find_all_agents` is a bunch of defunct code supporting `mngr message --include --exclude`, but the flags have been removed.

Also remove the `mngr message --provider` flag.

## Summary

Fix `mngr message` (alias `msg`) ignoring the providers named in agent addresses, and consolidate it onto the same discovery plumbing the other agent-address subcommands already use.

### Behavior change

- **Bug fix.** Previously, `mngr message` passed `--provider` straight into discovery and ignored any provider qualifier on the agent addresses themselves. `mngr msg agent@host.provider_a --provider provider_b` would query `provider_b` and report the agent missing.
- **Remove `--provider` from `mngr message`.** The set of providers queried during discovery is now derived from the addresses: the union of providers named by the addresses, or a full scan if any address omits its provider.
- No other subcommand has this bug. Every other agent-address subcommand (`stop`, `start`, `exec`, `archive`, `label`, `destroy`, `snapshot`, `limit`, `rename`, `connect`, `capture`) already derives provider scope from the addresses via `find_all_agents`, and none of them expose `--provider`.

### Internal cleanup (large impact)

- `mngr message` no longer maintains its own CEL-filter pipeline parallel to `find_all_agents`. The CLI now calls `find_all_agents` directly, like every other agent-address subcommand.
- `send_message_to_agents` is now an `AgentMatch`-in, `MessageResult`-out function. The CEL `include_filters` / `exclude_filters` / `all_agents` / `provider_names` parameters and the per-host CEL re-evaluation inside `_process_host_for_messaging` are gone. The dead `agent_to_cel_context` helper is deleted.
- `libs/mngr_uncapped_claude/orchestrator.py` (the only out-of-tree caller) is migrated; it constructs the `AgentMatch` directly from the local Claude agent it already has in hand, avoiding an unneeded discovery round-trip.

Net diff: +114 / −192 lines.

## Test plan

- [x] `just test-quick "libs/mngr -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'"` → 4021 passed, 1 skipped
- [x] `uv run pytest libs/mngr/imbue/mngr/api/message_test.py -v` (including the `@pytest.mark.tmux` tests) → 9 passed
- [x] `just test-quick libs/mngr_uncapped_claude` → 137 passed
- [x] `test_no_type_errors` and `test_no_ruff_errors` from `test_meta_ratchets.py` → pass
- [x] Sanity-checked `mngr message --help` (no `--provider` shown) and `mngr message nonexistent-agent -m hello` (prints "No agents found to send message to", exits 0)
- [ ] CI (offload + acceptance + release) green

### Manual things worth poking at

- `mngr msg my-agent -m hello` against a single agent on the default provider — basic regression check.
- `mngr msg agent1 agent2 -m hello` with two agents on **different** providers — confirms the union-of-providers logic actually queries both providers.
- `mngr msg agent@host.someprovider -m hello` — confirms the provider qualifier on the address restricts discovery to just that provider (the original bug case).
- `mngr msg agent_that_does_not_exist -m hello` — should still exit 0 with "No agents found to send message to" (preserved graceful behavior).
- `mngr msg agent@host.someprovider -m hello --on-error abort` against a nonexistent agent — should surface an `AgentNotFoundError` rather than silently exit 0.
- `mngr uncapped-claude` follow-up turn delivery (`_send_user_turn`) — exercises the migrated `mngr_uncapped_claude/orchestrator.py` caller against a real local Claude agent.


joshalbrecht:
lgtm, thanks for the cleanup!

PR #1849, opened 2026-06-02
Consolidate error hierarchy: all mngr errors inherit from MngrError

joshalbrecht:
## Summary

Completes the error-hierarchy consolidation (started with `HostError`) so that **every** mngr error inherits from `MngrError` as the single user-facing parent. Every error is now a `ClickException`: an uncaught instance renders as a clean `Error: ...` message instead of a Python traceback, and `except MngrError` handlers treat them as the user-facing errors they are.

Each formerly-`BaseMngrError`-only type was independently investigated (raise sites, catch sites, MRO, constructor/`ClickException` compatibility, CLI reachability, tests) and found safe to move.

### Types moved to `MngrError`
- **`AgentError`** and its subclasses `NoCommandDefinedError`, `AgentNotFoundOnHostError`, `SendMessageError`, `AgentStartError`. `AgentNotFoundError` / `DuplicateAgentNameError` already reached `MngrError` via an explicit mix-in, now dropped as redundant.
- `PluginSpecifierError`, `DiscoverySchemaChangedError`, `MalformedJsonlLineError` (`errors.py`)
- `TolerantPathError` (`utils/cel_utils.py`) — keeps its `TypeError` base
- `IssueSearchError` (`cli/issue_reporting.py`)
- `UncappedClaudeError` (`mngr_uncapped_claude`) — also fixes a prior inconsistency where its subclasses were already `MngrError` via `UserInputError` while the base was not

### Cleanups
- Dropped the now-redundant `AgentError` entry from the `except (MngrError, AgentError, ...)` guards in `mngr_mapreduce` launching/cli.
- Updated a stale comment in `message_test.py`.

### Behavioral notes
- Per discussion, **no** special re-raise carve-out was added for agent errors in the listing per-agent guard (unlike `HostConnectionError`).
- The only meaningful behavior change is at the CLI boundary (clean rendering) and in `list.py`'s ABORT path, where agent errors are now re-raised with their original type preserved instead of being wrapped in a fresh `MngrError`.

### Tests
- Added invariant tests asserting each moved type is a `MngrError`/`ClickException` (parametrized, mirroring the `HostError` tests), plus a CLI-render test for `AgentStartError`.
- Added `errors_test.py` for `mngr_uncapped_claude`.

Ran locally (all passing): `errors_test.py`, `message_test.py`, `list_test.py`, `discovery_events_test.py`, `cel_utils_test.py`, `issue_reporting_test.py`, `host_test.py`, full `mngr_mapreduce` and `mngr_uncapped_claude` suites, the affected `test_ratchets.py` files, and the import-layer meta ratchet. `ty` type check passed on push.

### DIFF
--- a/libs/mngr/changelog/mngr-fix-error-hierarchy-agent.md
+++ b/libs/mngr/changelog/mngr-fix-error-hierarchy-agent.md
@@ -1,0 +1,12 @@
+`AgentError` (and all of its subclasses, e.g. `NoCommandDefinedError`, `AgentNotFoundOnHostError`,
+`SendMessageError`, `AgentStartError`) now inherit from `MngrError` instead of `BaseMngrError`.
+The remaining `BaseMngrError`-only error types -- `PluginSpecifierError`,
+`DiscoverySchemaChangedError`, `MalformedJsonlLineError`, `TolerantPathError`, and
+`IssueSearchError` -- were moved the same way. This completes the consolidation of the error
+hierarchy under a single user-facing parent class: every mngr error is now a `ClickException`,
+so when one reaches the CLI it renders as a clean `Error: ...` message (plus any help text)
+instead of a Python traceback, and `except MngrError` handlers treat them as the user-facing
+errors they are.
+
+The now-redundant `MngrError` mix-in on `AgentNotFoundError` and `DuplicateAgentNameError`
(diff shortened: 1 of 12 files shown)


joshalbrecht:
merged as part of https://github.com/imbue-ai/mngr/pull/1852

PR #1851, opened 2026-06-02
Clean up leased OVH hosts on release + hourly cron

joshalbrecht:
## Summary

Fixes leased-host cleanup in the minds remote connector service. Previously, releasing a lease only flipped `pool_hosts.status` to `released` and set `released_at` -- the OVH VPS was never cancelled and its per-lease tags (`minds_env`, `mngr-host-id`) were never stripped, so released hosts leaked forever and were never recyclable for new pool bakes. The `cleanup_released_hosts.py` script existed but was never deployed/scheduled.

## What changed

- **Release route does the work inline (best-effort).** `release_host` now runs an idempotent, two-phase-commit-style chain: flip the row to a new `removing` status (the durable intent marker) -> strip the per-lease OVH tags (keeping `mngr-provider=ovh`) -> cancel the VPS in OVH (`deleteAtExpiration=true`, **by service name** so no tag resolution is needed) -> delete the row. Returns 200 once `removing` is committed (OVH flakiness never blocks the caller); a gone row reads as `already_released`; ownership/paid gates unchanged.
- **Direct OVH calls, no mngr in the image.** Adds a tiny `OvhOps` protocol + `HttpOvhOps` (official `ovh` SDK) + pure cleanup helpers in the self-contained `app.py`. The connector now receives an `ovh-<tier>` Modal secret (pushed by `minds env deploy`).
- **Hourly cron** (`cleanup_removing_pool_hosts`) mops up any row stuck in `removing` from a crashed/timed-out release (`FOR UPDATE SKIP LOCKED`, processes all such rows, logs per-host failures).
- **Runbook.** `cleanup_released_hosts.py` rewritten into a broad, dry-run-by-default operator tool that tag-scans the OVH account and normalizes+cancels every `mngr-provider` VPS (protecting `available`/`leased`-backed VPSes unless `--include-active`). Use it to clean the 8 existing live hosts.

End state of a cleaned host: cancelled, carrying only `mngr-provider=ovh` -> recyclable by the next bake.

## Tests

- `just test-quick apps/remote_service_connector` -> **238 passed, 0 failed**.
- New unit tests cover the release chain (tags stripped + VPS cancelled + row deleted), idempotent `removing`/missing-row cases, 200-on-OVH-failure, and the cron sweep (cleans only `removing`, keeps rows when OVH fails).
- `ty check` clean for all changed files; ratchets updated (`init_methods` 2->3 for the new `HttpOvhOps.__init__`, mirroring `HttpCloudflareOps`; `direct_subprocess` 2->1 since the runbook no longer shells out).

## Notes / open items

- The runbook's "broad across all tiers" stance assumes staging/production pools are empty now; it still protects `available`/`leased`-backed VPSes by default and is dry-run-first.
- OVH creds + DB DSNs for the runbook are sourced by the operator from Vault (it does not shell out to `vault` itself, to keep the connector free of a cross-app dependency).


joshalbrecht:
merged in https://github.com/imbue-ai/mngr/pull/1879

PR #1853, opened 2026-06-02
identify-suspicious-edge-cases skill + fixes found in testing it

evgunter:
pulled the actual skills into https://github.com/imbue-ai/mngr/pull/2194

i went over all the fixes, but a lot of the time i didn't have the context to know if the fix was right; leaving as draft since some of the fixes seemed scary (especially in concurrency-group)

---

## Summary

Adds a new Claude skill, `identify-suspicious-edge-cases`, in the same family as the existing `identify-*` skills (`identify-inconsistencies`, `identify-style-issues`, etc.).

It scans a given library for over-defensive edge-case handling and reports ranked candidates into `<library>/_tasks/suspicious-edge-cases/<date>.md`, in the standard `identify-*` format so findings feed into `create-fixmes`.

## Motivation

An intermittent, post-merge cleanup pass to counteract the tendency to *over-handle* edge cases: catching errors that should be allowed to crash, fallback `else` branches that paper over impossible states, and defensive guards that mask bugs. This aligns with the style guide ("Be very conservative with what exceptions are caught. Prefer to crash instead of catching errors.", mandatory `else` clauses that raise, `assert_never` for exhaustiveness).

## What it looks for

`else` clauses in if/elif chains, `except` clauses, fallback default values (`.get(k, default)`, `getattr(..., default)`, `or`-defaults), defensive guards (`if x is None`, `hasattr`/`isinstance`), and broadened return types.

For each candidate it asks: (1) is the logic semantically unassailable — and is *this specific handling* the right response, not just *some* handling? (2) can the case be avoided entirely via control flow or by leveraging the type checker (ty)? (3) could it silently produce wrong output?

## Reporting stance

Errs toward over-reporting. The remedy for a correct-but-flagged handler is to add a clarifying comment explaining why the logic is intentional — turning an unexplained edge case into documented, intentional logic.

---

## Issues fixed by agents while testing the skill

To demonstrate the skill in practice, agents ran it across several libraries/subdirs and fixed the highest-signal findings (plus some lower-signal ones). All are over-defensive handlers that could silently mask failures or produce wrong output. Grouped by project:

### `concurrency_group` (the scariest — flagged for careful review)

- `poll()` no longer fabricates a magic `1007` exit code for a run-thread that finished with neither a result nor an exception; it raises `ProcessInvariantError` instead of letting a fake status flow downstream.
- A submitted task that raises `BaseException` (e.g. `KeyboardInterrupt`) now completes its future via `set_exception` instead of leaving `result()` hung forever.
- `_shutdown_popen` raises `ProcessTerminationError` when a process cannot be killed, rather than returning a clean-looking `None` return code.
- Group teardown no longer swallows thread-join exceptions with a blanket `except Exception: pass`, and a dominant `BaseException` during exit no longer discards sibling `Exception` failures (they are aggregated into a `ConcurrencyExceptionGroup`).
- Force-kill on timeout now uses a real grace period and reports accurately when a process could not be killed (the message previously always claimed success).
- `ShutdownEvent()` constructed bare is now always safe to call `is_set()` on.
- New error types `ProcessInvariantError` and `ProcessTerminationError`.
- **Follow-up race fix** (cherry-picked from `mngr/review-tests`, where CI hit it on a tmux test): the stricter `poll()` invariant above exposed a memory-visibility race — `poll()` read `_completed_process` without synchronizing against the run thread's write, so under load it could observe a stale `None` and raise a spurious `ProcessInvariantError`. `poll()` now `join()`s the finished thread first (a happens-before edge that makes the result visible and re-raises any captured exception) before deciding the invariant was violated.

### `mngr`

- **config**: empty/whitespace `backend`/`plugin` values are no longer silently replaced by the section name; an untraversable `__extend` base path now raises instead of silently degrading to a plain assign; the `completion_writer` cache write no longer swallows unrelated `OSError`s from the build phase (only the final write is wrapped); `model_construct` tuple-field normalization now rejects unrecognized types (the prior "Pydantic will validate" comment was wrong).
- **utils**: `count_commits_between` raises instead of returning a `0` indistinguishable from the valid answer; `get_editor_command` raises instead of returning a known-absent `vim`; CEL filtering no longer catches broad `TypeError` (which masked real bugs as the wrong filter result); `LoggingConfig.merge_with` now merges on the set fields (its prior `is not None` guards never fired and clobbered unset fields with defaults); ANSI truecolor sequences render correctly; `install_hint`/`read_json_dict` no longer silently mishandle unknown platforms / non-object JSON.
- **hosts**: `get_uptime_seconds` and `get_ssh_connection_info` return `None` (typed `float | None` / `Path | None`) instead of misleading `0.0` / `Path("")` sentinels; submodule detection no longer swallows connection failures; env-file values are quoted with `shlex.quote`; several missing/corrupt agent-data paths now raise instead of fabricating defaults. The SSH-key return-type change threads `Path | None` through `SSHInfo`/`RemoteSSHInfo` and their callers.
- Corrected `non_issues.md`: fixed a typo and reworded the two CLI-options entries so they no longer falsely claim the style guide endorses omitting `Field(description=...)`.

### `mngr_notifications`

- The agent-state watcher no longer crashes or silently drops a WAITING notification when it reads a torn (partially-written) trailing line from the concurrently-appended events file — it uses `split_complete_lines` plus a `MalformedJsonLineWarner`, advancing its read offset only past complete lines.
- Required event fields (`agent_id`, `agent_name`) are read directly and raise on a malformed record instead of defaulting to `"unknown"` (which collapsed distinct agents together and produced "unknown is waiting" notifications).
- A real plugin-config type mismatch now raises instead of silently discarding the user's configured settings; likewise a foreign override type in `merge_with` raises instead of being dropped.
- `mngr notify` on an unsupported platform surfaces a clear error instead of exiting silently.
- File-read and alerter-invocation error handling was narrowed so genuine `OSError`s are no longer mislabeled as "no events" / "alerter not found".
- New library error module with `NotificationsError`-based types.

### SSH-key type change (`Path | None`) — downstream adaptations

The `mngr` hosts fix above changed SSH key paths from an empty-`Path("")` sentinel to a typed `Path | None`. Threaded through the providers:

- `mngr_forward`: the SSH tunnel (including the paramiko fallback) omits the `-i`/key when no mngr-owned key is present.
- `mngr_modal`: the key path threaded through the instance is now `Path | None`.
- `mngr_vps_docker`: now `Path | None`; the human-readable SSH command for a discovered host is built via `build_user_ssh_command`, which omits `-i` when there is no key path instead of emitting `ssh -i None`.
- `apps/minds`: parsing agent SSH info from `mngr list` JSON yields `None` for a missing `key_path` instead of constructing an empty path.

### `resource_guards`

- Corrected the `ResourceGuardMisconfiguration` docstring: an empty `@fixture_uses_resources()` call is allowed and is a no-op, so the docstring no longer lists "empty" as a misuse. Docstring-only.

### Skill generalization (`dev`)

- Generalized the existing `identify-*` skills (`identify-inconsistencies`, `identify-style-issues`, `identify-outdated-docstrings`, `identify-doc-code-disagreements`) to take a **target path** instead of a library name: each accepts a whole library or any subdirectory within one, derives the containing library for context, and writes findings into that library's `_tasks/` folder.


libs/mngr/non_issues.md
```diff
@@ -1,6 +1,6 @@
 The following are things that you often think are issues, but in fact are not:
 
 - default arguments in api/*.py top level command functions (e.g., list.py::list()). These are fine as they are the main entrypoints and having defaults makes sense for usability.
-- missing "is_" prefix for boolean options in CLI functions and data classes. The style guide explicitly says that this is ok.
-- missing Field and description for CLI options data classes. The style guide explicitly says that this is ok, since the descriptions are on the click decorators instead (and we don't want to duplicated them)
+- missing "is_" prefix for boolean options in CLI command functions and CLI-options data classes (e.g. ListCliOptions). These mirror user-facing CLI args, which the style guide explicitly exempts from the is_ convention. (Internal boolean fields on non-CLI data classes should still use is_.)
+- missing Field and description for CLI-options data classes (e.g. ListCliOptions). These classes intentionally mirror the click options, with the descriptions/defaults living on the click.option() decorators instead (and we don't want to duplicate them), as documented on the classes themselves.
```
evgunter: claude was confused about the original phrasing

libs/concurrency_group/imbue/concurrency_group/concurrency_group.py
```diff
@@ -217,9 +241,22 @@ def _wait_for_all_strands_to_finish_with_timeout(self, timeout_seconds: float) -
...
+                    # confirmed -- report whether we actually managed to terminate it.
+                    is_killed = True
+                    try:
+                        process.terminate(force_kill_seconds=PROCESS_FORCE_KILL_GRACE_SECONDS)
+                    except TimeoutExpired:
+                        is_killed = False
+                    if is_killed:
+                        outcome = "did not terminate in time and was killed."
+                    else:
+                        outcome = "did not terminate in time and could not be killed (it may still be running)."
```
evgunter, line 257: i assume this is wrong?

libs/concurrency_group/imbue/concurrency_group/local_process.py
```diff
@@ -112,15 +113,31 @@ def poll(self) -> int | None:
...
             if thread.exception_raw is not None:
                 thread.join()
-            return 1007
+            # If we get here the thread finished without a result and without an exception,
+            # which should be impossible: `run()` always assigns `_completed_process` unless it
+            # raises. Surface the invariant violation loudly rather than fabricating an exit code.
+            raise ProcessInvariantError(
+                f"Run thread for command `{' '.join(self._command)}` finished without producing a "
+                "result or recording an exception."
+            )
```
evgunter, line 133: suspicious

libs/concurrency_group/imbue/concurrency_group/thread_utils.py
```diff
@@ -74,6 +74,14 @@ def join(self, timeout: float | None = None) -> None:
         super().join(timeout)
         self.maybe_raise()
 
+    def join_without_raising(self, timeout: float | None = None) -> None:
+        """Wait for thread completion without re-raising any captured exception.
+
+        Use this when the caller collects strand failures separately (e.g. via
+        `maybe_raise`/`exception_if_not_suppressed`) and only needs to wait for the thread.
+        """
+        super().join(timeout)
```
evgunter, line 83: seems tempting to misuse

libs/mngr/imbue/mngr/config/key_resolver.py
```diff
@@ -109,9 +109,21 @@ def _walk_to_field(base: Any, path: tuple[str, ...]) -> Any:
...
+    which ``_apply_extend`` correctly treats ``__extend`` as assign. Reaching a
+    non-model, non-Mapping intermediate (e.g. a scalar) while path segments
+    remain is instead a structurally malformed ``__extend`` target: the user
+    asked to descend into ``a.b`` where ``a`` is a scalar. Silently returning
+    ``None`` there would downgrade the extend to an assign and drop the base
+    entries the user meant to extend, so it raises ``ConfigParseError`` to
+    surface the malformed path.
+
+    Raises ``ConfigParseError`` if an intermediate path segment cannot be
+    traversed because it lands on a non-model, non-Mapping value.
```
evgunter, line 124: i think i'm missing context here

libs/mngr/imbue/mngr/hosts/common.py
```diff
@@ -45,8 +45,14 @@ def build_ssh_transport_command(
     present in the known_hosts file. When known_hosts_file is provided, that file is
     used via UserKnownHostsFile. When None, the system default (~/.ssh/known_hosts)
     is used without setting UserKnownHostsFile.
+
+    When key_path is None, the ``-i`` flag is omitted entirely so ssh falls back
+    to the user's ssh-agent / ~/.ssh/config, rather than passing an empty key path.
```
evgunter, line 69: do we want this? (it seems related to some changes we intended to make, but i'm not sure)

libs/concurrency_group/imbue/concurrency_group/thread_utils.py
joshalbrecht: yeah, sketch

PR #1854, opened 2026-06-02
make restarting work for antigravity agents

evgunter:
## Summary

Stopped `antigravity` agents now resume their prior agy conversation on `mngr start`, instead of starting a fresh one.

- A second `PreInvocation` hook handler runs `capture_conversation_id.sh`, which reads agy's hook payload (it carries `conversationId`) and records the active conversation ID to a per-agent file (`antigravity_conversation_ids`).
- `AntigravityAgent.assemble_command` appends a shell-evaluated `--conversation <id>` that reads the last recorded ID at launch (the stored command is replayed verbatim on each start, so the decision must be shell-side, not Python-side). It uses `set --` / `"$@"` so it behaves identically under bash and zsh, and is guarded on agy's incremental `conversations/<id>.db` store still existing.
- The transcript streamer now discovers conversation IDs from the same capture-hook file instead of grepping agy's `--log-file` — the hook is the single source of truth, which also removes a latent bug (resumed conversations log `Resuming conversation`, not the `Resumed conversation` the old grep matched).

## Why `.db`, not `.pb`

Verified live (agy 1.0.4): `mngr stop` hard-kills agy, after which only the incremental `conversations/<id>.db` exists (the `.pb` snapshot is written only on a clean in-TUI exit). agy resumes from the `.db`, so the guard checks the `.db`.

## Verification

- Live agy 1.0.4: `--conversation <id>` resumes context (rejects unknown IDs); hook payload carries `conversationId` and is delivered to each handler independently; resume survives SIGKILL via the `.db`.
- End-to-end in tmux: plant a fact → capture hook writes the IDs file → SIGKILL (hard stop) → fresh relaunch via the real `.db`-guarded prelude → agy logs `Resuming conversation <id>` and recalls the fact.
- Resume prelude validated in both bash and zsh.
- Unit tests: capture script extraction/dedup/no-clobber, `assemble_command` resume snippet, streamer reading the IDs file, hook config. Full `mngr_antigravity` suite (148) passes locally.

## Deferred

Clone-resume (a cloned agent continuing the source's conversation) is left for a follow-up — agy's conversation store is global rather than per-agent, so it needs separate handling.


libs/mngr_antigravity/imbue/mngr_antigravity/resources/stream_transcript.sh
```diff
@@ -6,12 +6,15 @@
...
-# discovery step re-runs every poll cycle.
+# To scope the watch to *this* agent, we read the per-agent conversation-ids
+# file that the PreInvocation capture hook maintains
+# (capture_conversation_id.sh; see CONVERSATION_IDS_FILENAME in
+# antigravity_config.py). Every uuid recorded there is owned by this agent.
+# New conversation IDs can appear at any time -- /fork, /new, /switch, resume
+# -- so the discovery step re-runs every poll cycle. (Earlier versions grepped
+# agy's --log-file for `Created conversation <uuid>` lines; the hook is the
+# single source of truth now, which also avoids depending on agy's log
+# wording.)
```
evgunter, line 17: this feels cleaner than parsing, but could be less reliable

joshalbrecht:
lgtm, thaks!

PR #1857, opened 2026-06-02
unified interface for offline hosts with volumes

evgunter:
## What & why

The Claude plugin had custom logic to preserve some of its files (session JSONLs, transcripts, session-id history) when an agent is destroyed — duplicated across an SSH/rsync path for online agents and a separate Volume-walk path for stopped hosts. This PR standardizes a single way for any plugin to declare "keep these files on destroy", and along the way fixes a deeper conflation: the ability to **read/write a host's files** was tied to the host being **online**. A stopped host with a persistent volume can still be read (and written) — it just can't execute commands.

## Core abstraction

- **`HostFileReadInterface`** / **`HostFileWriteInterface`** — the read/write file ops extracted out of `OuterHostInterface` (which now extends both, so every online host is unchanged). This separates "can I read/write this host's files?" from "is it online?".
- **`OfflineHostWithVolume`** — implements both interfaces over a stopped host's persisted volume, addressing files by absolute paths under `host_dir` exactly as an online host does. The volume is resolved **lazily on first access** (so host discovery stays probe-free, e.g. no Modal listdir per host). `make_readable_offline_host()` wraps offline hosts at each provider's **construction site**, so a stopped host is readable whether reached via `get_host` (the destroy/GC path) or `to_offline_host`.
- **`OuterHost.list_directory`** — local FS walk, or remote SFTP `listdir_attr` over the same paramiko channel as file reads (with the same connection-error/retry handling).
- **`api/preservation.py`** — `PreservedItem`, `preserve_agent_data()`, `get_preserved_agent_dir()`. A plugin declares the files to keep once; the same declaration runs against an online host (rsync for dirs) or a volume-backed stopped host (file-by-file walk). Preserved files mirror the agent-state-dir layout under `<local_host_dir>/preserved/<name>--<id>/`.

## Consumers migrated onto the unified interfaces

- **Claude preservation** — `on_destroy` + `on_before_host_destroy` build one `_claude_preserved_items()` list and call `preserve_agent_data`; the ~10 duplicated SSH/Volume helpers are gone. New mirrored preserved layout (switch-forward; old data left in place); `subagent_proxy` reader updated.
- **`mngr file` get/list/put** — resolve a single readable/writable host instead of an `(online_host, volume)` pair; deleted the duplicate cross-platform listing script. (Online `list` no longer fills the opt-in `permissions` field — always `-`, matching what offline already showed; default columns unchanged.)
- **`api/events.py`** — discovery + reads go through one readable host; dropped the `find`/`cat`-over-SSH + events-scoped-Volume dual paths and the trailing-newline "sentinel-cat" workaround (`read_file` is byte-exact). Local tail thread preserved.
- A shared `try_resolve_readable_host` helper backs the `mngr_file` and `events` resolution.

**Left as-is:** `mngr_mapreduce` (volume-only by design — no online/offline fallback to collapse).

## Tests / review

New `api/preservation_test.py`, `OuterHost.list_directory` tests, docker offline read+write tests, migrated `mngr_file`/`events` tests (incl. a trailing-newline fidelity test proving the sentinel removal is safe). Architecture-verified (two follow-ups fixed: SFTP-listing connection-error classification, a stale comment) and autofix-clean. `ty` clean across all touched packages; ratchets pass. Full suite runs in CI.


libs/mngr/imbue/mngr/hosts/offline_host.py
```diff
@@ -363,3 +371,207 @@ def rename_agent(
...
+
+    The volume is resolved **lazily** -- on first file access, not at
+    construction. This keeps host *discovery* cheap: ``to_offline_host`` is
+    called in discovery fallbacks, and for some providers (e.g. Modal) resolving
+    the volume requires a network probe. Deferring the probe to the first read
+    means listing/discovering stopped hosts never pays for it; only code that
+    actually reads files does. The resolved volume (or its absence) is cached.
+    If no volume is available, reads behave as "nothing there": ``path_exists``
+    is False, ``list_directory`` is empty, and ``read_file`` raises
+    ``FileNotFoundError``.
```
evgunter: might not be what we want

joshalbrecht:
this looks roughtly right to me, but moving discussion to slack bc it's a bit complex

PR #1858, opened 2026-06-02
retry modal "app is locked" errors

evgunter:
## Problem

Modal acceptance tests flake intermittently with:

```
The selected app is locked - probably due to a concurrent modification
```

(example: PR #1837, `test_exec_failure_propagates_exit_code_on_modal`)

## Root cause

The `.modal` provider is **persistent**, so every `create_host` redeploys the `snapshot_and_shutdown` function into `self.app_name` via `modal deploy` (`instance.py`, `direct.py`). The app name is `{prefix}{provider_instance_name}` = `{prefix}modal` — not keyed to the host or test.

In a shared-environment offload run, the subprocess prefix is the shared-env timestamp, identical across every sandbox and test. So **all** subprocess `mngr create @.modal` tests resolve to the same app `{shared-prefix}modal`. Offload runs many sandboxes concurrently (different tests, plus speculative duplicates of the same test), and they all `modal deploy` into that one app at once. Modal serializes app mutations, so the losers fail with the app-lock error.

This is also a latent production bug: two parallel `mngr create x@.modal & mngr create y@.modal` for the same user race the same way.

## Fix

The lock is transient — it clears once the conflicting operation finishes — so `DirectModalInterface.deploy` now classifies it as a new retryable `ModalProxyAppLockedError` and rides through it with exponential backoff (mirroring the existing volume/permission retry patterns). Non-lock deploy failures still raise immediately without retry.

## Tests

- `test_is_app_locked_error` — message classification (incl. the real Modal wording, case-insensitivity, negatives).
- `test_deploy_retries_on_locked_app_then_succeeds` — fake `modal` on PATH fails once with the lock message, deploy retries and succeeds.
- `test_deploy_does_not_retry_on_non_lock_error` — non-lock failure raises immediately, exactly one attempt.


joshalbrecht:
lgtm, thanks for the fix!

PR #1859, opened 2026-06-02
bump bash strict-mode ratchet; fix ratchet scans crashing on symlinks

evgunter:
i'm not sure if the comments explaining the lack of `-e` are fully accurate

---

## Summary

Fixes two latent meta-ratchet failures that reproduce on a full local checkout of `main` (they pass in offload only because `.dockerignore` omits the relevant tracked paths from the sandbox build context).

**1. Content ratchet scans crash on a symlink-to-directory.**
`_get_all_files_with_extension` filtered files with `exists()`, which follows a symlink-to-directory and returns `True`. `read_text()` then raised `IsADirectoryError` on the tracked `.claude/skills/blueprint -> ../../.agents/skills/blueprint` symlink (added 2026-05-31), crashing `test_prevent_old_mng_name_in_file_contents` and any other content scan. Now filters on `is_file()`, which follows symlinks but excludes those resolving to a directory -- the exact predicate the sibling `find_bash_scripts_without_strict_mode` already uses for the same problem class. Added a regression test in `core_test.py` (verified it fails without the fix).

**2. Bash strict-mode ratchet was at 12 > snapshot(10) locally.**
Two minds verify scripts added 2026-05-29 use `set -uo pipefail` (no `-e`). This is a deliberate, matched choice: both scripts handle errors explicitly (a `fail` helper, `PIPESTATUS`, retry loops that depend on commands exiting non-zero, diagnostic blocks on failure), and the other three scripts in `apps/minds/scripts/` do use `set -euo pipefail`. `-e` would abort their explicit error handling. So both are treated as documented exemptions (mirroring the existing `.minds/template/*` accommodation): added/kept an explanatory comment in each, documented them in the test docstring, and raised the snapshot 10 -> 12. **No script runtime behavior is changed.**

## Test plan
- `test_meta_ratchets.py`: 18 passed (both previously-failing tests now pass; bash count is exactly 12).
- `core_test.py`: 25 passed, including the new `test_get_all_files_excludes_symlink_to_directory`.
- Confirmed the new test fails when the fix is reverted (`is_file()` -> `exists()`).
- `/verify-architecture` and `/autofix` both clean.

## Notes
- Why CI was green while `main` was locally broken: the offload sandbox's `.dockerignore` (generated from `.gitignore` at offload time) omits the `.claude/skills` symlinks and some `.sh` paths from the build context, so the symlink dangles (skipped) and several scripts are absent on disk. This is a structural CI-vs-local file-set divergence; the `is_file()` fix closes the crash but the broader divergence is left as-is (aligning the sandbox file set would be a separate infra change).


joshalbrecht:
I think just add FIXME's to those two comments and then lgtm

(the FIXME is roughly: "convert the script so that -e can be used")

PR #1860, opened 2026-06-02
in the mngr usage plugin, preserve data from destroyed agents

evgunter:
Builds on `mngr/preserve-hook` (not yet merged to main; PR base is that branch).

## What

Makes `mngr usage` preserve its usage files on destroy the same way `mngr claude` preserves session files, and makes the reader fold that preserved data back in so destroyed agents' spend still counts.

All changes are contained in the `mngr_usage` plugin -- no changes to `mngr_claude` or `mngr_claude_usage`.

### Write side (preserve on destroy)
- New `preserve_on_destroy` option on `UsagePluginConfig` (default `true`; set `false` to discard usage on destroy).
- Two hookimpls in `mngr_usage` -- `on_before_agent_destroy` (online) and `on_before_host_destroy` (offline/volume) -- copy each agent's `events/<source>/usage` directories (plus its `data.json`, for filtering) to `<local_host_dir>/preserved/<agent-name>--<agent-id>/`, reusing the shared `imbue.mngr.api.preservation` core. Agent-agnostic: fires for every agent type but is a no-op for agents with no usage events.
- A small `mngr_usage_meta.json` sidecar captures the agent's provider/host at preserve time (`data.json` doesn't record provider), so host-scoped filters work post-destroy.

### Read side (read-back)
- `gather_usage_snapshots` now folds in preserved usage by default. Preserved agents are filtered with the **same** provider + CEL predicates as live agents, evaluated against a minimal `AgentDetails` reconstructed from the preserved `data.json` + captured host metadata. So `--provider` / `--project` / `--local` / label filters apply uniformly to live and destroyed agents.
- A still-live agent that also has a preserved copy is de-duplicated (never double-counted).
- New `--preserved` / `--no-preserved` flag on `mngr usage` and `mngr usage wait` (default on) to opt out.

## Tests
- `preservation_test.py`: write-side (volume host copy + no-op without usage), discovery + provider/CEL filtering, merge fold-in + dedup, and `gather_usage_snapshots` include/exclude.
- `cli_test.py`: end-to-end CLI test asserting preserved usage shows by default and `--no-preserved` hides it.


joshalbrecht:
lgtm!

(not merging because this has a nonstandard base)

PR #1861, opened 2026-06-02
Clean up dead "release" branch CI refs; move release tests to a dedicated workflow

weishi-imbue:
## Summary

There is no `release` branch in this repo. Releases are cut from `main` as `v*` tags (`scripts/release.py` + `.github/workflows/publish.yml`). But `ci.yml` still had a whole release-branch apparatus -- a `release` push trigger, four `github.ref != 'refs/heads/release'` job guards, and two jobs (`test-release`, `test-docker-release`) gated to `refs/heads/release` push that, with no such branch, **never ran**.

This PR:
1. Removes the dead release-branch references from `ci.yml` (back to plain branch/PR CI).
2. Gives the release tests a real home: a dedicated **`Release Tests`** workflow you run **before** cutting a release.

## Why a dedicated workflow (not a tag-push job in ci.yml)

Running the release suite *at* `v*` tag-push time is too late to be useful: `publish.yml` publishes to PyPI on the same tag, in parallel, so tag-time tests can't prevent a bad release. The valuable moment is **before** tagging.

## Changes

- **`.github/workflows/release-tests.yml`** (new): contains `test-release` + `test-docker-release`, triggered by:
  - `workflow_dispatch` -- run against `main` to validate a commit **before** you cut a release (the real purpose).
  - push of a `v*` tag -- an automatic backstop record.
- **`.github/workflows/ci.yml`**: drop the `release` push trigger and the four `!= refs/heads/release` guards; the two release jobs move out. Net: `ci.yml` is now plain branch/PR CI.
- **`scripts/release.py`**: before the release confirmation prompt, print an advisory warning if the `Release Tests` workflow has not passed on the exact commit being tagged (with the `gh workflow run release-tests.yml --ref main` hint). Advisory only -- not a hard block.
- Doc/comment refreshes: `style_guide.md`, `conftest_hooks.py`, `test_docker_state_transitions.py`; and `test_meta_ratchets.py` drops `release` from the PR-branch skip guard.
- Changelog entries for the three touched projects: `dev/`, `libs/imbue_common/`, `libs/mngr/`.

## How you use it / where results show up

A tag push or manual dispatch isn't attached to a PR, but it produces a normal run in the **Actions tab** under "Release Tests":
```
gh workflow run release-tests.yml --ref main   # before cutting a release
gh run list --workflow release-tests.yml
gh run watch <run-id>
```

## Caveats (please review)

- **Not a hard publish gate.** `publish.yml` still triggers on `v*` and publishes in parallel; the `pypi` GitHub environment has no protection rules (verified: `protection_rules: []`), so nothing blocks a publish. Making it a true gate (publish `needs:` the release tests, or pypi-environment required reviewers) is a separate, deliberate decision.
- **These jobs have never run** (the old gate was unreachable), so the first real dispatch / tag may surface latent failures (bit rot, missing `OVH_*`/`VULTR_*` provider secrets, etc.). Best to do one manual dispatch to shake them out.
- On Linux, `test-release` runs the *entire* suite (all markers, 12 splits) with `IS_RELEASE=1` -- preserved verbatim from the original job; only the trigger changed.

Tracking: #1866.


joshalbrecht:
lgtm! feel free to fix merge conflicts and merge

PR #1862, opened 2026-06-02
record stop/destroy/cleanup failures

evgunter:
this is quite involved and i don't have high confidence that it doesn't break things

also some of the locking/threading errors through is pretty gross

---

## What this does

`mngr stop`, `mngr destroy`, and `mngr cleanup` no longer **hang on**, or **silently swallow**, cleanup problems. Every cleanup step is now bounded and classified as either **benign** (the target was already gone -> no error, exit 0) or a **real failure** (a resource is actually left behind). Real failures are **aggregated** (cleanup continues rather than failing fast), each tagged with a cause category, and surfaced via a **cause-specific exit code** plus a structured `failures` list in `--format json`.

Design: `specs/cleanup-error-aggregation.md`.

## Why

The branch started as a fix for an offload hang (a wedged `tmux list-panes` blocked `Host.stop_agents` forever). Investigating it surfaced a broader inconsistency: the stop path swallowed every non-timeout failure, and the destroy path logged provider failures as warnings and always exited 0 — so a partially-failed cleanup was indistinguishable from a clean one.

## How the behavior changes

**Before:** stop/destroy could hang on a wedged command, and otherwise swallowed failures and always exited 0.

**After:**
- Every stop/destroy step is bounded and best-effort; failures aggregate instead of aborting.
- **Benign vs. real:** a command that "failed" only because the target was already gone is not an error (exit 0). A resource that exists but couldn't be cleaned up is recorded.
- **Cause-specific exit codes** (extending `cli/exit_codes.py`): `2` timeout, `3` processes remain, `4` local state remains, `5` host/infrastructure remains, `6` provider inaccessible, `1` other. When several causes occur, the process exits with the most severe; JSON lists them all.
- **JSON output** gains a `failures` list (`category`, `message`, `agent_name`, `host_id`) and an `exit_code` field, replacing the old `errors` string list (breaking change for JSON consumers).
- Stopping an agent on an **offline host** is now a real `PROVIDER_INACCESSIBLE` failure (the host is unreachable, so the agents can't be verified stopped), not a silent skip.

## How it's classified

- **Shell commands** (stop path, `rm`): by stderr message-matching — tmux "can't find session/window/pane" / "no server running" and kill "No such process" (ESRCH) are benign; anything else is real. The six stop commands drop their `2>/dev/null`/`|| true` so stderr is visible. A timeout is just one cause (`TIMEOUT`), detected via the opt-in `raise_on_timeout` on `execute_idempotent_command` (normalizes the local killed-process and remote `socket.timeout` into one signal).
- **Provider `destroy_host`** (docker, local, ssh, modal, vps_docker, lima, imbue_cloud): by exception type — typed "not found" = benign; real errors -> `HOST_RESOURCE_REMAINS` (or `OTHER` for a record-write failure). Raised "couldn't even attempt" errors are classified `PROVIDER_INACCESSIBLE` at the orchestration boundary.

## Interface changes

`Host.stop_agents`, `Host.destroy_agent`, and `ProviderInstance.destroy_host` now return `list[CleanupFailure]`. `CleanupResult.errors` (string list) becomes `failures: list[CleanupFailure]` (with a derived `errors` property for human output).

## Testing

`just test-quick` across the mngr package + the four external provider packages: type-error and ratchet tests (280) pass; affected unit/integration test files (cleanup, host, cli stop/destroy/cleanup, create, gc, providers) pass — 1192 in the broad non-infra run. New tests cover the stderr classification (benign vs real, timeout, ESRCH-vs-EPERM), the exit-code mapping, aggregation, and the JSON `failures`/`exit_code` payload. The full offload suite (incl. tmux/docker integration) runs in CI.


libs/mngr/imbue/mngr/hosts/host.py
```diff
@@ -274,6 +274,19 @@ def _is_same_machine(a: OnlineHostInterface, b: OnlineHostInterface) -> bool:
...
+# hang indefinitely -- tmux occasionally fails to return under CI load, which
+# is why the test-cleanup helpers in utils/testing.py already bound every tmux
+# subprocess. Without a bound here, a single stuck `tmux list-panes` blocks
+# stop_agents forever (observed hanging an entire offload batch). These steps
+# are idempotent and best-effort: a timeout simply yields no PIDs for that
+# step, and the remaining steps (kill-session, env-scan fallback) still run, so
+# cleanup keeps making forward progress instead of stalling. Matches the 5.0s
+# bound already used by get_lifecycle_state for its tmux/ps calls.
+_STOP_AGENT_COMMAND_TIMEOUT_SECONDS: Final[float] = 5.0
+
```
joshalbrecht: just make this a tiny bit higher--I could see something taking 5 seconds if you get unlucky

10 seconds is probably enough that I don't mind anymore :-P

libs/mngr/imbue/mngr/hosts/host.py
```diff
@@ -2499,6 +2512,31 @@ def start_agents(self, agent_ids: Sequence[AgentId]) -> None:
                     if not result.success:
                         raise AgentStartError(str(agent.name), result.stderr)
 
+    def _run_bounded_stop_command(
+        self,
+        command: str,
+        timeout_seconds: float = _STOP_AGENT_COMMAND_TIMEOUT_SECONDS,
+    ) -> CommandResult:
```
joshalbrecht: I'm confused how this is related to stopping?

joshalbrecht:
I'm a little confused how anything related to "stop" is best effort?

libs/mngr/imbue/mngr/hosts/host.py
evgunter: it's just a wrapper to make things continue on timeout

libs/mngr/imbue/mngr/hosts/host.py
evgunter: should i also adjust the timeout in `get_lifecycle_state`?

libs/mngr/imbue/mngr/hosts/host.py
```diff
@@ -277,6 +279,19 @@ def _is_same_machine(a: OnlineHostInterface, b: OnlineHostInterface) -> bool:
...
+# descendant walk, and the MNGR_AGENT_ID env scan). A wedged tmux client can
+# hang indefinitely -- tmux occasionally fails to return under CI load, which
+# is why the test-cleanup helpers in utils/testing.py already bound every tmux
+# subprocess. Without a bound here, a single stuck `tmux list-panes` blocks
+# stop_agents forever (observed hanging an entire offload batch). On timeout the
+# step raises CommandTimeoutError (see _run_bounded_stop_command) rather than
+# degrading: continuing past a wedged command has no real teardown value. These
+# commands normally return near-instantly; the bound is generous headroom
+# (including for slower remote hosts) before declaring a command wedged.
+_STOP_AGENT_COMMAND_TIMEOUT_SECONDS: Final[float] = 30.0
```
evgunter: bumped to 30 rather than 10 because with fail fast we have at most one step that hits its timeout

libs/mngr/imbue/mngr/hosts/host.py
```diff
@@ -2616,29 +2822,45 @@ def _collect_session_pids(self, session_name: str) -> list[str]:
...
 
-    def _collect_pids_by_agent_id_env(self, agent_id: AgentId) -> list[str]:
+    def _collect_pids_by_agent_id_env(
+        self, agent_id: AgentId, failures: list[CleanupFailure], agent_name: AgentName | None
+    ) -> list[str]:
         """Find all PIDs whose MNGR_AGENT_ID environment matches agent_id.
 
+        Appends a failure to ``failures`` only if the scan itself times out; the command
+        is constructed to exit 0 (see the ``; true`` below) and suppresses per-pid grep
+        noise, so it does not otherwise produce a real failure.
```
evgunter, line 2956: hm

libs/mngr/imbue/mngr/hosts/outer_host.py
```diff
@@ -463,8 +463,17 @@ def _run_shell_command_local(
...
-        """Run a shell command on the local machine without going through pyinfra."""
+        """Run a shell command on the local machine without going through pyinfra.
+
+        When ``_raise_on_timeout`` is set, a timeout raises ``ProcessTimeoutError``
+        instead of being reported as an ordinary failed result. This mirrors the
+        remote SSH layer, which always surfaces a timeout as ``socket.timeout``,
+        so opt-in callers can treat a timeout as a hard failure uniformly across
+        backends. Left off by default because most callers want a timeout to look
+        like any other failed command (``success=False``).
+        """
```
evgunter, line 541: strange shape

evgunter:
now explicitly collects and surfaces not just timeouts but also a bunch of other errors

libs/mngr/imbue/mngr/interfaces/host.py
```diff
@@ -618,8 +619,12 @@ def provision_agent(
         ...
 
     @abstractmethod
-    def destroy_agent(self, agent: AgentInterface) -> None:
-        """Remove an agent and all its associated state from this host."""
+    def destroy_agent(self, agent: AgentInterface) -> list[CleanupFailure]:
```
joshalbrecht: sorry to be annoying about this PR, but this interface doesn't really make a ton of sense -- errors should cause exceptions to be raised, otherwise it's too easy to forget to deal with them

a raised exception can contain multiple independent causes (ie, be something like an exception group or like what concurrency groups are forced to do), but ultimately, yeah, it needs to raise

exception group is honestly probably the most natural affordance here

libs/mngr/imbue/mngr/interfaces/host.py
```diff
@@ -628,8 +633,12 @@ def start_agents(self, agent_ids: Sequence[AgentId]) -> None:
         ...
 
     @abstractmethod
-    def stop_agents(self, agent_ids: Sequence[AgentId], timeout_seconds: float = 5.0) -> None:
-        """Stop the specified agents gracefully within the given timeout."""
+    def stop_agents(self, agent_ids: Sequence[AgentId], timeout_seconds: float = 5.0) -> list[CleanupFailure]:
```
joshalbrecht: same comment as above

libs/mngr/imbue/mngr/interfaces/provider_instance.py
```diff
@@ -391,12 +392,17 @@ def start_host(
         ...
 
     @abstractmethod
-    def destroy_host(self, host: HostInterface | HostId) -> None:
+    def destroy_host(self, host: HostInterface | HostId) -> list[CleanupFailure]:
```
joshalbrecht: same comment as above

joshalbrecht:
this is moving in a better direction, but I do think it'd be worth addressing my one comment (then I'll re-review)

libs/mngr/imbue/mngr/interfaces/host.py
evgunter: an attempt has been made

evgunter:
addressed, although the state is still not great

joshalbrecht:
lgtm after we all looked at it

our conclusion is "slightly cursed but still an improvement"

PR #1865, opened 2026-06-02
Pin SSH auth to explicit key in build_ssh_transport_command

weishi-imbue:
## Summary
Currently `ssh -i /path/to/key` only adds the key to the candidate list. 
We want to force use the key. 
`-o IdentitiesOnly=yes -o IdentityAgent=none` pins authentication to the explicit `-i` path.

Without these flags, OpenSSH consults `SSH_AUTH_SOCK` first. On a macOS user session that routes to the 1Password biometric SSH agent, whose prompt cannot fire in a BatchMode child like `git push` or `rsync` — so ssh blocks indefinitely on the agent reply with no error surfaced upstream. This manifested as `mngr create` hanging silently after `Transferring git repository...`.

## Docs
macos: `man ssh_config`
image

linux:
https://manpages.debian.org/trixie/openssh-client/ssh_config.5.en.html

## Test plan
- [x] `libs/mngr/imbue/mngr/hosts/common_test.py -k build_ssh_transport_command`: 4 passed


joshalbrecht:
lgtm

this does cause some problems for people who want to use mngr more generally (with different SSH configurations), but I agree that it's closer to the behavior that we intend right now, so we can come back to that later if it matters

PR #1867, opened 2026-06-02
Fix minds deploy/recover/create bugs + runtime/secrets collision

joshalbrecht:
Bug fixes surfaced while standing up a fresh dev minds environment and exercising the imbue_cloud bake -> create -> release cycle, plus a `runtime/secrets` path-collision fix.

## minds env deploy / recover / create
- Deploy pushes the `ovh` per-env Modal Secret (was missing from every tier's `deploy.toml`), so `modal deploy rsc-<tier>` no longer fails with "Secret not found". Added a regression test.
- `minds env recover` runs non-interactively (`modal app stop -y`).
- `minds env recover` is re-runnable (Neon instant-restore treats a 409 "branch already exists" as already-restored).
- `minds env deploy` exits non-zero when a failed deploy rolls back (hidden `--from-failed-deploy` flag).
- `minds env activate` no longer dead-locks the recover flow (allows activating an env with its own pending recover-target).
- Fixed a workspace-create `ty` error / breakage from a leftover `gh_token` field.
- imbue_cloud fast->slow fallback now keys on a structured `{"event":"error","error_class":...}` JSONL record instead of substring-matching a class name that mngr never emits.
- imbue_cloud slow path rebuilds the FCT image and does not adopt the pre-baked agent; a failed `mngr create` now destroys the leaked host (gated by the existing debug retain flag).

## runtime/secrets collision
Host backups write `runtime/secrets/restic.env`, forcing `runtime/secrets` to be a directory. That broke the Cloudflare tunnel runner (read it as a file -> `IsADirectoryError`) and the Telegram injector (appended to it). `runtime/secrets` is now canonically a directory of per-secret `*.env` files:
- Cloudflare token -> `runtime/secrets/cloudflare_tunnel.env` (inject overwrites; `just forward-system-interface` writes it; FCT runner watches it).
- New `clear_tunnel_token_from_agent`, called on workspace disassociation after the tunnel is deleted, so the agent stops `cloudflared` instead of spinning against a deleted tunnel.
- Telegram token -> `runtime/secrets/telegram.env`.

The FCT-side runner change lives in the forever-claude-template repo (`libs/cloudflare_tunnel`) on the matching branch.


apps/minds/imbue/minds/cli/env.py
```diff
@@ -591,6 +602,49 @@ def _refuse_if_this_env_recover_target_exists(env_name: str) -> None:
...
+    recover-target (the activate-then-recover path), and only hard-refuse
+    when the pending recover-target(s) belong solely to *other* envs -- a
+    forgotten failed deploy the operator should clear first. When
+    activating an affected env, any other envs' targets are surfaced as a
+    warning rather than blocking.
+    """
+    try:
+        repo_root = find_monorepo_root()
+    except MindError:
+        return
```
joshalbrecht, line 633: I struggle to imagine this being a reasonable default vs just raising the error?

joshalbrecht:
merged in https://github.com/imbue-ai/mngr/pull/1879

PR #1868, opened 2026-06-03
minds: surface workspace creation failures on the creating page

gnguralnick:
## Problem

When workspace creation fails, the creating/onboarding page surfaced it almost invisibly: only a small faint-gray caption flipped to "Failed: …", while the heading still read "Setting up your workspace", the progress bar froze partway, and the rotating tips kept cycling — making it easy to miss that creation had failed (see the reported screenshot: a `git checkout failed for branch 'testing'` error shown in faint text under a "Setting up your workspace" heading).

## Change

Frontend-only (`apps/minds/.../desktop_client/`); the backend already reports `status: FAILED` + the error string via both the status poll and SSE.

- `templates/creating.html`: the loading section now holds a `progress-view` and a `failure-view` sub-block sharing one logs/details footer. The failure view shows a red **"We couldn't set up your workspace"** heading, the error message in a red box, and **Back to setup** (→ `/create`) / **Home** (→ `/`) buttons.
- `static/creating.js`: a `showFailure()` fires the instant failure is detected (poll or SSE), swaps to the failure view **from whatever onboarding screen the user is on**, stops the rotating tips, and stops the progress bar.

## Verification

- Drove the real template + JS through Playwright with a mocked `FAILED` status: failure state renders prominently and interrupts the onboarding question flow immediately (does not wait for the user to finish the questions). Happy path (progress bar + tips) confirmed unchanged.
- Existing creating-page pytest tests pass (5 passed).

## Notes

- No JS test runner exists in this repo, so the client-side behavior was verified manually with Playwright per the repo's interactive-component guidance; the Python render path is exercised by existing tests.
- Possible follow-up: "Back to setup" goes to a fresh `/create`, so the user re-enters their git URL / branch. `/create` supports `?git_url=` prefill, but the creating page's client doesn't currently carry the original form values.

Still a draft pending review of wording/behavior.


joshalbrecht:
lgtm, thanks for the fix!

PR #1869, opened 2026-06-03
mngr_claude: symlink ~/.claude/.credentials.json in the guest so a nested mngr can authenticate the agents it spawns

weishi-imbue:
## Symptom (Minds, Lima agent)

User creates a Lima-mode agent through Minds with Claude.ai already logged in on the host Mac. Agent provisions successfully. But the workspace chat panel's Claude session prompts `/login` and rejects every request with 401, as if no credentials existed. Re-running `claude --version` inside the workspace shows the CLI is installed, just unauthenticated. The agent's *primary* `claude` process (the one mngr_claude spawned at create time, with `--config-dir` pointing at the per-agent dir) works fine — only the chat panel's session, and anything else launched later by the in-VM mngr, is broken.

## Root cause

There are two mngrs in the picture:

1. **Outer mngr** in the Minds backend on the Mac. When you create an agent, it reads credentials from the macOS Keychain (`Claude Code-credentials`, the Claude CLI 2.x storage location) and stages them onto the Lima guest at the per-agent config dir: `/mngr/agents/<agent_id>/plugin/claude/anthropic/.credentials.json`. The primary `claude` it then spawns is told to look there, so it authenticates.

2. **In-VM mngr** running inside the workspace, used by `system_interface` (the chat panel, terminal tabs, anything that wants to spawn new claudes inside the workspace). This nested mngr uses the same `mngr_claude` plugin code, but it runs on Linux. It looks for credentials at the standard path: `~/.claude/.credentials.json`. On the guest that file is empty — Linux has no Keychain, and Claude CLI 2.x on the host no longer writes the file there either, so there's nothing for the outer mngr's rsync to have picked up. Empty file → 401 → `/login` prompt.

The primary claude is fine because it's pointed at the per-agent dir explicitly. Everything spawned later by the in-VM mngr inherits the broken default path.

## Fix

After staging credentials into the per-agent dir, point the guest's standard `~/.claude/.credentials.json` at the staged copy via symlink, in `_bridge_credentials_to_default_claude_home`:

```
mkdir -p -m 0700 ~/.claude && ln -sfn <per-agent-dir>/.credentials.json ~/.claude/.credentials.json
```

- Symlink (not copy) so host-side credential refreshes propagate without re-staging.
- `ln -sfn` is idempotent and replaces a stale link in place without dereferencing it.
- `mkdir -p -m 0700` matches the per-agent dir's permissions.
- Portable across macOS and Linux guests.

## Scope

- Fires on non-local hosts (`not host.is_local`) when a `.credentials.json` was staged for this agent. A Lima VM counts as non-local: mngr reaches it over SSH even though it runs on the same Mac, so the bridge applies there.
- No change to how the in-VM mngr looks up credentials — the symlink just presents the file at the path it already reads.
- Bumps `libs/mngr_claude` 0.2.10 → 0.2.11.

## Tests

Unit tests for the helper: non-local host with staged credentials emits the exact `mkdir … && ln -sfn …` command (target shlex-quoted); parametrized no-op covers both `is_local=True` and "no `.credentials.json` staged".


joshalbrecht:
I'm pretty confused by this, moving to slack

PR #1870, opened 2026-06-03
minds desktop: surface auto-update prompt and add dev-tools menu

weishi-imbue:
## Summary

Standalone extraction from #1317 so the desktop app's auto-update and developer-tooling fixes can be reviewed and merged on their own, ahead of the larger Lima / packaging work. **Touches only `apps/minds/electron/main.js`** plus a changelog entry; every change is functional on its own and the branch merges cleanly into current `main`.

### Auto-update
- Set ToDesktop's `updateReadyAction.showInstallAndRestartPrompt` to `"always"`. The runtime default is `"never"`, so users saw the initial "downloading in the background..." dialog and were **never prompted to install** the staged update.
- Only call `todesktop.init()` in packaged builds. In dev on macOS the constructor throws because `electron.autoUpdater` is undefined (Squirrel is not linked in the unsigned dev binary); dev launches now log and skip it.
- Add a `Check for Updates...` item to the macOS **application menu** (alongside About / Services / Quit) backed by `triggerUpdateCheck()`, which reports the outcome via dialog: update found / up to date / unavailable / error. The "unavailable" message is worded per build type -- dev builds say updates are only available in installed builds; packaged-but-unreleased (draft) builds explain the auto-updater is disabled until released to the latest channel.

### Developer tooling
- Add a `View` menu with `Toggle Developer Tools` (Alt+Cmd+I), zoom controls, and fullscreen. The default Electron DevTools shortcut crashes because the app uses `BaseWindow` + `WebContentsView` rather than a `BrowserWindow`, so the handler calls `toggleDevTools()` on the focused bundle's content view explicitly.
- Honor `MINDS_OPEN_DEVTOOLS=1` to auto-open detached DevTools on the content view at launch.
- Log startup env-setup failures to the console in addition to the error window.

### Deliberately excluded from this extraction
- **Content-view preload.** #1317 also added `preload.js` (the privileged `window.minds` IPC bridge) to the workspace content view. That is intentionally **not** in this PR: #1875 has since landed on `main`, giving the content view a deliberately locked-down `content-relay-preload.js` because it hosts foreign, untrusted workspace content and must not receive the privileged bridge. `main` now owns the content-view preload; this PR leaves it alone (the trusted chrome/sidebar/requests/modal views keep `preload.js`). Note for #1317: its `preload.js`-on-content-view change now conflicts with #1875 and should be reconciled there.
- **About-panel git-SHA block.** Excluded because it reads `electron/build-info.json`, generated only by the `build.js` packaging changes that are not part of this PR. It would be inert (and log a warning in packaged builds) without that work, so it stays with the build/packaging extraction.

## Test plan
- `node --check apps/minds/electron/main.js` passes.
- Merges cleanly into current `main` (`git merge-tree` reports no conflicts).
- Symbols used (`getMostRecentWindow`, `dialog`, `app.isPackaged`, `todesktop.autoUpdater`) already exist / are valid on `main`; the bundle from `getMostRecentWindow()` has the `.window` / `.contentView` shape the View-menu handler relies on (same guard pattern as existing callers).
- This is an Electron-main-process change; the repo has no JS test runner and the Python suite does not exercise `main.js`. Relying on CI for the full run.

### DIFF
--- a/apps/minds/changelog/mngr-extract-1.md
+++ b/apps/minds/changelog/mngr-extract-1.md
@@ -4,5 +4,4 @@ Desktop app auto-update and developer-tooling fixes (extracted from the larger m
 - Added a `Check for Updates...` item to the application menu that triggers a check and reports the result (update found / up to date / unavailable / error), with the unavailable message worded for the build type (dev vs unreleased draft).
 - Added a `View` menu with `Toggle Developer Tools` (Alt+Cmd+I), zoom controls, and fullscreen. The default Electron DevTools shortcut crashed because the app uses `BaseWindow` + `WebContentsView` rather than a `BrowserWindow`.
 - `MINDS_OPEN_DEVTOOLS=1` auto-opens detached DevTools on the content view at launch.
-- The workspace content `WebContentsView` now loads `preload.js`, on both its initial and its error-retry rebuild paths.
 - Startup env-setup failures are now logged to the console in addition to being shown in the error window.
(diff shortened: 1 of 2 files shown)


joshalbrecht:
lgtm, thanks!

PR #1872, opened 2026-06-03
Fix non-default branch checkout when creating a workspace from a remote repo

gnguralnick:
Creating a workspace from a remote repository failed whenever the requested branch was not the repository's default branch. Cloning forever-claude-template at branch `testing`, for example, died with:

```
error: pathspec 'testing' did not match any file(s) known to git
```

The remote clone path used `git clone --depth 1`, which implies `--single-branch` and so only fetches the default branch. The requested branch's ref was never downloaded, so the subsequent `git checkout <branch>` could not find it.

`clone_git_repo` now takes an optional `branch` and, when given, clones with `--single-branch --branch <branch>`: only that branch is fetched (still far cheaper than a full clone) but its complete, non-shallow history is present. The shallow (`--depth 1`) clone is removed entirely. Non-shallow matters beyond the checkout: `mngr create` mirror-pushes this clone into the agent container's bare repo, and git rejects a shallow source with "shallow update not allowed".

Requesting a branch that does not exist on the remote now fails cleanly at clone time rather than later at checkout.

Verified end-to-end against the real forever-claude-template `testing` branch, and added regression tests (single-branch non-shallow clone that is mirror-pushable; missing-branch raises `GitCloneError`).


joshalbrecht:
lgtm

PR #1873, opened 2026-06-03
Unify snapshot CLI targets; narrow event host discovery

qi-imbue:
## Summary

**`mngr snapshot create` / `list` / `destroy`** — Drop the `--agent` and `--host` option flags. All three subcommands now take a single positional `IDENTIFIERS` list parsed as agent-or-host targets, using the existing `parse_agent_or_host_address` disambiguation (bare names are agents, `@HOST[.PROVIDER]` and bare `host-...` IDs are hosts). Multiple identifiers that resolve to the same host are deduplicated; an ambiguous host name fans out to every match (preserving existing semantics).

  - `destroy` previously only accepted agents — it now also accepts host targets directly, which matches what the underlying operation does (snapshots are host-level).
  - The stdin `-` placeholder continues to work for all three.

**Snapshot target resolution is now one discovery pass.** Provider narrowing kicks in when every address pins a provider; the discovery-event-stream identifier optimization kicks in when no host-only addresses are mixed in. Previously the host-address path always did a full all-providers scan.

**`mngr event @host[.provider]`** — Discovery is now narrowed to the pinned provider, matching the agent path's behavior (which already went through `discover_by_address`). One-line change; user-visible only as a speed-up.

## Things to manually test

- [ ] `mngr snapshot create my-agent` and `mngr snapshot create @my-host` work; help renders cleanly.
- [ ] Mixed positionals: `mngr snapshot create agent1 @host2`.
- [ ] Stdin: `mngr list --ids | mngr snapshot create -`.
- [ ] `mngr snapshot destroy @my-host --all-snapshots --force` — the new host-positional path on destroy.
- [ ] `mngr snapshot destroy my-agent --snapshot snap-abc123 --force` — still works.
- [ ] Ambiguous-host error message: try `mngr snapshot create @SOMENAME` for a name that exists on multiple providers — should snapshot every match.
- [ ] Ambiguous-agent error: `mngr snapshot create AGENT_NAME` where the name exists on multiple providers — should error with the multi-line "Disambiguate using NAME@HOST.PROVIDER" hint.
- [ ] `mngr event @host.modal` against a real multi-provider setup — should noticeably skip providers other than `modal`.
- [ ] `--help` on `mngr snapshot`, `snapshot create`, `snapshot list`, `snapshot destroy` looks right and doesn't mention the removed flags.


joshalbrecht:
lgtm

PR #1874, opened 2026-06-03
Consolidate HostLocationAddress resolution and narrow provider discovery

qi-imbue:
## Summary
- `mngr create --from`'s cached host/agent loader now narrows discovery based on the source, the target host, and `--reuse` (if used): when all three pin a provider, only those providers are queried; if any consumer needs to search across providers (e.g. `--reuse` with a target that has no provider pin, or a `--from` that names an agent without a host), the loader falls back to a full scan exactly as today. Bare local sources and git URLs continue to skip provider discovery entirely.
- Internal: extracted the local-bare-path shortcut + discover-then-resolve flow into a new `imbue.mngr.api.find.resolve_host_location()` helper, used by `mngr rsync` and `mngr git push`/`mngr git pull`. (The same narrowing for those commands was independently added on main during a parallel polish PR; this branch just routes them through the shared helper instead of inlining the logic twice.) `resolve_host_location_address` (the lower-level entry point used by `mngr create`) is unchanged and remains the right call for sites that drive discovery themselves to share one result across multiple resolutions.

## Test plan
- [x] Unit tests for `_compute_loader_provider_filter` covering local source, provider-pinned source, agent-only source, provider-pinned target, new-host target, `--new-host` flag, `--reuse` with/without provider, source+target union, and git URL source.
- [x] Integration tests for `resolve_host_location` (local-bare-path shortcut, empty-address rejection, pinned-local-host round-trip).
- [x] `just test-quick libs/mngr` passes locally.
- [ ] Manually run `mngr create --from /local/path` and confirm no provider scan happens.
- [ ] Manually run `mngr create --from agent-name@host.modal` (target also pinned to Modal) and confirm only the Modal provider is queried.
- [ ] Manually run `mngr create --from agent-name` (no provider pin anywhere) and confirm the full provider scan still happens.
- [ ] Manually run `mngr create --reuse --name old-agent.modal` against an existing modal agent and confirm only Modal is queried.
- [ ] Manually run `mngr rsync ./src foo@bar.modal:/work` and `mngr git push @host.modal:/work` to confirm the routing through `resolve_host_location` preserves the existing single-provider behaviour.
- [ ] CI to run `just test-offload` and acceptance tests.


joshalbrecht:
lgtm

(not merging given the pending discussion on the below branch)

PR #1875, opened 2026-06-03
Make it possible for nested apps to open the permission request dialog window.

hynek-urban:
- Add a mechanism that allows workspace apps to open the permission dialog modal. (I think this should be relatively harmless in terms of security.)
- When the request was already approved or denied, do not render it again.

This PR accompanies a change in forever-claude-template that actually makes use of this: https://github.com/imbue-ai/forever-claude-template/pull/125/


joshalbrecht:
lgtm!

PR #1883, opened 2026-06-03
Migrate apps/minds desktop_client templates from Jinja macros to JinjaX

gdenisov:
## Summary

- Replace `_macros.html` + `{% extends 'base.html' %}` + `{% include %}` with JinjaX components. UI primitives (Button family, Card/CardRow, Notice, Spinner, TextInput, PageContainer, Opt, Associate) and layout (Base, AuthBase) are now `.jinja` components composed via `<Component>` tags.
- Each page becomes a PascalCase `.jinja` component (`templates/pages/Landing.jinja`, etc.). Auth pages and components are co-located under `templates/auth/`. The public `render_*` functions in `templates.py` keep their signatures and now delegate to `CATALOG.render("pages.Name", ...)`.
- The permissions dialog decomposes into 5 components (`PermissionsDialog`, `PermissionsHeader`, `PermissionsForm`, `PermissionsManualCredentials`, `PermissionsError`). The two latchkey subpages compose them directly.
- Extends the `/_dev/styleguide` page with examples for `CardRow`, `PageContainer`, `Opt`, and `auth.OauthIcon` on top of the existing primitives.
- Adds `jinjax>=0.45` to `apps/minds/pyproject.toml`. `.jinja` files ship in the wheel via hatchling's default packages-inclusion (verified with `uv build`).

## Before / After
- This should be a visual no-op as we're just changing how html gets injected into the components not which html.
- The PR includes a visual diff test script (to run locally) that renders 44 scenarios. You can see the full results here: 
[visual-diff.zip](https://github.com/user-attachments/files/28575619/visual-diff.zip)
- The test brought out these 4 differences all of which are legitimate:

|Before|After|
| - | - |
|image|image|
|This is a false positive due to differing spinner animation status||
|image|image|
|This is a false positive due to differing spinner animation status||
|image|image|
|Swapped handwritten button for a component one which results in minor style difference (good!)||
|image|image|
|Style Guide page legitimately changed content ||

## Test plan

- [x] `templates_test.py` -- 47 tests (38 existing + 9 new component-level tests for Button family, Notice variants, Card slot, Spinner sizes, `auth.OauthIcon` Google/GitHub paths).
- [x] `test_desktop_client.py` -- 120 integration tests through the FastAPI TestClient.
- [x] Full `apps/minds` quick run (`just test-quick "apps/minds -m 'not tmux and not modal ...'"`) -- 1079 pass.
- [x] Wheel smoke check -- `uv build apps/minds && unzip -l dist/minds-*.whl | grep '\.jinja'` lists 41 components.
- [x] Ratchet `prevent_silent_decode_error_catches` trimmed from `12 → 11` (snapshot was loose; actual count is 11).
- [ ] Manual UI smoke: launch the desktop client and exercise landing / create / welcome / sharing / workspace settings / `/_dev/styleguide` / a permission dialog. Confirm visible parity with main.

## Notes

- 5 pre-existing `webdav_test` failures are unrelated to this PR (confirmed by running `git stash` on the baseline -- they fail on `main` too). They look environment-specific.
- The recovery page (`render_recovery_page`) still builds HTML via Python string concatenation; out of scope for this PR.


joshalbrecht:
lgtm, thanks gleb!

PR #1884, opened 2026-06-03
tabcomplete for plugin add/remove

evgunter:
https://github.com/user-attachments/assets/2f0b20f8-383b-4e5f-8f98-d2d0e16d8874

https://github.com/user-attachments/assets/a05f533b-446d-445e-97cb-ee68a63a8748

i noticed that `mngr plugin remove concurrency-group` will run. should i make it validate that the thing being removed actually is a mngr plugin?

---

> [!NOTE]
> **Follow-up worth considering (not addressed here):** this PR only filters the tab-*completion* source for `mngr plugin remove` to real plugins. The `remove` **command itself** still validates against the raw uv-tool receipt extras (`_plugin_remove_impl`), so manually typing `mngr plugin remove imbue-common` would still "succeed" and could break the install. Tightening `remove`'s own validation to reject non-plugin packages (reusing `has_mngr_entry_points`) would close that footgun, but it's a behavior change to the command, so it's left out of this PR.

## Summary

Adds shell tab-completion for the positional arguments of `mngr plugin add` and `mngr plugin remove`.

- **`mngr plugin add <TAB>`** suggests installable plugin **package names** (e.g. `imbue-mngr-claude`, `imbue-mngr-modal`), drawn from the plugin catalog -- the same store the `mngr extras` install wizard uses.
- **`mngr plugin remove <TAB>`** suggests the plugin packages **currently installed** (from the uv-tool receipt), filtered to packages that actually register `mngr` entry points -- so non-plugin dependencies installed alongside plugins (e.g. workspace libraries like `imbue-common`, `concurrency-group`, `modal-proxy`) are not offered.

Both support prefix filtering and repeat the completion for each package when operating on several at once.

### Why these sources?

`add` and `remove` operate on PyPI **package names** (`imbue-mngr-modal`), whereas `enable`/`disable` operate on registered **plugin names** (`modal`). The existing `plugin_names` source is therefore wrong for `add`/`remove`:

- `add` takes installable packages you don't have yet -> draws from `plugin_catalog.get_installable_packages()`.
- `remove` takes packages you *do* have, and only ones that are actually plugins -> draws from the uv-tool receipt extras, filtered by `has_mngr_entry_points`.

## Implementation

Fits the existing cache-based completer (no per-keypress Click invocation):
- New `catalog_package_names` and `installed_plugin_package_names` fields on `CompletionCacheData`, with `catalog_packages` / `installed_packages` completion sources resolved in the lightweight `complete.py`.
- Positional spec entries `"plugin.add"` and `"plugin.remove"` in the writer.
- The installed-package list is computed by the cli-layer caller (`list.py`) and passed into the config-layer writer, mirroring how `topic_names` is threaded in, so the writer does not import the cli layer (which would break the import-linter layers contract).
- Promoted the mngr-entry-point check from a private helper in `cli/plugin.py` to a shared `uv_tool.has_mngr_entry_points` (single source of truth; the `plugin add` warning path now uses it too).

## Testing

- Unit tests in `completion_writer_test.py`, `complete_test.py`, and `uv_tool_test.py` (full list, prefix filter, variadic repeat, non-plugin filtering, best-effort receipt handling).
- Manually verified the real `complete.py` entrypoint with `COMP_WORDS`/`COMP_CWORD` set as the shell would, against the actual uv-tool receipt -- confirmed `remove` offers only real plugins and excludes `imbue-common` / `concurrency-group` / `modal-proxy`.

### DIFF
--- a/libs/mngr/changelog/mngr-plugin-autocomplete.md
+++ b/libs/mngr/changelog/mngr-plugin-autocomplete.md
@@ -1,0 +1,1 @@
+Added shell tab-completion for the positional argument of `mngr plugin add`. Pressing TAB now suggests installable plugin package names (e.g. `imbue-mngr-claude`, `imbue-mngr-modal`) drawn from the plugin catalog -- the same set the `mngr extras` install wizard offers. Completion repeats for each package when adding several at once.
--- a/libs/mngr/changelog/mngr-plugin-autocomplete.md
+++ b/libs/mngr/changelog/mngr-plugin-autocomplete.md
@@ -1,1 +1,6 @@
-Added shell tab-completion for the positional argument of `mngr plugin add`. Pressing TAB now suggests installable plugin package names (e.g. `imbue-mngr-claude`, `imbue-mngr-modal`) drawn from the plugin catalog -- the same set the `mngr extras` install wizard offers. Completion repeats for each package when adding several at once.
+Added shell tab-completion for the positional arguments of `mngr plugin add` and `mngr plugin remove`.
+
+- `mngr plugin add <TAB>` suggests installable plugin package names (e.g. `imbue-mngr-claude`, `imbue-mngr-modal`) drawn from the plugin catalog -- the same set the `mngr extras` install wizard offers.
+- `mngr plugin remove <TAB>` suggests the plugin packages currently installed (from the uv-tool receipt), which is exactly what `remove` accepts.
+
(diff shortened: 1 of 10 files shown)


joshalbrecht:
lgtm

mngr's autocomplete feels almost needlessly polished :)

evgunter:
it's how i remember what anything is called lol

PR #1886, opened 2026-06-03
identify-bad-tests skill

evgunter:
another example run: https://github.com/imbue-ai/mngr/pull/1895

---

## Summary

Adds a new `identify-*` skill, `identify-bad-tests`, modeled on the family added/edited on the base branch (`mngr/examine-errors-skill`), and then **applies it** to three targets so the reviewer can see what it produces.

### The skill

Scans a target path (a whole library or any subdirectory within one), resolves the scan scope and containing library the same deterministic way as the other `identify-*` skills, and reports ranked findings into the containing library's `_tasks/bad-tests/<date>.md` (the format `create-fixmes` consumes).

Its standard is the style guide's `# Testing` section plus the CLAUDE.md testing conventions; it does not restate them, it applies them. Each test is judged on two axes: **does it actually verify behavior?** (would it fail on a concrete bug; behavior vs. implementation; could it fail for the wrong reasons) and **is it well-formed?** (type/location/marker, test-grouping classes, naming, `parametrize` use, missing edge/branch cases, snapshot misuse, shared-fixture reuse and fixtures-in-conftest). The scan stage reads the `conftest.py` chain and its autouse fixtures so it understands the isolation baseline every test inherits. Raw patterns already counted by `test_ratchets.py` (`unittest.mock`, `monkeypatch.setattr`, `time.sleep`, inline imports) are deferred to the ratchets.

It deliberately reads the `_test.py` / `test_*.py` files (which repo conventions normally skip on the first pass).

### Demonstration: skill output applied to three libraries

Ran the skill on `libs/modal_proxy`, `libs/mngr_wait`, and `libs/mngr/imbue/mngr/utils` (a chunk of the large `mngr` lib). It produced 6 / 7 / 12 findings respectively, almost all genuine. The follow-up commits fix them:

- **modal_proxy**: the central `_translate_modal_error` was tested on only 2 of 8 branches; converter `case _` arms and two retry cases were untested; `_unwrap_*` used `model_construct` + a bare sentinel. Now parametrized exact-type branch coverage, real `modal.*` objects, and a new `errors_test.py`.
- **mngr_wait**: tautological by-construction state-set checks (true no matter which states were dropped) -> exact inline snapshots; no end-to-end `wait` CLI test -> `test_cli.py` exercising the exit-code contract via the real plugin manager; a test that *claimed* to simulate connection-error fallback but faked the path -> direct `poll_target_state` fallback tests.
- **mngr/utils**: assertions that checked key presence / substrings / return types instead of the actual conversion, branch name, or wired element id; non-unique sleep durations; a probabilistic uniqueness test -> seeded-RNG deterministic; relocated mis-placed unit tests.

Verification (local, external-resource markers excluded): **692 passed, 0 failed** (1 pre-existing skip) across the three scopes; all 165 ratchets pass; ruff clean.

The `_tasks/bad-tests/*.md` findings files are gitignored, so they are not in the diff (this is by design, matching the other `identify-*` skills).

## Full list of identified findings

The complete findings the skill produced for each scope (all marked `Decision: Accept`; the fix commits address them). The source `_tasks/bad-tests/*.md` files are gitignored, so they are reproduced here.

<details>
<summary><strong>libs/modal_proxy</strong> — 6 findings</summary>

1. **`_translate_modal_error` tested on only 2 of its 8 branches** — a swapped/dropped exception mapping would not be caught. Fixed by parametrizing all 8 branches with exact-type (`type(result) is expected`) assertions plus the fallback case.
2. **Both `_to_modal_stream_type` and `_to_file_entry_type` leave their `case _` default arm untested** — deleting/changing the "unsupported value" raise would go unnoticed. Added tests asserting `pytest.raises(ModalProxyError, match="Unsupported")`.
3. **`test_should_retry_volume_op` omits the `StreamTerminatedError` and `ProtocolError` retry branches** — dropping these transient-network retries would pass. Added rows asserting `True` for both.
4. **`test_unwrap_accepts_direct` uses `model_construct` + a bare `object()` sentinel** — never exercises real wrapping/validation. Switched to real `modal.*` objects where constructible offline.
5. **Inline `_FakeApp`/`_FakeImage`/`_FakeVolume`/`_FakeSecret` stubs are reusable interface stand-ins** that arguably belong in a shared `mock_*_test.py` (structural; documented for when a second consumer appears).
6. **`is_environment_not_found_error` has no direct test** — only transitively covered; empty-string / no-quotes / empty-name edges untested. Added `errors_test.py` parametrizing the regex contract.

</details>

<details>
<summary><strong>libs/mngr_wait</strong> — 7 findings</summary>

1. **Tautological enum-membership tests in `primitives_test.py`** — three tests assert values constructed *from* an enum are members of it (true by construction). Replaced the "subset" checks with exact-set inline snapshots (kept the meaningful `does_not_include_running` tests).
2. **No integration test exercises the `wait` CLI end-to-end** — exit-code mapping, already-in-state short-circuit, state-combining/defaults, duration parsing, stdin fallback all uncovered. Added `test_cli.py` driving the real plugin manager.
3. **`check_state_match` agent-target parametrize omits reachable branches** — `RUNNING_UNKNOWN_AGENT_TYPE`/`REPLACED` and the host-RUNNING-ignored-for-agent isolation case were untested. Added the missing cases.
4. **`api.py` `poll_target_state` / resolve fallback paths not directly tested** — `test_..._destroyed_after_connection_errors` faked the very `HostConnectionError`->offline fallback it claimed to cover. Added direct `poll_target_state` fallback tests.
5. **`resolve_wait_target` tests use an untyped `local_provider` parameter** — annotated as `LocalProviderInstance` to match the shared fixture.
6. **`validate_state_strings` tests miss dedup/normalization and error-message content** — added a dedup case and tightened the error match to assert the sorted "Valid states:" enumeration.
7. **`plugin_test.py` asserts only command count/name, not registration wiring** — a packaging/entry-point regression would pass. Covered end-to-end via the loaded plugin manager (folded into finding #2).

</details>

<details>
<summary><strong>libs/mngr/imbue/mngr/utils</strong> — 12 findings</summary>

1. **`build_cel_context` tests assert only key presence, not conversion** — would not catch `build_cel_context` ceasing to produce `celpy.celtypes` values. Assert on converted value types.
2. **`test_import` is a trivial smoke test with a tautological assertion** (`assert mngr`) and no return annotation. Tightened / annotated.
3. **`test_get_current_branch_returns_branch_name` asserts only generic string properties** — pinned a deterministic known branch name.
4. **Editor sleep scripts use non-unique sleep durations** — collision risk with the session-cleanup leak detector. Replaced with large unique durations per the style guide.
5. **`test_editor_session_is_running_returns_true_when_process_running` can race the editor exit** — clarified that the long sleep outlives the synchronous assertion.
6. **`test_format_arg_value_handles_complex_objects` asserts only substring membership** — switched to an exact inline snapshot.
7. **`test_check_bash_version_returns_bool` is a near-tautology** (asserts the annotated return type). Replaced with a case that exercises the version-comparison branch.
8. **`test_dependency_category_values` / `test_install_method_*` assert constructor inputs (tautological)** — removed the definition-restating ones.
9. **`test_render_test_detail_cast_player_init_script_is_present` checks substrings, not structure** — assert the script references the actual player id.
10. **`test_logging_suppressor_buffers_messages` uses loose `>=` lower-bound counts** — documented why exact counts can't be used under shared logging.
11. **`test_name_generator.py` is a `test_*.py` integration file holding pure unit tests** — renamed to `name_generator_test.py` (was missing its `_test.py` companion).
12. **`test_generate_agent_name_generates_unique_names` uses a probabilistic `>= 5` threshold** — made deterministic / behavior-asserting (parametrized variability test).

</details>

## Test plan

The skill itself is documentation; the test-fix commits are verified by the scoped test runs above and will be re-run in CI.


.claude/skills/identify-bad-tests/SKILL.md
```diff
@@ -0,0 +1,87 @@
...
+2. **Does it test behavior or implementation?** Would a legitimate, behavior-preserving refactor of the production code break this test? If yes, it is coupled to implementation details and is a candidate. The good version asserts on the observable effect (return value, persisted state, emitted output), not on how the result was produced.
+3. **Will it fail for the wrong reasons?** Could this test fail (or pass) due to timing, ordering, shared state, environment, or non-unique identifiers rather than the correctness of the code? Flag anything that is non-deterministic, not self-isolating, or dependent on another test having run first.
+
+## Reporting
+
+**Err on the side of over-reporting.** It is fine to report a test you are not certain is bad. The cost of a false positive is low: if the test turns out to be fine, the remedy is a one-line note explaining why it is adequate, which is itself useful documentation. So when in doubt, report it.
```
evgunter: may want to drop this

joshalbrecht: yeah, probably should drop that

joshalbrecht:
lgtm after fixing above thing we both commented on

PR #1887, opened 2026-06-04
rsync: resolve a relative agent destination :PATH against the agent's workdir

gnguralnick:
When an `AGENT:PATH` endpoint is used as a **destination** and its `:PATH` is *relative* -- e.g. `mngr rsync ./src/ my-agent:runtime/foo` -- that relative destination path was passed to rsync verbatim and resolved against the **caller's** working directory, not the agent's. (The local **source** side, `./src/`, is unaffected -- it still resolves against your shell cwd, as expected.)

Consequences for a relative agent **destination**:
- On a local-provider host it silently targeted the *caller's* checkout instead of the agent's worktree. When the source and the mis-resolved destination collapsed to the same directory, rsync transferred nothing ("0 files, 0 bytes").
- On a remote host the relative destination resolved against the SSH login dir, not the agent's worktree.

## Fix

`determine_resolved_path` now resolves a *relative* `:PATH` on an **agent** endpoint against that agent's working directory -- the same directory `mngr exec <agent>` / `mngr connect` drops you into. So `my-agent:runtime/foo` means `runtime/foo` inside the agent's worktree, whether the agent is the rsync source or destination, and regardless of where the command is run. This also makes `agent:relative/path` consistent with the existing by-name-only behavior (no `:PATH` already resolved to the workdir).

Unchanged:
- A bare local path / a relative **source** still resolves against the caller's cwd.
- An absolute `:PATH` is honored verbatim.
- A bare *host* endpoint (`@host:relpath`) still resolves against the SSH login dir, matching plain ssh/rsync.
- This affects every command that resolves a host-location address (rsync, git push/pull, create --from), which is intended.

## Verification
- New unit tests in `find_test.py` cover the relative-with-agent (join onto workdir), absolute-passthrough, and relative-without-agent cases.
- Verified live: with the patched resolver, `mngr rsync /tmp/src/ <agent>:runtime/probe/` lands in the agent's worktree (`/mngr/worktree/<name>/runtime/probe/`), not the caller's `/mngr/code/...`.
- Full fast `libs/mngr` suite: 4060 passed, 1 skipped (excludes acceptance/release/infra markers, which run in CI).
- Updated the `rsync` help metadata + regenerated doc.


joshalbrecht:
lgtm (paired with gabe)

PR #1888, opened 2026-06-04
Discovery provider-error resilience + latchkey mid-session refresh

joshalbrecht:
Implements both threads from `specs/discovery-provider-error-resilience.md` (the workspace-flicker debug follow-ups).

## Thread 1 — retain known hosts/agents through a transient provider discovery error
- A snapshot is authoritative only for providers that succeeded that poll. Consumers now retain previously-known agents/hosts whose provider is in `error_by_provider_name` instead of firing `destroyed`.
- A retained item is dropped only on an explicit destroy event or a later *successful* (non-errored) poll that omits it. Retention is indefinite while the provider keeps erroring — no timer, no cap, no schema change.
- New shared helper `partition_removed_agents_by_provider_error` (in `mngr`) is used by all consumers: `mngr_forward` (stream_manager), `mngr_latchkey` (discovery_stream), and minds (`forward_cli`).
- Minds marks retained workspaces **stale** (amber dot in the sidebar) while keeping them fully clickable; the non-UI consumers retain silently (debug log).
- `FullDiscoverySnapshotEvent` contract docstring flipped to the new rule.

## Thread 2 — latchkey refresh on mid-session config changes
- `mngr latchkey forward` now treats **SIGHUP** as "bounce the `mngr observe` child" (gateway + reverse tunnels stay up); SIGINT/SIGTERM remain shutdown.
- `LatchkeyForwardSupervisor.bounce()` SIGHUPs a live supervisor, or starts one if none is running (start-if-down).
- Minds bounces the detached latchkey forward on the same provider-set triggers that already bounce its own `mngr forward` observe: provider enable/disable, and imbue_cloud account add/remove on signin/signout/OAuth.
- Minds' own `mngr forward` lifecycle and the startup-ordering race are left unchanged (out of scope).

## Out of scope
forever-claude-template's `system_interface` consumer (picks this up via normal mngr vendoring), the "provider silently absent from `providers`" case, and the latchkey startup-ordering race.

## Tests
Per-consumer retention unit tests (errored-provider omission does not drop; a clean snapshot does), supervisor `bounce()` start-if-down test, `_build_workspace_list` stale-marking tests, and the shared-helper unit tests. Changelog entry per touched project (`mngr`, `mngr_forward`, `mngr_latchkey`, `minds`).


hynek-urban:
LGTM

PR #1889, opened 2026-06-04
agent isolation for antigravity

evgunter:
the auth propagation method causes popups saying 'A keychain cannot be found to store "antigravity."' on macos, which might spook users (i warned about it in the README but i can't imagine people will generally see that)

---

## Summary

Each `antigravity` agent now runs `agy` under its own per-agent `$HOME` (at `<agent_state_dir>/plugin/antigravity/home/`), giving each agent its own permission policy, model, and isolated config/transcript/session state instead of the all-or-nothing `--dangerously-skip-permissions` and shared global `~/.gemini`.

`agy` resolves its whole config/permission/auth/session tree from `$HOME/.gemini` and has no config-dir override env var, so relocating `$HOME` is the only lever that yields a per-agent `settings.json`. The per-agent HOME is provisioned and launched-under **unconditionally** (single code path) -- a locked-down vs open agent differs only in *data* (whether `settings_overrides` carries a `permissions` block).

Implements the design in `specs/agy-per-agent-permission-isolation/concise.local.md`.

## Changes

- **Config**: `settings_overrides` (free-form blob: `permissions`/`toolPermission`/`model`), `sync_home_settings` (base per-agent settings on a copy of the user's real settings, default true), `symlink_oauth_token` (default true; see Authentication).
- **Trust**: durable source-repo path persisted to the user's global settings (no re-prompt across agents/worktrees); transient per-agent workspace path written only to the per-agent settings. Consent gating unchanged in spirit -- never silently run an agent on untrusted code (interactive prompt / `--yes` / `auto_dismiss_dialogs`, else clean `SystemExit`).
- **Hooks**: now at `$HOME/.gemini/config/hooks.json`, executed directly; removed the `--add-dir` + `/tmp` hooks-symlink workaround.
- **Onboarding**: seed `cache/onboarding.json` to skip the first-run NUX.
- **Host-aware**: real `$HOME`/OS resolved on the host so token/settings/cache sharing works on remote hosts; heavy `ms-playwright-go` cache shared via symlink.
- **Shared core helper**: `find_git_source_path` extracted to `imbue.mngr.utils.git_utils` (both this plugin and `mngr_claude` delegate; no behavior change).

## Authentication (log in once, anywhere)

Each agent reads its token from `$HOME/.gemini/antigravity-cli/antigravity-oauth-token`. mngr creates that per-agent token as a **symlink to the shared** `~/.gemini` token -- **even when the shared token doesn't exist yet** (a dangling symlink). Because `agy` writes the token **in place** (verified -- it does *not* use temp-file + atomic rename), the first agent's login writes *through* the symlink to the shared path, which authenticates every other agent pointing at it and propagates refreshes. So there is no manual token handling and no background sync: sign in once in any agent (or run a normal host `agy` login) and the rest follow. Works on Linux (file token is native) and macOS (the login keychain isn't reliably readable from a relocated `$HOME`, so the symlinked file token is the cross-agent mechanism there too). `symlink_oauth_token = false` opts into full per-agent isolation. `mngr create` never blocks on a missing token.

## Manual testing (agy 1.0.5, macOS)

Drove real `agy` agents end-to-end to verify behavior, not just "no errors":

- **End-to-end via `mngr create`**: a provisioned agent authenticated via the mngr-seeded shared token (`evan.gunter@imbue.com`), skipped the NUX, used the `/tmp` workspace symlink, and reached the ready TUI -- no login/NUX/trust prompt intercepted input.
- **Permission enforcement (allow/deny)**: an agent with `permissions = { allow = ["command(ls)"], deny = ["command(uv)"] }` auto-ran `ls -la` with **no** prompt, and **hard-blocked** `uv --version` (*"execution of the `uv` command is explicitly set to denied in the agent's permission grants"*).
- **Per-agent isolation**: a second concurrent agent with the **inverse** policy (`allow uv` / `deny ls`) behaved inversely -- `uv` auto-ran, `ls` denied. Two agents, opposite policies, divergent behavior, same host.
- **Write-through symlink auth, end-to-end through mngr (no manual steps)**: from a clean state (no shared token), `mngr create` agent 1 provisioned a *dangling* symlink to the shared path; logging into agent 1 once wrote the 500-byte token *through* the symlink to the shared path (symlink intact -- agy writes in place, resolving the spec's refresh-clobbering risk); then `mngr create` agent 2 auto-authenticated with **no login** and answered a live prompt. So the only user step is the single first login, which propagates to all agents.
- **Auth mechanism findings**: the real file token is a 500-byte JSON `{"auth_method","token"}`; agy's keychain entry is an opaque 691-byte blob, so synthesizing a file token from the keychain is infeasible. A fresh per-agent `$HOME` with no file token falls through to login, confirming the keychain is not a reliable cross-`$HOME` source on macOS and the file token is the mechanism.

## Open items (deferred per spec)

- **Per-project permissions alternative (agy 1.0.5) -- evaluated and rejected; keeping per-agent `$HOME`.** agy 1.0.5 merges three permission scopes, which raised whether per-agent permissions could instead use a per-project file (`config/projects/<uuid>.json`) with no `$HOME` relocation. Verified end-to-end on 1.0.5 (with the `agy-config` agent): (1) a Project-scope `deny` *is* enforced on a tool call; (2) **but the *Global* scope (the CLI's `settings.json`) strictly outranks the *Project* scope in both directions** -- a Global `allow` beats a Project `deny`, and a Global `deny` beats a Project `allow`. Since the user's persisted "always allow X" grants live in the Global scope, a per-project lockdown *cannot* override them, so it can't provide authoritative lockdown at all. (3) The project uuid is random-on-first-open, cached by path in a shared `projects.json` that the GUI and every CLI invocation co-write -- a write-race on every `mngr create`. The per-agent `$HOME` approach instead writes the per-agent *Global* scope (`settings.json`, highest precedence) in an isolated fresh `config/`, giving authoritative, non-overridable, race-free lockdown. Net: 1.0.5 strengthened the case for `$HOME`; this fork is closed in its favor.

## Automated testing

`just test-quick "libs/mngr_antigravity -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'"` -- **150 passed, 0 failed**. (`mngr_claude` quick = 481, `git_utils` = 79 for the shared-helper extraction.) Type check (`ty`) and ruff lint clean.


libs/mngr/imbue/mngr/hosts/common.py
```diff
@@ -95,6 +95,53 @@ def is_macos() -> bool:
     return platform.system() == "Darwin"
 
 
+def symlink_or_copy_on_host(
```
joshalbrecht: this function just doesn't really make a ton of sense--why not one function for symlink and one for copy?

it's a little too strange to have a "do_x_or_y" function

joshalbrecht:
lgtm other than my comment!

PR #1890, opened 2026-06-04
harden scripts/ edge-case handling

evgunter:
Started as a `.gitignore` cleanup and grew (per request) to include fixes from an `identify-suspicious-edge-cases` test run against `scripts/`.

## .gitignore
- Add `**/*.local.sh` (mirroring `**/*.local.md`) so local scratch/helper shell scripts stay untracked; drop the now-redundant single-file `notify_user.local.sh` entry.
- Broaden the identify-* `_tasks/` rule from `*/*/_tasks/` to `**/_tasks/`, so the root-level `dev` project's `dev/_tasks/` output folder is ignored like the `libs/*/_tasks/` and `apps/*/_tasks/` ones.

## Coupled doc fix
The `make_cli_docs --check` change surfaced pre-existing broken doc links: `file`/`tmr` `see_also` still referenced `push`/`pull` (removed in `bba69a86b`, restructured into `rsync` + `git push/pull`). Repointed both at `rsync` and regenerated the docs (touches `libs/mngr_file`, `libs/mngr_tmr`, `libs/mngr` docs).

## Test-isolation fix
Importing `scripts/make_cli_docs` sets `MNGR_LOAD_ALL_PLUGINS=1` process-wide (it must, to load all providers for doc generation). `make_cli_docs_test.py` now pops that env var after import so it can't leak across the xdist worker, and `libs/mngr`'s `test_create_plugin_manager_blocks_disabled_plugins` now asserts up front that the var is unset (a tripwire that surfaces any future leak loudly instead of silently passing).

## scripts/ edge-case findings (from the suspicious-edge-cases review)

The full review surfaced 12 findings against `scripts/`. Resolutions below; the raw report lived at `dev/_tasks/suspicious-edge-cases/` (gitignored).

1. **`release.py::_get_pypi_version` swallowed every exception and returned `None`** — Fixed (blow up): removed the `except Exception: return None`; it now returns a plain `str` and lets HTTP/schema/JSON errors propagate. Caller drops the `(could not check)` / `is not None` handling. release.py needs PyPI access anyway, and "could not check" silently disabled the unpublished-release warning on a schema change.
2. **`release.py::_is_published_on_pypi` caught everything and defaulted to "published"** — Fixed (blow up): removed the `except Exception: return True`. "Assume published" on a non-network error could let a genuinely-new package skip its first-publication safeguards; now any error propagates.
3. **`josh/coordinator.py::process_tasks` catch-all `except Exception` masking bugs** — Reverted/kept as-is. After review we left the catch-all: `scripts/` (and this watcher in particular) is held to slightly different standards than the rest of the repo, and crashing the long-running poller on a transient handler error is worse here than degrading.
4. **`modal_nuke.py` chained `.get(..., "unknown")` fallbacks masked Modal schema drift** — Fixed: reads the real `--json` keys (`"App ID"`, `"Name"`) directly and raises a clear `ModalSchemaError` naming the unexpected schema if a key is missing, so the destructive `modal app stop` / `modal volume delete` path never runs against an `"unknown"` placeholder. (`Description` stays a lenient `.get(..., "")` — it's display-only.)
5. **`make_cli_docs.py::_format_option_type` dead `option.type is not None` guard + `"text"` fallback** — Fixed: dropped the unreachable guard (click options always carry a non-`None` type) and the silent `"text"` fallback.
6. **`make_cli_docs.py::get_relative_link` emitted a broken link for unresolved See-Also refs** — Fixed: raises a clear error (naming command + unresolved ref) when a `see_also` ref resolves to neither a command nor a topic, so `--check` catches it instead of publishing `[mngr help foo](mngr help foo)`. (This is what surfaced the `push`/`pull` doc breakage above.)
7. **`make_cli_docs.py::generate_subcommand_docs` redundant `hasattr(command, "commands")` guard** — Fixed: dropped the `hasattr` half (the `click.Group` type guarantees `.commands`), kept the `not command.commands` empty-group early return.
8. **`sync_common_ratchets.py` `section = "Unknown"` fallback before any header** — Fixed: a check/test function that resolves to no `# --- ... ---` section now raises instead of silently syncing a bogus `# --- Unknown ---` section monorepo-wide.
9. **`make_cli_docs.py::_infer_argument_description` heuristic substring fallbacks** — Documented (accepted as best-effort): added a clarifying comment that these are heuristics for arguments lacking explicit metadata, and that supplying `arguments_description` in `CommandHelpMetadata` bypasses them.
10. **`junit_test_summary.py` if-chains ending in `UNKNOWN`/`PASSED` defaults** — Documented: added comments noting a childless `<testcase>` is by definition a pass, and that `final_status`'s `UNKNOWN` only arises for a zero-attempt record the parser never creates.
11. **`warm_cli_example.py` broad `except Exception` + `os.chdir` swallow (example file)** — Fixed/documented: the `os.chdir` failure now warns to stderr instead of silently running in the wrong cwd; the top-level `except Exception` boundaries got a comment noting they're the deliberate process-entrypoint pattern.
12. **`modal_sandbox_list_bug_repro.py::main` unreachable trailing `return 1`** — Left as-is (throwaway repro script; low priority, noted in the report).

New focused tests for `modal_nuke` and `make_cli_docs` cover the schema-drift and See-Also-resolution paths.


scripts/modal_nuke.py
```diff
@@ -36,13 +36,37 @@
...
+
 
 def _get_app_id(app: Mapping[str, str]) -> str:
-    return app.get("App ID", app.get("app_id", app.get("id", "unknown")))
+    return _require_key(app, _APP_ID_KEY, "app")
 
 
 def _get_volume_name(volume: Mapping[str, str]) -> str:
-    return volume.get("Name", volume.get("name", "unknown"))
+    return _require_key(volume, _VOLUME_NAME_KEY, "volume")
```
evgunter, line 69: not sure if it's desired that these raise instead of returning "unknown" now

joshalbrecht:

PR #1891, opened 2026-06-04
make find_git_worktree_root stop swallowing errors

evgunter:
## Summary

Fixes the rare flaky failure of `test_create_plugin_manager_blocks_disabled_plugins` (surfaced in #1890's CI, but pre-existing on `main`) by fixing the production code rather than working around it in the test.

## Root cause analysis

The failure symptom was `assert pm.is_blocked("modal") -> False`, i.e. `read_disabled_plugins()` returned an empty set. The only path to that is `find_git_worktree_root` returning `None` (so the project config dir is never resolved and the project config layer is silently skipped).

`find_git_worktree_root` caught **any** `ProcessError` from `git rev-parse --show-toplevel` and returned `None`. Crucially, `ProcessError` also covers `ProcessSetupError` (failed to spawn) and `ProcessTimeoutError` — so a non-zero exit, a timeout, *or* a spawn failure all silently became `None`, which every caller interprets as "not in a git repository". A transient or environmental git problem would therefore silently drop the project-scope config with zero diagnostics — which is exactly why the flake was rare, load-dependent, and impossible to reproduce or explain.

## The fix

Stop swallowing. Only git's own "not a git repository" result (the one case where `None` is the correct, expected answer that all callers rely on) maps to `None`. Every other failure is re-raised with git's actual return code and stderr, so the real cause surfaces loudly instead of causing confusing downstream misbehavior. Added a unit test for the raise-on-unexpected-failure behavior.

## What changed vs. the first attempt

The first attempt routed the affected tests around git via `MNGR_PROJECT_CONFIG_DIR`. That hid the symptom without fixing the (unconfirmed) cause and left the error-swallowing in production. Per review feedback, that workaround was reverted in favor of this production fix.

The reported test keeps `@pytest.mark.flaky` as a diagnostic safety net: the underlying cause is still unconfirmed, so a rare recurrence shouldn't block unrelated PRs — but with loud failure in place, offload's first-attempt failure will now capture git's real error for us to inspect.

## Testing

- `just test-quick "libs/mngr/imbue/mngr/utils/git_utils_test.py libs/mngr/imbue/mngr/config/pre_readers_test.py libs/mngr/imbue/mngr/config/loader_test.py libs/mngr/imbue/mngr/main_test.py"` -> 257 passed.
- `just test-quick "libs/mngr -k 'no_type_errors or ratchet'"` -> 55 passed.


joshalbrecht:
lgtm after tests pass

PR #1892, opened 2026-06-04
improve cron recipes: cd into project, workaround for cron auth issues on macOS

evgunter:
## Summary

Updates `libs/mngr_usage/docs/cron_recipes.md`:

- The agent-spawning recipes (`warm-window.sh`, `dispatch-task.sh`) now `cd "$PROJECT_DIR"` (a placeholder project path) before `mngr create`. cron starts in `$HOME` (usually not a git repo), and mngr resolves project-scoped config from the cwd's git worktree root -- so running inside the project both gives `create` a git root to branch from and makes the project's settings (`create_templates`, labels, etc.) apply. Dropped the now-redundant `--from ":$PROJECT_DIR"` from `dispatch-task.sh` since cd makes the create source default to the project's git root.
- Reworked the Scheduling section: explained the bare-cwd caveat alongside the bare-`PATH` one, and added a separate macOS `PATH` example including `/opt/homebrew/bin` (Apple Silicon Homebrew), derived from `scripts/install.sh` (uv-tool install of `mngr`, `claude` via the install script, brew/apt deps `jq`/`git`/`tmux`).
- Added a **macOS LaunchAgent** section as the recommended alternative to `cron` on macOS. cron jobs run outside the GUI (Aqua) login session and can't reach the login Keychain, where Claude Code stores its credentials -- so cron-launched agents come up "Not logged in" and do nothing. A user LaunchAgent loaded into the Aqua session has Keychain access and authenticates with the normal Claude login (no API key on disk). Includes a plist skeleton (`StartInterval`, `EnvironmentVariables` PATH, log paths), `launchctl bootstrap`/`bootout` load/unload commands, the same cwd/config caveat, and the runs-only-while-logged-in tradeoff. (Motivation and facts verified while debugging a real macOS setup.)

Note on the design choice: `--from` only sets the agent's *source repo*; the project config layer is resolved purely from the cwd's git worktree root (`resolve_project_config_dir` -> `find_git_worktree_root`), with no input from `--from`. That's expected behavior, and the reason these recipes need to cd rather than rely on `--from`.

## Testing

`just test-quick "libs/mngr_usage -m '...'"` -- 115 passed, 0 failed.

### DIFF
--- a/libs/mngr_usage/changelog/mngr-refine-cron-recipes.md
+++ b/libs/mngr_usage/changelog/mngr-refine-cron-recipes.md
@@ -1,0 +1,13 @@
+Refined the cron automation recipes doc (`docs/cron_recipes.md`):
+
+- The agent-spawning recipes (`warm-window.sh`, `dispatch-task.sh`) now `cd
+  "$PROJECT_DIR"` before creating an agent, using a placeholder project path.
+  cron starts in `$HOME` (usually not a git repo), and mngr resolves
+  project-scoped config from the cwd's git worktree root -- so running inside
+  the project is what gives the new agent a git root to branch from and applies
+  the project's settings (`create_templates`, labels, etc.). Dropped the now
+  redundant `--from ":$PROJECT_DIR"` from `dispatch-task.sh` (cd makes the
+  create source default to the project's git root).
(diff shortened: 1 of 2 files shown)


joshalbrecht:
lgtm

PR #1893, opened 2026-06-04
improve mngr_file tests

evgunter:
Fixes the seven findings from the `identify-bad-tests` review of `libs/mngr_file`. Test-only changes; no production behavior change.

## Changes
1. Replaced `test_put_and_get_file_on_localhost` (which bypassed the put/get command code and only exercised the underlying host interface) with a real end-to-end CLI round-trip test driving `file_put` (stdin) and `file_get` (JSON) against `@localhost`. This gives the command modules their first integration coverage.
2. Strengthened the localhost listing test to create a uniquely-named file and directory and assert each appears with the correct `file_type` and `size`, instead of asserting the listing is merely non-empty. Made the recursive test self-isolating with `uuid4()` names.
3. Replaced tautological `FileEntry` constructor assertions with immutability and optional-default (omitted, not passed-`None`) tests.
4. Pinned the full `PathRelativeTo` / `FileType` serialized-value mappings via `inline_snapshot` so any rename/add/remove is caught.
5. Tightened CLI argument-validation tests to assert a usage error (exit code 2 + `Missing argument`) rather than just a non-zero exit code.
6. Added a comment clarifying the `--help` tests are intentionally shallow smoke checks (registration/behavior covered elsewhere).
7. Parametrized the three near-identical `_is_volume_accessible_path` tests into one.

## Verification
- `just test-quick "libs/mngr_file -m 'not acceptance and not release'"`: 127 passed, 0 failed.
- `uv run ty check` on all changed files: passed.

### DIFF
--- a/libs/mngr_file/changelog/mngr-bad-tests-mngr-file-local.md
+++ b/libs/mngr_file/changelog/mngr-bad-tests-mngr-file-local.md
@@ -1,0 +1,8 @@
+Improved the quality of the mngr-file test suite (test-only changes, no behavior change):
+
+- Added a real end-to-end test that drives `mngr file put` and `mngr file get` through the CLI and asserts the content round-trips, replacing a test that bypassed the command code and only exercised the underlying host interface.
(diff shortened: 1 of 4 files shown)


joshalbrecht:

PR #1894, opened 2026-06-04
improve mngr_pi_coding unit tests

evgunter:
Fixes test-quality issues in `libs/mngr_pi_coding/imbue/mngr_pi_coding/plugin_test.py` identified by `/identify-bad-tests`. No production behavior changed.

## Findings fixed
1. **`on_before_provisioning` test asserted nothing and was not isolated** — it read the real `~/.pi/agent/auth.json` (the temp-home setup was dead) and never checked the warning. Now redirects `HOME` to a temp dir and asserts the missing-credentials warning, plus a new positive case that asserts no warning when an auth file is present.
2. **Remote auto-install test never verified installation** — it only checked `provision()` didn't raise, which holds whether or not pi is installed. The stub host now records executed commands and the test asserts `npm install -g ...` actually ran.
3. **Abstract-method check probed `__abstractmethods__`** — replaced with `inspect.isabstract(PiCodingAgent) is False`, the observable "is this instantiable" property.
4. **Local symlink tests only checked `is_symlink()`** — now also assert the link targets via `os.readlink`.
5. **`pi_agent` fixture lived in the test file** — moved to `conftest.py` (per the all-fixtures-in-conftest convention); added a `log_warnings` capture fixture.

The full findings report (including lower-priority notes on the stub-host location and the remaining constant/getter tests) lives in `libs/mngr_pi_coding/_tasks/bad-tests/` (gitignored).

## Testing
`just test-quick libs/mngr_pi_coding` — 77 passed, 0 failed.


joshalbrecht:

PR #1895, opened 2026-06-04
improve concurrency_group tests

evgunter:
Acts on the 14 findings from the `identify-bad-tests` review of `libs/concurrency_group` (file under `_tasks/bad-tests/`). Test-only; no production code changed.

## Key fixes (tests that passed without verifying behavior)
- **`test_executor_respects_max_workers`**: was `assert max_concurrent <= 2`, which passes even if all work serialized. Now asserts `== 2` and fails if the synchronizing barrier breaks.
- **`test_run_background_thread_safety`**: the subprocess finished before the worker threads started, so concurrent access was never exercised. Now blocks the process on a signal file, hammers `poll`/`read` while it runs, and asserts on the observed output.
- **`test_run_background_interleaved_stdout_stderr`**: asserted `sorted(...)`, hiding any per-stream reordering. Now asserts per-stream order directly.
- **Suppressed/unchecked failed-thread tests**: would pass even if the thread never ran. Now confirm the failing thread actually executed.
- **`test_all_failure_modes_get_combined`**: dropped the timing-dependent exact count (`== 4`); asserts the always-present failure kinds and polls for the reap.

## Flakiness / hygiene
- Removed flaky wall-clock upper-bound timing assertions (`test_run_background_real_time_queue`, `test_concurrency_group_does_not_raise_when_within_timeout`).
- `test_nesting_in_the_same_thread_just_works` now asserts an observable effect instead of only not raising.
- Collapsed two duplicate `_shutdown_popen` tests into one that checks the SIGTERM returncode.
- Clarifying comments on the strand-cleanup tests; removed unused `tmp_path` params; `test_run_background_with_cwd` uses the `tmp_path` fixture; long-lived placeholder subprocesses use a single globally-unique sleep duration (`LONG_SLEEP_SECONDS`).

## Testing
`just test-quick libs/concurrency_group` -> 161 passed. Ratchets pass. Timing-sensitive tests run 6x with no flakes.

### DIFF
--- a/libs/concurrency_group/changelog/mngr-bad-tests-concurrency-group-local.md
+++ b/libs/concurrency_group/changelog/mngr-bad-tests-concurrency-group-local.md
@@ -1,0 +1,11 @@
+Hardened the concurrency_group test suite (no production behavior change):
+
+- `test_executor_respects_max_workers` now asserts that two workers actually run concurrently (`== 2`) and fails loudly if the synchronizing barrier breaks, instead of only checking the upper bound.
+- `test_run_background_thread_safety` now blocks the subprocess on a signal file so the concurrent poll/read access is genuinely exercised while the process is running, and asserts on the observed output rather than only "no errors".
+- `test_run_background_interleaved_stdout_stderr` now asserts per-stream line order directly instead of sorting it away.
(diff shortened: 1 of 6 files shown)


joshalbrecht:

PR #1901, opened 2026-06-04
improve mngr_claude tests

evgunter:
## Summary

Output of `/identify-bad-tests libs/mngr_claude` followed by fixing the findings. Reviewed all 11 test files in `libs/mngr_claude` against the style guide's testing bar and rewrote tests that passed without establishing correctness. Full findings recorded in `libs/mngr_claude/_tasks/bad-tests/` (gitignored).

Key fixes:
- **Seven `on_before_provisioning` tests asserted nothing** ("did not raise") -- now assert observable effects: config left untouched, missing-credentials warning present/absent (via `capture_loguru`), untrusted worktree rejected.
- **`does-not-extend-trust` tests** asserted the presence of a key the test itself wrote -- now assert the exact set of trusted projects.
- **Transcript truncation tests** used `<=` upper bounds -- now assert exact length + the `...` ellipsis marker, plus short-input non-truncation cases; "skips event" tests now prove only the bad event is dropped and cover the missing-`timestamp` branch.
- **Command-assembly / install-command tests** compared hand-rebuilt exact shell strings -- now assert shlex-parsed tokens / load-bearing flags.
- **Skill-install skip tests** relied on mtime equality (can false-pass) -- now assert file content unchanged; added remote-install coverage. Custom-type resolution test now sets a non-default field to prove it survives the merge.
- **Headless grace-period test** now asserts the poll actually re-checked; removed dead `_patch_agent_as_stopped` calls and a fragile wall-clock timing assertion.
- **claude_config no-op tests** assert content is byte-for-byte unchanged; effort-callout test isolated to the effort dialog.
- **Release/acceptance tests** assert real effects (preserved session JSONL, gone-session exit log) instead of brittle `"Done."` log strings; removed a duplicate `temp_source_dir` fixture; replaced magic `sleep` literals with a named constant.

Also removed an introduced `unittest.mock.patch` of the function under test (used the real no-credentials environment instead, consistent with the sibling unit test).

## Test plan
- `just test-quick libs/mngr_claude` (non-infra markers): **486 passed**.
- Ratchets: **55 passed** (no mock/monkeypatch count increase).
- `ty check` + `ruff`: clean.
- Release/acceptance/tmux tests edited conservatively from production code (cannot run locally); will run in CI.

### DIFF
--- a/libs/mngr_claude/changelog/mngr-bad-tests-mngr-claude-local.md
+++ b/libs/mngr_claude/changelog/mngr-bad-tests-mngr-claude-local.md
@@ -1,0 +1,11 @@
+Test-quality hardening across the mngr_claude test suite (no user-visible behavior change). Replaced assertions that passed without verifying correctness with ones that check real effects:
+
+- Seven `on_before_provisioning` tests that asserted nothing ("did not raise") now assert observable effects (config left untouched, missing-credentials warning present/absent, untrusted worktree rejected).
+- `does-not-extend-trust` provisioning tests now assert the exact set of trusted projects instead of the presence of a key the test itself wrote.
+- Transcript-converter truncation tests now assert exact lengths and the ellipsis marker (not just an upper bound), and the "skips event" tests now prove only the bad event is dropped (a known-good event survives) plus cover the missing-`timestamp` branch.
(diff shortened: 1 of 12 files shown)


joshalbrecht:

PR #1905, opened 2026-06-04
Consolidate to a single discovery observer for the minds app

joshalbrecht:
## Summary

Removes the duplicate `mngr observe --discovery-only` process under the minds app so there is a **single source of truth** for discovery.

Previously two observers ran: one spawned by `mngr latchkey forward` (for reverse tunnels) and one by minds' own `mngr forward --service system_interface`. PRs #1885/#1888 worked around the resulting cross-talk by isolating them onto separate event logs and adding provider-error retention. This closes the loop by eliminating the second observer entirely.

## Changes

- **mngr**: new `mngr forward --observe-via-file` tails the shared discovery events file in-process (new public `tail_discovery_events_file` helper reusing the snapshot-emit + tail phases of `run_discovery_stream`) instead of spawning its own observe. The PR-1885 `MngrConfig.events_base_dir_override` plumbing is fully reverted — the discovery log always lives under the default host dir, and `--events-dir` with `--discovery-only` is now a `click.UsageError`.
- **mngr_forward**: `ForwardStreamManager` gains a tail-file discovery mode; per-agent `mngr event` streams are unchanged; SIGHUP is a no-op in this mode. `--observe-via-file` is mutually exclusive with `--no-observe` and works with `--service` or `--forward-port`.
- **mngr_latchkey**: `mngr latchkey forward`'s observe writes to the standard discovery log again (it is now the sole observer); the private `discovery-observe` dir + rmtree are removed.
- **minds**: launches `mngr forward --observe-via-file`; provider-set changes refresh discovery solely by bouncing the `mngr latchkey forward` supervisor; the now-redundant `EnvelopeStreamConsumer.bounce_observe` and its call sites are removed.

## Behavior

- Exactly one `mngr observe --discovery-only` runs per minds session; user-visible behavior is unchanged.
- The tailer emits the latest cached snapshot immediately and is corrected by the live observer's next snapshot; it tolerates the file being absent at startup and being rotated/truncated.

## Testing

- Unit tests for `tail_discovery_events_file` (cached-snapshot emission, tailing appends, absent-file wait).
- Integration test that `mngr forward --observe-via-file` picks up agents from a log written by another process and spawns no observe subprocess.
- CLI validation tests for the new flag combinations; updated config/discovery tests for the reverted override.
- Locally: targeted unit/integration tests for all four touched projects, ratchets, and `ty` type checks pass. Full offload suite runs in CI.


joshalbrecht:
roughly lgtm

isn't my favorite, but should be a bit less confusing than having two mngr observe processes running

PR #1906, opened 2026-06-04
Disable OVH VPS backups by purging qemu at provision time

joshalbrecht:
## Summary

OVH automated backups drive the image's `qemu-guest-agent` to freeze the guest filesystem, which causes serious runtime problems on the agent. This disables OVH-side backups at the OVH provider level by purging qemu on every provisioned VPS.

## Changes

- **`bootstrap.py`**: new `purge_qemu_packages` helper that detects qemu (`dpkg -l | grep qemu`) and runs `apt-get purge --auto-remove 'qemu*'` over root SSH. Detect-first so it's a clean no-op on a qemu-free image (avoids an apt glob error). Reuses the existing 300s apt timeout. Raises `VpsProvisioningError` on failure.
- **`backend.py`**: calls `purge_qemu_packages` as the final outer-bootstrap step in `OvhProvider._provision_vps`, after `install_required_outer_packages`. Runs on both the fresh-order and recycle paths (rebuilding the OS reinstalls the agent). A failure aborts provisioning so the normal create-cleanup tears down the VPS — no host is left running with backups enabled.
- **Regression test** (`ordering_test.py`): asserts the order/cart flow never configures a backup option (only `vps_datacenter` / `vps_os` / `vps_install_rtm=no`).
- **Bootstrap tests** (`bootstrap_test.py`): cover the detect-then-purge command shape and the fatal-on-failure behavior.
- **OVH README** + changelog entries.

## Scope / notes

- New provisions only; existing live hosts are not swept (they pick up the purge when next recycled).
- imbue_cloud pool hosts are baked through this provider, so they're covered; their later container-rebuild slow path runs on an already-purged OS and doesn't reinstall OS-level qemu.
- The minds restic `BackupProvider` feature is out of scope and untouched.

## Testing

- `just test-quick libs/mngr_ovh` (filtered): **203 passed**, including the 3 new tests and all ratchets (no `unittest.mock` ratchet bump — tests were appended to existing files).


joshalbrecht:
lgtm

PR #1907, opened 2026-06-04
Redact pool DSN from minds pool command logging

joshalbrecht:
## What

\`minds pool {list,create,destroy}\` echoed the full admin argv -- including \`--database-url\`, the Neon pool DSN (which embeds the DB username + password) -- into the \`Running: ...\` log line whenever \`--database-url\` was passed explicitly. This leaked production DB credentials into logs/terminals.

## Fix

- Mask secret-bearing flag values before rendering the command for logging. The real subprocess still receives the unredacted value.
- Centralize the masking logic that \`mngr forward\`'s \`--preauth-cookie\` redaction already used into a shared \`imbue.minds.utils.secret_redaction.redact_secret_flag_values\` helper, and reuse it in both \`cli/pool.py\` and \`desktop_client/forward_cli.py\`.
- The helper masks both \`--flag value\` and \`--flag=value\` forms, every occurrence.

## Tests

- New \`utils/secret_redaction_test.py\` (space/joined forms, multiple occurrences, multiple flags, no-op, dangling flag, no mutation).
- New \`pool_test.py\` case asserting a representative DSN never survives into the loggable admin command.
- Existing \`forward_cli\` redaction tests still pass (its \`_redact_secrets\` now delegates to the shared helper).


joshalbrecht:
lgtm

PR #1908, opened 2026-06-04
improve mngr_claude_usage tests

evgunter:
Acts on the findings from `libs/mngr_claude_usage/_tasks/bad-tests/` (an `/identify-bad-tests` run). Test-only changes.

## Changes
- **Real agent instead of ad-hoc stubs (highest-value):** the `on_before_provisioning` filter test now uses a real non-Claude `BaseAgent` on the real `local_host` instead of `BaseModel` stubs + `# ty: ignore`. If the `isinstance(agent, ClaudeAgent)` guard regresses, provisioning actually runs and the `assert not ... exists()` checks catch it cleanly, rather than crashing on a fake's missing methods.
- **Correct test placement:** moved the `claude_statusline.sh` shim subprocess test out of the unit `plugin_test.py` into a new `test_shim.py` integration file, with a `bash`-availability skip (parity with the writer's `jq` skip).
- **Fixtures in conftest:** moved shared `writer_path` / `events_file` fixtures into `conftest.py`.
- **Stronger assertions:** validate the full nanosecond-precision UTC ISO 8601 timestamp shape (was a bare `"T" in ts`); assert every distinct event survives exactly once in the concurrent-append test (was just a line count).

## Deliberately not changed
One finding (`test_capture_tolerates_malformed_json` silently tolerating a malformed user-authored `settings.json`) was left alone: `_capture_existing_statusline_command` runs at *provisioning* time, so whether it should raise/warn on bad user config (per the style guide) vs. tolerate it is a production-behavior contract decision, not a test fix. Flagged for a maintainer decision.

## Tests
- `just test-quick libs/mngr_claude_usage` -> 82 passed
- ratchets -> 55 passed


joshalbrecht:

PR #1910, opened 2026-06-04
automatically create publish graph + bump pinned versions in release script

evgunter:
## Why

The original ask: the branch `wz/fix-mngr-pin-stragglers` manually realigned stale `imbue-mngr-*==` pins that a release left behind — can `release.py` do this itself?

Root cause: `release.py` drove off a **hand-maintained `PACKAGES` allowlist** (`scripts/utils.py`). Any `libs/*` package nobody remembered to add was invisible to the release tooling — never bumped, never pin-aligned, never offered for publication. Two consequences:

1. **Stale pins.** A dozen internal `==` pins drifted (e.g. `mngr_usage` pinned `imbue-mngr==0.2.6`, `mngr_latchkey` pinned `imbue-common==0.1.17`). These are masked day-to-day by the `[tool.uv.sources]` workspace redirect and only break `apps/minds/scripts/build.js`'s override-free `uv lock` — a path CI never runs.
2. **Plugins unpublished by omission.** Several standalone-usable plugins were never published. `imbue-mngr-schedule` is even listed in the install catalog, so the wizard offered it while it 404'd on PyPI.

## What

Invert the model to match the intended behavior ("scan everything; offer anything not deliberately unpublished"):

- **Auto-discover** the publish graph: every `libs/*` package is a publish candidate unless listed in `UNPUBLISHED_PACKAGES` (the single opt-out, shared with the install wizard). `internal_deps` and topological order are derived from the pyprojects.
- **Generalize pin alignment** (`update_internal_dep_pins`) over every workspace member (`libs/` + `apps/`) across `dependencies`, `optional-dependencies`, and `dependency-groups`. Publishable wheels force `==` on publishable runtime deps; everywhere else only existing `==` pins are realigned.
- **Generalize verification**: `verify_pin_consistency` now catches stale/missing pins anywhere (fails CI instead of only at build-time); `validate_package_graph` asserts the publish graph is *closed*; a new `test_every_lib_is_classified` ratchet guarantees no lib can fall into limbo again.
- **Fix new-package detection** to cover cascade-reached packages, so an unpublished package that depends on `imbue-mngr` (cascades every release) is offered for first publication instead of silently bumped.
- **Classify** the previously-invisible libs (per maintainer decision): publish `mngr_{usage,claude_usage,ovh,imbue_cloud,latchkey,forward,schedule,uncapped_claude}` (all standalone-usable); keep `mngr_{mapreduce,tmr,claude_subagent_proxy}` and `skitwright` unpublished.
- **Apply the alignment now**: realign the 12 stale pins and add the 8 `==` pins the newly-publishable wheels require.

## Consequence to be aware of

After this merges, the next `release.py` run will **offer to first-publish 8 packages**. Each is gated by an interactive y/N, and PyPI allows only one pending Trusted Publisher per account at a time — so publication is one-at-a-time, not automatic. Nothing publishes without explicit confirmation.

## Verification

- `scripts/` + `plugin_catalog` tests: 203 passed.
- `uv run ty check`: clean (one pre-existing unrelated `celpy` warning).
- End-to-end `release.py patch --dry-run`: all 8 newly-discovered packages correctly listed under "New packages (first publication)".
- Pin alignment is idempotent; `verify_pin_consistency()` returns 0 errors.

Note: two `test_meta_ratchets.py` checks (`test_prevent_bash_without_strict_mode`, `test_prevent_old_mng_name_in_file_contents`) fail locally, but on **pre-existing tracked files in `main`** (a committed `.claude/skills/blueprint` symlink and `.minds/template/*.sh`) — this PR touches none of them and does not edit the snapshot.


joshalbrecht:
lgtm

will let you merge this one though (and hit publish after to take care of the manual step)

PR #1912, opened 2026-06-04
mark ssh as optional + improve flags for mngr dependencies

evgunter:
## Summary

Reworks `mngr dependencies` so that **which dependencies count** and **whether to install** are two orthogonal options, instead of being conflated into the old `-c`/`-a`/`-i` flags. Also reclassifies `ssh` from a core to an optional dependency.

### New flags (replacing `-c`/`-a`/`-i`)

| option | values | meaning |
|---|---|---|
| `--scope` | `core` \| `all` (default `all`) | Which deps determine the exit code (and which `--install auto` targets). `core` exits non-zero **only** if a core dep is missing — missing optional deps are tolerated. |
| `--install` | `none` \| `interactive` \| `auto` (default `none`) | Whether/how to install missing deps. The interactive prompt is unchanged. |

Mapping: old `-c` → `--scope core --install auto`; old `-a` → `--install auto`; old `-i` → `--install interactive`.

The requested mode (`mngr dependencies --scope core`) exits 1 **only** if a core dependency is missing:

```
$ PATH=<core-only> mngr dependencies --scope core ; echo $?   # only optional (ssh/rsync/...) missing
0
$ PATH=<core-only> mngr dependencies ; echo $?                # default scope=all
1
```

### `ssh` demoted to optional

Remote connectivity runs through paramiko (pure-Python, no `ssh` binary). The `ssh` binary is only needed to attach an interactive session to a remote agent and as the transport for rsync/git over SSH — all remote-only features, putting it in the same class as `rsync`/`unison`. Core deps are now `git`, `tmux`, `jq`. `mngr connect` to a remote agent now raises a clear `BinaryNotInstalledError` if `ssh` is missing.

### Other changes
- `scripts/install.sh`: `mngr dependencies -i` → `mngr dependencies --install interactive --scope core` (warns only on a missing *core* dep; preserves the old core-only warn semantics while still letting users install everything at the prompt).
- Updated the README dependency list, the isolated-venv test fixture, and all affected tests.

## Testing
- `just test-quick` on the affected unit/integration tests (deps, check_deps, install_script, connect): **111 passed**.
- `uv run ty check` on changed files: clean.
- Ratchets: pass.
- Manually verified the new exit-code semantics on a controlled PATH (shown above).


joshalbrecht:
lgtm

I've never seen a system that doesn't have ssh, but sure :-P

PR #1915, opened 2026-06-04
standardize how plugins share test autouse fixtures + enforce with meta test

evgunter:
## What

Every mngr plugin now isolates `$HOME` in its tests via the single shared mechanism `register_plugin_test_fixtures(globals())`, and a new meta test enforces it.

This started from a request to add a test that each mngr plugin inherits mngr's autouse test fixtures. Investigating revealed the inheritance was happening **three** different ways (and 8 plugins not at all), so this PR standardizes on one and dedups the underlying fixture definitions.

## Why

The autouse `setup_test_mngr_env` fixture redirects `$HOME` to a temp dir so tests can't read/write the real `~/.mngr` / `~/.claude.json`. It was reachable via:
- `pytest_plugins = ["imbue.mngr.conftest"]` (5 plugins)
- `register_plugin_test_fixtures(globals())` (12 plugins)
- neither (**8 plugins**) — their tests ran against the real home dir (confirmed via a probe: `mngr_vultr` saw `HOME=/Users/ev`)

On top of that, `setup_test_mngr_env` and ~15 sibling fixtures were **defined twice** (in `imbue.mngr.conftest` and `imbue.mngr.utils.plugin_testing`), which had already drifted (e.g. a load-all vs. blocking `plugin_manager`).

## Changes

- **Single-source** the shared plugin fixtures in `imbue.mngr.utils.plugin_testing` (relocating `stub_mngr_log_sh`, `mngr_transcript_lib_sh`, `project_config_dir`, `temp_work_dir`, `local_host`, `per_host_dir`). `mngr`'s own `conftest.py` now registers that set instead of redefining ~19 duplicate fixtures, keeping only the 3 fixtures that *intentionally* differ for mngr-core (blocking `plugin_manager` + `enabled_plugins`, `mngr_test_id`'s tmux-cleanup tracking, `temp_mngr_ctx`'s provider-cache reset).
- **Migrate all 25 mngr plugins** to `register_plugin_test_fixtures(globals())`. `mngr_claude` keeps `pytest_plugins = ["imbue.mngr_modal.conftest"]` *only* to share modal's test fixtures (orthogonal to HOME isolation) and drops its base `setup_test_mngr_env` so modal's credential-loading override still wins; its now-vestigial `enabled_plugins` override is removed.
- **New meta test** `test_every_mngr_plugin_isolates_home_in_tests`: any project with a `[project.entry-points.mngr]` table must call `register_plugin_test_fixtures` in a conftest. (`mngr_mapreduce` / `mngr_vps_docker` register no such entry point — they're support libs, not plugins — and are correctly excluded.)

Internal test-infrastructure change only; no user-facing behavior change.

## Verification

- `mngr` full fast suite: 4063 passed, 1 skipped
- 11 simple-migrated plugins: 1456 passed; `mngr_claude`: 481 passed; `mngr_modal` (non-real): 386 passed
- meta-ratchets ruff / ty / import-layer / new isolation test: pass
- Empirically confirmed before/after: flagged plugins ran against real `$HOME`; after, HOME is redirected; `mngr_claude` still loads `MODAL_TOKEN_ID`


libs/mngr/imbue/mngr/conftest.py
```diff
@@ -100,47 +89,27 @@ def _remove_deprecated_urwid_module_aliases() -> None:
...
+#
+# Most of mngr's test fixtures (HOME isolation, temp host/profile/config dirs,
+# git-repo helpers, the autouse setup_test_mngr_env, the shell-stub fixtures,
+# etc.) are defined once in imbue.mngr.utils.plugin_testing and registered here
+# via register_plugin_test_fixtures -- the exact same entry point every mngr
+# plugin uses. Single-sourcing them this way keeps mngr-core and its plugins
+# from drifting apart.
+#
+# A few fixtures below intentionally OVERRIDE the plugin_testing versions
+# because mngr-core needs stricter behavior than plugins do:
```
evgunter: weird

joshalbrecht:
lgtm, good catch!

PR #1921, opened 2026-06-04
harden minds/config suspicious edge-case handling

evgunter:
not sure if we want this

---

## Summary

Cleanup pass (`/identify-suspicious-edge-cases`) over `apps/minds/imbue/minds/config`, then fixes for the findings.

- **`parse_agents_from_mngr_output`** (`config/data_types.py`): now raises `MalformedMngrOutputError` instead of (a) silently returning `[]` when mngr's stdout is empty/blank and (b) raising a bare `KeyError` when the parsed JSON object lacks an `agents` key. `mngr list --format json` always serializes its result as `{"agents": [...]}` (zero agents is `{"agents": []}`), so empty stdout signals broken upstream output, not "no agents" -- consistent with the function's existing non-JSON branch and documented philosophy.
- **`load_client_config` / `load_deploy_config`** (`config/loader.py`): catch the precise `tomllib.TOMLDecodeError` and `pydantic.ValidationError` rather than the broad `ValueError`, so an unrelated `ValueError` is no longer mislabeled as a config parse/validation failure.

Added unit tests for the new raising behavior.

## Test plan

- `just test-quick apps/minds/imbue/minds/config` -- 28 passed
- `just test-quick "apps/minds -k 'test_no_type_errors or ratchet'"` -- 55 passed


joshalbrecht:
lgtm -- these changes actually look good

PR #1923, opened 2026-06-04
Apply extra IAM tags (minds_env) when recycling OVH VPSes

joshalbrecht:
## What

When the OVH provider recycles a cancelled-but-still-running VPS instead of ordering a fresh one, it now (over)writes the bake's extra IAM tags (e.g. `minds_env=<env>`) on the recycled host, alongside the existing `mngr-host-id` swap. Fresh-order behavior is unchanged.

## Why

The recycle path (`try_recycle_cancelled_vps`) previously swapped only `mngr-host-id` and skipped extra tags entirely (the old comment in `backend.py` said so explicitly). Since OVH's provider name is shared across all envs/tiers (`mngr-provider=ovh`), a staging bake can recycle a VPS that previously belonged to another env — and the resulting pool host would carry **no** `minds_env` tag, or a stale one.

That broke two things:
- `minds env destroy` enumerates an env's VPSes by the `minds_env` IAM tag (`apps/minds/.../envs/providers/ovh_tags.py`), so a recycled pool host was invisible to env-scoped teardown.
- Operators couldn't tell which env actually owns a recycled host.

Discovered while baking a staging pool host on this branch, which recycled `vps-d7fa1e9d` and produced a row with no `minds_env` tag.

## Changes

- `recycle.py`: `try_recycle_cancelled_vps` takes `extra_tags` and applies them (overwrite) under the recycle lock in `_try_recycle_one`, after the host-id swap; abort + release lock on attach failure.
- `backend.py`: thread parsed `MNGR_VPS_EXTRA_TAGS` into `_maybe_claim_recycled_vps`; updated the now-stale comments.
- `recycle_test.py`: new `test_recycle_claim_overwrites_extra_tags_for_new_owner`; existing call sites updated for the new required kwarg.

## Tests

`just test-quick libs/mngr_ovh` — 205 passed (incl. type-checks + ratchets).

## Note (not in this PR)

The already-baked staging pool host (`vps-d7fa1e9d`) was retro-tagged with `minds_env=staging` by hand. Separately, `_sync_mngr_into_template` does not reset `vendor/mngr/` after a `--mngr-source` bake even though `host-pool-setup.md` claims it does — left as-is here.


joshalbrecht:
lgtm

PR #1924, opened 2026-06-04
document minds/utils edge cases + unify logging level metadata

evgunter:
Started as the output of the `/identify-suspicious-edge-cases` pass over `apps/minds/imbue/minds/utils`, then grew a follow-on cleanup of `logging.py` (per review discussion).

## Part 1: document the intentional edge cases

All four suspicious edge-case handlers found turned out to be intentional and correct (three are pinned by existing tests in `output_test.py`). Per the skill, the remedy for a correct-but-undocumented handler is a clarifying comment:

- **`emit_event` JSON no-op** (`output.py`): the `case OutputFormat.JSON: pass` silently drops every event. `--format json` is a valid global CLI option, and `minds run`'s only structured output flows through `emit_event`, so `minds run --format json` prints nothing. Comment explains JSON is a single-document format with no place for a stream of events, and that streaming consumers should use `jsonl`.
- **`emit_event` HUMAN message guard** (`output.py`): events lacking a `message` key (e.g. `mngr_forward_started`) are intentionally suppressed in HUMAN mode because they are machine-only signals. Comment added.
- **`logging.py` level metadata**: comments explain why `ConsoleLogLevel.NONE` has no loguru level (the caller filters it before lookup) and why uncolored/unknown levels render plain.

## Part 2: unify the logging level metadata (`logging.py`)

The original code split per-level knowledge across two places -- a `_LEVEL_MAP` dict (enum to loguru level name) and a sequential `if level_name == ...` chain in `_format_user_message` that carried the colors and `WARNING:`/`ERROR:` prefixes (and sidestepped the if/elif-without-else ratchet). These are now one source of truth:

- A single `_LEVEL_STYLES` table keyed by `ConsoleLogLevel`, whose values are a `_LevelStyle` **`FrozenModel`** holding the loguru name, color, and message prefix per level.
- `setup_logging` derives the sink threshold from it; `_format_user_message` renders via a derived `_STYLE_BY_LOGURU_NAME` reverse map (keyed by loguru level name, since the sink sees loguru names, not our enum), with a `None` default for loguru-only levels (`SUCCESS`/`CRITICAL`) that minds never emits.
- The per-line rendering -- a pure derivation of a level's own `color`/`prefix` -- lives on the frozen object as `_LevelStyle.loguru_format()`. The cross-table lookup and unknown-level fallback (not a property of any single style) stay in the free `_format_user_message` function.

Output is byte-identical for every level (verified directly against the previous literals, including the loguru-only levels). No behavior change.

## Verification

- In-scope tests (`apps/minds/imbue/minds/utils`): 23 passed, 0 failed.
- `if/elif-without-else` ratchet still passes at 0; `ty check` clean.
- Architecture verification and autofix gates run (clean).

**Design concern (tracked separately via PR comment):** `minds run --format json` succeeding silently with empty output is a footgun. A follow-up could make streaming-only commands reject `--format json` with a clear error.


apps/minds/imbue/minds/utils/output.py
```diff
@@ -22,13 +22,23 @@ def emit_event(
...
             event = {**data, "event": event_type}
             sys.stdout.write(json.dumps(event) + "\n")
             sys.stdout.flush()
         case OutputFormat.JSON:
+            # ``emit_event`` is for streaming, per-event output. JSON denotes a
+            # single self-contained document, into which a stream of events has
+            # no place, so emitting events is a deliberate no-op here. Commands
+            # that produce a JSON document build it separately (see
+            # ``cli/env.py``); structured consumers of streaming commands such
+            # as ``minds run`` should select ``jsonl`` rather than ``json``.
```
evgunter, line 41: should this error/warn?

joshalbrecht:
this change is mid, we can probably discard it

PR #1927, opened 2026-06-04
fix unquoted args in antigravity

evgunter:
## Problem

Passing a model name (or any value with spaces/parens) as an explicit CLI arg to an `antigravity`/`agy` agent breaks `mngr create`:

```
mngr create <name> --type antigravity ... -- --model "Gemini 3.5 Flash (Medium)" ...
# bash: syntax error near unexpected token `('
```

The assembled launch command contained `agy --model Gemini 3.5 Flash (Medium) ...` with the value spliced in **unquoted**, so the shell that evaluates the command word-split it and parsed `(Medium)` as a subshell.

## Root cause

`BaseAgent.assemble_command` (`libs/mngr/imbue/mngr/agents/base_agent.py`) joined the post-`--` `agent_args` into the command string with plain spaces, without shell-quoting. `agent_args` are raw argv (the OS shell already stripped their quotes at invocation), so they must be re-quoted before being spliced into a shell-evaluated string.

## Fix

Quote each `agent_args` element with `shlex.quote`. `cli_args` are intentionally left unquoted: string-form `cli_args` configs are split with a quote-preserving (non-POSIX) shlex (`config/data_types.py:split_cli_args_string`) and so already arrive shell-safe. Fixing the base method fixes `antigravity` and every other agent type that inherits it.

The same quoting rule already existed (duplicated) in the `mngr_claude` plugin's `assemble_command` override, which doesn't call `super`. The second commit factors the rule into a shared `quote_agent_args` helper in `base_agent.py` that both call sites use, so they cannot drift (no behavior change).

Note: the supported way to set an antigravity model is `settings_overrides` (a `model` key in the per-agent `settings.json`), which never touched the CLI and was never affected. This fix covers the case where a model is passed explicitly as a CLI argument anyway.

## Tests

- `base_agent_test.py`: direct unit test for `quote_agent_args`, plus an `assemble_command` regression test (`--model "Gemini 3.5 Flash (Medium)"` -> `agy --model 'Gemini 3.5 Flash (Medium)'`).
- `plugin_test.py` (antigravity): end-to-end launch command quotes the value and never contains the raw unquoted form.

Known residual (not addressed, consistent with the existing convention): **list-form** `cli_args = ["--model", "Gemini 3.5 Flash (Medium)"]` is still not quoted. Closing that would require distinguishing already-quoted string-form tokens from raw list tokens.


joshalbrecht:
lgtm

feels a little dangerous, but also seems like it is correct...

PR #1931, opened 2026-06-05
in antigravity, only set WAITING when all background tasks are done

evgunter:
built on https://github.com/imbue-ai/mngr/pull/1889

---

## What

The antigravity `Stop` hook previously cleared the `active` lifecycle marker on any `fullyIdle:true`, so an agent flipped to **WAITING** the moment the root agent's turn ended -- even if a subagent or backgrounded task it launched was still running. This gates the clear so the agent reports **RUNNING** until the *root* agent itself is fully idle. It also centralizes "which conversation is the agent's main one" and uses that to fix a related resume bug.

## How (lifecycle marker)

- `PreInvocation` runs **`set_active_marker.sh`**: touches the marker and, when the marker was absent (this invocation opens a new turn), records the conversation id as the turn's **root** in `root_conversation`. agy always runs the root agent's invocation before it spawns subagents, so the turn opener is the true root; re-recording each turn keeps `/clear`, `/fork`, `/switch`, and resume correct.
- `Stop` runs **`clear_active_marker_when_idle.sh`**: clears the marker only when the payload's conversation id matches the recorded root **and** reports `"fullyIdle":true` (liveness fallback: clear on `fullyIdle:true` if no root was ever recorded).

### Why the subagent gate (review feedback)

Subagents share the agent's `hooks.json` and `$MNGR_AGENT_STATE_DIR`, and each fires its **own** `Stop` with `fullyIdle:true` when it finishes -- which can arrive while the root agent is still working. The Stop payload has **no** root/parent flag; the only discriminator is `conversationId`. Confirmed live (agy 1.0.5): subagent conv `55afd8ce` emitted `fullyIdle:true` while root conv `f0941390` was still emitting `fullyIdle:false`.

## Also: conversation resume (same root tracking)

`assemble_command`'s resume prelude now reads the main conversation id from `root_conversation` instead of `tail -n 1` of the conversation-ids file. This fixes a latent bug: subagents also append to the conversation-ids file via the shared capture hook, so its last line (and thus the resumed conversation) could be a subagent rather than the agent's own. The conversation-ids file is now purely the *set* for transcript scoping (its only consumer, `stream_transcript.sh`, reads it `sort -u`), and `capture_conversation_id.sh` is simplified to append each distinct id once.

## Verification (live, agy 1.0.5)

Wired the actual scripts into a probe `hooks.json` and drove agy with a Claude model, watching the marker file over time:

- **Background task:** root emitted interim `fullyIdle:false` then final `fullyIdle:true`; marker held then cleared.
- **Mid-flight message:** sent a user message *while* a 90s background task was running. The marker stayed held (the mid-flight turn's Stop was `fullyIdle:false` because the bg task was still pending), the root conversation never changed, and the marker cleared only after the task finished.
- **Subagent:** live subagent spawns were repeatedly blocked by agy backend capacity ("high traffic"), so this case is proven by a **deterministic replay** of the real captured root+subagent interleaving through the actual scripts (the `test_real_root_subagent_interleaving_only_clears_on_root_done` integration test): the subagent's `fullyIdle:true` leaves the marker present; only the root's clears it.
- **Resume prelude:** verified it reads `root_conversation` and sets `--conversation` (and emits nothing when absent) under both bash and zsh.

## Review gates

- `/verify-architecture` and `/autofix` will be re-run for the subagent/resume changes.


danielmewes:
I think we also need to handle the case where the agent launches subagents. Those will also call the hook with fullyIdle=true when they complete, even though the main agent isn't done yet.

joshalbrecht:
lgtm

PR #1933, opened 2026-06-05
improve top-level mngr tests

evgunter:
Reviewed the test files directly under `libs/mngr/imbue/mngr/` (top-level only; subdirectories are covered by separate tasks) and fixed tests that passed without establishing correctness.

## Changes
- **`test_install.py::test_version_output`** -- removed the branch that accepted both a working `--version` and the (already-fixed) `package_name="mngr"` failure mode; now asserts success unconditionally and drops the stale NOTE. `main.py` already uses `package_name="imbue-mngr"`.
- **`uv_tool_test.py`** -- deleted four tautological `ToolRequirement` constructor tests that only re-tested pydantic field storage; the field semantics are already covered by the command-building tests.
- **`plugin_catalog_test.py`** --
  - deleted `test_catalog_signals_are_signal_check_instances` (type-enforced tautology: `signal: SignalCheck | None` is validated by pydantic);
  - rewrote `test_get_installable_packages_covers_all_published_packages` to pin explicit anchors (a known published package present, every `UNPUBLISHED_PACKAGES` member absent) instead of reconstructing the production filter;
  - strengthened `test_catalog_has_entries` from `len > 0` to asserting the core entry points (`claude`, `opencode`, `tutor`) are present.
- **`primitives_test.py`** -- `test_discovered_agent_create_time_returns_datetime_from_string` now asserts the full parsed `datetime`, not just `.year`.
- **`test_upgrade.py`** -- removed the permanently-skipped `test_preexisting_agent_state_discovered`. It was written against a non-existent on-disk layout; the real disk-discovery behavior (`Host.discover_agents()` reading `$MNGR_HOST_DIR/agents/<id>/data.json`) is already covered by the ten `test_discover_agents_*` tests in `hosts/host_test.py`, and the fresh-install `mngr list` wiring by `test_install.py::test_list`, so a rewrite would only duplicate that.

The pervasive `error.attr == passed_value` assertions in `errors_test.py` were considered and deliberately left: those error classes are hand-written `ClickException` subclasses (not pydantic), so the assertions verify real `__init__` wiring and the public attribute contract that `except` handlers consume.

## Testing
- `uv run ruff check` passes on all changed files.
- Fast unit tests pass: `just test-quick "libs/mngr/imbue/mngr/uv_tool_test.py libs/mngr/imbue/mngr/plugin_catalog_test.py libs/mngr/imbue/mngr/primitives_test.py"` -> 72 passed.
- `test_install.py` / `test_upgrade.py` are `@pytest.mark.release` (isolated-venv subprocess). Release tests do NOT run on PR branches (only on release/main), so they will not be exercised by this PR pipeline. The one release test whose assertion this PR tightened, `test_install.py::test_version_output`, was run locally (`just test ...::test_version_output`) and PASSED in 10.8s; the deleted upgrade test required no execution.


joshalbrecht:

PR #1934, opened 2026-06-05
Add full Anthropic model lineup + inline pricing to LiteLLM proxy

joshalbrecht:
## Summary

The LiteLLM proxy's model list topped out at `claude-opus-4-7`, so agents requesting `opus` (which now resolves to `claude-opus-4-8`) got HTTP 400 `Invalid model name` — this is what broke on staging. This PR:

- Adds the full current Anthropic Claude lineup to the proxy: `claude-opus-4-8` (latest), `claude-opus-4-6`, `claude-opus-4-5`, `claude-opus-4-1`, `claude-sonnet-4-5`, and the bare `claude-haiku-4-5` alias — alongside the existing `claude-opus-4-7`, `claude-sonnet-4-6`, dated Opus 4 / Sonnet 4, and dated Haiku 4.5 ids.
- Registers per-token pricing (input / output / cache-write / cache-read) inline via `litellm_params` for every model, mirrored from litellm's price map. This keeps cost tracking accurate on the pinned litellm 1.83.0, whose bundled price map predates opus-4-7/4-8 (no stable litellm release has opus-4-8 pricing yet — only the 1.88.0 pre-releases), so no litellm upgrade is needed.
- Mirrors the same list in the local-dev `litellm_proxy/config.yaml` and adds `apps/modal_litellm/config_drift_test.py`, which fails CI if the two representations ever diverge.
- Bumps the `modal_litellm` yaml-usage ratchet 5 -> 10 with justification (the drift test parses litellm's externally-mandated YAML config; litellm cannot use TOML).

## Verification

- All 57 `apps/modal_litellm` tests pass (`just test-quick apps/modal_litellm`); `ty check` clean on changed files.
- Offline: a litellm `Router` built from the new model list registers `claude-opus-4-8` in `litellm.model_cost` with the exact inline pricing on litellm 1.83.0.
- Live: redeployed `dev-josh-1` (`minds env deploy`, both apps healthy). `GET /v1/models` now lists all 11 models; `POST /v1/messages` with `claude-opus-4-8` returns 200 (was 400); the resulting spend log shows `$0.00036` for a 22-in / 10-out call — exactly `22 x $5/1M + 10 x $25/1M`, confirming cost tracking end to end.

## Notes

- Staging and production still need their own `minds env deploy` from this code to pick up the new model list.
- Once litellm ships a stable release with opus-4-8 in its bundled price map (>= 1.88.0), the inline pricing block can optionally be dropped.


joshalbrecht:
lgtm

PR #1935, opened 2026-06-05
Remote latchkey

hynek-urban:
(Accompanied by https://github.com/imbue-ai/forever-claude-template/pull/140.)

- When creating remote agents, install Latchkey, sync the credentials and run `latchkey gateway` on the VPS.
- Expose the new gateway instance as a "secondary gateway" to the agents (to be used when user's machine is offline) via an SSH tunnel between the VPS and the agent container.
- Keep syncing changes to permissions and credentials.

There are some things I'll do in follow-ups:

- use a dedicated encryption key and make sure it cannot leak through logs
- make the secondary gateway and the tunnel between the VPS and the container robust to restarts, crashes, ...
- protect the remote gateway with a password
- prevent refresh tokens from being consumed in the secondary gateway


joshalbrecht:
lgtm!

we could use host.docker.internal, but the SSH approach is fine for now

and yes, all of the follow ups make sense to me as well

PR #1943, opened 2026-06-05
add tmux config

evgunter:
not super attached to this but seems like it could be nice

---

## Summary

Adds a nested `[tmux]` configuration section (parsed and merged exactly like `[retry]`/`[logging]`) so tmux users can customize how mngr creates and attaches to agent sessions without resorting to a full `connect_command` override.

Three options:

- **`tmux.attach_args`** — extra tmux *client* flags inserted before the `attach` subcommand when connecting to an agent (`tmux <attach_args> attach ...`). The motivating case is `["-CC"]` for iTerm2 control mode (native tabs/windows); `-u` / `-2` also work. Applied to **both** local attach and the remote SSH wrapper. The remote resize helper's stdout is now also redirected (`>/dev/null 2>&1`) so it can't corrupt the `-CC` control-protocol stream.
- **`tmux.user_config_path`** — an extra tmux config file sourced into every mngr session (guarded by `test -f`). Unlike the auto-generated `~/.mngr/tmux.conf`, this file is never overwritten, so it's a stable place for mngr-session-specific tmux config.
- **`tmux.primary_window_name`** (default `agent`) — mngr now names the agent's primary window and targets it **by name** instead of the literal `:0` index, so targeting is independent of the user's tmux `base-index` setting. This is the idea from the abandoned draft #1155, generalized. Threaded through the start command, `base_agent.tmux_target`, and the provider listing scripts (modal / vps_docker / imbue_cloud).

Also extracts `build_local_attach_argv` as a pure helper so the local attach argv is unit-testable without monkeypatching the exec call.

Docs: new "Customizing mngr's tmux sessions" section in `docs/tmux_users.md`.

## Manual verification of `-CC`

Verified by running mngr's actual code paths against an isolated local tmux server (under a PTY via `script`, with `env -u TMUX` + a dedicated socket so the real server is never touched):

- Local: `tmux -CC attach` enters control mode — emits the iTerm2 control handshake (`\033P1000p`, `%begin`/`%end`, `%session-changed`).
- Remote: mngr's actual generated SSH-wrapper script runs `tmux -CC attach` and its stdout **starts** with the control handshake, uncorrupted by the activity-loop / resize-helper machinery (the stdout-redirect fix holds). SSH `-t` just pipes stdout, so the remote path streams the same clean protocol.

The only thing not verified is iTerm2 *rendering* that protocol as native tabs (no GUI in this environment) — but we emit the correct, uncorrupted control stream, which is the part mngr controls.

## Other notes

- On the **base-index** fix: in tmux 3.6a I could not reproduce the exact "can't find window: 0" failure #1155 described (naming a window doesn't move it off index 0 on a fresh server). The name-based-targeting change is still strictly safer than `:0` and covers any `base-index`/`pane-base-index`/version combination, so it's kept regardless.
- In-flight agents started before this change have an unnamed primary window, so name-based targeting would miss them after an upgrade. `get_lifecycle_state` now **self-heals** this on first contact: when the by-name probe misses but the session exists, mngr renames the session's lowest-index window to `primary_window_name` once and re-probes (remote-safe, idempotent, fires only on the miss path). This is a one-time migration affordance, not part of normal operation. New agents are unaffected.

## Tests

`just test-quick` over the full `libs/mngr` unit + integration suite (excluding tmux/modal/docker/acceptance/release markers): **4092 passed, 1 skipped, 0 failed**. Ratchets and `test_no_type_errors` pass; provider packages type-check clean. Full CI (incl. acceptance/offload) runs on this PR.

(Per repo convention, the interactive `-CC` checks above are not crystallized into pytest — tmux/control-mode behavior is timing-dependent and useless in CI.)


joshalbrecht:
roughly lgtm, though curious for @weishi-imbue and @qi-imbue thoughts here too

qi-imbue:
LGTM but I'll bikeshed that `user_config_path` is better called `additional_config_path` because naming it `user_config_path` makes it sound like it would **replace** the role of ~/.tmux.conf

PR #1947, opened 2026-06-05
Fix two bad tests under apps/minds/scripts

evgunter:
Fixes the two findings from the identify-bad-tests scan of \`apps/minds/scripts\`.

## 1. \`build_test.py\` no longer hard-fails when \`node\` is absent
\`test_bundled_limactl_is_signed_with_virtualization_entitlement\` shells out to \`node\` to evaluate \`todesktop.js\`. It captured \`shutil.which("node")\` as Optional but then \`assert\`ed it non-None, so a runner without node failed with an opaque \`AssertionError\` instead of skipping. Now guarded with \`@pytest.mark.skipif(_NODE_BINARY is None, ...)\`, and a non-zero node exit surfaces \`stderr\` via \`pytest.fail\` instead of hiding it inside a bare \`CalledProcessError\`.

## 2. Renamed \`test_deployments.py\` -> \`run_deployment_tests.py\`
That file is a click CLI orchestrator (run via \`just minds-test-deployment*\`), explicitly "not a pytest wrapper", but its \`test_\` name matched pytest's \`test_*.py\` collection glob -- so the minds suite imported it for zero tests and readers grepping for the deployment-test suite landed on an 835-line script with none. Renamed it and updated the 7 justfile recipes + the \`specs/minds-deployment-tests.md\` reference.

Local checks: \`build_test\` passes, \`apps/minds/scripts\` collection is clean (renamed file no longer collected), minds \`test_ratchets.py\` (55) pass, and the orchestrator CLI still runs (\`--help\`, recipe dry-run).


evgunter:
duplicate

PR #1960, opened 2026-06-05
improve mngr/config/ tests

evgunter:
Fixes test-quality issues identified by `/identify-bad-tests libs/mngr/imbue/mngr/config`. Based on `mngr/review-more-tests`.

## Production / isolation hardening
- `config/plugin_registry.py`: add `reset_plugin_config_registry()`, wired into the autouse `plugin_manager` fixture, so the module-global `_plugin_config_registry` no longer leaks registrations across tests in a worker.
- `get_or_create_user_id`: raise `ConfigParseError` on `MNGR_USER_ID` mismatch instead of a bare `assert` (which `python -O` strips).

## Test-quality fixes
- Removed tautologies (`assert ... or True`, default-equals-default and constructor round-trip micro-tests) and "didn't raise" tests with no assertions.
- Replaced loose `len()>0`/`isinstance(list)` completion-cache assertions with concrete membership checks; the OSError-swallow test now verifies the debug log.
- Added missing branch coverage: set/frozenset and dict-vs-non-dict settings narrowing, the `~/.mngr` completion-cache default, JSON `null`->None, local-disables and lower-layer-survives precedence directions, duplicate/empty plugin-config registration.
- Renamed pytest-collected `Test*` helper classes.
- Shared a typed `read_completion_cache` helper and an `isolated_load_config_pm` fixture via `config/conftest.py`; restored explicit registry resets where the autouse fixture's loaded-agents/orphan state is the thing under test.
- Removed an obsolete `@pytest.mark.flaky` on `test_apply_plugin_overrides_creates_disabled_plugin`: the subprocess-leak mis-attribution it worked around (the `sleep 30` in `concurrency_group_test`) is now fixed upstream (shutdown tests `poll_until` the subprocess is reaped; already in `main`).

Local: 416 config tests + 1238 agents/api/plugins/providers tests pass; ratchets and `ty` clean.


libs/mngr/imbue/mngr/config/conftest.py
```diff
@@ -1,10 +1,53 @@
...
+
+def read_completion_cache(cache_dir: Path) -> CompletionCacheData:
+    """Read the completion cache JSON from ``cache_dir`` as typed CompletionCacheData.
+
+    Unknown keys are dropped so the helper stays robust to additive schema
+    changes in the on-disk format. Shared by the unit and integration completion
+    tests so both read the cache through one typed accessor.
+    """
+    data = json.loads((cache_dir / COMPLETION_CACHE_FILENAME).read_text())
+    return CompletionCacheData(**{k: v for k, v in data.items() if k in CompletionCacheData._fields})
```
evgunter: shouldn't be in conftest + ideally would be real instead of a special function for testing

PR #1973, opened 2026-06-05
improve mngr_vps_docker tests

evgunter:
Addresses the bad-test findings for `libs/mngr_vps_docker`. Test-only changes; no production code touched.

## What changed

- **`_emit_docker_build_output` tests** now capture log output and assert the BUILD-level line is emitted (stripped) for non-empty input and nothing for whitespace-only input, rather than only asserting "does not raise".
- **`_is_retryable_rsync_error`** is now exercised by a single parametrized test in `instance_test.py` covering one representative stderr string for each of the eight retryable connection patterns plus negatives. Removed the duplicated, incomplete copies (and the duplicate `_redact_secret_env` tests) from `_outer_helpers_test.py`.
- **snapshot_helper.service load test** asserts the resource is non-empty and contains expected systemd directives instead of discarding the result.
- **cloud-init generation** is pinned with a full `inline_snapshot` of the rendered user_data, so the load-bearing YAML indentation and key placement are guarded (loose substring checks could pass on malformed output).
- **`list_persisted_agent_data` round-trip test** asserts the read call count does not scale with agent count (2 vs 5) and documents that the call-count assertion deliberately guards the network round-trip budget.
- **Tautological constructor round-trip tests** removed from `host_store_test.py`, `config_test.py`, `primitives_test.py`. The remaining default/wire-value contract tests carry a comment marking them deliberate change-detectors.
- Tightened the `init_methods_in_non_exception_classes` ratchet from 1 to 0 (recorded count was stale).

## Testing

`just test-quick libs/mngr_vps_docker` -> 196 passed, 0 failed. Ratchets pass. cloud-init snapshot verified without xdist.

### DIFF
--- a/libs/mngr_vps_docker/changelog/mngr-bad-tests-mngr-vps-docker-local.md
+++ b/libs/mngr_vps_docker/changelog/mngr-bad-tests-mngr-vps-docker-local.md
@@ -15,7 +15,8 @@ Test-quality cleanup of the mngr_vps_docker unit tests (no production code chang
 - `cloud_init_test.py`: replaced the loose bag-of-substrings generation checks
   with a single full `inline_snapshot` of the rendered user_data, so the
   load-bearing YAML indentation and key placement (the embedded SSH private key
-  in particular) are pinned exactly.
+  in particular) are pinned exactly, plus a companion test that parses the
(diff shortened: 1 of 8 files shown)


joshalbrecht:
lgtm

PR #1977, opened 2026-06-05
improve remote_service_connector tests

evgunter:
the commit [deb192d](https://github.com/imbue-ai/mngr/pull/1977/commits/deb192dd2696ea8e0753b80f8532d3e3381e800a) only moves things except one change flagged by the comment below. most of the real changes are in the other commits

---

Acts on the `/identify-bad-tests` findings for `apps/remote_service_connector`.

## Strengthened weak tests (findings #1-#6)
- `FakeCloudflareOps` now persists created Access service tokens, so the create/list/delete round-trip is exercised for real. The service-token round-trip tests assert the created token is surfaced by the listing (secret omitted) instead of asserting an always-empty list (which was hard-coded into the fake).
- Migration schema-drift guards match column names on word boundaries, so a dropped column (e.g. the bare `access` column, previously masked by the `access_key_id` substring) now fails the test.
- Split the mislabeled `make_tunnel_name` truncation test into a prefix-strip test plus a new test that actually exercises truncation to 16 characters.
- The KV "create namespace when missing" test asserts the create POST fired exactly once.
- `derive_s3_secret_access_key` is pinned against a literal SHA-256 digest instead of recomputing the production algorithm in the assertion.
- Documented the intentional "returns without raising" contract for two no-positive-observable tests.

## Reorganized test layout (finding #7)
- Moved the `TestClient`-driven end-to-end tests (routes, auth, host-lease, paid-list CRUD, bucket endpoints) out of the unit `app_test.py` into a new integration file `test_remote_service_connector_routes.py`. Pure-function and single-layer (ctx, http-ops, supertokens, paid-gate function) unit tests stay in `app_test.py`.
- The two cleanup-sweep tests now build their fake backend directly via `make_fake_pool_backend()` rather than through the route test client, keeping them self-contained unit tests.
- The split is a pure code move (plus that small refactor): the union of collected tests across both files is identical to before (203 items, verified disjoint).


apps/remote_service_connector/imbue/remote_service_connector/app_test.py
```diff
@@ -1898,13 +914,14 @@ def call(self, method: str, path: str, data: object, need_auth: bool) -> object:
...
     ops.set_delete_at_expiration("vps-gone.vps.ovh.us", True)
 
 
-def test_cleanup_sweep_keeps_row_when_ovh_cancel_fails(monkeypatch: pytest.MonkeyPatch) -> None:
+def test_cleanup_sweep_keeps_row_when_ovh_cancel_fails() -> None:
     """A removing row whose OVH cancel fails is kept for the next run."""
-    _client, backend = _make_pool_test_client(monkeypatch)
+    backend = make_fake_pool_backend()
     backend.ovh_ops.fail_on_cancel = True
     backend.add_removing_host(host_id=UUID("00000000-0000-0000-0000-0000000000b1"), version="v0.1.0")
```
evgunter, line 926: actual change

apps/remote_service_connector/imbue/remote_service_connector/app_test.py
```diff
@@ -1568,10 +1583,10 @@ def test_route_create_and_list_service_tokens(monkeypatch: pytest.MonkeyPatch) -
...
     assert resp.status_code == 200
     listed = resp.json()
-    # FakeCloudflareOps.list_service_tokens returns an empty list by design (it
-    # doesn't persist created tokens), so the listing is empty -- the test
-    # still covers the endpoint + ForwardingCtx.list_service_tokens path.
-    assert listed == []
+    # The created token is surfaced by the listing; the secret is only returned
+    # at creation time, never on a later listing.
+    assert [t["name"] for t in listed] == ["my-token"]
+    assert listed[0]["client_secret"] is None
```
evgunter: enabled by changes in apps/remote_service_connector/imbue/remote_service_connector/testing.py

joshalbrecht:
I think hold off on this one--I have some other in-flight changes that I would rather not conflict with this

once they're merged, worst case we just re-run this one

PR #1980, opened 2026-06-05
improve mngr_robinhood tests

evgunter:
Addresses an `/identify-bad-tests` audit of `libs/mngr_uncapped_claude`. All changes are test-only (no production behavior change); each closes a gap where a real regression would have passed CI.

## What changed
- **`_build_pass_env_vars`**: now tested for its actual purpose -- forwarding the parent env while dropping the per-agent `MNGR_*` / `LLM_USER_PATH` vars -- instead of merely asserting non-empty.
- **`output_modes`**: added coverage for the `tool_result` stream-json branch (both `is_error` values), assistant `tool_use` content blocks, and `_parse_input_preview` (empty / valid-JSON / unparseable). The user-event test now asserts the full converted message, not just the envelope type.
- **`raw_transcript`**: added coverage for tool-use input-preview rendering + truncation, usage conversion (incl. empty-usage -> None), tool_result output truncation, list-form content flattening, and mixed text + tool_result user messages.
- **`arg_partition`**: `test_rejected_flags_raise` now asserts the per-flag rejection reason; added a test for the inline `--flag=value` form.
- **`orchestrator`**: strengthened the `monotonic_ms_since` test to guard the ms scaling; documented the deliberate timing margin in the polling-ticker test.
- **New `test_cli.py`**: first end-to-end coverage of the CLI's click wiring and exit-code contract via the shared `cli_runner` fixture -- asserts the command is registered, each rejected flag exits 2, and a bad `--output-format` exits 2. Covers the no-spawn failure paths without a live claude agent.

A full live acceptance test of a real turn was intentionally deferred (the lighter integration-test level was chosen for finding #5).

## Testing
`just test-quick libs/mngr_uncapped_claude` -- 164 passed, 0 failed (includes `test_no_type_errors` and `test_ratchets`).

### DIFF
--- a/libs/mngr_robinhood/changelog/mngr-bad-tests-mngr-uncapped-claude-local.md
+++ b/libs/mngr_robinhood/changelog/mngr-bad-tests-mngr-uncapped-claude-local.md
@@ -15,3 +15,7 @@ changes close gaps where a real regression would have slipped past CI:
   for the inline `--flag=value` form of pass-through claude value flags.
 - Strengthened the `monotonic_ms_since` test to guard the milliseconds scaling, and documented the
   deliberate timing margin in the polling-ticker test.
+- Added a lightweight integration test (`test_cli.py`) that drives the real `mngr robinhood`
+  command through the top-level CLI for the no-spawn failure paths: it asserts the command is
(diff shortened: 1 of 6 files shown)


joshalbrecht:
lgtm

PR #1986, opened 2026-06-05
Revive & strengthen Modal e2e tests; fix first-create Modal bootstrap bug

evgunter:
## Summary

Started as an `identify-bad-tests` pass over `libs/mngr/imbue/mngr/e2e/`, and uncovered (and fixed) a real product bug in the Modal create path along the way.

### Product fix
- **First `mngr create --provider modal` in a fresh Modal environment failed to bootstrap the environment.** `api/create.py::create()` resolved the new-host provider via `get_provider_instance(...)` without `is_for_host_creation=True`, so the Modal backend raised `ProviderEmptyError` ("Modal environment ... does not exist yet") before `resolve_target_host` (which *does* pass the flag) could bootstrap it. Worse, `get_provider_instance` caches by `(name, ctx)`, so the non-bootstrap instance would poison the cache. Fixed by passing `is_for_host_creation=True`, matching the sibling call just below.

### Bad-test findings fixed
1. Deleted `test_basic.py` — its 10 tests were each an equal-or-weaker duplicate of a per-feature e2e test. Verified via the tutorial matcher that no tutorial blocks are orphaned.
2. Deleted `test_create_modal_retry` — a silent char-for-char duplicate of `test_create_provider_modal` (its tutorial block is comment-only).
3. Tightened two over-loose error regexes (`test_create_with_template_modal_disabled`, `test_create_address_syntax_existing_host`); the latter also never actually reached the host-resolution path it claimed to test (it died at the agent-type gate).
4. Documented the intentional overlap among the three `--transfer=none` in-place tests.

### Modal e2e suite revival
The whole `test_create_modal.py` file had silently rotted: every create omitted `--type`, so each died at the "No agent type provided" gate before reaching Modal (the offload release suite has not run in CI in 3.5 months, so nobody noticed). Gave each create an explicit `--type` (`command`+`sleep` for generic tests, `claude` for the three claude-specific ones), strengthened the previously success-only tests to verify the feature took effect (work_dir / `mngr exec` probes for volume, host env, image), set `EDITOR=true` so `--edit-message` doesn't hang, and added `@pytest.mark.flaky` + larger timeouts for slow Modal cold-starts / claude installs.

## Testing
- Validated **16 of 17** modal tests against **real Modal** locally (sequential; `-n>1` trips SSH-banner races). The 17th (`no_connect_message`) can't run locally without `ANTHROPIC_API_KEY`.
- Findings #4/#5 (failure-path) validated locally.
- `/autofix` clean (no issues found).


evgunter:
duplicate

PR #1994, opened 2026-06-06
improve edge-case handling in top-level minds

evgunter:
Acts on the `identify-suspicious-edge-cases` findings for the top-level `apps/minds/imbue/minds` modules (subdirectories are handled by their own queue tasks). Each finding is its own commit.

## Findings addressed

1. **`_ensure_mngr_settings` crashed the import-time bootstrap on a corrupt `config.toml`.** It re-parsed `config.toml` with an unguarded `tomllib.loads`, while `read_active_profile_dir` reads the same file inside a try/except. Since `_ensure_mngr_settings` runs in `apply_bootstrap` as an import-time side effect on every `minds` invocation, a corrupt config raised a raw `TOMLDecodeError` before the CLI parsed args. Now routes through the guarded `_resolve_active_settings_path`.

2. **`reconcile_imbue_cloud_providers_from_sessions` silently skipped a malformed `accounts.json`.** A decoded-but-wrong-shape index or malformed entry was dropped with no log, which would silently reconcile nothing and leave the user unable to `mngr create` — the exact drift the function exists to repair. Now logs a warning (mirroring the plugin's own loader) before skipping.

3. **`is_imbue_cloud_provider_enabled_for_account` crashed the UI on a corrupt `settings.toml`.** Every other branch biases toward the documented "enabled" default, but the TOML parse was unguarded. Now falls back to the same default (with a warning).

4. **`cli_entry`'s `ctx.invoked_subcommand or "unknown"` was dead code.** The group has no `invoke_without_command=True`, so the callback only runs once click resolves a subcommand. Replaced the fallback with an assertion.

5. **`resolve_minds_root_name` production-fallback blast radius (comment only).** A set-but-invalid `MINDS_ROOT_NAME` deliberately resolves to the production data dir; added a comment documenting this and the stricter gate destructive callers use. No behavior change.

## Tests

New unit tests cover the reconcile happy-path / wrong-shape / malformed-entry cases and the provider-enabled disabled-block / malformed-TOML / unreadable-path fallbacks. `bootstrap_test.py` (42), `main_test.py`, `test_ratchets.py`, and `test_claude_version_alignment.py` pass locally. The 11 failures in the broader local run are pre-existing `desktop_client/` environment issues (missing `restic` binary, macOS tmp symlink 404s), unrelated to these changes.


apps/minds/imbue/minds/bootstrap.py
```diff
@@ -74,6 +74,15 @@ def resolve_minds_root_name() -> str:
...
         )
+        # Intentional: a set-but-invalid value (including a typo'd dev/ci env
+        # name) falls back to production (~/.minds/) rather than raising. This
+        # keeps every caller pointed at *some* consistent host_dir, but it does
+        # mean a mistyped env name silently resolves to the production data dir.
+        # That blast radius is accepted because the warning above surfaces it and
+        # because callers for whom landing on production would be destructive
+        # (`minds env deploy/destroy`, `minds run`) gate on the stricter
+        # is_minds_root_name_set_to_active_env (which treats invalid as "not
+        # activated") instead of trusting this fallback.
```
evgunter: not sure if this justification is correct

apps/minds/imbue/minds/bootstrap.py
```diff
@@ -586,7 +601,20 @@ def is_imbue_cloud_provider_enabled_for_account(email: str, *, root_name: str |
...
+    try:
+        raw_settings = settings_path.read_text()
+    except OSError as e:
+        logger.warning("Could not read mngr settings {} for provider-enabled check: {}", settings_path, e)
+        return True
+    try:
+        parsed = tomllib.loads(raw_settings)
+    except tomllib.TOMLDecodeError as e:
+        logger.warning("Malformed mngr settings {} for provider-enabled check: {}", settings_path, e)
+        return True
```
evgunter: also unsure about this

apps/minds/imbue/minds/cli_entry.py
```diff
@@ -30,7 +30,13 @@
...
     """minds: run and manage your own persistent, specialized AI agents."""
     console_level = console_level_from_verbose_and_quiet(verbose, quiet)
-    command_name = ctx.invoked_subcommand or "unknown"
+    # This group has no ``invoke_without_command=True``, so click only runs this
+    # callback once it has resolved a subcommand (a bare ``minds`` with no
+    # subcommand prints help and exits before reaching here). The old
+    # ``or "unknown"`` fallback was therefore dead code that would silently
+    # mislabel logs if the invariant ever broke; assert it instead.
+    assert ctx.invoked_subcommand is not None, "cli group callback ran without a resolved subcommand"
+    command_name = ctx.invoked_subcommand
```
evgunter, line 38: huh

apps/minds/imbue/minds/bootstrap.py
```diff
@@ -74,6 +74,11 @@ def resolve_minds_root_name() -> str:
             MINDS_ROOT_NAME_PATTERN,
             DEFAULT_MINDS_ROOT_NAME,
         )
+        # Intentional: a set-but-invalid value (e.g. a typo'd env name) falls
+        # back to production rather than raising, so every caller has *some*
+        # consistent host_dir. Callers for whom landing on production would be
+        # destructive (`minds env deploy/destroy`, `minds run`) instead gate on
+        # the stricter is_minds_root_name_set_to_active_env.
```
evgunter: not sure if this justification is correct

apps/minds/imbue/minds/bootstrap.py
```diff
@@ -606,7 +610,18 @@ def is_imbue_cloud_provider_enabled_for_account(email: str, *, root_name: str |
...
+    try:
+        raw_settings = settings_path.read_text()
+    except OSError as e:
+        logger.warning("Could not read mngr settings {} for provider-enabled check: {}", settings_path, e)
+        return True
+    try:
+        parsed = tomllib.loads(raw_settings)
+    except tomllib.TOMLDecodeError as e:
+        logger.warning("Malformed mngr settings {} for provider-enabled check: {}", settings_path, e)
+        return True
```
evgunter: also unsure about this

apps/minds/imbue/minds/bootstrap.py
```diff
@@ -646,7 +646,19 @@ def list_disabled_provider_names(*, root_name: str | None = None) -> list[str]:
     settings_path = _resolve_active_settings_path(root_name)
     if settings_path is None or not settings_path.exists():
         return []
-    parsed = tomllib.loads(settings_path.read_text())
+    # Like the other branches here, a read/parse failure degrades to "nothing
+    # disabled" rather than crashing this passive providers-panel enumeration on
+    # a corrupt settings.toml.
```
evgunter: same concern

joshalbrecht:
these changes mostly seem bad -- I'm tempted to close unless you feel like there's something valuable in here that we should pull out?

evgunter:
there's some added warnings and some DRY cleanup; i commented on everything i thought was suspicious, which was indeed like half of the changes, but i can just drop those ones?

apps/minds/imbue/minds/bootstrap.py
```diff
@@ -173,18 +173,12 @@ def _ensure_mngr_settings(root_name: str) -> None:
...
-    if not profile_id:
-        return
-    settings_dir = mngr_host_dir / "profiles" / profile_id
-    if not settings_dir.exists():
+    # Use the guarded helper rather than re-parsing config.toml inline, so a
+    # corrupt config.toml degrades to None instead of raising out of the
+    # import-time bootstrap.
+    settings_path = _resolve_active_settings_path(root_name)
+    if settings_path is None:
         return
```
evgunter, line 181: apparently this behavior is more consistent with the rest of the bootstrappnig

PR #2001, opened 2026-06-06
Harden suspicious edge-case handling in mngr_file file command

evgunter:
## Summary

Hardens suspicious edge-case handling in the `mngr file` command internals (`libs/mngr_file`), following the `/identify-suspicious-edge-cases` review. Each finding is its own commit.

1. **Narrow host-availability catch** (`target.py` `_get_host_access`): caught the root `MngrError` when probing host availability, silently degrading to volume access at `trace` level on *any* library error (auth/config/provider bugs). Now catches only `HostNotFoundError` and logs the fallback at `warning`; everything else propagates.
2. **Make `base_path` optional instead of fabricating offline sentinels** (`target.py` + consumers): `/mngr-host-dir` and `/unknown` were synthesized for offline targets but only ever read when online. `ResolveFileTargetResult.base_path` is now `Path | None` (None offline), computed only in the online branch; a new `host_base_path` property (mirroring `host`) is used by the get/put/list consumers in their `is_online` branches. The `/unknown` `work_dir` ternary is dropped (WORK already guards work_dir). Offline use of a base path now surfaces an explicit error rather than a plausible-but-wrong path.
3. **Raise on unknown list field** (`list.py` `_get_field_value`): the `case _: return ""` default would silently render a blank column if a new validated field lacked a formatter. Now raises `MngrError`; test updated.
4. **Log malformed list lines at warning** (`list.py`): subprocess output that doesn't split into 6 fields was dropped at `trace` (effectively silent); raised to `warning` per the style guide.
5. **Log unparseable size** (`list.py`): a non-integer size was silently coerced to `None`; keep the fallback but log at `warning`.
6. **Document un-`lstat`-able skip** (`_LIST_SCRIPT`): the embedded script silently skips entries it cannot `lstat`; documented the behavior, including that permission-denied entries are silently omitted.

## Testing

`just test-quick "libs/mngr_file -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'"` — 129 passed, 0 failed (includes ratchets and the type-error check). The warning branches for #4 and #5 are covered by existing `parse_list_output` tests; new unit tests cover both `host_base_path` branches.


evgunter:
main changed enough that this is basically no longer relevant

PR #2010, opened 2026-06-06
Finish the mngr-backed Agent SDK

joshalbrecht:
Completes the previously-stubbed control surfaces of the mngr-backed Agent SDK
(`imbue.mngr_robinhood.agent_sdk`) per `blueprint/finish-agent-sdk/plan-finish-agent-sdk.md`,
so it is a faithful drop-in for `claude_agent_sdk` on the existing interactive-tmux + session-JSONL transport.

## Implemented (and unskipped for the mngr target)
- **`can_use_tool` + `hooks`** — a local HTTP bridge in the SDK process; the agent is launched with a `--settings` file whose PreToolUse/PostToolUse/UserPromptSubmit hook commands POST each event to the bridge, which runs the in-process Python callback (on an anyio portal) and returns claude's hook JSON. Permission allow/deny/`updated_input` all work; denials surface in `ResultMessage.permission_denials`. The `--settings` hooks merge with mngr's readiness hooks rather than replacing them.
- **`interrupt()`** — turn messages are now streamed incrementally, so `interrupt()` (called between yields) stops the agent mid-turn; the drain observes the death and emits the terminal `ResultMessage`. The next `query()` restarts-with-resume.
- **`set_model` / `set_permission_mode`** — restart the agent on its resumed session under the new options (previously a no-op).
- **`fork_session`** — adopts the resumed session into a fresh agent launched with `--fork-session` (new session id), delivering the turn only after the re-mirrored adopted history is skipped.
- **`get_server_info()`** — real commands / output style from a one-shot `claude` stream-json `system/init` probe, cached per session.
- **`total_cost_usd`** — computed from per-turn token usage × a per-model price table.

## Not implemented (documented, real-SDK-only)
- Partial-message streaming (`include_partial_messages` -> `StreamEvent`) and the real `system/init` tools list (kept as the documented built-in set). Hermeticity stays as-is (real host settings are needed for auth).

## Testing
- Offline: full `mngr_robinhood` suite passes (233 passed, incl. ratchets and type checks); added offline unit + localhost-HTTP integration coverage for the bridge, pricing, server-info probe, and option overrides (coverage 68%, above the 65% gate).
- Live (mngr target, real API, haiku): verified the can_use_tool allow/deny/updated_input + hook (pre/post/user-prompt/deny-decision) cases, `interrupt`, `fork_session`, `set_model`, `set_permission_mode`, `get_server_info`, and `total_cost_usd`. The full both-target `just test-sdk-live` was not run in one pass (paid); a CI run still excludes the `sdk_live` suite.


joshalbrecht:
merged as part of https://github.com/imbue-ai/mngr/pull/2018

PR #2014, opened 2026-06-07
Consolidate host setup (single source of truth) + fix imbue_cloud slow-path runsc

joshalbrecht:
## Summary

Consolidates host-level provisioning into a single source of truth so the imbue_cloud slow (rebuild) path re-applies all idempotent host setup, with a pinned Docker version, and fixes the deferred slow-path runsc gap.

- New `mngr_vps_docker/host_setup.py`: ordered, idempotent, config-gated steps (pinned Docker install, optional gVisor `runsc`, sshd tuning, base packages, optional qemu purge) + `apply_host_setup_on_outer()` SSH runner. `cloud_init.py` renders its first-boot `runcmd` from the same steps.
- Docker pinned to `29.5.1` (Debian 12) via the official apt repo with `--allow-downgrades`; `runsc` pinned to a dated gVisor release (download + checksum). SSH host-key injection stays first-boot-only and is excluded from the re-runnable steps.
- OVH `_provision_vps` now applies the shared host setup over SSH (OVH has no cloud-init). This closes a real gap: OVH never installed `runsc`, so `[providers.ovh] install_gvisor_runtime=true` was a silent no-op. The OVH `install_required_outer_packages` / `purge_qemu_packages` helpers are removed (folded into the shared list).
- imbue_cloud slow path re-applies host setup on the leased VPS after teardown and before rebuild (fatal on failure); `ImbueCloudProviderConfig` now extends `VpsDockerProviderConfig` and forwards `docker_runtime` / `install_gvisor_runtime` / `default_start_args` onto the delegated provider. minds bootstrap writes the runsc + hardening values into the per-account `[providers.imbue_cloud_<slug>]` block.

Plan: `blueprint/consolidate-host-setup/plan-consolidate-host-setup.md`. Supersedes and deletes `specs/imbue-cloud-slow-path-runsc.md`.

## Rollout note

Re-bake the pool so fast-path hosts gain pinned Docker + runsc (the fast path does no host setup); the slow path self-heals everything else.

## Follow-ups / to confirm

- The exact pinned strings (`docker-ce=5:29.5.1-1~debian.12~bookworm`, gVisor release `20260602`) must be confirmed against the live apt/gVisor repos before deploy.
- End-to-end runsc verification (`docker inspect --format '{{.HostConfig.Runtime}}'` -> `runsc`) is tracked separately.

## Test

- Unit: `host_setup_test.py` (step list gating/order, pinned versions, SSH-host-key exclusion, base64 wrapping, runner success/failure), updated `cloud_init_test.py`, `instance_test.py` (`_build_delegated_vps_config` forwarding), `bootstrap_test.py`. `ty` + ruff clean; affected-package ratchets pass.


joshalbrecht:
merged as part of https://github.com/imbue-ai/mngr/pull/2032

PR #2017, opened 2026-06-07
Fix duplicated paragraphs in robinhood live streaming after horizontal rules

joshalbrecht:
## Problem

`uv run mngr robinhood --stream-plain-text -p "..."` (and `--include-partial-messages`) sometimes printed duplicated paragraphs in the live stream. The duplicates consistently appeared immediately after a markdown horizontal rule (`---`).

## Root cause

The live stream is reconstructed by polling the spawned agent's `stream_buffer` (a reverse-mapped capture of the tmux pane) and computing an incremental delta in `compute_stream_delta`. The buffer body is assumed to be strict-append, but Claude's TUI **reflows already-rendered text** as later text streams in — most visibly it collapses the blank line around a horizontal rule when the following paragraph arrives.

When that happens the new body is no longer a clean prefix-extension of what was already emitted, so the function fell into its divergence branch, which emitted everything past the **character-level** longest common prefix. The reflow moves that divergence point *before* already-printed content, and since plain-text stdout cannot be unprinted, the already-emitted region was printed a second time.

## Fix

The divergence branch now uses a whitespace-aware scan (`_unemitted_suffix_start`) that treats whitespace runs as equivalent, so a reflowed-but-already-printed region is recognized as already emitted and only genuinely new content is emitted. At worst a little already-printed whitespace (a collapsed blank line) is left stale; no visible content is duplicated.

## Tests

- New unit tests in `stream_buffer_test.py` covering the blank-line-collapse-after-already-emitted-paragraph case, the reflow-plus-new-tail case, and a full poll-sequence that asserts the paragraph is emitted exactly once.
- Updated `orchestrator_test.py`'s reflow test (whitespace at a reflow boundary is now collapsed rather than re-emitted; the cross-turn prefix-preservation property is unchanged).
- Full local `mngr_robinhood` suite: 272 passed; ratchets: 56 passed. (Full offload suite runs in CI — Modal creds unavailable locally.)

### DIFF
--- a/blueprint/tmux-window-size/plan-tmux-window-size.md
+++ b/blueprint/tmux-window-size/plan-tmux-window-size.md
@@ -1,0 +1,44 @@
+# Plan: configurable tmux window size for agents (and `mngr robinhood`)
+
+## Overview
+
+- Add per-session control over the inner agent's tmux window **width**, **height**, and **resize behavior**, set once at session creation and never affecting other tmux sessions/windows.
+- Motivation: `mngr robinhood`'s approximate live streaming reverse-maps the agent's tmux pane, so the agent's pane width is baked into the streamed text as hard line wraps. A wide, fixed pane removes most of that wrapping. This complements (does not replace) the streaming dedup fix already on this branch.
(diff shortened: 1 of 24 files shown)


joshalbrecht:
merged as part of: https://github.com/imbue-ai/mngr/pull/2018

PR #2021, opened 2026-06-08
Bump Latchkey + implement a few related fixes.

hynek-urban:
- Bumping the Latchkey version to 2.15.4 to:
  - support permission checks based on request bodies
  - fix Notion integration
  - allow access to well-known private Slack API methods (e.g. `subscriptions.thread.mark`)
  - hopefully fix at least some of the issues with Google Cloud Project creation that Kanjun observed
- Do not automatically deny requests in case of failures
- Do not automatically refresh tokens in remote Latchkey instances


joshalbrecht:
lgtm!

PR #2022, opened 2026-06-08
no dialog when using enterprise authentication in antigravity

evgunter:
## Summary

`build_onboarding_seed` hardcoded `enterpriseOnboardingComplete: False`. That seed exists solely to suppress agy's first-run NUX, but agy runs a *separate* onboarding flow for enterprise accounts gated on that flag. So consumer-authenticated users got the NUX skipped, while enterprise-authenticated users were still intercepted by the enterprise onboarding flow on their first message.

This marks all three completion flags `True` so the seed suppresses the NUX regardless of how the user authenticated. Seeding the flag for an account type the user doesn't have is inert — the file only records that the first-run flow has been seen.

## Changes

- `antigravity_config.py`: `enterpriseOnboardingComplete` now `True`; docstring explains why.
- Updated assertions in `antigravity_config_test.py` and `plugin_test.py`.
- Added changelog entry.

## Verification

Ran the affected unit tests: `antigravity_config_test.py` (22 passed) and `plugin_test.py::test_provision_writes_onboarding_seed` (1 passed).

Note: the fix rests on the reported behavior that agy's enterprise onboarding flow keys on `enterpriseOnboardingComplete`; not live-tested against an actual enterprise account.


joshalbrecht:
lgtm

PR #2025, opened 2026-06-08
Final minds workspace-create fixes (lima runsc/docker, backups, region, destroyed hosts)

joshalbrecht:
Final cleanup pass on the standardized minds workspace-provider stack. Plan: \`blueprint/finalize-minds-provider-fixes/plan-finalize-minds-provider-fixes.md\`.

## Changes

1. **Lima drops runsc** (config-only, in the forever-claude-template worktree): the lima provider now runs the agent container under stock \`runc\` instead of gVisor. The Lima VM already provides a hardware-virtualization boundary, so runsc only added download size + provisioning complexity. docker/vultr/ovh still use runsc. (This edit lives in the FCT clone under \`.external_worktrees/\`, committed on the same branch there; it is not part of this monorepo PR's diff.)

2. **Lima pins docker** (\`libs/mngr_lima\`): installs a pinned \`docker-ce\` (\`5:29.5.1-1~debian.12~bookworm\`) from Docker's official apt repo, the same way/version as the remote VPS providers, instead of Debian's unpinned \`docker.io\`.

3. **Robust, non-blocking backups** (\`apps/minds\`): restic backup setup runs after workspace readiness, bounds each \`mngr exec\` to 60s, retries the whole operation for ~5 minutes, and only notifies on final failure. Never blocks create; litellm key path and \`mngr create\` unchanged.

4. **Explicit per-provider region** (\`apps/minds\`, \`libs/mngr_imbue_cloud\`, \`apps/remote_service_connector\`): removed the soft \"preferred region\" everywhere. The create form now shows an explicit region dropdown (advanced settings) for imbue_cloud + vultr, defaulting to last-used (per-provider in \`~/.minds/config.toml\`) → background geo guess → hardcoded default. Passed as \`-b region=\` (imbue_cloud, hard) / \`-b --vps-region=\` (vultr).

5. **Hide DESTROYED hosts** (\`apps/minds\`): \`host_state\` is threaded from discovery (full snapshot + delta events) into the resolver; destroyed workspaces drop off every active surface and destroy no longer reports a false \"failed\".

## Testing

Per-project unit/integration suites run green locally (minds desktop_client, mngr_lima, mngr_imbue_cloud, remote_service_connector). Full offload run + CI pending.


joshalbrecht:
merged as part of https://github.com/imbue-ai/mngr/pull/2032

PR #2026, opened 2026-06-08
Fix volume data loss; fail loudly and scope discovery when a provider is unavailable

gnguralnick:
## Problem

A Docker host whose container was stopped could no longer be restarted, failing with:

> Start step of host restart failed: ... cannot access path .../volumes/vol-<hex>: no such file or directory

Root cause (traced via on-disk state + production logs): when the Docker daemon was briefly unavailable during a GC-running command (a daemon restart), the Docker provider's `discover_hosts` swallowed the failure and returned `[]`. GC's `_discover_hosts_for_gc` only skips a provider whose discovery *raises*, so it recorded the provider with an empty host list. `gc_volumes` then treated **every** volume as orphaned and deleted the per-host `host_dir` directories, so the host's isolated `volume-subpath=volumes/vol-<hex>` mount could no longer bind and the container wouldn't start.

## Fix

**1. Root cause — Docker raises instead of returning `[]`.** `discover_hosts` / `_list_containers` now raise `ProviderUnavailableError` when the daemon is unreachable. An empty host list now unambiguously means "this provider has zero hosts", never "the lookup failed", so GC (which skips a provider whose discovery raises, at its own `_discover_hosts_for_gc` boundary) never runs `gc_volumes` against a false-empty list.

**2. Discovery fails loudly instead of silently skipping.** Following review, multi-provider discovery is now "dumb": `_run_discovery` no longer swallows `ProviderUnavailableError`, so it propagates out of `discover_hosts_and_agents`. A provider that might hold a target must not be silently dropped — e.g. `mngr message my-agent`, meant to reach every instance, must not quietly miss one on a down provider. Enumerate-all commands (`message`, `limit`, `snapshot`, `create`) now fail with a clear "provider is not available" error rather than doing a silent partial job. `snapshot._discover_all_hosts` also stops swallowing provider errors into `[]` for the same reason.

**3. Targeted commands scope discovery so an *unrelated* down provider can't fail them.** `mngr rsync`, `mngr git push`/`pull`, and `mngr event <host>` now resolve only the provider(s) that could hold the target — via the `.PROVIDER` qualifier and/or the agent name (the latter resolved through the discovery event stream) — instead of a blind multi-provider scan. So `mngr rsync ./x agent@host.local` keeps working when an unrelated Docker daemon is down (Docker is never queried), while a command whose target really is on the down provider fails with a clear error.

`gc` and `list` are unchanged — both already isolate per-provider failures at their own boundaries (`gc` skips a provider it can't reach; `list` reports it as unavailable).

## Tests
- `instance_test.py`: daemon-offline unit tests assert `discover_hosts` / `discover_hosts_and_agents` raise `ProviderUnavailableError`.
- `test_gc_docker.py`: assert `_discover_hosts_for_gc` *skips* an offline Docker provider (the data-loss-prevention path).
- `list_test.py`: the discovery orchestrator test asserts `discover_hosts_and_agents` *propagates* `ProviderUnavailableError` for an enumerate-all call.
- Local: relevant unit tests for the touched modules pass; `ruff` and `ty` clean. The full suite (incl. `docker_sdk`/acceptance) runs in CI.

## Note
This branch is behind `origin/main`; review against `origin/main` (where the diff is the docker-restart files only). CI runs the full suite.


libs/mngr/imbue/mngr/api/discover.py
```diff
@@ -74,6 +75,17 @@ def _discover_provider_hosts_and_agents(
...
+    except ProviderUnavailableError as exc:
+        # A provider whose backend is unreachable must not block discovery for
+        # every other provider. Per the ProviderUnavailableError contract,
+        # multi-provider commands continue with the providers that *are*
+        # available, so e.g. `mngr rsync <local-agent>` still works when an
+        # unrelated Docker daemon happens to be down. Skip this provider and
+        # contribute no hosts -- the same outcome callers saw before the Docker
+        # provider was changed to raise (it previously returned []), and
+        # consistent with how `mngr list` isolates per-provider failures.
+        logger.warning("Skipping provider {} during discovery (unavailable): {}", provider.name, exc)
```
gnguralnick: this should raise? why swallow here

gnguralnick: the higher-level operation calling this one should deal with this. at the level of the particular provider this is still an error

joshalbrecht:
this looks much better!

I fixed the merge conflicts, lgtm assuming the tests pass

joshalbrecht:
sigh, and ofc the tests didn't pass, fixing in a separate branch here: https://github.com/imbue-ai/mngr/pull/2133

PR #2027, opened 2026-06-08
Shut down local minds on app quit + landing Start/Stop controls

gnguralnick:
## Summary

Adds on-quit shutdown of running minds whose host can be stopped/started, plus per-row Start/Stop controls on the landing page for the `apps/minds` desktop app, with container liveness sourced from the existing discovery snapshot rather than a dedicated poll.

Which providers expose host shutdown is gated by a **single predicate** — `provider_backend_supports_shutdown` (currently the local `docker` / `lima` backends) — so the rest of the machinery is provider-agnostic and ready to widen when remote providers gain host shutdown support. The names reflect this (`mind_liveness.py`, `MindLiveness`, `compute_mind_liveness_by_agent_id`, `/api/minds/*`); the "local-only, for now" restriction lives only in that one predicate.

- **Quit-time shutdown prompt** (`electron/main.js`, `content-relay-preload.js`, `shell.html`): when quitting with shutdown-capable minds still running, a native prompt offers **Cancel** / **Leave running** / **Shut down all**. "Shut down all" stops every running mind via a single synchronous `POST /api/minds/stop-hosts` (which runs one `mngr stop <ids…> --stop-host` — mngr stops the hosts concurrently via its own executor), shows progress on an in-app "quitting" page, and offers **Retry** for any that didn't stop, then stops this env's mngr docker state container so nothing minds-related is left running. Cancelling leaves the app fully intact. If the running-minds check itself fails, the user is asked **Quit anyway / Cancel** rather than silently quitting. Programmatic shutdowns (SIGTERM/SIGINT) skip the prompt.
- **Landing page Start/Stop** (`templates.py`, `Landing.jinja`): each shutdown-capable mind shows a Running / Stopped / Unknown badge and a per-state control — Stop (with confirmation) when running, Start when stopped; other minds keep the existing Restart button. The "server not responding" health badge is suppressed while a container is stopped.
- **Liveness from discovery, folded into the workspace list** (`mind_liveness.py`, `backend_resolver.py`, `app.py`): container status is read straight from the global discovery snapshot's `host_state` (the same state `mngr list` discovery already tracks for every mind), classified to RUNNING/STOPPED/UNKNOWN — there is no second `mngr list` loop. It rides each workspace entry in the existing landing SSE `workspaces` payload (`supports_shutdown` + `liveness`) rather than a separate event channel. A user-initiated Start/Stop sets a short-lived (90s TTL) **optimistic host-state override** on `MngrCliBackendResolver`, so the badge and quit prompt flip immediately; the next discovery snapshot confirms (or corrects) it. The landing JS reconciles optimistic `STOPPING`/`STARTING` transients against the authoritative payload. An externally-driven stop/start is reflected on the next snapshot.
- **Backend endpoints** (`app.py`, `cli/run.py`): `POST /api/agents/{id}/stop-host` and `POST /api/agents/{id}/start-host` run **synchronously** (the `mngr` command to completion, in the request's threadpool worker) and return the real outcome. `POST /api/minds/stop-hosts` is the quit-time bulk stop (one `mngr stop … --stop-host`, returns which minds are still running). `GET /api/minds/running` reads in-memory discovery state (no subprocess) so the quit dialog is instant. `POST /api/minds/stop-state-container` stops this env's mngr docker state container.

Design notes: a "shutdown-capable mind" = a workspace on a provider whose host minds can stop/start — `provider_backend_supports_shutdown` is the one place encoding that (docker/lima today). The optimistic override lives on the resolver as a generic per-host layer (not local-specific), so future remote Start/Stop inherits the immediate-flip behavior; it only ever holds RUNNING/STOPPED, so it cannot affect the DESTROYED-only `list_active_workspace_ids` filter. The bulk stop relies on mngr's built-in concurrent multi-host stop rather than firing N requests; on a partial `mngr stop` failure (which raises after stopping every reachable host) the endpoint reports "still running" by recomputing liveness from discovery — it may briefly over-report a host that did stop until discovery catches up, and a Retry re-stop is idempotent. The docker **state container** stop is inherently docker-specific. See `blueprint/local-mind-shutdown/plan-local-mind-shutdown.md` (note the plan predates the discovery-derived-liveness and provider-agnostic-reframe revisions).

## Testing

- `ty` type-check and `ruff` (lint + format) clean on all changed files.
- Local `just test-quick` over the `desktop_client` suite passes (821 passed); a handful of restic / backup cases fail only because this sandbox has no `restic` binary, and the `webdav_test.py` cases are skipped locally (a known TMPDIR-override conflict in this sandbox) — both are environmental, unrelated to this change, and pass in CI. The focused `mind_liveness_test.py` + `mind_controls_test.py` pass (22).
- Unit-tested: the single shutdown-capability predicate, host-state → liveness classification, the resolver optimistic override (override wins until discovery agrees, drop-on-agreement, clear-reverts, doesn't perturb the active-workspace filter), the stop/start/bulk-stop/running endpoints' auth + availability behavior, and the landing-page Start/Stop/Unknown button rendering.
- imbue-code-guardian autofix gate: clean.
- **Not yet done:** the interactive Electron flows (quit prompt, native Stop/Start dialogs, the "quitting" progress page, landing-button SSE behavior, the new "couldn't check" prompt) have not been exercised by launching the running app — verified via code review, type/syntax checks, and Python render tests only. The synchronous stop/start endpoints block a Starlette threadpool worker for the duration of the `mngr` command (fine for a single-user desktop). The 90s override-TTL-expiry branch has no direct unit test (would need a clock seam). The full `just test-offload` CI run on the latest commit is the authoritative check.


apps/minds/electron/main.js
```diff
@@ -1319,6 +1442,185 @@ function sleepInterruptible(ms) {
...
+    }
+    let settled = false;
+    let isAccepted = false;
+    const settle = () => { if (!settled) { settled = true; resolve(isAccepted); } };
+    const timer = setTimeout(() => { try { req.abort(); } catch { /* noop */ } settle(); }, LOCAL_MINDS_HTTP_TIMEOUT_MS);
+    req.on('response', (response) => {
+      isAccepted = response.statusCode < 400;
+      response.on('data', () => {});
+      response.on('end', () => { clearTimeout(timer); settle(); });
+      response.on('error', () => { clearTimeout(timer); settle(); });
```
gnguralnick: seems unsafe to swallow these errors?

apps/minds/electron/main.js
```diff
@@ -1319,6 +1442,185 @@ function sleepInterruptible(ms) {
...
+      resolve();
+      return;
+    }
+    let settled = false;
+    const settle = () => { if (!settled) { settled = true; resolve(); } };
+    const timer = setTimeout(() => { try { req.abort(); } catch { /* noop */ } settle(); }, LOCAL_MINDS_HTTP_TIMEOUT_MS);
+    req.on('response', (response) => {
+      response.on('data', () => {});
+      response.on('end', () => { clearTimeout(timer); settle(); });
+      response.on('error', () => { clearTimeout(timer); settle(); });
```
gnguralnick: similar as above

apps/minds/electron/main.js
```diff
@@ -1319,6 +1442,185 @@ function sleepInterruptible(ms) {
...
+// takeover screen (which the quit sequence has already flipped to) via
+// `status-update`. Returns true to proceed with the quit, false to cancel it.
+// On a failure to stop everything, offers Retry / Quit anyway / Cancel.
+async function stopAllLocalMindsThenDecide(running) {
+  let remaining = running;
+  while (true) {
+    updateQuittingStatus(remaining.length === 1 ? 'Stopping 1 mind…' : `Stopping ${remaining.length} minds…`);
+    await Promise.all(remaining.map((mind) => postLocalMindStop(mind.id)));
+    const deadline = Date.now() + LOCAL_MIND_STOP_DEADLINE_MS;
+    let stillRunning = await getRunningLocalMinds();
```
gnguralnick: i see - the approach is to just settle the direct stop call but repoll after settling to see if it actually worked. this feels a bit clunky

apps/minds/imbue/minds/desktop_client/app.py
```diff
@@ -1888,6 +1914,13 @@ def _on_health_change(agent_id: AgentId, status: AgentHealth) -> None:
                         json.dumps(_system_interface_status_payload(tracker, str(aid), status))
                     )
 
+            # Send the initial local-mind liveness snapshot (derived from
+            # discovery host state) so the landing page can render Start/Stop
+            # controls. Subsequent ticks diff against this to push only changes.
+            last_local_mind_states: dict[str, LocalMindState] = compute_local_mind_state_by_agent_id(backend_resolver)
```
gnguralnick: it seems weird to have so much extra state for local mind liveliness. why don't we just have general mind liveliness stored somewhere and use that appropriately? i feel like we must already have that.

i believe this relates to some of the optimistic overriding but idk this approach just seems awkward

apps/minds/imbue/minds/desktop_client/app.py
```diff
@@ -2573,6 +2614,189 @@ def _handle_restart_host_api(
...
+# A "local mind" is a workspace whose host runs on a local provider backend
+# (docker / lima); stopping one frees the user's machine while preserving data
+# and leaving it fully restartable. Stop = ``mngr stop --stop-host`` on the
+# host (same teardown the host-restart tier uses); Start = ``mngr start`` (boots
+# the stopped container). Both set an optimistic host-state override on the
+# resolver so the landing page and quit prompt flip at once; the next discovery
+# snapshot then confirms (or corrects) it.
+
+
+class _LocalMindAction(UpperCaseStrEnum):
```
gnguralnick: right now all this stuff just happens to apply only to local minds, because we don't currently surface for other providers whether they support stopping the hosts. this will change eventually. so in advance of that can we reframe all of this to not be speciifc to local minds? let's just have there be like one place that confirms that only local minds (currently) support shutdown, but all the machinery and stuff should be named appropriately and setup such that we can just update that condition function at the point when we have a proper condition for checking whether other providers support shutdown

apps/minds/imbue/minds/envs/docker_cleanup.py
```diff
@@ -152,6 +153,82 @@ def remove_state_container(
...
+    with cg:
+        try:
+            stop_result = cg.run_process_to_completion(
+                command=["docker", "stop", container_name],
+                timeout=_DOCKER_CMD_TIMEOUT_SECONDS,
+                is_checked_after=False,
+                env=dict(os.environ),
+            )
+        except ProcessError as e:
+            logger.info("Docker CLI unavailable; skipping state-container stop for {} ({})", container_name, e)
```
gnguralnick, line 184: is it really safe to just swallow this?

gnguralnick: similar for other errors in this file

joshalbrecht:
lgtm, thanks gabe!

PR #2029, opened 2026-06-08
fix builtin-exception ratchet regex + fix some real violations

evgunter:
## Summary

The \`PREVENT_BUILTIN_EXCEPTION_RAISES\` ratchet regex had a trailing \`\b\` after the opening paren:

\`\`\`
raise\s+(ValueError|...|KeyboardInterrupt)\(\b
\`\`\`

The \`\b\` requires a word character immediately after \`(\`, so the ratchet only matched raises whose first argument starts with a word char (e.g. \`raise OSError(msg)\`, \`raise IOError(f"...")\`) and **missed the most common forms**: \`raise ValueError("literal")\`, \`raise OSError()\`, etc. Removing \`\b\` is the fix.

Removing it surfaced ~60 previously-hidden raises (28 in test files, 32 in production).

## Changes

- **Regex fix:** removed the trailing \`\b\` from the pattern.
- **Test files excluded:** the ratchet now skips test files (\`TEST_FILE_PATTERNS\`), since tests legitimately raise built-in exceptions to simulate error conditions. (Note: this diverges from the sibling \`except\`-family ratchets, which scan tests; the raise case is different because simulating errors in tests is common and fine.)
- **Production fixes (29):** replaced surfaced raises across 10 projects with custom exception types inheriting from the package base exception and the built-in (the established pattern, e.g. \`class ParseSpecError(MngrError, ValueError)\`). Several reuse existing exceptions (e.g. minds \`TelegramCredentialExtractionError\`).
- **Grandfathered (3):** the three \`cron_runner.py\` raises are left as raw \`RuntimeError\` because that file must not import \`imbue.*\` at module scope (Modal runs it standalone), so package exceptions are unavailable. \`mngr_schedule\`'s recorded count is now 3.
- **Re-snapshotted** the affected ratchet counts (imbue_common 6→0, skitwright 1→0, mngr_imbue_cloud 3→0, mngr_schedule 6→3).

## Verification

- All 37 \`test_prevent_builtin_exception_raises\` tests pass.
- Targeted unit tests for every changed module pass (308 total).
- \`ruff check\`, \`ruff format --check\`, and \`ty check\` (workspace-scoped) all clean.


apps/minds/scripts/create_telegram_bot.py
```diff
@@ -98,7 +98,7 @@ def _fetch_telegram_web_api_credentials() -> tuple[int, str]:
             if cred_match:
                 return int(cred_match.group(1)), cred_match.group(2)
 
-        raise ValueError("Credential pattern not found in any bundle chunk")
+        raise TelegramCredentialError("Credential pattern not found in any bundle chunk")
```
evgunter, line 101: not sure if these were not `TelegramCredentialError` for a reason

apps/minds/imbue/minds/telegram/bot_creator.py
```diff
@@ -108,7 +109,7 @@ def fetch_telegram_web_api_credentials() -> tuple[int, str]:
             if cred_match:
                 return int(cred_match.group(1)), cred_match.group(2)
 
-        raise ValueError("Credential pattern not found in any bundle chunk")
+        raise TelegramCredentialExtractionError("Credential pattern not found in any bundle chunk")
```
evgunter, line 112: not sure if these were not `TelegramCredentialExtractionError` for a reason

joshalbrecht:
lgtm, thanks for the fix!

PR #2034, opened 2026-06-08
i have solved config layering and from now on there will be no issues ever

evgunter:
https://github.com/imbue-ai/mngr/pull/2208 should go in right after this

and now claude layering works properly! here it is telling me how to fix the thing that made my `defaultMode` (a member of `permissions`) be ignored:

https://github.com/user-attachments/assets/ff6bc150-0f3b-441b-9bbb-db019a568fb9

(and showing that with `__assign` it continues to be ignored)

---

## Summary

mngr's layered config now merges through a single, explicit **operator algebra** instead of an ad-hoc deep-merge / assign-by-default. The generic part of that algebra is extracted into a standalone, dependency-free **`overlay` library** built on a typed-node core, and mngr's **entire config model** — `MngrConfig`, agent types, `parent_type` inheritance, and Claude `settings_overrides` — is routed through it by one merge pipeline. Unifying the merge under one model is what makes the behavior predictable; it also closes a class of silent-drop bugs and the original Claude-hooks-leak bug (the hooks now live in, and `settings_overrides` folds onto, the same per-agent `settings.json`).

The library's semantics are documented in `libs/overlay/README.md` (framework-agnostic); mngr's integration pipeline is documented in the `imbue.mngr.config.overlay_merge` module docstring.

## Operator semantics for layered config

mngr config (and Claude `settings_overrides`) now merge with **explicit, predictable** operators:

- **bare `key`** → assign, replacing the value below, with a **narrowing guard** that hard-errors if the assign would silently drop a non-empty aggregate sibling (e.g. replacing `permissions = {allow:[...]}` would drop a `defaultMode`).
- **`key__extend`** → merge onto the value below (list concat, set union, **recursive** dict merge; nest `__extend` to merge deeper). Never narrows.
- **`key__assign`** → assign **without** the narrowing warning — the per-key opt-out (coexists with the global `allow_settings_key_assignment_narrowing` flag).
- **`Static*` markers** (`StaticTuple`/`StaticList`/`StaticDict`) mark an aggregate **atomic**, so replacing it is a value-set, not narrowing. mngr's `StringDerivedTuple`/`ScalarTuple` are now `StaticTuple` subclasses.

Values now **accumulate across config scopes and `parent_type` inheritance** (per-key; non-overlapping keys from every scope survive, same-key `__extend`s combine) rather than a higher scope wiping the whole value. The narrowing guard is **recursive** — a bare key nested inside an `__extend` is caught, not just top-level keys.

## The `overlay` library (`libs/overlay`)

The generic, model-free part of the algebra is a new **standalone, dependency-free** workspace library that mngr depends on. It owns one thing: how N precedence-ordered config layers combine (assign vs. extend, narrowing detection, deferred resolution) as a pure dict→dict algebra — nothing about pydantic, TOML, or mngr's schema.

Its core is a **typed-node algebra**: `Default` / `Assign` / `Extend` wrappers (plus `Static*`) instead of parsing operators out of key strings at every step. A single `lift` pass converts the suffix surface syntax (`key__extend` / `key__assign`, unchanged for users) into nodes; `lower` converts back at storage boundaries. The operator lives in the value's **type**, so the algebra dispatches on and rewrites the outermost wrapper but **never re-parses** — which structurally removes a latent hazard where a stacked suffix (`a__extend__assign`) would have an inner operator silently reactivate on a later pass. The public API is `merge(lower, higher) -> patch` (raises an aggregated `NarrowingError`) and `merge_narrowing_allowed(lower, higher) -> (patch, narrowings)` (returns them for the caller to surface or discard), over one private core. Both take typed-node `Patch`es — so a caller can fold many layers (`merge(acc, lift(layer))`) or feed wrapper-marked dicts directly; the consumer applies `lift` at its boundary. The node algebra is fully property-tested (associativity, recursive narrowing) at 100% coverage.

## mngr integration: the whole config model on overlay

mngr keeps suffix-keyed dicts as its storage/boundary representation (JSON-safe, unchanged contract); the node algebra is the internal engine, bridged by `lift`/`lower`. The whole config-model merge runs through one pipeline, `overlay_merge.merge_models_via_overlay`: serialize the sparse override (`model_dump(exclude_unset=True)`) and the full base → pre-process (`SettingsPatchField` → `__extend`, two-level container marking, `None`-drop) → overlay `combine` / `merge_narrowing_allowed` → reparse into `type(base)`. This **replaces the field-by-field pydantic merges**:

- `AgentTypeConfig.merge_with`, `MngrConfig.merge_with` / `merge_with_narrowings`, and `parent_type` inheritance (`agent_config_registry`) all delegate to it. The per-model `merge_with` overrides are removed; the merged result is covered by feature tests (loader, `data_types`, CLI, provider/plugin configs) plus direct-assertion pipeline tests.
- **Cross-scope `settings_overrides` narrowing is now caught at config-load.** Because `settings_overrides` accumulates across scopes, a higher scope whose *bare* key replaced a non-empty aggregate set by a lower scope previously dropped it silently (the provision-time guard only saw drops of the *home* `settings.json` base). The loader now forwards `merge_with_narrowings`' settings paths into the same flag-gated narrowing error used for every other field. **(Intentional behavior change.)**
- The `parent_type` no-override case now returns a fresh `==`-equal instance rather than the same object (identity is irrelevant for an immutable frozen model). **(Intentional behavior change.)**

**Scope note:** all config-load narrowing now flows through the overlay merge — the model-walking `detect_settings_narrowing` is removed, as is the older string-suffix combine engine in `overlay.merge` (only the leaf/narrowing primitives `apply_extend` / `would_assignment_narrow` / `narrowing_paths` remain, with the typed-node algebra in `node_merge` as the sole engine). Removing the `allow_settings_key_assignment_narrowing` flag (raise-immediately) is deliberately **deferred** pending a team decision on narrowing philosophy.

**Simplification sweep:** a multi-agent review drove a follow-up cleanup — field markers (`SettingsPatchField` / `RegistryField`) unified behind one metadata helper in a leaf module (dropping the resolver-callable threading and the hard-coded container-field set), the three serialize→operator-dict mark/unmark passes collapsed into one recursive walk, sub-model detection made runtime (`isinstance`), and the now-vestigial refactor scaffolding removed: the frozen old-logic equivalence guards (which had pinned new == old in CI) and the superseded string-suffix engine, ~1.7k lines net. A later review pass deduped shared helpers further: the narrowing-error message is now one builder shared by config-load and the Claude provision fold; the `settings.local.json` gitignore guard moved into the `claude_subagent_proxy` plugin (its only caller); a `write_json_dict_via_host` companion to `read_json_dict_via_host` absorbs the host JSON-dict writes; and the duplicated stop-hook-guard / gitignore-acceptance / hook-shape-validation logic collapsed to single helpers.

## Claude `settings_overrides` + the hooks-leak fix

`settings_overrides` is the user-facing application of the same algebra. It's a schema-free `dict[str, Any]` (annotated `SettingsPatchField`) folded onto the per-agent `settings.json` at provision via the node algebra (`_build_settings_json`): a bare key assigns under the narrowing guard, `key__extend` merges onto the home value, `key__assign` opts out — exactly the operators above. This replaces the previous deep-merge-by-default, and fixes #1647 (a `permissions__extend.allow__extend` now preserves `defaultMode`) and the scope-wipe regression (a project-scope `model` and a user-scope `permissions` now both apply). One intentional fix rides along: a **deferred `key__assign`** now reliably suppresses the narrowing guard at provision (previously the no-warn intent could be dropped at config-load).

The original bug that motivated all of this: in a **plain (non-mngr)** Claude session a user saw

```
Stop hook error: ... mkdir: cannot create directory '/events': Permission denied
```

because mngr wrote its readiness/credential/permission hooks (and the subagent-proxy's hooks) into the project's `.claude/settings.local.json`, which plain `claude` in that directory also reads — so mngr's hooks fired outside mngr where `$MNGR_AGENT_STATE_DIR` is unset.

**Fix:** mngr bakes all of its own hooks into the **per-agent config-dir `settings.json`** (`$MNGR_AGENT_STATE_DIR/plugin/claude/anthropic/settings.json`) — the "user" settings layer Claude reads from `$CLAUDE_CONFIG_DIR`, which a plain `claude` run in the work dir never reads (it reads `~/.claude`). This is the same file `settings_overrides` folds onto. Hooks are rebuilt fresh every provision (no cross-version accumulation), and in normal mode mngr passes **no `--settings` of its own**, so a user's own `--settings` passes through and Claude layers it natively (deep-merging dicts, concatenating same-event hooks).

- `use_env_config_dir` mode (no per-agent config dir): mngr loads its hooks **and the resolved `settings_overrides` patch** from a private managed `--settings` file, which Claude layers highest-precedence over the user's shared config — so `settings_overrides` is applied in this mode too (folded onto mngr's hooks as the base, so its narrowing guard there only catches an override dropping mngr's own hooks, not the shared-config values Claude layers itself). Because mngr already passes that `--settings`, a user-supplied `--settings` (in `cli_args`/`agent_args`) is now **rejected** at provision with a clear error (mngr can't reliably merge a second one — a `--settings` value may be inline JSON, not a file); put those settings in `settings_overrides`, or set `use_env_config_dir=False`. That mode is not yet used in production.
- The `settings.local.json`-must-be-gitignored requirement is narrowed: it now applies only when the `claude_subagent_proxy` plugin actually rewrites user Stop hooks (the only remaining writer of that file), enforced at the point it writes.
- This stops **new** leaks; it does not strip hooks a prior mngr already wrote into existing `settings.local.json` files — clean those up manually if present.

## Drive-by: `mngr config` no longer self-blocks on a narrowing config

**(Preexisting in `main`, fixed here.)** The settings-narrowing guard runs while loading the merged config — which `mngr config set` / `mngr config unset` also do — so the very command you'd use to *fix* a narrowing config failed with the narrowing error itself (a catch-22). `mngr config` is now exempt from the guard (it must be able to load a config in order to edit it); every other command still enforces it.

Relatedly, an expected error raised during provisioning (e.g. a settings-narrowing `ConfigParseError` while building the per-agent `settings.json`) now surfaces as a clean one-line error instead of an "Unexpected error" + traceback — the `create` flow unwraps it from the `ConcurrencyExceptionGroup` that the provisioning concurrency group wraps it in (via the group's existing `only_exception_is_instance_of` / `get_only_exception` helpers).

---

## Tests

New/updated coverage across `overlay` (node algebra incl. lift/lower round-trips, recursive narrowing, associativity; 100%), `mngr` config (the `merge_models_via_overlay` pipeline, operators, `Static*`, narrowing, deferred `__assign`, the shared narrowing-message), and `mngr_claude` / `claude_subagent_proxy` (the node-based `settings_overrides` fold in both modes, the `use_env_config_dir` `--settings` rejection, hook-leak behavior, gitignore gating). Affected packages green locally after merging `main` (`mngr` config + hosts, `mngr_claude` ~586, `claude_subagent_proxy` 176, plus `codex`/`opencode`/`pi_coding`); ratchets passing, `ruff`/`ty` clean. Full monorepo suite runs in CI.


joshalbrecht:
this *does* seem reasonable, except that now I think this will break my config:

```
[agent_types.coder]
parent_type = "claude"
cli_args = """--settings '{"hooks": {"SessionStart": [{"hooks": [{"type": "command", "command": "read INPUT; SID=$(echo \\"$INPUT\\" | jq -r \\".session_id // empty\\"); [ -n \\"$SID\\" ] && [ -n \\"${CLAUDE_ENV_FILE:-}\\" ] && echo \\"export MNGR_CLAUDE_SESSION_ID=$SID\\" >> \\"$CLAUDE_ENV_FILE\\" || true"}]}]}}'"""
```

seems like we should probably mix explicit user CLI settings in to our file? Yes it's absolutely hideous, but I don't really see a better way to do this and still be correct?

happy to discuss this one too

(code looks fine other than the above question)

evgunter:
[CLAUDE] The per-model `merge_with` methods on the config sub-models (`RetryConfig`, `LoggingConfig`, `ProviderInstanceConfig`, `PluginConfig`, `CommandDefaults`, `CreateTemplate`) are now **vestigial in production**: the config merge is fully routed through the overlay pipeline (`config/overlay_merge.py::merge_models_via_overlay`), whose generic `combine` subsumes them (`_merge_container_dict`, their only caller, was removed).

They are deliberately **kept for now** because they double as the *live frozen reference* that the `config/overlay_merge_test.py` equivalence tests compare the new overlay path against, proving the refactor is behavior-identical.

**They should be deleted before this PR merges.** The clean removal: inline frozen `_reference_*` copies of their bodies into the equivalence test (mirroring the existing `_reference_assign_scalar` / `_reference_merge_container_dict` pattern already in that file), then sweep the seven now-orphaned `PluginConfig.merge_with` subclass overrides that live across the separate plugin projects. That was scoped out of this PR as a low-priority cross-project cleanup.

(`MngrConfig.merge_with` / `merge_with_narrowings` and `AgentTypeConfig.merge_with` stay — the former is the loader's entry point, the latter the public agent-type merge; both now thin-delegate to the overlay pipeline.)

libs/overlay/imbue/overlay/nodes.py
```diff
@@ -0,0 +1,73 @@
...
+    List concat, set union, recursive patch merge. Never narrows (an extend is a
+    superset). Carries a single ``payload`` (a leaf or a nested ``Patch``).
+    """
+
+    payload: Any
+
+
+# A node is one of the three operator wrappers; a patch maps field names to nodes.
+Node = Default | Assign | Extend
+Patch = dict[str, "Node"]
```
evgunter: does this need to be str? `Node` is obviously defined here

libs/overlay/imbue/overlay/operators.py
```diff
@@ -0,0 +1,111 @@
...
+from imbue.overlay.errors import OverlayError
+from imbue.overlay.pure import pure
+
+# Operator suffix on a leaf key indicating "extend the current value".
+# Lowercase form used in the common surface syntax (TOML, override paths, keys).
+EXTEND_SUFFIX: Final[str] = "__extend"
+
+# Uppercase form for env-var path segments. Matches the all-uppercase
+# convention for env-var path segments.
+EXTEND_SUFFIX_ENV: Final[str] = "__EXTEND"
```
evgunter: i'm wondering if there's a clean way we could let people register their own extend vs assign suffixes. also what is EXTEND_SUFFIX_ENV used for

libs/mngr/imbue/mngr/config/data_types.py
```diff
@@ -131,81 +150,20 @@ def split_cli_args_string(cli_args: str) -> tuple[str, ...]:
...
-    {"agent_types", "providers", "plugins", "commands", "create_templates"}
+# The fully-qualified paths of ``MngrConfig``'s top-level container dict fields, whose
+# cross-scope merge is per-key (instead of assign-by-default). Specified as *paths*
+# rather than bare field names so the narrowing walker matches only the actual
+# top-level field and never a same-named field nested inside a sub-model (e.g. a plugin
+# config's own ``commands`` dict, which must be narrowing-checked as a leaf aggregate,
+# not per-key-recursed -- a bare-name match silently skipped its dropped keys).
+_CONTAINER_DICT_FIELD_PATHS: Final[frozenset[tuple[str, ...]]] = frozenset(
+    {("agent_types",), ("providers",), ("plugins",), ("commands",), ("create_templates",)}
 )
```
evgunter: could flatten this since all the fields are in fact top-level

libs/mngr/imbue/mngr/config/data_types.py
```diff
@@ -131,81 +150,20 @@ def split_cli_args_string(cli_args: str) -> tuple[str, ...]:
...
-    if not isinstance(override_value, dict):
-        return True
-    if any(key not in override_value for key in base_value):
-        return True
-    return any(would_assignment_narrow(sub_base, override_value[key]) for key, sub_base in base_value.items())
+# The bare top-level field names, derived from the paths. Used by the overlay merge,
+# which only ever inspects the top-level ``MngrConfig`` dict (it never recurses into a
+# sub-model), so a name match there is unambiguous. All container paths are single
+# segment (top level), so the first element is the field name.
+_CONTAINER_DICT_FIELD_NAMES: Final[frozenset[str]] = frozenset(path[0] for path in _CONTAINER_DICT_FIELD_PATHS)
```
evgunter: sus

libs/mngr/imbue/mngr/config/loader.py
```diff
@@ -364,43 +367,71 @@ def get_or_create_profile_dir(base_dir: Path) -> Path:
...
-    processed_sources: Sequence[tuple[_SettingsSource, MngrConfig]],
+) -> None:
+    """Mark ``source`` as the owner of every path this layer assigns.
+
+    Called after a layer is folded in (and after its narrowings are attributed against
+    the prior provenance), so each path maps to the highest-precedence source that has
+    assigned it so far.
+    """
+    for path in _assigned_paths(parsed_layer):
+        provenance[path] = source
```
evgunter, line 459: these are kind of silly but seem acceptable

libs/mngr/imbue/mngr/api/create.py
```diff
@@ -353,7 +354,18 @@ def create(
...
+                    except ConcurrencyExceptionGroup as group:
+                        # provision_agent runs its body inside a concurrency group, which
+                        # re-raises any body exception wrapped in a ConcurrencyExceptionGroup
+                        # (and flattens a nested group's single leaf up to this level). Surface
+                        # a single expected error -- e.g. a settings-narrowing ConfigParseError
+                        # raised while building settings.json -- cleanly rather than as an
+                        # "Unexpected error" with a full traceback.
+                        if group.only_exception_is_instance_of(MngrError):
+                            raise group.get_only_exception() from group
+                        raise
```
evgunter, line 368: flagging!

libs/mngr/imbue/mngr/cli/common_opts.py
```diff
@@ -177,7 +177,10 @@ def setup_command_context(
     if strict is None:
         strict = resolve_strict_from_env()
 
-    # Load config (is_interactive will be resolved below)
+    # Load config (is_interactive will be resolved below). The ``mngr config`` command is
+    # exempt from the settings-narrowing guard: it must be able to load a config that would
+    # otherwise narrow in order to *edit* it (otherwise `mngr config set`/`unset` -- the way
+    # to fix a narrowing config -- would themselves fail with the narrowing error).
```
evgunter, line 183: this was a preexisting bug

libs/mngr/imbue/mngr/api/create.py
joshalbrecht: this is an improvement, thanks

libs/mngr/imbue/mngr/cli/common_opts.py
joshalbrecht: yup, this makes sense

libs/mngr_latchkey/imbue/mngr_latchkey/config_test.py
```diff
@@ -20,62 +9,3 @@ def test_default_values_are_none() -> None:
...
-    """A populated override replaces the base value for every field it sets."""
-    base = LatchkeyPluginConfig(directory=Path("/base/dir"), latchkey_binary="/base/bin")
-    override = LatchkeyPluginConfig(
-        directory=Path("/override/dir"),
-        latchkey_binary="/override/bin",
-    )
-    merged = base.merge_with(override)
-    assert merged.directory == Path("/override/dir")
-    assert merged.latchkey_binary == "/override/bin"
-    assert merged.enabled is True
```
joshalbrecht, line 35: this PR seems to have removed a bunch of settings related tests from other modules (fine), but it's not clear that those tests were all effectively re-created somewhere else?

ie, are we sufficiently happy with the settings test coverage that we have for mngr config?

joshalbrecht:
it's a great PR title :)

I wouldn't be opposed to a bit more testing though, and I'm also curious for @qi-imbue 's thoughts!

(overall lgtm except the comments about testing and our confidence level in this)

libs/mngr_latchkey/imbue/mngr_latchkey/config_test.py
evgunter: the merge_with tests that were removed were only done so when they tested exactly the same thing as the base level tests. i think i'm pretty happy with the coverage

PR #2035, opened 2026-06-08
Confirm message submission on enqueue OR hook, concurrently

gnguralnick:
## Summary

Sending a message to a Claude agent could block for a very long time when the agent was busy, even though the message had already been accepted into the agent's queue. `send_enter_via_tmux_wait_for_hook` waited up to the full submission timeout (90s) on the `UserPromptSubmit` `tmux wait-for` signal before falling back to checking the enqueue log. That hook only fires when a prompt reaches the model — for a message sent to a busy agent, that's when the agent finally *dequeues* it, which can be many minutes later (a ~473s enqueue→dequeue gap was observed in real transcripts). Any caller that blocks on the send over HTTP (e.g. a front-end hitting `mngr`'s send path) then exceeds front-door proxy timeouts and 504s, even though the message was queued and will be processed.

## Change

Claude records an `enqueue` event in its transcript log the instant a message enters its queue. `send_enter_via_tmux_wait_for_hook` now watches a fresh enqueue **concurrently** with the hook's `tmux wait-for` signal — in a single remote command — and succeeds on whichever lands first:

- For a **busy** agent the enqueue confirms acceptance in well under a second.
- The hook still covers prompts that never enqueue a model turn — the `/clear` and `/compact` TUI-local commands, whose signal is fired from `SessionStart` (so this preserves the fix in #1560).

It's one bash command (background the full-timeout hook waiter, which writes a sentinel on success; poll the sentinel + a fresh enqueue until the deadline), so there's no dangling process and the register-the-waiter-before-Enter ordering — which keeps the one-shot signal from being missed — is preserved. The enqueue path is scoped to agents that supply a queue-log path (Claude); other TUIs keep the original hook-only wait. Removes the now-unused two-phase `_get_last_queue_timestamp` helper.

## Test plan

- [x] Unit tests in `tui_agent_test.py`: the send dispatches a single host command that watches the hook alone when no queue-log path is given, and watches both the hook and a fresh enqueue when one is; a non-success result surfaces as failure. (`uv run pytest libs/mngr/imbue/mngr/agents/tui_agent_test.py` — 7 passed.)
- [x] Verified the inner bash control flow with a local `tmux`/queue-log simulation across all paths: enqueue-appears, hook-fires, neither (timeout), and a stale pre-existing enqueue (must not count).
- [x] End-to-end against a real local Claude agent: with the agent busy in a 30s foreground command, a send returned in **1.63s** with the message confirmed *queued* in the TUI (its `UserPromptSubmit` had not yet fired), then the queued message was dequeued and processed normally. The idle-agent send was also fast (1.92s).
- [ ] CI: full suite.


libs/mngr/imbue/mngr/agents/tui_utils.py
```diff
@@ -280,40 +272,82 @@ def _send_enter_and_wait_for_signal(
...
+    )
 
 
-def _get_last_queue_timestamp(
+# FIXME: this is fairly Claude-specific for an otherwise agent-agnostic module.
+# `queue_log_path_template` plus the hardcoded `"operation":"enqueue"` grep + jq
+# below assume Claude Code's transcript-event-log schema. It would be cleaner to
+# generalize: have the agent supply how to detect its own "message accepted"
+# marker (e.g. a command or matcher) so this module stays agent-neutral. (The
+# coupling predates this function -- it was previously in _get_last_queue_timestamp.)
```
joshalbrecht: yeah.. we probably should clean this up properly while we're here!

probably not too much more effort than this current change anyway

joshalbrecht:
I think we should fix this one the right way

generally looks good beyond my comment

PR #2037, opened 2026-06-08
Fix loading-window position jump and preserve MRU on startup

gdenisov:
## Summary

Fixes the startup loading-window position jump and improves multi-window restore. Four related changes in `apps/minds/electron/main.js`:

1. **Apply saved bounds before the loading screen renders.** Between `createBundle()` and `runStartupSequence()`, load saved state and call the existing `restoreWindowBounds()` on the initial window. The window is created with `show: false`, so bounds applied before the first `show()` are invisible — no flash at the default-centered position.
2. **Persist window state in MRU order.** `saveSessionState()` now iterates `mruWindows` instead of the `bundles` Set, so entry 0 is the user's most-recently-focused window. For multi-window users the loading screen opens at the bounds of the last window they actually interacted with.
3. **Open lesser-MRU windows without stealing keyboard focus.** Added an opt-in `showInactive` mode on `openNewWindow` (consulted by `wireBundleShowLogic`) so the additional restored windows use `showInactive()` rather than `show()`.
4. **Re-raise initialBundle so it stays visually on top.** `showInactive()` blocks keyboard focus but not z-order on macOS, so registered a `once('show')` listener on each restored window that calls `initialBundle.window.focus()` (equivalent to `moveTop()` on macOS) to bring the loading window back to the front. The focus event handler's mruWindows manipulation is a net no-op since initialBundle was already at position 0 after the explicit MRU rewrite.

Bonus: after the restore loop, `mruWindows` is explicitly rewritten to `[initialBundle, ...restoredBundles]`. `createBundle`'s `unshift` would otherwise reverse the saved order in the in-memory MRU list, which would silently corrupt the next `saveSessionState` write.

Plan: `blueprint/loading-window-position/plan-loading-window-position.md`.

## Test plan

- [ ] Fresh install (delete `window-state.json` first): loading window appears centered at default 1200x800. No jump. Unchanged behavior.
- [ ] Single-window: drag to a non-default position and resize, quit. On relaunch the loading window appears immediately at that location and size with no jump when content loads.
- [ ] Multi-window, focus check: open 2+ windows, focus the second-to-last one last, quit. On relaunch the loading window appears at the last-focused window's bounds; the other windows restore behind it; the last-focused window stays visually on top (not just keyboard-focused) after restore.
- [ ] MRU persistence across restarts: relaunch, focus a different window, quit, relaunch again. Loading window should appear at the newly most-recently-focused window's bounds, not the original.
- [ ] Deleted-workspace edge case: with multi-window state, delete the workspace at entry 0, relaunch. Loading window appears at entry 0's bounds; content window then settles at entry 1's bounds (small secondary jump is accepted).
- [ ] Display-gone case: save bounds on a secondary monitor, disconnect it, relaunch. Existing `restoreWindowBounds()` clamping should place the loading window on the primary display.
- [ ] Unauthenticated start: with a saved state file but logged out, loading window appears at entry 0's bounds and the welcome page loads in the same window without a jump.

CI note: `test-docker-electron` has been failing pre-existing on `main` across recent runs (`playwright TimeoutError`), unrelated to this branch — the e2e test uses a fresh `userData` so `loadSessionState()` returns `[]` and this branch's code is inert in that path.


joshalbrecht:
lgtm

PR #2039, opened 2026-06-08
reset stale activity markers on restart/resume

gnguralnick:
A Claude agent restarted or resumed mid-turn reported the `RUNNING` lifecycle state forever.

`get_lifecycle_state` keys `RUNNING` vs `WAITING` off the `active` marker file. That marker is set on `UserPromptSubmit` and removed only by the `Stop` / idle `Notification` hooks. A turn abandoned by an abnormal exit (container restart, OOM, crash) never runs the `Stop` hook, so `active` is left stale and the agent is stuck reporting `RUNNING` even though it is idle. This affects every consumer of lifecycle state (`mngr ls`, notifications, the `WAITING`-gated wait-for-agent flows).

## Change

The `SessionStart` hook now, on `startup`/`resume` only (a fresh, not-mid-turn Claude process):
- clears the stale `active` and `permissions_waiting` markers, so the lifecycle state self-heals on the next (re)start;
- touches a new `claude_process_started` marker whose mtime is a clean restart boundary — any transcript event older than it belongs to a turn the current process did not run, which lets downstream consumers (e.g. an activity indicator) distinguish an abandoned mid-turn tail from work still in flight.

`compact` is deliberately excluded, because auto-compaction fires mid-turn while Claude is genuinely active and must not move the boundary or clear `active`. The four `SessionStart` `source` values are per the Claude Code hook docs (`startup`, `resume`, `clear`, `compact`).

The shared "clear active markers and emit an activity event" shell snippet (previously inlined in the Notification idle hook) is extracted into a constant and reused by both the Notification idle hook and the new `SessionStart` hook so they stay byte-identical.

## Testing

- Added a behavioral test that runs the generated `SessionStart` hook shell across all four `source` values and asserts the markers are reset (and `claude_process_started` recorded) on `startup`/`resume` but not on `compact`/`clear`.
- Updated the `SessionStart` hook-structure assertions.
- `libs/mngr_claude`: 515 passed (unit/integration, infra-marked excluded), ratchets pass, `ruff` and `pyright` clean.


joshalbrecht:
lgtm once tests pass

PR #2041, opened 2026-06-08
Fix requests panel X button unclickable on startup auto-open

gdenisov:
## Summary
- Fixes a race condition where the inbox/requests modal auto-opens at app startup before chrome.js has registered its `onModalStateChanged` IPC listener. The initial `modal-state-changed: { open: true }` event is dropped, the body never gets the `modal-open` class, and the titlebar's drag region intercepts clicks (notably the X button) in the y=0..TITLEBAR strip.
- Fix: fold modal-state replay into the existing `primeViewWithCachedChromeState` helper that runs on the chrome view's `did-finish-load`. Alongside the cached `workspaces`, `auth_status`, and `requests` events it already primes, it now also sends `modal-state-changed: { open: bundle.modalVisible }` (scoped to the chrome titlebar view; the sidebar doesn't listen for it).

## Test plan
- [ ] Launch the app fresh with a pending request so the requests panel auto-opens; verify the X button closes the modal on first click.
- [ ] Open the requests panel manually (non-startup path); verify X still works (regression check).
- [ ] Open and close the panel several times to confirm `modal-open` toggles correctly.

Co-authored-by: Sculptor <sculptor@imbue.com>


joshalbrecht:
lgtm

PR #2043, opened 2026-06-08
Override docker_runtime to runc in Electron e2e test

gdenisov:
## Summary

- FCT's `[providers.docker]` block sets `docker_runtime = "runsc"` to harden the local-docker provider with gVisor. CI runners (and most dev machines) do not have runsc installed, so `docker run --runtime runsc` fails immediately with "unknown or invalid runtime name: runsc" and the test times out waiting for the agent navigation URL.
- Set `MNGR__PROVIDERS__DOCKER__DOCKER_RUNTIME=runc` via `monkeypatch.setenv` right after the existing Modal override; the Electron child inherits it through `_build_electron_env`. FCT's own settings.toml comment names this exact env var as the supported escape hatch for CI / Modal.

## Test plan

- [ ] CI runs `test-docker-electron` to green on this branch.
- [ ] Local run of `just minds-test-electron` on a machine without runsc completes.


joshalbrecht:
lgtm, thanks for the fix!

not sure how CI passed before this...

gdenisov:
My Claude believes it passed because the mngr changes have been run with a version of the FCT pre [providers.docker] update

PR #2045, opened 2026-06-09
minds: fail fast on workspace-create failure + typed docker runtime error

evgunter:
depends on https://github.com/imbue-ai/mngr/pull/2043 or similar fix

the changes in apps/minds/imbue/minds/desktop_client/e2e_workspace_runner.py and apps/minds/imbue/minds/desktop_client/e2e_workspace_runner_test.py seem kind of sketchy

---

## Background

While investigating why `test-docker-electron` fails on every branch, I found the proximal cause: the e2e test clones forever-claude-template `main`, whose `[providers.docker]` block now sets `docker_runtime = "runsc"`, but CI runners have no gVisor registered, so `docker run --runtime runsc` fails with exit 125. (That proximal fix is handled separately on `gleb/electron-runc-override`, which adds an in-test `MNGR__PROVIDERS__DOCKER__DOCKER_RUNTIME=runc` override.)

This PR fixes the two deeper problems that made that failure *opaque* — a 10-minute hang surfacing a generic Playwright timeout instead of the real one-line cause.

## Changes

**e2e: fail fast instead of hanging.** `create_workspace_via_electron` previously only waited for the workspace-ready redirect, so *any* `mngr create` failure blocked the full 600s navigation budget before timing out opaquely. It now races the workspace-ready URL against the create flow's `#failure-view`, raising `WorkspaceCreationFailedError` with the surfaced `#error-message` the moment creation fails. (Affects only the e2e/snapshot path — the shipped app already surfaces the failure view to users.)

**mngr: typed, actionable docker-runtime error.** The docker provider now raises `DockerRuntimeNotRegisteredError` when the configured `docker_runtime` (e.g. `runsc`) is not registered with the daemon, instead of a raw exit-125 `ProcessError` that dumps the whole `docker run` command line with the cause buried inside. As an `MngrError`, it renders cleanly to stderr and `mngr create --format jsonl` emits `error_class: "DockerRuntimeNotRegisteredError"`.

Adds unit tests for both and per-project changelogs (`libs/mngr`, `apps/minds`).

## CI note

This branch intentionally does **not** carry the `runc` override, so `test-docker-electron` stays red until `gleb/electron-runc-override` merges to `main`; merging `main` in afterward turns it green. (Coordinated this split deliberately.)


libs/mngr/imbue/mngr/errors.py
```diff
@@ -309,6 +309,30 @@ def __init__(self, provider_name: ProviderInstanceName, timeout_seconds: int) ->
...
+            f"Docker runtime '{runtime_name}' is not registered with the Docker daemon "
+            f"for provider '{provider_name}'.",
+        )
+        self.user_help_text = (
+            f"Install and register the '{runtime_name}' runtime with Docker (e.g. gVisor's "
+            f"runsc), or select the default runtime by setting docker_runtime to 'runc':\n"
+            f"  mngr config set --scope user providers.{provider_name}.docker_runtime runc\n"
+            f"or per-invocation:\n"
+            f"  MNGR__PROVIDERS__{provider_name.upper()}__DOCKER_RUNTIME=runc"
+        )
```
evgunter, line 333: is this the recommendation we want to make?

joshalbrecht: yup, that looks fine for now

joshalbrecht:
lgtm!  thanks for the root fix

PR #2046, opened 2026-06-09
fix recursive ~/.claude symlinks

evgunter:
## Summary

Fixes two bugs in `mngr_claude` that wrote cruft into the user's shared `~/.claude/` on every *local* agent provision.

### Bug 1: recursive self-referential symlinks
`_sync_user_resources` used plain `ln -sf` as an idempotent command. On the second and later provisions the destination was already a symlink-to-directory, so `ln` dereferenced it and nested a new link **inside the shared source** (e.g. `~/.claude/skills/skills -> ~/.claude/skills`, `~/.claude/commands/commands`, `~/.claude/plugins/cache/cache`, etc.). Each local provision re-triggered it.

Fix: all sync symlinks now use `ln -sfn` (`--no-dereference`) — directory, child, and individual-file links (including credentials and `keybindings.json`), so re-provisioning replaces the dest symlink instead of following it. `-n` is portable across GNU and BSD/macOS `ln`.

### Bug 2: skill-provisioned agents' primary skill leaked into every local agent
`_install_skill_locally` wrote the agent's primary skill (`code-guardian`, `fixme-fairy`) to global `~/.claude/skills/<name>/SKILL.md`. `_sync_user_resources` then symlinked all of `~/.claude/skills/` into *every* local agent's config dir, so the skill showed up in the skill list of every agent.

Fix: the skill is now installed into the agent's **own per-agent config dir** (`$CLAUDE_CONFIG_DIR/skills/<name>/SKILL.md`), strictly agent-scoped — mirroring the already-agent-scoped remote install. To let a per-agent real skill file coexist with the user's other (symlinked) skills, `skills/` is now synced via **child-level symlinks** (the same pattern `plugins/` already uses) rather than one dir-level symlink; the install breaks any same-named child symlink before writing so it can't corrupt the shared source. The local install is now **silent** (no interactive prompt), matching the always-silent remote install.

## Tests
- New: `_sync_user_resources` is idempotent across two provisions with no self-referential loops (covers both the dir-level branch and the child-level branch).
- New: `_install_skill_locally` writes into the per-agent config dir, and breaks a child symlink into shared skills instead of following it.
- Updated the existing `_install_skill_locally` tests for the new per-agent-dir signature; removed the now-obsolete interactive-prompt test.
- Verified locally (other test slices run in CI; a foreign session held the global pytest lock during this run): `skill_agent_test.py` 33 passed, `plugin_test.py` 199 passed, `test_ratchets.py` 56 passed, `ty check` clean on all four files. Also confirmed empirically that `ln -sf` reproduces the loop and `ln -sfn` fixes it.

## Notes / tradeoffs
- This only prevents *new* loops; pre-existing loops on machines in the wild are not self-healed (deliberate, per design discussion). A safe cleanup one-liner exists if needed.
- Child-level `skills/` symlinks mean a skill the user adds to `~/.claude/skills/` after provisioning won't live-propagate into already-provisioned agents (same tradeoff `plugins/` already has); it re-syncs on the next provision.
- Remote install is unchanged (remote hosts are isolated machines, so no cross-agent leak there).


joshalbrecht:
lgtm

PR #2047, opened 2026-06-09
fix release tests

evgunter:
Addresses the 72 new failures in the 2026-06-08 Release Tests run (analysis in `~/mngr/release_test_new_failures.local.md`), which diff against the v0.2.11 (2026-06-06) baseline.

### 1. `settings.local.toml` parse regression (dominant — ~65 of 67 new e2e failures)
The e2e fixture wrote a malformed `settings.local.toml` with a duplicated `type = "claude"` key under `[commands.create]` (introduced by the "273 agents" commit, after baseline). TOML refuses to overwrite an already-set key, so every `mngr` invocation in the e2e/tutorial release suite failed to parse its config and exited non-zero. Removed the duplicate line.

### 2. Two stale e2e config-set tests (genuine, platform-independent)
`test_config_set_unknown_key_fails` and `test_config_set_rejects_unknown_key` asserted the project `settings.toml` does not exist until a command writes it. The same commit made the fixture intentionally pre-seed that file with the pytest opt-in key (needed so tests that write project config pass the opt-in guard). Updated both tests to verify a rejected `config set` leaves the file unchanged and never persists the bad key, instead of asserting the file is absent.

### 3. robinhood streaming tests exited 2
`test_streaming.py` drives a real claude agent in tmux but lacked `@pytest.mark.tmux`, so the resource-guard PATH wrapper blocked tmux (exit 127), making the robinhood subprocess exit 2. Added `@pytest.mark.tmux` plus `timeout(660)` (a real agent run far exceeds the default 30s; the value sits just above the 600s subprocess timeout so robinhood's own captured error surfaces first). Confirmed in a clean env that robinhood streaming itself works.

## Validation
Release tests do not run in CI, so these were validated locally on macOS:
- 63 of the 67 e2e new-failure tests pass after the parse fix; the 2 stale tests above now pass; `test_scripting.py` is 6/6.
- The remaining 2 (`test_config_edit_editor_failure`, `test_create_with_env`) fail **only on local macOS** for environment reasons — `/bin/false` is absent on macOS (it exists on the Linux CI image), and a tmux pane-capture timing check that was present and passing on the Linux baseline. Both are resolved on the actual Linux release platform by fix #1. Left unchanged rather than risk a macOS-portability change that can't be validated against Linux locally.
- Both robinhood streaming tests get past the exit-2 guard (validated with a real ANTHROPIC_API_KEY).


joshalbrecht:
lgtm

PR #2048, opened 2026-06-09
fix missing dependency for release script

evgunter:
## Problem

The `publish` workflow has been failing at the **Verify versions and pin consistency** step since commit `e46ef4593` made `scripts/utils.py` import `UNPUBLISHED_PACKAGES` from `imbue.mngr`:

```
File ".../scripts/utils.py", line 13, in <module>
    from imbue.mngr.plugin_catalog import UNPUBLISHED_PACKAGES
ModuleNotFoundError: No module named 'imbue.mngr'
```

`scripts/verify_publish.py` was invoked with a bare `uv run`, which only syncs the root `imbue` project. The root project does not depend on `imbue-mngr` (it's a workspace *member*, not a dependency), so the package never lands in the CI `.venv` and the top-level import explodes. It passed locally only because dev `.venv`s are synced with `--all-packages`.

Example failing run: https://github.com/imbue-ai/mngr/actions/runs/27176781221/job/80227307251

## Fix

Use `uv run --all-packages` for all three `verify_publish.py` invocations so the workspace package is installed. This is the minimal change and preserves the single-source-of-truth design (the constant stays in `plugin_catalog.py`). The job's next step already builds every `libs/*` package, so an all-packages environment is in character.

A follow-up could decouple the release tooling from the runtime package entirely (e.g. a `force-include`d data file read by both the wizard and the script), but that's a larger change.

## Verification

In throwaway venvs simulating CI's clean environment:
- bare `uv run` → `ModuleNotFoundError` (reproduces the failure)
- `uv run --all-packages` → exit 0, full verification output


.github/workflows/publish.yml
```diff
@@ -47,15 +51,15 @@ jobs:
...
           else
-            uv run scripts/verify_publish.py
+            uv run --all-packages scripts/verify_publish.py
           fi
 
       - name: Build all packages
         run: |
           set -euo pipefail
-          pkg_dirs_output=$(uv run scripts/verify_publish.py --list-package-dirs)
+          pkg_dirs_output=$(uv run --all-packages scripts/verify_publish.py --list-package-dirs)
```
evgunter, line 58: `--all-packages` is only needed because it uses `UNPUBLISHED_PACKAGES`

joshalbrecht:
lgtm

PR #2050, opened 2026-06-09
aws template + scalar tuple stuff

evgunter:
## What

Adds an `aws` create-template to this repo's `.mngr/settings.toml` so `mngr create -t aws <name>` brings up a dev agent on an AWS EC2 host — the AWS analogue of the existing `modal`/`docker` dogfooding templates. Validated end-to-end against real EC2 + depot.

Exercising the path surfaced that the underlying AWS/`mngr_vps_docker` remote-build flow was not actually functional (the depot path was WIP/broken), so this PR also fixes the root-cause bugs, each with unit tests.

## The template (`.mngr/settings.toml`)

- `[create_templates.aws]` + a shared `[providers.aws]` block (region `us-west-2`, instance type `t3.large`, `auto_shutdown_minutes = 120`, `builder = "DEPOT"`). Only the operator-specific `allowed_ssh_cidrs` lives in gitignored `.mngr/settings.local.toml`; the two blocks merge per-field (`ProviderInstanceConfig.merge_with` honors `model_fields_set`).
- `build_arg` uses the real-source-tree shape (context `.`, **full history, no `--git-depth`**): the VPS backend builds remotely and seeds the in-container `/code/mngr/.git` via a thin-pack `git push`, which needs the base objects a shallow clone lacks (else "pack has N unresolved deltas / index-pack abnormal exit").
- Requires `DEPOT_TOKEN` + `GH_TOKEN` exported at create time. Also switches the `modal` template's `pass_env` to `pass_env__extend` (plain assignment would clobber an inherited `pass_env` like a profile's `ANTHROPIC_API_KEY`).

## Backend bug fixes (general, not AWS-specific)

Real bugs that real runs surfaced, fixed in the libraries (not papered over in the template):

- **env forwarding for streaming SSH** (`libs/mngr/.../hosts/outer_host.py`) — the old code prefixed env as a bare `KEY=VAL command`, which only applies to the first simple command, so for a compound `install && build` the var was lost before `build` ran. Now uses `export KEY=VAL && command` (extracted to a pure `_prepend_env_exports` helper). This made remote `depot build` fail with "missing API token" despite `DEPOT_TOKEN` being passed via `env`.
- **depot CLI invocation** (`libs/mngr_vps_docker/.../container_setup.py`) — the installer drops the CLI at `$HOME/.depot/bin` (off the non-interactive shell PATH), but the build invoked it by bare name → "command not found". Now invokes it by absolute path (double-quoted `"$HOME/.depot/bin/depot"`), and the idempotent install check uses `test -x` of that path.
- **provider-config type coercion** (`libs/mngr/.../config/loader.py`) — `_parse_providers` used `model_construct`, storing raw TOML scalars without coercion, so an enum field like `builder = "DEPOT"` stayed the string `"DEPOT"` and broke `builder is DockerBuilder.DEPOT` for a single-layer provider block (silently falling back off depot). Now uses `model_validate`, which coerces while still recording only the provided keys in `model_fields_set` (so per-field merging is unchanged). This also coerces nested-model provider fields (SSH static `hosts` → `SSHHostConfig`), subsuming `mngr/aws`'s dedicated coercion helper; wrapped in `ConfigParseError` to preserve a friendly parse-time error.

## Fixes found while validating against the merged `mngr/aws`

- **Wrong instance-type key** (`.mngr/settings.toml`) — the block used `default_plan`, which is not a field on `AwsProviderConfig` (the real field is `default_instance_type`). The loader silently strips unknown keys, so the host would have launched as the `t3.small` default instead of the intended `t3.large`. Fixed to `default_instance_type = "t3.large"`. (Found by autofix.)
- **Settings-narrowing on `allowed_ssh_cidrs`** (`libs/mngr/.../config/data_types.py`, `libs/mngr_aws/.../config.py`) — `mngr/aws` flipped the AWS `allowed_ssh_cidrs` default from empty to a non-empty `["0.0.0.0/0"]`. Combined with the committed `[providers.aws]` block + a developer's `settings.local.toml` tightening the CIDR to their own IP, the settings-narrowing guard rejected the local override as "dropping" `0.0.0.0/0` — failing **every** `mngr` command for anyone who tightened the value (the security-conscious case; it doesn't hit CI because `settings.local.toml` is gitignored). Fix: generalize the existing `StringDerivedTuple` narrowing exemption into a `ScalarTuple` marker base (+ a reusable `ScalarStrTuple` annotated type) that the guard treats as scalar-replace, and declare `allowed_ssh_cidrs` as `ScalarStrTuple` — combining CIDRs across config layers is never the intent, so a local override cleanly replaces the default. The marker is applied by `model_validate`, which is why the `_parse_providers` switch above is load-bearing here.

## Cleanup

- **Pure unknown-field check** (`libs/mngr/.../config/loader.py`) — renamed `_check_unknown_fields` → `_drop_unknown_fields` and made it return the cleaned config instead of mutating its argument in place; all call sites reassign the result.

## Merge note (`mngr/aws`)

This branch merged `mngr/aws`, which independently swapped the empty-`allowed_ssh_cidrs` behavior from fail-closed (raise) to warn-and-proceed. That made an earlier "actionable empty-CIDR error" (and the `AwsVpsClient.provider_name` field that keyed it) obsolete; both were dropped in the merge resolution.

## Verification

- End-to-end `mngr create -t aws` on real EC2 (us-west-2): provision → depot remote build → full-history source-sync push → agent start, all green; instance auto-destroyed.
- `mngr list`/`mngr create` with the committed `[providers.aws]` + a local `allowed_ssh_cidrs` now succeed (previously failed with a narrowing error) and resolve to the local IP.
- Unit + regression tests for every fix; the `data_types`/`loader`, `mngr_aws`, and `mngr_vps_docker` suites are green. `test-docker-electron` flaked once on a 600s Playwright navigation timeout (unrelated to this diff — touches no electron/minds code) and passed on re-run.

## Changelogs

Per-project entries for `dev/`, `libs/mngr`, `libs/mngr_aws`, and `libs/mngr_vps_docker`.


libs/mngr/imbue/mngr/hosts/outer_host.py
```diff
@@ -209,6 +209,24 @@ def _drain_ssh_stderr_into(state: _SSHStderrState) -> None:
...
+    path) rather than a bare ``KEY=VAL command`` prefix: a variable-assignment
+    prefix only applies to the single simple command it precedes, so it would
+    NOT survive a compound ``command`` containing ``&&`` / ``||`` / ``|`` (e.g.
+    ``install && depot build`` would lose the var before ``depot build``).
+    ``export`` sets it in the shell environment for the whole command.
+    """
+    if not env:
+        return command
+    exports = " ".join(f"export {shlex.quote(f'{k}={v}')} &&" for k, v in env.items())
+    return f"{exports} {command}"
```
evgunter, line 314: hm

.mngr/settings.toml
```diff
@@ -27,7 +27,25 @@ provider = "modal"
...
-pass_env = ["GH_TOKEN"]
+pass_env__extend = ["GH_TOKEN"]
+pass_host_env = ["EDITOR"]
+target_path = "/code/mngr"
+
+# Run an agent on this repo on an AWS EC2 host (the analogue of the `modal` template above).
+# Differs from modal because the VPS backend resolves `--file=` inside the uploaded context,
+# so modal's `.mngr/dev/build/` keyframe tarball can't be used: the context is `.` (the
+# worktree), which the mngr_vps_docker build flow clones + overlays with uncommitted changes
+# and the Dockerfile's `COPY . /code/mngr/` consumes. The image builds on depot's remote
```
evgunter, line 38: will be changed in https://github.com/imbue-ai/mngr/pull/2092

libs/mngr/imbue/mngr/config/data_types.py
```diff
@@ -54,23 +56,62 @@
...
+    return ScalarTuple(value)
+
+
+# A ``tuple[str, ...]`` config field that is semantically a single scalar value: a
+# higher-precedence settings layer that sets it replaces the whole value rather than
+# narrowing it. Use for fields where combining entries across config layers is never
+# the intent (e.g. an AWS provider's ``allowed_ssh_cidrs``). The narrowing exemption
+# only takes effect under ``model_validate`` (which runs the after-validator);
+# ``_parse_providers`` validates provider blocks, so provider-config fields qualify.
+ScalarStrTuple = Annotated[tuple[str, ...], AfterValidator(_coerce_to_scalar_tuple)]
```
evgunter, line 112: funky

libs/mngr_vps_docker/imbue/mngr_vps_docker/_outer_helpers_test.py
```diff
@@ -434,8 +434,14 @@ def test_build_image_on_outer_with_depot_uses_depot_build(monkeypatch: pytest.Mo
...
-    # Depot install + depot build, with --load (so the image lands on the daemon)
-    assert "depot build --load -t depot-image" in cmd
+    # depot build, with --load (so the image lands on the daemon), invoked by
+    # absolute path: the installer drops the CLI at $HOME/.depot/bin, which is not
+    # on the non-interactive shell's PATH, so a bare `depot` would not be found.
+    # Double-quoted so the remote shell expands $HOME at run time.
+    assert '"$HOME/.depot/bin/depot" build --load -t depot-image' in cmd
+    # The idempotent install check must also reference the same absolute path
+    # rather than a bare `command -v depot` (which never sees the off-PATH binary).
+    assert 'test -x "$HOME/.depot/bin/depot"' in cmd
```
evgunter: huh

joshalbrecht:
> Requires DEPOT_TOKEN + GH_TOKEN exported at create time.

Why do we need that?  Is this just for our own use / not required for anyone else? Is it really just for speeding up the builds?

joshalbrecht:
> The loader silently strips unknown keys

We should probably stop silent stripping? That seems bad / wrong

.mngr/settings.toml
joshalbrecht: approved there

joshalbrecht:
lgtm after conflicts and above comments

libs/mngr/imbue/mngr/config/data_types.py
joshalbrecht: agreed

I guess it works though

evgunter:
> > Requires DEPOT_TOKEN + GH_TOKEN exported at create time.

> Why do we need that? Is this just for our own use / not required for anyone else? Is it really just for speeding up the builds?

yeah, it's just in our own `.mngr/settings.toml`, and the builds are truly painful without caching from depot

evgunter:
the loader stripping was preexisting, so i'll make a new PR for it

PR #2053, opened 2026-06-09
Resolve plugin enablement from the agent's work_dir during provisioning

evgunter:
## Problem

Plugin enablement is enforced by blocking disabled plugins on the plugin manager, but that blocked-set is built **once, from the directory `mngr create` was invoked from**. The agent, at runtime, resolves config (and thus which plugins are enabled) from its **own work_dir**. When the two diverge, a plugin the work_dir disables can still fire its provisioning hooks.

This surfaced as the `claude_subagent_proxy` plugin writing its Claude `PreToolUse:Agent` / cleanup / reap hooks into a worktree whose own `.mngr/settings.toml` disables the proxy. At subagent-spawn time those hooks run `mngr create --type mngr-proxy-child` in that worktree, where the proxy is disabled and the child type is unregistered → `Unknown agent type 'mngr-proxy-child'` → the dispatcher loops forever.

Triggers where invocation-cwd ≠ work_dir config: creating from a temp dir, an `MNGR_PROJECT_CONFIG_DIR` override, `--from` another repo, or an in-place create like `mngr robinhood` (which provisions onto `cwd` with `create_work_dir=False`). The real incident came from `mngr robinhood` release-test runs that set `MNGR_PROJECT_CONFIG_DIR` to a temp config (no proxy disable) while cwd was a monorepo worktree (proxy disabled) — so the proxy defaulted to enabled at provision time and stamped its hooks into the worktree.

## Fix

- `create()` reads the agent's **work_dir** project/local config (through the host interface, so it works for local and remote work_dirs) and blocks whatever the work_dir disables, right before provisioning hooks fire. Blocking only adds, so this suppresses a plugin the work_dir disables but does not re-enable one already disabled at startup (the rarer converse case).
- The shared test plugin manager (`plugin_testing.plugin_manager`) now disables `claude_subagent_proxy` by default, so its real provisioning side effects no longer fire for agents provisioned in tests. The plugin's own tests invoke its hookimpls directly and are unaffected.
- `disabled_plugins_from_raw_layers` factored out of `read_disabled_plugins` so the disabled-set computation is shared across cwd-rooted and work_dir-rooted reads.

## Validation

- New unit tests for `disabled_plugins_from_raw_layers` (precedence + enable/disable).
- New integration test: an in-place create (`create_work_dir=False`) onto a repo whose config disables a plugin does **not** fire that plugin's provisioning hooks (while pre-block hooks still fire and the agent is still created).
- Guard test that the proxy is blocked in the shared test plugin manager.
- Manually reproduced the original bug (`cd /tmp && mngr create x --from <monorepo>` installed proxy hooks into a proxy-disabled worktree) and confirmed this branch's code no longer installs them.
- `libs/mngr` config + api + `mngr_claude` plugin tests: 1329 passed locally. Proxy unit tests: 22 passed. Ratchets: 56 passed.


evgunter:
i don't think we actually want this

PR #2054, opened 2026-06-09
support pi for real

evgunter:
agent comes up and reports lifecycle states:

https://github.com/user-attachments/assets/2a08a79c-1517-4c7b-8467-f46df243e230

(resuming sessions after `mngr stop` also works)

(there is no subagent support yet; the base harness does not support them) [CLAUDE: pi still ships no *built-in* subagents, but mngr now syncs the `agents/` definitions dir, so an installed subagent *extension* (e.g. `pi-subagents`) works under mngr.]

---

## Summary

Brings the `pi-coding` agent-type plugin from baseline-isolation to **antigravity-level parity**, per `specs/agent-plugin-parity/spec.md`. Before this PR pi-coding had per-agent config isolation, auth/settings sync, and install management, but reported WAITING forever, had no transcript, started fresh on every restart, delivered input by simulating tmux keystrokes, and stalled at pi 0.79's trust dialog.

pi exposes **no shell-hook surface** (the lever `mngr_claude`/`mngr_antigravity` use). Its only lifecycle-event surface is the in-process TypeScript **extension** API. So all the dynamic behavior is driven by a single mngr-owned extension (`resources/mngr_pi_lifecycle.ts`) that the plugin provisions and loads with `pi -e`. No background shell supervisor.

### What's added
- **RUNNING vs WAITING marker.** The extension touches `$MNGR_AGENT_STATE_DIR/active` on pi's `agent_start` and clears it on `agent_end`. No child/root gating is needed: pi has no in-process subagent tool (one agent loop per process), and a nested `pi` the agent spawns via the bash tool never loads this extension (it's loaded only via the explicit `-e` flag on the mngr launch), so it can't touch the marker.
- **Readiness sentinel.** Written on pi's `session_start` (which fires at TUI startup, pre-prompt, even with no model). `wait_for_ready_signal` polls **only** this file — *not* the `"pi v"` banner, which prints before the session can accept input (gating on it dropped the first message).
- **Message delivery by extension injection.** mngr appends each message (JSON-encoded, one per line) to a per-agent `pi_inbox` file; the extension's watcher injects it into the live session via `pi.sendUserMessage`. This replaces typing into the TUI via tmux `send-keys` + Enter (which pi intermittently swallowed, forcing a retry loop), keeps the TUI viewable (`mngr connect`), and behaves identically on local and remote hosts. `PiCodingAgent` therefore subclasses `BaseAgent` directly — none of `InteractiveTuiAgent`'s paste pipeline is used.
- **Raw + common transcripts**, emitted directly from `message_end` events (no JSONL re-parsing). `event_id` is seeded from the existing line count so ids stay unique across a resumed restart. `mngr transcript` reads the common envelope.
- **Conversation resume** via `pi --session <recorded file>`. The extension records the *main* session's file; `assemble_command` shell-evaluates a prelude that resumes it. Preferred over `--continue`, whose cwd-keyed "most recent" can be a nested pi's session.
- **Trust-dialog handling (pi 0.79+).** pi prompts "Trust project folder?" when the workspace (cwd or an ancestor) has pi project trust inputs -- a `.pi` config dir in the cwd or a `.agents/skills` dir in the cwd/an ancestor (not CLAUDE.md/AGENTS.md, which are loaded only once trusted). mngr pre-seeds pi's `trust.json` for the workspace so the agent never stalls, gated like claude/agy: silent under `mngr create --yes` or `auto_dismiss_dialogs`, an interactive prompt otherwise, and it extends the grant automatically when the source repo is already trusted.
- **Remote auto-install + subagent definitions.** On remote hosts (when allowed) pi is auto-installed from npm (`@earendil-works/pi-coding-agent`); local hosts still defer to the user unless `--yes`. The `agents/` resource dir is synced alongside `skills`/`prompts`/`extensions`/`themes` so an installed subagent extension finds its definitions.

### Design notes / robustness (validated against pi's docs, source, and community prior art)
- The extension **imports nothing from pi packages** (only Node built-ins + local structural types), dodging the documented jiti bare-specifier resolution hazard under global installs.
- Every handler fails safe (wrapped); throwing handlers don't disrupt pi's loop. **Async injects are `.catch`-guarded** so a rejected `sendUserMessage` can't surface as an unhandled rejection that crashes pi (Node's default `--unhandled-rejections=throw`). The extension sends no steer messages (avoiding the `-p` hang).
- The extension is loaded only via the explicit `-e` flag (placed under `commands/`, not the auto-discovered `extensions/` dir), so a nested pi never runs it. The inbox watcher seeds its offset at load (before the readiness sentinel), so a resumed restart never re-injects prior messages and none sent right after readiness is missed.

### Verification
- **Live end-to-end** on real pi 0.79 (under tmux, Anthropic model), driven entirely through the new injection path: the readiness sentinel fires; the marker flips RUNNING→WAITING around a real turn; the common transcript captured `user → assistant(tool=bash) → tool_result → assistant(ACK)` with the expected field-level shapes; stop/start resumed full context (the model recalled a planted secret) with transcript ids continuing across the restart; the trust dialog was pre-dismissed.
- **Release e2e** (`test_pi_agent_e2e.py`, real pi) drives the full mngr CLI lifecycle and validates the transcript record shapes against **real** pi events — the same contract the synthetic node-harness asserts (so the harness's assumed event shapes can't silently drift).
- **Node-harness behavioural tests** for the extension (marker gating, nested-session discrimination, transcript envelope incl. graceful degradation on unknown content/roles, restart id-uniqueness, readiness, session-file recording, inbox-watcher offset/skip-malformed/order, and async-rejection safety), gated on Node availability — the `.ts` is a resource, so it does not affect Python coverage.
- **Python unit tests** for `assemble_command`, the resume prelude, env knobs, transcript-mixin wiring, provisioning, trust seeding, and the inbox send/confirm path. Package suite: **116 passed**; `ty` clean. (The release e2e runs separately with a real key.)

### Not pi-relevant (handled by being absent)
pi has no onboarding NUX and no default tool-permission gate, so those parity dimensions are no-ops here.

### Spec
Extends `specs/agent-plugin-parity/spec.md` (dimension D) with a note on a related premature-idle case: the RUNNING/WAITING marker tracks the agent's turn/loop, not work it **detaches** from that loop (a `cmd &` or a `run_in_background`-style tool). The CLIs' idle signals correctly gate *in-loop* pending work (agy's `fullyIdle`-plus-root match; pi's foreground tools blocking the turn), but detached work is loop-scoped for claude, agy, and pi alike — claude's `Stop` hook excludes `CLAUDECODE=1` bash tasks, so it doesn't solve this for background bash either.

## What's left (deferred follow-ups)
These match the gaps every port past claude still carries:
- **Session preservation on destroy** (`on_destroy` / `on_before_host_destroy`) — claude-only so far.
- **Scheduled-deploy contributions** (`get_files_for_deploy` / `modify_env_vars_for_deploy`) — needed only if pi agents run under `mngr schedule`.
- **`waiting_reason` field generator** for the `mngr list` column.
- **Live streaming snapshot** (the in-progress assistant-text view).
- **Per-agent permission policy** — pi runs tools without a confirmation gate; a permission-gate extension would be additive.
- Cloning an agent does not carry the source conversation forward (shared limitation across ports).
- `auto_allow_permissions`-style flags are N/A (no default gate to bypass).


libs/mngr_pi_coding/imbue/mngr_pi_coding/resources/mngr_pi_lifecycle_test.py
```diff
@@ -0,0 +1,471 @@
...
+
+import pytest
+
+from imbue.mngr_pi_coding.plugin import _LIFECYCLE_EXTENSION_NAME
+from imbue.mngr_pi_coding.plugin import _load_resource
+
+# Node driver: load the extension, register its handlers via a fake `pi`, then
+# replay a JSON list of events. Each event is `{event, payload?, sessionId?,
+# sessionFile?}`; the fake ctx returns the given session id/file. Assertions
+# live in Python (below) against the files the extension writes.
```
evgunter, line 30: there's also a real e2e test

libs/mngr_pi_coding/imbue/mngr_pi_coding/plugin.py
```diff
@@ -1,28 +1,150 @@
...
+# Note: the ``npm`` dir (where ``pi install`` materialises npm-package extensions
+# under ``npm/node_modules``) is deliberately NOT synced. pi re-resolves the
+# ``packages`` list in the synced ``settings.json`` on every startup and
+# auto-installs any missing ones into the per-agent ``$PI_CODING_AGENT_DIR/npm``,
+# so npm-package extensions (e.g. ``npm:pi-subagents``) become available without
+# copying ``node_modules`` around, and each agent keeps an isolated install. The
+# only cost is a per-agent ``npm install`` (~1s) on first launch, which needs
+# network and so would not work on a fully-offline host; if that latency or the
+# offline case ever matters, copy ``npm`` into the per-agent dir here (copy, not
+# symlink -- a shared ``node_modules`` would race across concurrent startups).
```
evgunter, line 52: flagging

libs/mngr_pi_coding/imbue/mngr_pi_coding/plugin.py
```diff
@@ -1,28 +1,150 @@
...
+
+# After inboxing a message, wait up to this long for the turn to start (the
+# ``active`` marker to appear) as delivery confirmation. Covers the extension's
+# poll interval plus pi accepting the injected message.
+_TURN_CONFIRM_TIMEOUT_SECONDS: float = 16.0
+
+
+def _load_resource(filename: str) -> str:
+    """Load a resource file (e.g. the lifecycle extension) shipped in the wheel."""
+    return importlib.resources.files(_pi_resources).joinpath(filename).read_text()
```
evgunter, line 97: probably we should distribute our extension like normal people instead of doing this

libs/mngr_pi_coding/imbue/mngr_pi_coding/plugin.py
```diff
@@ -129,8 +356,91 @@ def get_pi_config_dir(self) -> Path:
...
+
+        ``MNGR_AGENT_STATE_DIR`` (where the marker, sentinel, and transcripts
+        live) is injected by the host; the extension reads it directly. The
+        remaining vars tell the extension which agent-type subdirectory to use
+        and whether to emit each transcript layer.
+        """
         env_vars["PI_CODING_AGENT_DIR"] = str(self.get_pi_config_dir())
+        env_vars["MNGR_PI_AGENT_TYPE"] = _PI_AGENT_TYPE
+        env_vars["MNGR_PI_EMIT_COMMON_TRANSCRIPT"] = "1" if self.agent_config.emit_common_transcript else "0"
+        env_vars["MNGR_PI_EMIT_RAW_TRANSCRIPT"] = "1" if self.agent_config.emit_raw_transcript else "0"
```
evgunter, line 372: feels like we might be able to pass info to the extension in a nicer way

libs/mngr_pi_coding/imbue/mngr_pi_coding/plugin.py
joshalbrecht: that's fine -- logically you cannot copy dependencies to another host anyway, because it can cross OS boundaries (ie, one host is OSX, another is linux)

libs/mngr_pi_coding/imbue/mngr_pi_coding/plugin.py
joshalbrecht: I'm not sure -- what are the downsides of this approach?

I actually don't hate this

libs/mngr_pi_coding/imbue/mngr_pi_coding/plugin.py
joshalbrecht: this honestly didn't seem terrible to me

joshalbrecht:
lgtm!

leaving this for you to merge so that we can talk about the follow ups that matter -- I think these would be good:

- more thorough e2e release tests (like, a bunch, to make sure that everything works + continues working)
- Live streaming snapshot (the in-progress assistant-text view) (and we should generically talk about how to get this working across different agents, ie, what the mngr-side interface is for it)
- waiting_reason field generator for the mngr list column (is this a thing?)
- Session preservation on destroy (on_destroy / on_before_host_destroy) — claude-only so far.
- Cloning an agent does not carry the source conversation forward (shared limitation across ports).

claude suggested these but idk that I care that much:
- Scheduled-deploy contributions (get_files_for_deploy / modify_env_vars_for_deploy) — needed only if pi agents run under mngr schedule.
- Per-agent permission policy — pi runs tools without a confirmation gate; a permission-gate extension would be additive.
- auto_allow_permissions-style flags are N/A (no default gate to bypass).

evgunter:
all of those followups are already on my list!

PR #2055, opened 2026-06-09
support opencode for real

evgunter:
agent can receive messages, marks lifecycle states, and is resumable

https://github.com/user-attachments/assets/3b01af0c-3c68-4d02-b55d-bb01d5e871fb

(there are no background tasks to complicate lifecycle states)

---

Brings the `opencode` agent type from a bare `BaseAgent` shell up to first-class `InteractiveTuiAgent` support, at roughly `mngr_antigravity` parity (bring-up milestones 1-4 from `specs/agent-plugin-parity/spec.md`).

OpenCode is architecturally unlike Claude Code / Antigravity -- a client-server app with SQLite-backed sessions and no POSIX-sh hook mechanism -- so this uses OpenCode's own extension points: an in-process TypeScript plugin and its config-dir env vars.

## What's included (milestones 1-4)
- **Lifecycle RUNNING/WAITING** -- an in-process TS plugin (auto-loaded from the per-agent config dir) maintains the `active` marker. Subagent-aware: the clear is gated on the *root* session, so task-tool subagents (child sessions) keep the agent RUNNING until the whole turn finishes. This is simpler than Antigravity's `fullyIdle` matching because OpenCode's per-session status already encodes "is the root done".
- **Conversation resume** across stop/start via `opencode --continue` (resumes the most recent root session from the per-agent SQLite store, which survives the hard kill `mngr stop` performs).
- **Transcripts** -- the plugin writes the raw transcript in-process (no `sqlite3` dependency on remote hosts); a backgrounded converter produces the common transcript `mngr transcript` reads.
- **Per-agent isolation** via `OPENCODE_CONFIG_DIR` + `XDG_DATA_HOME` (the preferred config-dir shape; no `$HOME` relocation), shared auth via a symlink to `~/.local/share/opencode/auth.json` (one login authenticates all agents), per-agent model/permission policy.
- New config: `config_overrides`, `sync_global_config`, `symlink_auth`, `auto_allow_permissions`, `emit_common_transcript`.

## Verification
Verified live against **opencode 1.16.2**: marker flips RUNNING->WAITING around a turn; `--continue` resumes the same root session (no new root); raw + common transcripts produced with the correct schema; dotted `~/.mngr/...` cwd accepted (no symlink workaround needed, unlike agy); `TUI_READY_INDICATOR = "ctrl+p commands"` confirmed. Unit/integration tests: **101 passed**, 98% coverage on `imbue.mngr_opencode` (`ty` + `ruff` clean). The TS plugin's marker logic is verified live (not unit-testable in CI without bun); the converter is unit-tested via subprocess.

## Not yet done (to be addressed in follow-ups)
Carried gaps, same as `mngr_antigravity`:
- Session preservation on destroy (online + offline).
- Scheduled-deploy file/env contributions (`mngr schedule`).
- `waiting_reason` listing field generator.
- Live streaming snapshot.
- Clone carrying the source conversation forward.
- Subagent idle-gating is verified by design + live for the root path, but not exercised live with a real subagent (the test-tool path is hard to force with the free test model).

## Heads-up
OpenCode **self-upgrades**, so the installed version is a moving target (it upgraded 1.2.10 -> 1.16.2 mid-development). The integration is written to tolerate both event shapes (`session.status` and the deprecated `session.idle`). Version pinning / install management (parity dimension P) is a natural follow-up.


joshalbrecht:
summary about the self upgrade means that we should allow users to prevent that properly, like with claude:
- https://opencode.ai/docs/config/#autoupdate
- https://github.com/anomalyco/opencode/issues/1793

It's important (even for our own testing and sanity) to have version-locked coding agents. Anything else is kinda madness in CI

joshalbrecht:
largely lgtm, again leaving open to talk about follow ups (probably want the same follow ups here as for pi agent)

one extra idea for follow ups on these: after the follow ups, would probably be good to make an agent go investigate whether there are any missing features, divergences, bugs, etc in these integrations. I suspect a round or two of that could really help polish the extension!

(and after that a bit of "actually trying it out" would probably go a long ways too, just in case there's something obviously busted)

evgunter:
i have already done several rounds of looking for missing features and divergences, and some of the resulting fixes are coming soon!

evgunter:
i also have tried them out--i flagged a few PRs recently as having been made with the new agent types

PR #2056, opened 2026-06-09
support codex for real

evgunter:
supports sending messages, lifecycle state detection including waiting for subagents, and (not pictured) resuming. no visible indication is given when subagents finish, but the video shows the state remaining `RUNNING` briefly after the turn with subagents stops, and then switching to `WAITING`

https://github.com/user-attachments/assets/41188bdc-8e39-407f-b98c-cf87f0267df7

---

## Real support for the `codex` agent type (OpenAI Codex CLI)

Brings the `codex` stub up to milestones 1-4 of the [agent-plugin parity](specs/agent-plugin-parity/spec.md) work, as a new `libs/mngr_codex/` plugin. Codex is the closest CLI yet to Claude Code (same hook model, a config-dir override env var, file auth, resume-by-id, append-as-you-go session JSONL), so this follows the `mngr_claude` shape. The full source-verified investigation is in [`specs/agent-plugin-parity/codex-investigation.md`](specs/agent-plugin-parity/codex-investigation.md).

### What's implemented (milestones 1-4)
- **Per-agent `CODEX_HOME` isolation** — each agent gets its own config dir (injected only on the codex process), user's real `$HOME` untouched. No HOME relocation, no workspace symlink (codex accepts the dotted `~/.mngr/...` cwd).
- **Shared auth** — per-agent `auth.json` symlinks to the user's `~/.codex/auth.json` (codex writes it in place and reloads-before-refresh, verified against source), so one login authenticates every agent and refreshes propagate. `cli_auth_credentials_store = "file"` is pinned so codex never uses a `CODEX_HOME`-keyed keyring entry.
- **Lifecycle marker (RUNNING/WAITING), subagent-aware** — the subtle part. Codex subagents (`spawn_agent` multi-agent) run **asynchronously**: the root `Stop` hook fires while subagents are still running, with no ordering guarantee and no `fullyIdle` signal (verified against codex source). So mngr does **not** just clear on `Stop`; it recomputes the `active` marker under a portable `mkdir` lock so it stays RUNNING until the root turn **and** every in-flight subagent are done — present iff `codex_root_active` (set on `UserPromptSubmit`, cleared on the root `Stop` for the recorded root session) **or** any `codex_subagents/<agent_id>` (maintained by the `SubagentStart`/`SubagentStop` hooks). Whichever of the root `Stop` or the final `SubagentStop` runs last performs the clear. (Unlike claude, whose Task subagents are *synchronous* — so `mngr_claude` can just decline to hook `SubagentStop`; codex can't.) A recorded root `session_id` guards against a nested `codex` process sharing the home. Limitation: backgrounded OS processes are untrackable (no codex hook).
- **Conversation resume** — the hook captures the root `session_id`; `assemble_command` shell-evaluates `codex resume <id>` (codex's rollout JSONL survives the `mngr stop` SIGKILL).
- **Transcripts** — raw rollout JSONL streamed verbatim + converted to the agent-agnostic common transcript (`response_item` history, ignoring `event_msg` display-duplicates; tool calls paired by `call_id`).
- **Trust + hook-bypass, consent-gated** — seeds the work dir as a trusted project (skips codex's folder dialog) and passes `--dangerously-bypass-hook-trust` so mngr's hooks run. Because trusting the workspace also lets codex load repo-local `.codex/hooks.json`, the bypass is gated together with workspace trust (prompt / `--yes` / `auto_dismiss_dialogs`), mirroring `mngr_claude`/`mngr_antigravity`.
- **Readiness** — codex's `SessionStart` fires lazily (on first prompt, not launch — openai/codex #15269), so readiness uses the `InteractiveTuiAgent` banner poll on `/model to change`.
- **Update surfacing (notify + opt-in auto-update)** — codex's own startup "Update available!" prompt is **blocking** and would intercept the first pasted message, so mngr pins `check_for_update_on_startup = false` and surfaces updates itself instead. At provision it compares `codex --version` to the `latest_version` codex already recorded in `~/.codex/version.json` (**no network call** — codex refreshes that file on its own ~20h schedule), and when outdated either runs `codex update` (the opt-in `auto_update`), prompts interactively, or logs a non-blocking notice. `codex update` self-detects the install method (brew/npm/standalone), so mngr needs no per-method logic; `--yes` never triggers a global upgrade. New config: `check_for_updates` (default `true`), `auto_update` (default `false`).

`codex` is moved out of mngr core (the `default_plugins` `BaseAgent` stub) into the plugin catalog as a recommended INDEPENDENT-tier agent type. The codex **tutorial e2e tests** (`test_create_basic.py`, `test_agent_types.py`) are updated accordingly — codex is a real plugin now, not a fakeable command stub, so they create it with `--no-auto-start` (real-run coverage lives in the plugin's own release test); this also fixed a pre-existing duplicate `type = "claude"` key in the e2e fixture's settings template that those tests had been masking.

### Verification
- **Unit/integration:** full `mngr_codex` suite green (167 passed, coverage ≥95%); affected core tests migrated and passing; `ty` type-check and ratchets clean.
- **Live end-to-end** (`mngr create … codex`, real binary): provision → readiness detection → message send → marker toggled to WAITING on completion → `session_id`/`transcript_path` captured → raw+common transcripts produced (correct user/assistant/tool mapping, no duplicates) → **`mngr stop`/`start` resumed the prior conversation with context intact**. Update probe verified live against the real binary (installed-vs-cached version parsed correctly; outdated → notice/prompt/`codex update` per config).
- **Automated release test** (`test_codex_agent_e2e.py`, run locally — release tests don't run in CI): drives the real `mngr` CLI end to end — create → reply captured in transcript → `mngr stop`/`start` → follow-up answered from pre-restart context (proving resume carried the conversation) → destroy. Fully isolated (`MNGR_HOST_DIR`, a throwaway `CODEX_HOME` seeded from the real auth, and a private tmux server); skipped unless `codex` + auth are present.

### Model note
Codex picks the account's default model, and a ChatGPT-account login rejects some `*-codex` slugs; set the `model` config field (e.g. `gpt-5.5`) if codex's default fails. See the lib README.

### Not yet implemented (follow-ups, relative to `mngr_claude`)
- Session preservation on destroy (online + offline)
- Deploy/scheduling contributions (`mngr schedule`)
- Field generators (`waiting_reason`) — feasible for codex since it fires `Notification`/`PermissionRequest`, unlike agy
- Streaming snapshot (live in-progress view)
- Installing codex when it is absent (update *surfacing*/auto-update is implemented above; auto-installing a missing binary is not)
- An **app-server-backed agent variant** (a second agent type, like `mngr_claude`'s `headless_claude`): drive codex over the `codex app-server` JSON-RPC protocol for programmatic messaging (no `tmux send-keys`) + a `codex --remote` TUI viewer (still watchable in tmux) + clean `initialize`-based readiness. We built the **TUI first** because — like `claude -p` — the app-server lacks full feature parity on a ChatGPT login: the backend gates some `*-codex` models on the client identity (`originator`), which the first-party TUI gets but an honestly-identifying programmatic client does not, and the only way to close that gap is to spoof `clientInfo.name='codex-tui'`, which **OpenAI's ToS disallow**. So the app-server is a *complement* for the API-key/honest-entitlement cases, not a replacement. Design, verified mechanics, and the ToS caveat are documented **in the plugin** at `libs/mngr_codex/README.md` ("Future direction: an app-server-backed agent variant").


libs/mngr/imbue/mngr/e2e/tutorial/test_agent_types.py
```diff
@@ -108,15 +108,14 @@ def test_create_codex_positional(e2e: E2eSession) -> None:
         # you can specify the agent type as the second positional argument to create:
         mngr create my-task codex
     """)
+    # codex is a real agent-type plugin now (not a command-driven stub), so it
+    # can't be faked with a `command` override. Create it without launching
+    # (--no-auto-start) and auto-approve workspace trust (-y), which exercises the
+    # positional-argument type resolution without needing a codex binary or auth
+    # on this host. The real codex run is covered by the mngr_codex release test.
```
evgunter, line 149: hm

libs/mngr_codex/imbue/mngr_codex/test_codex_agent_e2e.py
```diff
@@ -0,0 +1,279 @@
...
+    # so a mis-set path can never take down the user's real tmux server.
+    assert tmpdir_str.startswith(_TMUX_TMPDIR_PREFIX), (
+        f"refusing to kill-server for unexpected TMUX_TMPDIR {tmpdir_str}"
+    )
+    socket_path = tmux_tmpdir / f"tmux-{os.getuid()}" / "default"
+    kill_env = os.environ.copy()
+    kill_env.pop("TMUX", None)
+    kill_env["TMUX_TMPDIR"] = tmpdir_str
+    subprocess.run(["tmux", "-S", str(socket_path), "kill-server"], capture_output=True, env=kill_env)
+    shutil.rmtree(tmux_tmpdir, ignore_errors=True)
```
evgunter, line 114: might be an overreaction (it killed my tmux server at some point)

libs/mngr_codex/README.md
```diff
@@ -0,0 +1,137 @@
...
+preservation on destroy, deploy/scheduling contributions, field generators (`waiting_reason`),
+the streaming snapshot, and installation/version management.
+
+## Future direction: an app-server-backed agent variant
+
+This agent drives the codex **TUI** by `tmux send-keys` (paste + Enter), with banner-poll
+readiness. That works, but it's fragile (screen-scraping) and codex's `SessionStart` fires
+lazily, so there's no clean pre-input readiness signal. Codex offers a much cleaner surface we
+should adopt in a **follow-up** as a *second* agent type (mirroring `mngr_claude`'s
+`claude` + `headless_claude`): the **app-server**.
```
evgunter: flagging

libs/mngr_codex/README.md
```diff
@@ -0,0 +1,137 @@
...
+So the app-server variant must **identify honestly** (mngr's own client name) and use whatever
+models that identity is legitimately entitled to. For the gated `*-codex` models in app-server
+mode, authenticate with an **API key** (OpenAI's documented path for programmatic workflows) --
+do not spoof the TUI. The genuine `codex` TUI agent in this plugin remains the legitimate way
+to use `*-codex` models on a ChatGPT-subscription login (it really *is* the TUI).
+
+(Driving codex programmatically via the app-server on a single user's own ChatGPT login is
+itself fine -- it's a first-party feature and OpenAI staff have called such use "permissive";
+the line we don't cross is identity-spoofing to defeat the model gate, plus the usual no
+credential-sharing / no multi-tenant-proxying / no rate-limit-bypass.)
```
evgunter, line 169: this is kind of burying the lede--we didn't do this because, like `claude -p`, the app server version doesn't have full feature parity (unless you tell it you're the codex tui, which seems not allowed)

specs/agent-plugin-parity/codex-investigation.md
```diff
@@ -0,0 +1,194 @@
+# Codex CLI investigation (for `mngr_codex`)
+
+Verified findings about the OpenAI Codex CLI (Rust `codex`, **v0.138.0**), mapped to the
+[parity dimensions](./spec.md). Source claims are checked against the openai/codex repo at
+tag `rust-v0.138.0` (commit `c18e9f4`) and confirmed live where noted. This is the design
+basis for the `libs/mngr_codex/` plugin.
```
evgunter, line 6: probably should delete this file

joshalbrecht:
idk what this means but it sounds scary:

> (codex's rollout JSONL survives the mngr stop SIGKILL).

joshalbrecht:
this seems needlessly complicated:

> Update surfacing (notify + opt-in auto-update) — codex's own startup "Update available!" prompt is blocking and would intercept the first pasted message, so mngr pins check_for_update_on_startup = false and surfaces updates itself instead. At provision it compares codex --version to the latest_version codex already recorded in ~/.codex/version.json (no network call — codex refreshes that file on its own ~20h schedule), and when outdated either runs codex update (the opt-in auto_update), prompts interactively, or logs a non-blocking notice. codex update self-detects the install method (brew/npm/standalone), so mngr needs no per-method logic; --yes never triggers a global upgrade. New config: check_for_updates (default true), auto_update (default false).

at the very least, our default should probably be "do NOT update you"

in fact, we may want to switch all of our coding agents to that -- if they're run via mngr, they should always have auto updates disabled (the whole point of mngr is programmatic interaction, which means that it automatically updating on you is too weird)

joshalbrecht:
however, auto *installation* probably should be implemented here (and for other agents), bc it's important for bootstrapping hosts to actually be able to work

joshalbrecht:
(same comment about follow ups as the other harnesses as well)

joshalbrecht:
also, I couldn't really review the code for this branch -- the diff is way too large in github? seems like something went wrong

evgunter:
> idk what this means but it sounds scary:
> > (codex's rollout JSONL survives the mngr stop SIGKILL).

it just means the transcript doesn't die if the session gets killed instead of ending normally

evgunter:
diff is fixed. i'm not sure about the codex updating thing--if the default were to ignore updates, i probably just wouldn't get the updates if i were using codex exclusively through mngr, which is not what i want; and more broadly i think as a user i'd be surprised for mngr to modify the default behavior in that way

PR #2057, opened 2026-06-09
Coerce static SSH host entries during config loading

DanverImbue:
## Problem

A customer reported that statically-configured SSH hosts didn't work: after adding an SSH provider and a host to `settings.toml`, `mngr` "wasn't able to resolve the host." Investigation showed every host-enumerating command (`mngr list`, `mngr connect`, `mngr create <agent>@<host>.<pool>`, `mngr exec`, ...) crashed while building the SSH provider instance with:

```
AttributeError: 'dict' object has no attribute 'key_file'
```

## Root cause

`_parse_providers` builds provider configs with `model_construct` (which skips validation/coercion, so unset top-level fields stay `None` for config-layer merging). `SSHProviderConfig.hosts` (`dict[str, SSHHostConfig]`) is the **only** nested-model provider-config field, so its entries were left as raw TOML dicts. `FrozenModel` does not set `revalidate_instances`, so `MngrConfig.model_validate` never re-coerced them either. The backend then did `host_config.key_file` on a `dict` and crashed. The *dynamic*-hosts path (`SSHHostConfig.model_validate`) was unaffected — only the static `settings.toml` path was broken, and no test exercised it.

## Fix

`_parse_providers` now coerces nested pydantic-model provider-config fields after `model_construct`, via a small generic helper (matching the existing `_normalize_tuple_fields_for_construct` precedent). Only fields present in the raw config whose annotation references a model are touched, so scalar fields and the None-for-unset merge semantics are unaffected. A malformed host entry now surfaces a clear `providers.<pool>.hosts` config error instead of a late `AttributeError`.

## Tests

- Added `test_static_hosts_from_settings_dict_are_coerced` — drives the real parse path (`parse_config` → `build_provider_instance` → `get_host`), reproducing the crash and asserting host entries become `SSHHostConfig`. (Existing SSH tests hand in already-built `SSHHostConfig` objects, which is why they missed this.)
- Verified end-to-end via the real `load_config` + `get_all_provider_instances` (the call that crashed): the pool builds, the host discovers and resolves, and `key_file` expands correctly.
- Local: full `providers/ssh` dir (47), full `config` dir (414), and `utils/test_ratchets.py` (56) all pass.

## Docs

- New `libs/mngr/docs/core_plugins/providers/ssh.md` (host schema, dynamic hosts, the `NAME@HOST.PROVIDER` usage form, limitations).
- Registered `ssh` in `libs/mngr/docs/concepts/providers.md` with a working config example.

### DIFF
--- a/libs/mngr/changelog/danver-investigate-ssh.md
+++ b/libs/mngr/changelog/danver-investigate-ssh.md
@@ -1,0 +1,15 @@
+Fixed a crash that made statically-configured SSH hosts unusable, and added documentation for the SSH provider.
+
+- Static `[providers.<pool>.hosts.<host>]` tables defined in `settings.toml` previously crashed every
+  host-enumerating command (`mngr list`, `mngr connect`, `mngr create <agent>@<host>.<pool>`, ...) with
+  `AttributeError: 'dict' object has no attribute 'key_file'`. Provider configs are built with
+  `model_construct` (to keep unset top-level fields `None` for config-layer merging), which does not
+  coerce nested values, so each host entry stayed a raw dict instead of an `SSHHostConfig` and blew up
+  as soon as the backend touched it. The config loader now coerces nested pydantic-model fields (the
+  only one today being the SSH provider's `hosts` map) after `model_construct`, so static SSH hosts load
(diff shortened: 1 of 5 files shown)


joshalbrecht:
lgtm, thanks for the fix!

PR #2058, opened 2026-06-09
Prevent fake Latchkey usages

hynek-urban:
Set `LATCHKEY_DISABLE_COUNTING` in the minds snapshot script. (Maybe it will help reducing fake recorded Latchkey usage events once a snapshot gets re-created.)


joshalbrecht:
lgtm!

PR #2059, opened 2026-06-09
Preserve known_hosts_file when expanding SSH key_file

DanverImbue:
## Summary

Fixes a pre-existing bug in the SSH provider where `known_hosts_file` was silently dropped whenever `key_file` was set on a statically-configured host, then consolidates the duplicated path-expansion logic.

`SSHProviderBackend.build_provider_instance` expanded the `key_file` path by rebuilding `SSHHostConfig` while re-listing only `address`/`port`/`user`/`key_file`. Any host that configured **both** `key_file` and `known_hosts_file` lost `known_hosts_file` (it became `None`), which silently disabled strict host-key checking for that host.

## What changed

- **The fix:** introduced `SSHHostConfig.with_expanded_key_file()` — a pure frozen-object method that updates only `key_file` via the type-safe `model_copy_update`, preserving every other field (notably `known_hosts_file`, and any future fields).
- **Consolidation:** both the static path (`backend.py:build_provider_instance`) and the dynamic path (`instance.py:_read_dynamic_hosts`) previously hand-rolled this expansion independently — the static copy was the source of the bug, the dynamic copy was correct only by coincidence. Both now call the single shared method, so neither can silently drop a field if `SSHHostConfig` gains one in the future.

## History note

This was originally opened as a stacked follow-up on `danver/investigate-ssh` (which fixed static SSH host-config loading — the bug here was unreachable before that fix). That branch has since merged to `main`, so this PR now targets `main` directly; its diff is just the seven SSH files.

## Testing

- Red-first regression test in `backend_test.py` (static path) and a matching test in `instance_test.py` (dynamic path), both asserting `known_hosts_file` is preserved and `key_file` is expanded.
- New `config_test.py` unit-tests `with_expanded_key_file()` directly, including a guard test asserting every field except `key_file` is preserved.
- `just test-quick "libs/mngr/imbue/mngr/providers/ssh"`: 51 passed.
- Ratchets: 56 passed; repo-wide `test_no_type_errors`: passes.


joshalbrecht:
lgtm!

PR #2060, opened 2026-06-09
Address Josh's PR #1937 review comments

qi-imbue:
## Summary

Follow-ups on the three review comments Josh left on https://github.com/imbue-ai/mngr/pull/1937:

- `mngr gc --provider X` against an empty/unavailable provider now still runs gc against any other selected providers, but exits non-zero so the user sees that their explicit request was not fully honored. The skipped providers are listed in the summary. Post-destroy gc (which uses `--all-providers` internally and tolerates skips) is unchanged.
- Dropped the `-a` / `--all` flag from `mngr message`; the tutorial and CLI examples now broadcast via `mngr list --ids | mngr msg -`. Users who relied on `-a` should switch to the pipe form (optionally with `--include` / `--exclude` to scope the broadcast).
- Dropped `--no-ensure-clean` from the agent-type e2e tutorial tests. The e2e fixture now gitignores the per-test project config directory (where `mngr config set` writes), so the working tree stays clean. Removed `@pytest.mark.rsync` from those tests (its only purpose was satisfying the resource guard for the rsync path that `--no-ensure-clean` happened to trigger) and `@pytest.mark.modal` (which the tests do not actually exercise).
- Drive-by: fixed a pre-existing duplicate `type = "claude"` line in the e2e fixture's seeded `settings.local.toml` that was breaking every release-tier tutorial test with "Cannot overwrite a value".

## Test plan

- [x] `just test-quick libs/mngr` (4331 passed, 3 skipped)
- [x] `just test libs/mngr/imbue/mngr/cli/test_gc.py` (4 passed, including the new `test_gc_exits_non_zero_when_explicit_provider_is_empty`)
- [x] `just test libs/mngr/imbue/mngr/cli/message_test.py` (20 passed; `--all`-related tests replaced by a regression check that `-a` / `--all` are now rejected with "No such option")
- [x] `just test libs/mngr/imbue/mngr/api/message_test.py` (10 passed; API parameter retained for now)
- [x] `just test libs/mngr/imbue/mngr/e2e/tutorial/test_agent_types.py -k 'not yolo'` (5 passed locally; the yolo test trips on macOS lacking `/bin/true` — pre-existing, unrelated to this PR, will run fine in Linux CI)
- [ ] Wait for CI release tests to confirm the tutorial e2e tests still pass end-to-end on Modal/Docker.


libs/mngr/imbue/mngr/cli/gc.py
qi-imbue, line 1: This does the job, although boolean returns are a little bit confusing...

qi-imbue (commented):
The gc fix is a bit meh, but no idea how to do it better. Everything else looks quite good

libs/mngr/imbue/mngr/cli/gc.py
```diff
@@ -326,30 +345,41 @@ def _emit_jsonl_summary(result: GcResult, dry_run: bool) -> None:
...
+            # An explicitly-requested provider that is empty (e.g. a fresh
+            # Modal per-user environment with nothing to collect) or
+            # unavailable is still recorded as an error: the user asked us to
+            # gc it specifically and we did not. Continue so the remaining
+            # providers' work still happens; the CLI exits non-zero at the end.
             try:
                 providers.append(get_provider_instance(name, mngr_ctx))
             except ProviderEmptyError as e:
                 logger.debug("Skipping provider {} (empty -- nothing to gc): {}", name, e)
+                skipped_errors.append(f"provider {name} is empty: {e}")
```
joshalbrecht: should this be logger.error?

joshalbrecht:
lgtm! only one small suggestion

libs/mngr/imbue/mngr/cli/gc.py
qi-imbue: The `logger.debug` _below_ should be a `logger.error`, and this one should not even be recorded in `skipped_error` at all IMO - an empty provider with nothing to do is fine for GC

PR #2062, opened 2026-06-09
Fix file sharing

hynek-urban:
- Fix a bug that primarily manifested on macOS that caused file sharing to fail for Glenn in the first place.
    - (`wsgidav` converts mounts root points to lowercase)
- Allow people to easily override the requested path (something that Glenn attempted to do but couldn't).
- Let agents know when they're trying to request access to a file path that cannot be fulfilled (because it lies outside of the mounted root).

browse_buttons


joshalbrecht:
lgtm!

PR #2064, opened 2026-06-09
Apply create-template setting/setting__extend entries to config (#1914)

### LINKED ISSUE #1914: Create-template `setting`/`setting__extend` entries are silently dropped (never applied to config)
## Summary

Create-template `setting` / `setting__extend` entries are silently dropped — they never reach the resolved `MngrConfig`. A template can declare `setting__extend = ["providers.docker.docker_runtime=runsc"]`, the value flows into the create command's `--setting` param, but it has **no effect** on the config the command actually uses. Direct CLI `-S/--setting` flags work fine; only the *template-provided* ones are lost.

This contradicts `apply_create_template`'s own contract, whose docstring states template settings use "the same operator suffix recognised in TOML, `--setting`, and env vars" — i.e. they're meant to behave exactly like `--setting`.

## Root cause

In `libs/mngr/imbue/mngr/cli/common_opts.py`, `setup_command_context` applies settings to the config **exactly once**, and it does so **before** create templates are applied:

1. ~line 209-217: `apply_settings_to_config(mngr_ctx.config, initial_opts.setting, ...)` — `initial_opts` is parsed straight from the CLI args, so `initial_opts.setting` is the **CLI `-S` flags only**.
2. ~line 234: `apply_create_template(ctx, updated_params, mngr_ctx.config)` appends the template's `setting__extend` entries into `updated_params["setting"]` (→ `opts.setting`).
3. Nothing ever re-applies `opts.setting`. `apply_settings_to_config` is called only at step 1.

So template-contributed `setting` entries land in `opts.setting` after the single application point and are never merged into `mngr_ctx.config`. `get_provider_instance` / the command body then read provider config from `mngr_ctx.config`, which never saw them.

## Why it went unnoticed

The only prior use of a template `setting__extend` for provider config was `providers.docker.is_enabled=true` (e.g. forever-claude-template's docker create-template). That is a **no-op regardless of this bug**: `get_provider_instance(name)` (`libs/mngr/imbue/mngr/api/providers.py`, ~line 28-37) builds an explicitly-named provider directly from `config.providers[name]` with **no `is_enabled` check**. The `is_enabled` filter lives only in `list_provider_names_to_load` (~line 148), which is used for provider *discovery/listing*, not for explicit `@host.<provider>` targeting.

So an explicit create like `mngr create system-services@host.docker --template docker` never cared whether `is_enabled=true` actually applied — the provider is resolved by name either way. `docker_runtime` is the first template-set provider field whose absence is observable (the container comes up under `runc` instead of `runsc`), which is how the latent bug surfaced.

## Reproduction

Direct `-S` works:

```
mngr create system-services@x.docker -S providers.docker.docker_runtime=runsc
# -> docker run ... --runtime runsc   (correct)
```

Via a create template, it does not. With:

```toml
# .mngr/settings.toml
[providers.docker]
is_enabled = false

[create_templates.docker]
provider = "docker"
setting__extend = ["providers.docker.is_enabled=true", "providers.docker.docker_runtime=runsc"]
```

```
mngr create system-services@x.docker --template docker
# -> docker run ...   (NO --runtime; docker_runtime=runsc silently ignored)
```

Note: `mngr config get providers.docker.docker_runtime` is **not** a reliable probe here — `config get`/`config list` don't traverse provider-subclass fields (they only surface base `ProviderInstanceConfig` fields like `is_enabled`), so they report "Key not found" even when the value is set. Confirm via the actual resolved provider config / the emitted `docker run` command instead.

## Impact

- Any create template that tries to configure anything via `setting`/`setting__extend` is silently ineffective. The failure is silent (no error, no warning), so misconfiguration is easy and hard to diagnose.
- Concretely blocked: setting provider-config fields (e.g. `docker_runtime`, `builder`, resource defaults) per-template. Workaround is to put provider config in the `[providers.<name>]` block instead (which loads via the normal config file layer), but template `setting__extend` should work as documented.

## Suggested fix

Apply the template-contributed settings after `apply_create_template`, e.g. run `apply_settings_to_config` on the post-template `updated_params["setting"]` (rather than only the pre-template `initial_opts.setting`). Preserve precedence so direct CLI `-S` still wins over template settings (same-key, later-in-list wins in `apply_settings_to_config`, so order CLI entries after template entries). Add a regression test: a create template that sets an observable config value via `setting__extend` and assert it lands in the resolved config.

## Affected code

- `libs/mngr/imbue/mngr/cli/common_opts.py` — `setup_command_context` (single `apply_settings_to_config` call, applied pre-template) and `apply_create_template` (docstring promises `--setting`-equivalent behavior)
- `libs/mngr/imbue/mngr/api/providers.py` — `get_provider_instance` (no `is_enabled` gate) vs `list_provider_names_to_load` (has the gate), which is why the prior `is_enabled=true` use masked the bug

(Separately and minor: `mngr config get`/`config list` not traversing provider-subclass config fields is a cosmetic gap that made this harder to diagnose.)

joshalbrecht:
## Summary

Fixes #1914. Create-template `setting`/`setting__extend` entries flowed into the create command's `--setting` param but were never merged into the resolved `MngrConfig`, so they had no effect (e.g. `providers.docker.docker_runtime` set by a template was silently ignored). Direct CLI `-S` worked; only template-provided settings were lost.

Root cause: `setup_command_context` applied `--setting` to the config exactly once, *before* `apply_create_template` ran, so the template-appended setting entries never reached the config.

## Changes

- `cli/common_opts.py`: fold the post-template `setting` list into the config at the end of the create pipeline. The combined list is re-applied against the pristine pre-`--setting` config, so CLI `-S` still wins over template settings (CLI entries are last in the list) and `__extend` operators are not doubled.
- Template (or command-default) `setting` entries targeting `commands.*` or `create_templates.*` now raise `ConfigParseError` -- those sections are resolved before template settings are applied, so the value could never take effect. Direct CLI `-S` for those keys is unaffected.
- Template `setting` entries go through the same settings-narrowing guard as `--setting`.
- `cli/config.py`: `config get` and `config list --all` now dump with `serialize_as_any=True`, so provider-subclass fields (e.g. `docker_runtime`) are visible instead of reporting "Key not found".

## Testing

- New unit tests in `common_opts_test.py`: template `setting__extend`/bare `setting` reach the config; provider-subclass field via template; CLI `-S` wins over template; `commands.*`/`create_templates.*` rejection; CLI `-S commands.*` still works on create; template-setting narrowing raises.
- New unit tests in `config_test.py`: `config get`/`config list --all` surface a provider-subclass field.
- `just test-quick` on both files: 160 passed. `ruff check`/`ruff format`/`ty check` clean on changed files.
- Manually verified the real CLI `mngr config get providers.local.host_dir` now returns the value.

Design doc: `blueprint/template-setting-not-applied/plan-template-setting-not-applied.md`.

### DIFF
--- a/blueprint/template-setting-not-applied/plan-template-setting-not-applied.md
+++ b/blueprint/template-setting-not-applied/plan-template-setting-not-applied.md
@@ -1,0 +1,78 @@
+# Plan: Fix create-template `setting`/`setting__extend` entries being silently dropped (issue #1914)
+
+## Overview
+
+- **Bug:** A create template can declare `setting__extend = ["providers.docker.docker_runtime=runsc"]`. The value flows into the create command's `--setting` param but never reaches the resolved `MngrConfig`, so it has no effect. Direct CLI `-S/--setting` works; only template-provided settings are lost.
+- **Root cause:** In `libs/mngr/imbue/mngr/cli/common_opts.py`, `setup_command_context` applies settings to the config exactly once (lines 209-217, `apply_settings_to_config(... initial_opts.setting ...)`), and does so *before* `apply_create_template` runs (line 234). The template appends its `setting__extend` entries into `updated_params["setting"]` *after* that single application point, so they land in `opts.setting` but are never merged into `mngr_ctx.config`. `get_provider_instance` reads provider config from `mngr_ctx.config`, which never saw them.
+- **Why latent so long:** The only prior template `setting__extend` provider use was `providers.docker.is_enabled=true`, which is a no-op regardless of this bug — `get_provider_instance` (`api/providers.py:57`) resolves an explicitly-named provider directly from `config.providers[name]` with no `is_enabled` gate (that gate lives only in `list_provider_names_to_load`, `:166`). `docker_runtime` is the first template-set provider field whose absence is observable.
+- **Fix approach (decided in Q&A):** Keep the existing line-209 CLI `-S` application so `apply_config_defaults`/`apply_create_template` still see CLI settings. Then, at the end of the pipeline (create only), re-run `apply_settings_to_config` on the **originally-loaded** config with the combined `template-then-CLI` setting list, replacing the config returned in `mngr_ctx`. This preserves "CLI `-S` wins over template" precedence and avoids double-applying `__extend` (because we re-apply against the original config, not the already-modified one).
+- **Guardrail:** Template-contributed `setting` keys that target `commands.*` or `create_templates.*` cannot take effect (those are consumed *before* the re-application point), so they raise `ConfigParseError` rather than failing silently. Validation is scoped to template-contributed entries only — direct CLI `-S commands.*` keeps working.
+- **Secondary fix:** `config get` and `config list --all` cannot surface provider-subclass fields (e.g. `docker_runtime`) because `MngrConfig.providers` is typed as the base `dict[..., ProviderInstanceConfig]` and `model_dump(mode="json")` serializes by declared type, dropping subclass-only fields. Add `serialize_as_any=True` at the two `config.py` call sites. This made the bug hard to diagnose (the natural probe reported "Key not found").
+
+## Expected behavior
+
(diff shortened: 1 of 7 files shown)


joshalbrecht:
lgtm

PR #2065, opened 2026-06-09
mngr_lima: run agents directly in the VM as root (drop docker-in-VM)

joshalbrecht:
## Summary

Replaces the Lima provider's docker-in-VM mode with **direct-in-VM execution as root**, and consolidates the agent setup so a Lima workspace gets the *same* dependency install as a Dockerfile-built one.

- Deletes the entire `is_host_in_docker=True` code path from `mngr_lima` (no Docker daemon, in-VM image build, nested container, outer-host/container-SSH/port-forward/snapshot-helper machinery).
- Adds `LimaProviderConfig.is_run_as_root`: mngr injects a root client key, enables key-based root login, and SSHes in as root, so the agent (uid 0) can `apt install` / write anywhere with no `sudo` — matching the docker/vps_docker providers (the VM is the isolation boundary).
- `is_run_as_root` requires the btrfs additional-disk layout (`is_host_data_volume_exposed=false`); the invalid combination with the 9p bind-mount layout is rejected at config construction.
- Removes the docker-mode config fields and severs the `mngr_lima → mngr_vps_docker` dependency.
- btrfs-based backups keep working: `/mngr` stays on a btrfs disk and the root agent snapshots it directly (`btrfs_local`), so the snapshot-helper/trigger-volume is no longer needed.

The shared-setup consolidation lives in **forever-claude-template** (separate repo, committed on the same branch in a worktree): the Dockerfile's setup is factored into `scripts/setup_system.sh`, `install_dependencies.sh`, `build_workspace.sh`, which the Dockerfile `RUN`s at its `COPY` boundaries and the Lima create template runs directly in the VM via `extra_provision_command` after the repo is synced in.

## Notes / manual verification

Per the plan, Lima end-to-end behavior is verified **manually** (Lima is too slow for CI): agent runs as root + `apt install` with no sudo, and `btrfs_local` backups work. The unit suite (128 passed, 67% cov) covers config validation, the root-SSH resolution, the run-as-root provisioning script, and host-record round-trips.

Design doc: `blueprint/lima-direct-vm-no-docker/plan-lima-direct-vm-no-docker.md`.


joshalbrecht:
lgtm

PR #2068, opened 2026-06-09
Convert the desktop client sidebar into a floating menu

gdenisov:
## Summary

Converts the minds desktop **sidebar into a floating "workspace menu"** that matches the Figma "Space switcher menu" design, and folds it into the existing modal infrastructure.

### UX
- Floating dark panel (rounded, shadowed) anchored under the titlebar's Projects button, with a colored dot per workspace.
- Each row reveals a **settings gear on hover** (and, in Electron, an "Open in new window" button); the current workspace's row shows them at all times.
- Bottom of the menu: **"New workspace"** (→ /create) and **"Manage account(s)" / "Log in"** — the account control moved here out of the titlebar.
- Behaves like a modal: click outside or Escape to dismiss.
- Uniform spacing grid (4px panel padding, 2px between entries, bare full-width divider), menu anchored 2px left / 2px below the trigger button. Panel background is flat black *for now* while the color treatment is iterated.

### Architecture
- **Three WebContentsViews per window instead of four**: the sidebar no longer has its own view — `/_chrome/sidebar` loads into the same `modalView` that hosts `/inbox`, reusing its transparent background, Escape handling, and titlebar-drag suppression.
- **Caller-driven positioning**: the chrome page passes the trigger button's `getBoundingClientRect` + an offset; the menu anchors at `trigger.bottom-left + offset` (no baked-in `is_mac` logic). Move the button and the menu follows.
- **One shared row builder** (`window.mindsSidebarRow.buildRow`) consumed by the Electron menu (sidebar.js), the browser menu (chrome.js), and the styleguide — the row carries no outer positioning, so the parent owns spacing. Removes the previously-duplicated `renderWorkspaces` row markup.
- macOS titlebar left padding widened 72→76px so the first button's hover clears the traffic lights.
- An incoming permission request no longer clobbers an open menu (auto-open is gated on no modal already being visible).

### Out-of-scope
- `@pytest.mark.flaky` on `libs/mngr_modal/.../test_exec_cwd_override_on_modal` for a Modal sandbox-setup race (`snapshot_and_shutdown` / ephemeral-app quota) unrelated to this branch.

### Follow-ups (deferred)
- Rename the internal "sidebar" identifiers to "workspace menu" (it's no longer a sidebar) — separate PR.
- De-duplicate the settings-gear SVG: the row uses a hand-scaled 16px copy of the same glyph the home page renders via `Icon24` (24px); unify by exposing `ICONS_24` to JS.

## Test plan
- `apps/minds/imbue/minds/desktop_client/` subtree: 863 passing; ruff + ratchets clean.
- Manual: open the menu from a workspace and from Home; hover non-current rows (gear fades in); click outside / Escape to close; "New workspace" / account rows navigate; trigger a permission request while the menu is open (menu stays, badge updates).


joshalbrecht:
I don't have any particular complains about the code, but see slack for discussion

PR #2070, opened 2026-06-09
Fix leaked Docker state containers from test runs

joshalbrecht:
## Problem

Local test runs leak Docker **state containers** (the singleton Alpine file-server the docker provider uses for its named-volume state). Four were found running ~41h, named \`mngr_<uuid>-docker-state-<user_id>\`, \`provider=docker\`.

Two compounding causes:
1. The state container was created lazily by *any* read-only operation touching the state volume (e.g. \`discover_hosts\` -> \`list_all_host_records\`). So \`mngr list\`/\`gc\`/\`cleanup\` and tests running \`mngr\` subprocesses created a persistent state container even with zero docker hosts.
2. The session-cleanup safety net's prefix detector had an off-by-one (\`container_name[4:]\` stripped only \`mngr\`, not \`mngr_\`), so it never matched \`mngr_<hex32>-\` test containers and never swept them.

## Fix

**Source (mirrors the Modal backend's \`is_for_host_creation\` gate):** the state container is the docker analog of the Modal per-user environment. \`DockerProviderBackend.build_provider_instance\` now raises \`ProviderEmptyError\` when the state container doesn't already exist and \`is_for_host_creation\` is \`False\`, so the provider loader skips docker for read-only commands instead of materializing a state container. Only \`mngr create\` creates state. Read-only commands no longer spin up a state container on a machine with zero docker hosts (intended behavior change).

**Test infra:**
- Replaced the buggy \`_looks_like_test_prefix\` with a regex-based \`looks_like_mngr_test_container_name\` in \`utils/env_utils.py\` (unit-tested in \`env_utils_test.py\`).
- Added a \`worker_docker_state_prefixes\` registry that the docker fixtures populate; \`session_cleanup\` now **fails** the suite (after cleaning) on a surviving state container under one of this worker's own prefixes, and **warn-and-cleans** unattributable containers from other concurrent workers/sessions (parallel-safe).
- Added a \`docker_sdk\` regression test: read-only construction raises \`ProviderEmptyError\` and creates no state container.

## Verification (local)

- \`ty check\`: pass; ruff: clean.
- docker unit + \`docker_sdk\` (incl. new regression test): pass.
- docker lifecycle (40) + gc (4): pass; state-container count stable (no leaks, no false-positive failures).
- mngr ratchets (56) + env_utils unit (29): pass.
- The fixed safety net swept the 4 pre-existing leaked containers during a run.

A blueprint plan documenting the investigation/design is in \`blueprint/docker-state-container-leak/\`.


joshalbrecht:
lgtm

PR #2071, opened 2026-06-09
improve antigravity lifecycle and messaging by using statusLine

evgunter:
## Summary

Makes `mngr message` to an `antigravity` (`agy`) agent **confirmed** rather than blind, and replaces the fragile `set_active_marker.sh` / `clear_active_marker_when_idle.sh` marker-hook pair with agy's `statusLine` mechanism.

A single mngr-owned `statusline.sh`, seeded into the per-agent `settings.json` as agy's `statusLine` command (applied **last**, so it wins over `settings_overrides`), is now the single source of truth for agent lifecycle. agy invokes it on every agent-state change (JSON payload on stdin), and it:

1. Maintains the `active` marker that drives RUNNING vs WAITING — busy **iff** `agent_state` is not in `{idle, initializing, authenticating, ""}` (a denylist).
2. Records the root `conversation_id` in `root_conversation` for resume (the payload always reports the root, even while a subagent runs).
3. Fires `tmux wait-for -S "mngr-submit-<session>"` — the signal `mngr message` now waits on to confirm submission.

agy allows only one `statusLine` command, so a user-supplied one (in `settings_overrides` or the synced global settings) can't coexist with mngr's; the provisioner logs a warning naming the discarded command rather than dropping it silently.

agy's top-level `agent_state` already aggregates subagent activity (stays `working` continuously while a subagent runs), so this single check replaces the hand-rolled marker pair.

Submission switches from `send_enter_best_effort` to `send_enter_via_tmux_wait_for_hook` (the proven Claude strategy), waiting on `mngr-submit-{session_name}` with `queue_log_path_template=None` (the upstream fallback is hardcoded to Claude's `jq`/enqueue schema, which conflicts with agy's no-`jq` constraint). The conversation-id capture `PreInvocation` hook is retained — it is the only place subagent ids surface (for transcript scoping), since the statusLine payload only reports the root. Readiness is intentionally still gated by the banner poll.

## Live verification (agy 1.0.7, `Claude Sonnet 4.6 (Thinking)` to dodge Gemini quota)

Created a real antigravity agent via `mngr create` and exercised the full loop:

- **Provisioning correct**: `settings.json` carries the mngr-owned `statusLine`; `hooks.json` has only the capture `PreInvocation`; `statusline.sh` installed in `commands/`.
- **Confirmed submission**: `mngr message` returned in **14.2s** (waited for the busy-state signal), not an instant blind Enter.
- **Lifecycle**: agent showed **RUNNING continuously while the Coder subagent executed**, then **WAITING** when fully done; `mngr list` matched. The capture hook recorded both the root and subagent conversation ids.
- **Resume**: `mngr stop` + `mngr start` resumed the root conversation with **live context** — a follow-up ("first line of the haiku?") was answered from memory without re-reading.
- **Statusline renders** in the TUI (`* working` / `- idle`).

Observations (flagged, not blocking):
- Brief WAITING flickers occur at *step boundaries* within a turn (a ~3s root idle before the subagent spawned, and while the subagent was blocked on its own approval dialog). The subagent's actual execution window stayed continuously RUNNING, consistent with the planning finding; the flickers are genuine idle/human-wait gaps. This is the deliberate semantic of trusting `agent_state` directly.
- The rendered statusline currently shows `agent_state` (model field was empty in sampled payloads); cosmetic only.
- `--dangerously-skip-permissions` does not propagate to agy subagents (pre-existing, unrelated).

## Known edge (follow-up, not in scope)

A model that *refuses* a prompt (quota exhausted) never enters a busy state, so the submit-wait times out even though the prompt was enqueued. A `jq`-free, agy-specific enqueue fallback (grep `history.jsonl`) would close this.

## Testing

`just test-quick libs/mngr_antigravity` — 180 passed (incl. new `statusline_test.py`, updated config/plugin tests, type-errors, ratchets).


libs/mngr_antigravity/imbue/mngr_antigravity/plugin.py
```diff
@@ -534,6 +551,25 @@ def _provision_agy_home(self, host: OnlineHostInterface, host_home: Path, host_u
...
+        user_statusline = per_agent_settings.get("statusLine")
+        if user_statusline is not None:
+            logger.warning(
+                "Antigravity agent {} has a user-provided statusLine ({!r}) that mngr is overriding with "
+                "its lifecycle statusLine. agy allows only one statusLine command, and mngr's drives "
+                "RUNNING/WAITING detection and message-submission confirmation, so it must win. Remove the "
+                "statusLine from your settings/settings_overrides to silence this warning.",
+                self.name,
+                user_statusline,
+            )
```
evgunter: flagging

libs/mngr_antigravity/imbue/mngr_antigravity/conftest.py
```diff
@@ -1,12 +1,34 @@
...
+    """
+    messages: list[str] = []
+    handler_id = logger.add(lambda msg: messages.append(msg.record["message"]), level="WARNING", format="{message}")
+    try:
+        yield messages
+    finally:
+        try:
+            logger.remove(handler_id)
+        except ValueError:
+            pass
```
evgunter: huh i thought this was a shared fixture

libs/mngr_antigravity/imbue/mngr_antigravity/plugin.py
joshalbrecht: why can't the user statusline be blended / wrapped like we do for claude?

libs/mngr_antigravity/imbue/mngr_antigravity/conftest.py
joshalbrecht: that is suspicious, yes

likely unexpected warnings should fail tests anyway...

joshalbrecht:
the followup does seem worth fixing (as long as it's not too crazy)

joshalbrecht:
lgtm except for the above comments!

libs/mngr_antigravity/imbue/mngr_antigravity/plugin.py
evgunter: ah, this is actually outdated--it's already fixed!

libs/mngr_antigravity/imbue/mngr_antigravity/conftest.py
evgunter: fixed (shared fixture hadn't landed yet when this was written)

PR #2072, opened 2026-06-09
mngr: generalize TUI submission-confirm to an agent-supplied marker command

gnguralnick:
## What

Resolves the FIXME in `libs/mngr/imbue/mngr/agents/tui_utils.py` that flagged Claude-specific coupling in an otherwise agent-agnostic module.

The shared message-submission-confirm helper `send_enter_via_tmux_wait_for_hook` confirms a submission either via a tmux `wait-for` hook signal or, concurrently, by watching for a fresh "message accepted" event in the agent's transcript log. That second path was hardcoded to Claude Code's schema (a `queue_log_path_template` plus an `"operation":"enqueue"` grep + `jq` pipeline), coupling a generic module to one agent type.

## How

- `tui_utils` now takes an opaque, agent-supplied `accept_marker_command`: a shell snippet that prints a lexicographically-monotonic "accepted" token (empty if none yet). The module baselines it before Enter and watches for a newer value concurrently with the hook signal, holding no knowledge of any agent's marker schema. Renamed `_build_signal_or_enqueue_command` -> `_build_signal_or_marker_command` and dropped its now-unused `agent` parameter.
- `ClaudeAgent._build_accept_marker_command` (in `mngr_claude`) now builds the Claude-specific enqueue/grep/jq/transcript-path probe, so the coupling lives where the agent-specific knowledge belongs.
- The generated remote shell command for Claude is **byte-identical** to before -- this is a pure relocation, with no runtime behavior change.
- Updated docstrings/comments, the bare-tmux ratchet comment, and added a changelog entry for both `mngr` and `mngr_claude`.

## Testing

- `ClaudeAgent`'s probe is now covered by two behavioral tests (run the real probe against a fixture transcript: one asserts it returns the latest enqueue event's timestamp, skipped when `jq` is absent; one asserts an empty token when the log is missing), replacing a brittle substring-assertion test.
- `libs/mngr` tui_agent / tui_utils tests reworked to assert agent-neutrality (an arbitrary probe is embedded verbatim).
- Verified locally (`just test-quick`): `libs/mngr_claude` plugin_test 200 passed; `libs/mngr` tui_agent + tui_utils + ratchet tests 75 passed; `ty` type-check, `ruff format`, `ruff check` all clean.
- Full offload + acceptance suites left to CI.


gnguralnick:
Recreating cleanly: the branch had origin/main auto-merged in by the stop-hook orchestrator, which muddied the stacked diff. Reset the branch to just the 4 relocation commits on top of mngr/confirm-send-on-enqueue and opening a fresh PR.

PR #2073, opened 2026-06-10
mngr: generalize TUI submission-confirm to an agent-supplied marker command

gnguralnick:
Stacked on `mngr/confirm-send-on-enqueue`. Diff is exactly the 4 relocation commits (8 files under `libs/mngr` and `libs/mngr_claude`).

## What

Resolves the FIXME in `libs/mngr/imbue/mngr/agents/tui_utils.py` flagging Claude-specific coupling in an otherwise agent-agnostic module. The submission-confirm helper `send_enter_via_tmux_wait_for_hook` confirms either via a tmux `wait-for` hook signal or, concurrently, by watching the agent's transcript log for a freshly-recorded "message accepted" event. That second path was hardcoded to Claude Code's schema (a `queue_log_path_template` plus an `"operation":"enqueue"` grep + `jq` pipeline).

## How

- `tui_utils` now takes an opaque, agent-supplied `accept_marker_command`: a shell snippet that prints a lexicographically-monotonic "accepted" token (empty if none yet). The module baselines it before Enter and watches for a newer value concurrently with the hook signal, holding no knowledge of any agent's marker schema. Renamed `_build_signal_or_enqueue_command` -> `_build_signal_or_marker_command` and dropped its now-unused `agent` parameter.
- `ClaudeAgent._build_accept_marker_command` (in `mngr_claude`) now builds the Claude-specific enqueue/grep/jq/transcript-path probe.
- The generated remote shell command for Claude is **byte-identical** to before -- a pure relocation of the coupling, no runtime behavior change.
- Updated docstrings/comments, the bare-tmux ratchet comment, and added a changelog entry for both `mngr` and `mngr_claude`.

## Testing

- `ClaudeAgent`'s probe is covered by two behavioral tests (run the real probe against a fixture transcript: one asserts it returns the latest enqueue event's timestamp, skipped when `jq` is absent; one asserts an empty token when the log is missing), replacing a brittle substring-assertion test.
- `libs/mngr` tui_agent / tui_utils tests reworked to assert agent-neutrality (an arbitrary probe is embedded verbatim).
- Verified locally (`just test-quick`): `libs/mngr_claude` plugin_test 200 passed; `libs/mngr` tui_agent + tui_utils + ratchet tests 75 passed; `ty` type-check, `ruff format`, `ruff check` all clean.
- Full offload + acceptance suites left to CI.


joshalbrecht:
lgtm, will let you merge into the base branch

PR #2074, opened 2026-06-10
minds: let an agent spawn peer minds via the latchkey minds-api-proxy

gnguralnick:
## Summary

Enables an agent inside an mngr/minds workspace to spawn peer minds via a latchkey-gated `POST $LATCHKEY_GATEWAY/minds-api-proxy/api/create-agent`.

- Rides the latchkey gateway's existing `minds-api-proxy` extension and the central per-startup `MINDS_API_KEY`: the proxy injects `Authorization: Bearer <key>` on the agent's behalf — the agent never sees the credential, and nothing new is persisted to disk.
- `/api/create-agent`, its status, and its logs endpoints accept the central key as an alternative to the browser session cookie.
- Materializes a `minds` detent scope with three named permissions (`minds-create`, `minds-status`, `minds-logs`) gated to `domain=latchkey-self.invalid` AND path prefix `/minds-api-proxy/api/create-agent`, so a user clicking `any` can't escape into other gateway-self endpoints (per-agent `/api/v1` proxy paths, the permission-requests extension, ...). First spawn pops the standard "Peer minds" permission dialog; subsequent spawns by the same agent are silent.
- Latchkey UNKNOWN credentialStatus is now treated as proceed-and-trust in the grant flow. This covers catalog scopes that aren't registered latchkey services at all (like `minds`, served by a gateway extension): `latchkey services info minds` fails and degrades to UNKNOWN, which previously popped a spurious "Manual credential setup required" panel.
- Idempotent startup migration backfills the schema into existing per-host permissions files before the gateway starts.

History note: the branch originally implemented this with a persisted `auth/api_token` plus `latchkey services register minds` / `auth set minds`. Main meanwhile landed the `minds-api-proxy` + in-memory central key infrastructure (commits `d3c63cae9`, `a5ea1a810`, `a6cee64aa`), so the merge commit (`2d09cabed`) rearchitects the feature onto that path and drops the parallel mechanism (no persisted token, no `register_service`/`auth_set_header` CLI wrappers).

## Test plan

- [ ] CI passes (full offload suite + acceptance)
- [ ] Manually verify: fresh install → spawn a peer mind from an agent shell, observe the "Peer minds" permission dialog the first time and silent success thereafter
- [ ] Manually verify: status and logs endpoints reachable from within the agent through the minds-api-proxy path
- [ ] Manually verify: existing installs get the schema migration backfilled on next `minds run`


joshalbrecht:
superseded

PR #2076, opened 2026-06-10
unify agent transcript schemas + add shared e2e harness

evgunter:
Unifies the cross-cutting pieces of the three "for real" agent-plugin ports.

**Base branch:** `mngr/ts-plugins-merged` — a synthetic branch that is just the three ports merged onto `main` (no unification). Because of that base, **this PR's diff is the unification layer only** (the three ports are reviewed separately in the PRs below).

**The three ports (content of the base branch):**
- pi-coding: imbue-ai/mngr#2054
- opencode: imbue-ai/mngr#2055
- codex: imbue-ai/mngr#2056

## What this PR adds (the diff)

**1. Single canonical spec.** The three branches each edited `specs/agent-plugin-parity/spec.md` divergently; consolidated to one doc and updated it so opencode (HTTP client/server lever) and codex (third shell-hooks subagent-gating shape; update-dialog suppression; mngr-side update notify) are documented as real ports. No named agent type is a stub any more.

**2. Canonical common-transcript envelope.** `imbue.mngr.agents.common_transcript_records` is the single source of truth for the `user_message`/`assistant_message`/`tool_result` records every agent emits. Live conformance tests prove **all five emitters** (claude, antigravity, opencode, pi-coding, codex) produce records matching it, so the independently written emitters can't silently drift. The opencode one is also the first CI-runnable check of its in-process TS emitter.

**3. Parameterized release tests.** Shared lifecycle harness (`imbue.mngr.agents.agent_release_testing`) drives the common create → WAITING → message → transcript → stop/start resume → destroy arc; each plugin's release test is now a thin per-agent profile, so every agent is held to the same contract executably.

## Validation
- All unit/integration tests + ratchets green; import-layer contract satisfied (harness lives in `agents/`).
- **All three release tests pass end-to-end** against real binaries: opencode (free model), pi (with `ANTHROPIC_API_KEY`), and codex — codex's earlier post-restart-send failure was the update-available dialog blocking the paste, durably fixed by the plugin's `check_for_update_on_startup = false` pin (codex commit `afb010c21`).


joshalbrecht:
love this!

lgtm

not merging bc idk how you want to handle the branches, but consider it approved

joshalbrecht:
maybe one question here worth discussing -- do we want to chat about the exact common format? Are there other common formats or standards out there that we could leverage?

like, is there an RFC for this stuff yet? :)

PR #2078, opened 2026-06-10
gcp

evgunter:
first run will fail with something like this:

Screenshot 2026-06-13 at 10 56 55 PM

(at least if you merge https://github.com/imbue-ai/mngr/pull/2143 to make it red instead of normal)

then:

Screenshot 2026-06-13 at 10 57 56 PM

---

## GCP Compute Engine provider (`mngr_gcp`)

Adds a `gcp` provider backend that runs agents in Docker containers on GCE VMs, built on the shared `mngr_vps_docker` base exactly like the AWS EC2 provider. Base branch is `mngr/aws` (landing order AWS → GCP).

### ⭐ First-run onboarding (fixes the screenshot above)

The very first `mngr create --provider gcp` requires a one-time `mngr gcp prepare` (privileged firewall setup). **Previously** that need surfaced *mid-create*, buried under a `Host creation failed, attempting cleanup...` line (the screenshot) — after an SSH key had already been uploaded and was then deleted. A messy, alarming first impression for an entirely expected setup step.

**Now** the firewall rule's existence is checked as a read-only **pre-flight, before any provider write** (SSH key upload / instance creation). A user who hasn't run `prepare` gets the clean, actionable message immediately — nothing created, nothing cleaned up:

- In the shared `mngr_vps_docker` base, the `_validate_provider_args_for_create` hook now runs at the top of `create_host`, ahead of the first provider write, instead of partway through `_provision_vps`. (Benefits every VPS provider; the hook's docstring now documents the "before any write" contract.)
- GCP's override adds a `resolve_firewall()` read-only pre-flight that raises the `mngr gcp prepare` pointer when the rule is missing.
- Tests lock in the ordering invariant (validate runs before `upload_ssh_key`) and the GCP missing-firewall pre-flight.

Relatedly, `project_id` is no longer hard-required: when unset it falls back to the project ADC resolves (`gcloud config set project` / `GOOGLE_CLOUD_PROJECT`), and `mngr create` logs the inferred project. The "nothing resolved" and prepare-pointer errors now name the exact commands (`mngr config set …`, `gcloud config set project …`) and drop the dev-only `uv run` prefix.

### What's here
- **`libs/mngr_gcp/`**: `config.py` (`GcpProviderConfig`, ADC-only — no credential material), `client.py` (`GcpVpsClient` over `google-cloud-compute`), `backend.py` (`GcpProvider` + `GcpProviderBackend` + hookimpl), `cli.py` (`mngr gcp prepare`), full unit tests, ratchets, release test, README.
- **Credentials**: Application Default Credentials only (`google.auth.default()`); the only identifier is the non-secret `project_id` (optional — falls back to the gcloud/ADC default).
- **Firewall / IAM split** (matches AWS): privileged firewall creation lives in a one-time `mngr gcp prepare`; the hot `create` path only resolves the rule read-only and points at `prepare` if missing, so developers can create with instance-only IAM.
- **Auto-delete safety net**: `scheduling.max_run_duration` + `instance_termination_action=DELETE` (the GCE-native analog of AWS `InstanceInitiatedShutdownBehavior=terminate`).
- **Discovery** via the `mngr-provider` label; **SSH** via per-instance `ssh-keys` metadata + a direct cloud-init root-key write.

### Empirically validated (real VMs in `mngr-498908`)
- Standalone SSH milestone: VM boots, no-TOFU host key matches, root SSH works, `/var/run/mngr-ready` appears, Docker 29.5.1 installed.
- Full `mngr create` → exec → list → destroy lifecycle (the release test, run locally) — **passes**, no leaked resources.

### Two findings that drove shared-base changes
1. **GCE Debian images don't run cloud-init** (the guest agent ignores `user-data`). GCE **Ubuntu LTS** images do, so `default_source_image` defaults to `ubuntu-2204-lts`. To make the shared Docker install work there, `mngr_vps_docker.host_setup` now derives the Docker apt repo + pinned version from `/etc/os-release` at run time — **byte-identical on Debian** (Vultr/OVH/AWS unaffected), correct on Ubuntu.
2. **Guest-agent vs cloud-init race** could leave root without the key. `generate_cloud_init_user_data` gained an optional `authorized_user_public_key` written straight into root's `authorized_keys`; `_provision_vps` now always passes it (additive/idempotent for the other backends).

A real API bug was also caught and fixed: `InstancesClient.list()` takes `filter` on a `ListInstancesRequest`, not as a flattened kwarg (the unit-test fake was corrected to mirror the real API).

### Known limitation: `mngr stop` / `mngr start` are container-level (no VM-level stop/start yet)

`mngr stop` / `mngr start` stop and start the agent's **Docker container** over SSH — they do **not** stop the GCE VM, so a "stopped" agent's instance keeps running and billing. This is the current shared `mngr_vps_docker` behavior for every VPS provider, and the README now documents it (Usage note + Implementation details + Future improvements).

True VM-level stop/start (GCE `instances.stop`/`start`, ephemeral-IP rebind on resume, and offline label-based discovery of stopped instances — a Modal-like idle-paused-but-resumable lifecycle that actually stops compute billing) is a **follow-up**. The AWS-only `mngr/aws-stop` branch adds it for EC2 entirely within `mngr_aws` (it overrides `stop_host`/`start_host` and adds `stop_instance`/`start_instance` to its client; the shared base is untouched), so **GCP is not covered by it** and needs its own parallel implementation in `mngr_gcp`. Tracked under "Future improvements" in the GCP README.

### Notes
- Touches `libs/mngr_vps_docker` (OS-aware Docker, optional root-key injection, the earlier pre-create validation), root `pyproject.toml` (coverage), and regenerates `mngr create` CLI docs. Changelog entries added for each touched project.


libs/mngr_gcp/imbue/mngr_gcp/config.py
```diff
@@ -0,0 +1,210 @@
...
+DEFAULT_SERVICE_ACCOUNT_SCOPES: tuple[str, ...] = ("https://www.googleapis.com/auth/cloud-platform",)
+
+# Global Ubuntu 22.04 LTS image family. GCE image families are global (unlike
+# AWS AMIs, which are per-region), so a single string suffices -- no per-region
+# map. Ubuntu (not Debian) is the default deliberately: the stock GCE
+# ``debian-cloud`` images do NOT ship/run cloud-init, so the ``user-data``
+# metadata carrying the mngr bootstrap is silently ignored on them. The GCE
+# Ubuntu LTS images run cloud-init with the GCE datasource, so the shared
+# ``mngr_vps_docker`` cloud-init flow works unchanged. The family always
+# resolves to the latest published image.
```
evgunter: flagging

evgunter: (one assumes that a workaround is possible, but i wasn't sure if i should attempt it)

libs/mngr_gcp/README.md
```diff
@@ -0,0 +1,227 @@
...
+- **No automatic snapshot-on-create**: unlike `mngr_modal`, this provider does not snapshot GCE instances automatically. `GcpVpsClient.create_snapshot` / `list_snapshots` / `delete_snapshot` are implemented (boot-disk snapshots); you can call them manually via `mngr snapshot`, or write a plugin that hooks `on_host_created`.
+- **Stop/start operate on the container, not the GCE VM**: `mngr stop` / `mngr start` stop and start the agent's Docker container (Docker over SSH), not the GCE instance itself, so the instance keeps running and `compute.instances.stop` / `compute.instances.start` are not needed (they are intentionally absent from the IAM list above). This is the current shared `mngr_vps_docker` behavior across all VPS providers. VM-level stop/start is future work -- see below.
```
evgunter, line 195: will fix this + make the auto shutdown behavior not be "destoy without a trace" in a followup PR

libs/mngr_vps_docker/changelog/mngr-gcp.md
```diff
@@ -0,0 +1,12 @@
...
+
+- The pinned Docker install step now derives the apt repo (`download.docker.com/linux/$ID`) and the full apt version suffix (`~$ID.$VERSION_ID~$VERSION_CODENAME`) from `/etc/os-release` at run time, instead of hardcoding the Debian 12 / bookworm strings. The script *text* changed (the os-release source moved up, a `DOCKER_APT_VERSION` shell var was added, the repo URL is now `linux/$ID`), but on Debian 12 it renders the identical apt version (`5:29.5.1-1~debian.12~bookworm`) and repo URL (`linux/debian`) as before, so the existing Debian-family outers (Vultr / OVH / AWS) are behaviorally unchanged; the change lets the same step work on GCP's Ubuntu LTS images. `PINNED_DOCKER_APT_VERSION` is retained as the fully-rendered Debian 12 apt version string for any caller or test that needs the exact Debian value rather than the runtime-derived suffix.
+
+## cloud-init: optional direct root-key injection
+
+- `generate_cloud_init_user_data` gains an optional `authorized_user_public_key` parameter. When set, the key is written straight into root's `authorized_keys` by cloud-init, independent of the existing copy-from-default-user (`admin` / `ec2-user` / `ubuntu` / `debian` / ...) step.
```
evgunter: hm

libs/mngr_gcp/imbue/mngr_gcp/testing.py
```diff
@@ -11,20 +11,23 @@
...
 
 # Optional prefix release tests use for their agent names so leaked instances
-# (should the scanner ever fail) are still visually identifiable as test-owned.
-# Cleanup logic does NOT depend on this -- ``GcpVpsClient.create_instance``
-# labels pytest-launched instances with ``mngr-pytest-launched=true`` and the
-# conftest scanner filters on that label.
-GCP_TEST_NAME_PREFIX: Final[str] = "test-gcp-"
+# (should the scanner ever fail) are still visually identifiable as mngr-created
+# test instances. No "gcp" in the prefix: a leaked instance is already in a GCP
+# project, so naming it "gcp" would be redundant. Cleanup logic does NOT depend
```
evgunter: some comments here and elsewhere which should be cleaned up

dev/changelog/mngr-gcp.md
```diff
@@ -0,0 +1,7 @@
+## GCP provider support: root-level changes
+
+- Top-level coverage configuration adds `--cov=imbue.mngr_gcp` so the new package contributes coverage data.
+- `scripts/make_cli_docs.py` adds `gcp` to `SECONDARY_COMMANDS` so the `mngr gcp` operator command group gets generated docs (required by `help_formatter_test`).
+- `uv.lock` updated to add the new `imbue-mngr-gcp` workspace package and its dependencies (`google-cloud-compute`, `google-auth`, and their transitive deps).
```
evgunter, line 5: need to check if we can lean on the cli more than the sdk (if we would indeed want to do that)

libs/mngr/imbue/mngr/plugin_catalog.py
```diff
@@ -190,6 +190,18 @@ class CatalogEntry(FrozenModel):
...
+        package_name="imbue-mngr-gcp",
+        description="GCP Compute Engine provider backend plugin for mngr",
+        tier=PluginTier.INDEPENDENT,
+    ),
+    CatalogEntry(
+        entry_point_name="ovh",
+        package_name="imbue-mngr-ovh",
+        description="OVH Cloud VPS provider backend plugin for mngr",
+        tier=PluginTier.INDEPENDENT,
+    ),
```
evgunter, line 204: was ovh being omitted deliberate?

libs/mngr_aws/imbue/mngr_aws/test_release_aws.py
```diff
@@ -283,7 +284,6 @@ def test_provider_lifecycle_create_exec_and_destroy(
         # --force skips the destroy confirmation, so no stdin input needed.
         # Result is intentionally not checked: best-effort cleanup.
         _run_mngr(aws_test_settings_dir, temp_git_repo, "destroy", agent_name, "--force", timeout=120)
-        time.sleep(20)
```
evgunter, line 286: reran the release tests and it sure seems like this wasn't doing anything. cool

joshalbrecht:
> GCE Debian images don't run cloud-init (the guest agent ignores user-data)

what the hell?  But also, it doesn't really matter / isn't a reason to change the base image -- we still need to deal with the case where someone selects debian 12 instead (namely, because we do with minds anyway). The contract for our providers is that they bootstrap the stuff they need if it is missing, so, should probably just have GCP ensure that it works with either of those images?

dev/changelog/mngr-gcp.md
joshalbrecht: hmm, why is that?

(to be clear, I'm not opposed, just curious)

libs/mngr/imbue/mngr/plugin_catalog.py
joshalbrecht: "deliberate" is a bit more intentionality than I would attribute to it :-P

I didn't include it because I doubted anyone cared about it, but now that we have others, might as well

it's kinda jank though -- they have such limited VPS supply that people won't be able to, in practice, use it

libs/mngr_aws/imbue/mngr_aws/test_release_aws.py
joshalbrecht: lol

libs/mngr_gcp/imbue/mngr_gcp/config.py
joshalbrecht: see my other comment on the main body of the PR (I think there's an easy enough fix to this, and we should do it -- just run the script and make sure it's idempotent? worst case it runs twice and the second time is effectively a no-op?)

libs/mngr_gcp/imbue/mngr_gcp/testing.py
joshalbrecht: please do!

libs/mngr_vps_docker/changelog/mngr-gcp.md
joshalbrecht: this seems good / non-crazy?

like, this one of the nice reasons to have cloud-init (it means that there's no need for TOFU (trust on first use))

unless GCP gives you the SSH identity for the root user or something in a different way? But idk...

joshalbrecht:
lgtm after all of the above fixes and comments!

libs/mngr_vps_docker/changelog/mngr-gcp.md
evgunter: it seems sensible, i was just not sure if it belonged in the base class or just gcp

dev/changelog/mngr-gcp.md
evgunter: wasn't sure if there was a general principle behind why we don't use the sdk for claude (but it's moot here because it would indeed be annoying to just use the cli)

libs/mngr_gcp/imbue/mngr_gcp/testing.py
evgunter: done, hopefully

libs/mngr_vps_docker/imbue/mngr_vps_docker/instance.py
```diff
@@ -997,11 +1029,22 @@ def _provision_vps(
...
+        # cloud-init backends, which set the key pre-sshd.
+        with log_span("Waiting for expected VPS host key"):
+            self._wait_for_expected_host_key(
+                vps_ip, vps_host_public_key, timeout_seconds=self.config.ssh_connect_timeout
+            )
+
+        logger.log(
+            LogLevel.BUILD.value, "Waiting for host bootstrap to complete (Docker installation)...", source="vps"
+        )
+        with log_span("Waiting for host bootstrap (Docker install)"):
```
evgunter, line 1049: not ideal to have this be in the shared core, but it seems even worse to have a hook here that's only used by gcp

evgunter:
debian supported, hackily

libs/mngr_gcp/imbue/mngr_gcp/startup_script.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def generate_gce_startup_script(
+    host_private_key: str,
+    host_public_key: str,
+    install_gvisor_runtime: bool,
+    auto_shutdown_seconds: int | None = None,
+    authorized_user_public_key: str | None = None,
+) -> str:
+    """Generate a GCE ``startup-script`` for VPS provisioning (the cloud-init analog for GCP).
```
joshalbrecht, line 16: this doesn't make sense -- the abstraction is improperly inverted (mngr_vps_docker should NOT know about gce)

evgunter: nice catch, fixed

PR #2079, opened 2026-06-10
antigravity: use keychain on macos + do crimes to still support transcript based on the protobuf sqlite db

evgunter:
all the protobuf stuff is black magic i have no confidence in beyond the tests

---

## Summary

Two related fixes for the antigravity (`agy`) plugin, both caused by recent (~2-week-old) agy changes, plus the e2e release test that exercises them. Brings in `mngr/unify-ts-plugins` (the shared release harness + the agy e2e test) as the base, excluding that branch's commit that dropped the test.

### 1. macOS keychain barrier (the original report)
The per-agent `$HOME` relocation that isolates agy's config also hid `$HOME/Library/Keychains`, where agy's embedded-Chromium `os_crypt` resolves the login keychain holding its "Antigravity Safe Storage" key. With no keychain found, macOS raised a **modal** "A keychain cannot be found to store Antigravity Safe Storage" dialog that **blocks agy until dismissed** — hanging any unattended run and popping on every fresh agent interactively. Fix: `_provision_macos_keychain` symlinks the per-agent home's `Library/Keychains` to the user's real one on macOS (no-op on Linux). Mirrors the existing playwright-cache symlink and claude's linux-file/macos-keychain split. Validated live: the agent now creates/starts/runs a turn with zero keychain interaction.

### 2. Transcript pipeline ported to agy's SQLite store
agy **1.0.4 (2026-06-01)** switched its interactive conversation store from a per-conversation JSONL transcript (which `stream_transcript.sh` tailed, and which agy no longer writes) to a **protobuf SQLite `.db`** — so the streamer captured nothing on current agy (print mode still writes the JSONL, which masked the break). agy publishes no `.proto` schema, so the schema was **recovered from the binary's embedded `FileDescriptorProto`s** (documented + repeatable in `libs/mngr_antigravity/dev/README.md`).
- `stream_transcript.sh` is now a thin **python3-guarded** supervisor around a new self-contained decoder (`decode_agy_transcript.py`) — a dependency-free protobuf wire-walk (no `protobuf` lib, no shipped descriptors).
- It reads new terminal steps from each conversation `.db` and emits the **same record shape** the old JSONL had, so `common_transcript.sh` is unchanged apart from now accepting agy's clean (un-enveloped) user text.
- Decodes USER_INPUT / PLANNER_RESPONSE / ERROR text **and assistant tool calls** (name + args). Tool *results* are out of scope (agy records command output in unmapped step types; file-edit CODE_ACTION steps don't occur in practice).

## Validation
- Unit tests: 201 passing, incl. the keychain provision test, the decoder (decode + streaming semantics against synthetic `.db` fixtures) and the bash wrapper + python3 guard end-to-end.
- Decoder verified against real `~/.gemini` conversation `.db` files (user/assistant text + run_command tool calls).
- Release test (`test_antigravity_agent_full_lifecycle`): the keychain fix already unblocked creation/startup live; the full e2e re-run is pending the agy account quota reset (~15h) and will confirm the transcript port end-to-end.

## Decoder hardening + testing (follow-up pass)

A confidence pass on the "black magic": the protobuf wire-walk was treated as untrusted and adversarially probed/fuzzed, which surfaced three robustness bugs — all now fixed with regression tests. Two of the three are anti-*corruption*, not just anti-crash:

- **Out-of-range `created_at` no longer aborts the whole pass.** `time.gmtime` raises `OverflowError`/`OSError` for a `seconds` value outside the platform `time_t`; that exception escaped `run_once` (which catches only `sqlite3.Error`), so one bad step emitted nothing and — since the offset never advanced — blacked out *every* conversation on *every* cycle, permanently. `_iso_timestamp` now degrades a bad timestamp to `""`. This is best-effort and the one genuine "degrade instead of crash": `created_at` is isolated, informational metadata (the step's content is decoded from a separate field), so it cannot silently wrongify content — but it is unreachable under the *correct* schema, so if it ever fires it signals drift/corruption (the schema-verification release test is the real canary). Switchable to "skip the offending step" if a stricter fail-fast posture is preferred.
- **Truncated fixed-width (32/64-bit) fields are now detected as truncation.** Previously they yielded a short slice and *advanced the offset* — silently emitting corrupt data that was never retried. Now they raise `_TruncatedError` like the length-delimited path, so the step is skipped and retried once complete. Strictly reduces corruption risk.
- **A corrupt per-conversation offset file resets to the start** instead of crashing (`"--5"` passes a naive `isdigit` check but `int()` rejects it). This just makes one edge case consistent with the existing reset-to-`-1` behavior for all other unparseable offsets; the re-read's duplicates are deduped downstream by `event_id`.

Testing:
- 6 new decoder regression tests (decoder unit file 18 → 24). Full antigravity suite **213 passing**, ratchets **56 passing**, `ty` clean.
- **Real-db validation against agy 1.0.8**: decoded the live `~/.gemini` conversation stores with **zero decode exceptions** — clean text/timestamp/type round-trips, and (critically) the **`ChatToolCall` name/args path**, which the schema-verification release test *cannot* reach because its proto (`codeium_common.proto`) isn't recovered from the binary. Decoded `run_command` tool calls with well-formed JSON args, confirming `_TOOL_CALL_NAME`/`_TOOL_CALL_ARGS` against real data.
- Confirmed the schema-verification release test (`test_decoder_field_map_matches_installed_antigravity_binary`) passes against the installed agy 1.0.8 binary.
- Observation (not a bug): real transcripts contain step types 21 (`run_command`) and 23 (`checkpoint`) but no type 5 (`CODE_ACTION`), so the `CODE_ACTION` → `tool_result` pairing never fires in practice — tool *calls* surface, tool *results* (command output) do not. Matches the documented "out of scope" note above.
- Next: drive a *live* (actively-written) `.db` to exercise the streaming / non-terminal-status gating path, which synthetic fixtures and static real dbs can't fully cover.

## Notes
- **PR base is provisional**: this branch stacks `mngr/agy-statusline` + a merge of `mngr/unify-ts-plugins` + this work, so the diff is large. Retarget if a different base is preferred.
- Draft until the e2e re-run confirms green.


libs/mngr_antigravity/changelog/mngr-agy-keychain.md
```diff
@@ -0,0 +1,32 @@
...
+hid that directory, so agy found no keychain and macOS raised a modal "A keychain cannot be
+found to store Antigravity Safe Storage" dialog -- which blocked agy until dismissed,
+hanging any unattended run and popping on every fresh agent interactively.
+
+Provisioning now symlinks the per-agent home's `Library/Keychains` to the user's real one
+on macOS (Linux has no such keychain and Chromium falls back to its file-based store, so
+nothing changes there). agy is already in the keychain item's ACL from interactive logins,
+so it reads the key with no prompt. This mirrors the existing playwright-cache symlink --
+another HOME-relative, machine-shared resource -- and the claude-style "straightforward on
+Linux, keychain on macOS" split.
```
evgunter, line 14: i'm not sure if this has security implications!

evgunter: claude has reassured me that it does not, but still flagging

libs/mngr_antigravity/changelog/mngr-agy-keychain.md
```diff
@@ -52,3 +52,14 @@ Simplified the common-transcript converter's user-message handling to match agy'
 store: it now passes through the clean typed text agy records in `CortexStepUserInput.query`,
 dropping the speculative `<USER_REQUEST>...</USER_REQUEST>` envelope stripping that existed
 only for the retired agy-1.0.0 JSONL format.
+
+Hardened the SQLite decoder against malformed/drifted protobuf so a single bad step can no
+longer take down transcript capture. A `created_at` timestamp outside the platform range now
+degrades to an empty timestamp instead of raising an uncaught error that aborted the entire
+decode pass (which, since the offset never advanced, blacked out every conversation on every
+cycle -- a likely failure mode if a future agy renumbers a field into the timestamp slot).
```
joshalbrecht: I kinda hate protobufs, why is this even possible

joshalbrecht:
I'm approving this, BUT -- it's probably worth having a claude take a closer look to see if we're doing stupid things there. I thought the whole point of protobufs was to do a better job with migrations like this...

libs/mngr_antigravity/changelog/mngr-agy-keychain.md
evgunter: i think claude was overselling how likely this is

evgunter:
it's better iff you actually give people the schema, is my understanding

PR #2080, opened 2026-06-10
Protect remote latchkey gateway with the shared gateway password

hynek-urban:
- Protect remote Latchkey gateway with a password.
- It's the same password that agents already use to access the primary gateway so it should "just work" from their perspective.
- Reusing the gateway password feels relatively fine security-wise. It's just an opaque autogenerated string.


hynek-urban:
Succeeded by https://github.com/imbue-ai/mngr/pull/2101.

PR #2084, opened 2026-06-10
make is-gitignored util; use it in subagent-proxy

evgunter:
## Problem

Launching a mngr Claude agent from a git-tracked repo (e.g. `~/code-guardian`) left an untracked `.claude/agents/mngr-proxy.md`, tripping clean-tree stop hooks. Root cause: the `claude_subagent_proxy` plugin is **opt-out** (enabled whenever its entry point is discoverable in the `mngr` environment, unless explicitly disabled in config), and in PROXY mode (default) its provisioning hook unconditionally writes that agent-definition file into the agent's worktree. DENY mode has the same issue with `.claude/skills/mngr-subagents/SKILL.md`.

## Changes

- **Relocate artifacts under a `mngr-proxy/` subdirectory** so a single `.gitignore` line covers each:
  - PROXY agent def: `.claude/agents/mngr-proxy.md` -> `.claude/agents/mngr-proxy/proxy.md` (discovery is unaffected -- Claude Code scans `.claude/agents/` recursively and identifies the subagent by its frontmatter `name:` field).
  - DENY skill: `.claude/skills/mngr-subagents/SKILL.md` -> `.claude/skills/mngr-proxy/SKILL.md` (the skill is correspondingly renamed `mngr-subagents` -> `mngr-proxy`; `SKILL.md` filename is required by Claude Code).
- **Gitignore guard:** at provisioning the plugin now refuses to write either artifact when its path is not gitignored, raising `UnignoredProxyArtifactError` that points at both the gitignore fix and the repo-scope disable command (`mngr config set --scope project plugins.claude_subagent_proxy.enabled false`).
- **Test infra:** the proxy `FakeHost` now honors `cwd` and reports non-zero exits via `success` (matching real `OnlineHostInterface`) instead of `check=True`-raising, so the git-based guard is exercisable in unit tests.

## Core git helper (refactor)

The gitignore check started as inline logic in mngr_claude's `settings.local.json` guard. Since neither the check nor the symlink handling is claude-specific, it was generalized and moved into core **`mngr.api.git`**, where both that guard and the new proxy guard now consume it:

- `check_path_gitignore_status(host, repo_path, relative_path)` -> `(GitignoreStatus, Path)`: reports whether any repo-relative path (which need not exist yet) is gitignored. Returns `SKIP` / `IGNORED` / `NOT_IGNORED`.
- `check_path_repo_gitignore_status(...)`: same, but returns `ONLY_GLOBAL` when only the user's global excludes cover the path (for preflight checks whose result must hold on a remote host / fresh clone). This replaces the earlier `require_repo_rule` flag, so the set of returned statuses no longer depends on an argument.
- **Symlinks resolved at any depth:** a small POSIX-sh resolver walks up to the longest existing ancestor and `realpath`s it, then re-appends the missing remainder -- canonicalizing a not-yet-existing path without `realpath -m` (GNU-only; macOS/BSD lacks it). This also fixes a latent bug in the original leading-only resolution, where a non-leading symlink made `git check-ignore` exit 128 ("beyond a symbolic link"), which was misread as `NOT_IGNORED`.

## Note on scope

This makes worktree placement *clean* (ignorable + guarded); fully moving the artifacts out of the worktree remains a separate, larger followup (documented in the README) blocked on Claude Code discovery limitations.

## Tests

- New `libs/mngr/imbue/mngr/api/git_test.py` cases for both helpers: any-rule (SKIP / IGNORED / NOT_IGNORED), repo-rule (`ONLY_GLOBAL`, repo-rule IGNORED, NOT_IGNORED pass-through), and a mid-path (non-leading) symlink.
- `libs/mngr_claude_subagent_proxy` full quick suite (incl. guard tests) and `libs/mngr_claude` plugin tests (incl. preflight/symlink/only-global) all pass; ratchets pass; `ty` clean for `mngr`, `mngr_claude`, `mngr_claude_subagent_proxy`.
- Manually verified the rendered error message and the NOT_IGNORED / IGNORED / non-git SKIP behaviors (and the symlink resolver across edge cases) against real git repos.


joshalbrecht:
lgtm after conflicts fixed

PR #2085, opened 2026-06-10
mngr create --reuse: scope agent lookup to the address's host name

gnguralnick:
## Problem

`mngr create system-services@babatest.docker --reuse --update --new-host` fails with:

> Multiple agents found with name 'system-services'. Use address syntax (e.g. 'system-services@HOST.PROVIDER') to target a specific host.

even though the address *already* uses host syntax (`@babatest.docker`).

## Root cause

`_try_reuse_existing_agent` scoped the lookup by provider and by `target_host_ref` only. Under `--new-host`, `_parse_target_host` returns `NewHostOptions` (not a `DiscoveredHost`), so `target_host_ref` is always `None` and the host name from the address (`babatest`) was never used. The lookup therefore matched *any* `system-services` on the `docker` provider. Docker discovery includes stopped containers, so once two or more same-named agents existed (e.g. several minds workspaces, since the minds desktop client names every workspace's primary agent the constant `system-services` and relies on the host name for identity — see `Workspace identity = host name, not agent name`), the lookup became ambiguous and raised.

## Fix

Pass `address.host_name` into `_try_reuse_existing_agent` and scope the candidate set to that host (by `HostId` or `HostName`):

- Fresh host: nothing matches → returns `None` → caller creates a new agent.
- Existing host re-create: matches exactly the agent on that host → reuse/update in place.
- Shared name with no host in the address: still genuinely ambiguous → still raises the disambiguation error.

## Tests

`libs/mngr/imbue/mngr/cli/create_test.py`:

- `test_try_reuse_existing_agent_scopes_to_address_host_name_when_new_host` — multiple `system-services` on different docker hosts + an address naming a fresh host → returns `None` (the regression guard).

- `test_try_reuse_existing_agent_raises_on_ambiguous_match_without_host_scope` — multiple same-named agents, no host in the address → still raises, so host scoping did not swallow real ambiguity.

Existing reuse tests updated for the new parameter.

Local run: `create_test.py` (fast markers) — 117 passed, 0 failed; reuse subset — 7 passed. `ty check` clean on both changed files. Full suite via CI.


joshalbrecht:
lgtm

libs/mngr/imbue/mngr/cli/create.py
```diff
@@ -885,6 +885,7 @@ def _create_agent(
             agent_name=agent_opts.name,
             provider_name=address.provider_name,
             target_host_ref=target_host if isinstance(target_host, DiscoveredHost) else None,
```
weishi-imbue, line 887: I think `target_host_ref` is now redundant once you pass in `address.host_name`. Everything can be now derived from the address.host_name. (See at #1694)

joshalbrecht: yeah, I don't understand why we're not just passing in `address`

that has the complete information

PR #2087, opened 2026-06-10
support fetching github data for up to 1000 agents in kanpan

evgunter:
thanks to my `mngr_usage` test, i have over 100 PRs

---

## Problem

The kanpan GitHub data source fetches every board PR in one `gh api graphql` search call with `first: 100` — GitHub's hard per-page cap. With more than ~100 agents, the matched PRs overflow a single page and the overflow agents silently render "Create PR" instead of their real PR. The old code detected this via `pageInfo.hasNextPage` and surfaced an error, but never fetched the rest.

## Fix

`fetch_board` now cursor-paginates: it follows `pageInfo.endCursor` and accumulates every page, up to GitHub's ~1000-result search ceiling (10 pages of 100), beyond which it surfaces an explicit error rather than silently dropping data.

Each page request is wrapped in a bounded `tenacity` retry with exponential backoff that fires on transient GitHub failures — HTTP 403 secondary rate limits (whose REST-style error body carries no `data.s`), 5xx, and unparseable bodies. Because the retry wraps a **single page**, a failure on page N retries only page N's cursor; pages already fetched are kept and never re-requested. (A primary `RATE_LIMITED` error needs a multi-minute wait, so it falls through to the existing board-level 60s cooldown rather than being usefully retried inline.)

## Notes

- A successful search always returns a `data.s` object; its absence is the discriminator for "this page request failed at the transport level, retry it."
- Parsing is split into `_parse_board_page` (parsed-dict → nodes/cursor/errors) and `_build_prs_from_nodes` (accumulated nodes across all pages → one PrInfo per requested (repo, branch), preserving the OPEN > MERGED > CLOSED dedup across page boundaries).

## Testing

- `just test-quick libs/mngr_kanpan -m 'not acceptance and not release'`: 395 passed, 0 failed (includes ratchets).
- `uv run ty check` on the changed files: clean.


joshalbrecht:
lgtm!

PR #2090, opened 2026-06-10
add agent-type aliases like agy and pi

evgunter:
built with an opencode agent! (except the autofix, which wants claude code)

---

## What

Lets `mngr create my-agent agy` work as a short form of `mngr create my-agent antigravity` (and `pi` for `pi-coding`).

## Design

Aliases are a **name-resolution layer above** the agent-type registries, not duplicate registry entries:

- New `register_agent_aliases` pluggy hook: plugins return `{alias: canonical_type}`.
- An incoming name is normalized through the alias map to its canonical type **before** any registry lookup (in `resolve_agent_type`, `is_known_agent_type`, the `mngr create` capture point, and config `parent_type` parsing).
- Therefore an alias is never itself a distinct agent type:
  - hidden from `mngr plugin list --kind agent-type` (canonical types only),
  - tab-completable for `--type` (via `list_selectable_agent_type_names`),
  - the persisted `data.json` records the **canonical** type (also fixes the canonical-keyed process-name lookup for aliased agents).
- Defining a custom `[agent_types.X]` block whose name collides with a built-in alias is rejected at config load.

## Disabled-plugin attribution fix

mngr now records which plugin registered each agent type, so the "plugin is disabled, enable it with `mngr plugin enable ...`" message names the **real** registering plugin instead of assuming the type name equals the plugin name. This fixes the message for types whose name differs from their plugin's entry-point name (e.g. the `pi-coding` type registered by the `pi_coding` plugin).

## Tests

New unit tests cover alias resolution, alias-not-a-registered-type, completion vs. plugin-list inclusion, collision/missing-target skipping, alias `parent_type`, the custom/alias collision error, and disabled-plugin attribution. The antigravity/pi-coding plugin tests assert their alias hooks.

Bumped the `cast` ratchet by 2 (with an explanatory comment): the casts isolate pluggy's untyped `HookImpl.function()` boundary, needed to attribute each registration to its owning plugin — the same blessed pattern as `api/create.py`.


libs/mngr/imbue/mngr/config/loader.py
```diff
@@ -767,11 +769,22 @@ def _parse_agent_types(
...
 
     for name, raw_config in raw_types.items():
+        # A custom type name must not collide with a plugin-registered alias:
+        # that would make the same name mean two different things, so reject it.
+        if is_agent_alias(name):
+            raise UserInputError(
+                f"Agent type '{name}' conflicts with a built-in alias for "
+                f"'{normalize_agent_type_name(name)}'. Rename your custom "
+                f"[agent_types.{name}] block to a name that is not an alias."
+            )
```
evgunter: seems least ambiguous, but does mean that adding a plugin could cause a config error in config that was previously fine

joshalbrecht:
I mean... you *could* do this without the aliases (just make a new config entry)

but lgtm I guess -- it is nice :)

joshalbrecht:
(not merged bc non-standard base branch)

PR #2091, opened 2026-06-10
tabcomplete settings

evgunter:
i know tabcomplete is getting excessively fancy but i don't remember the names of things so it's very helpful

https://github.com/user-attachments/assets/0644be8f-f03f-436c-b584-cc5a6413a0eb

deprecation warning, printed once every 24h until the user updates (this is a one-time thing; the code in the user's ~/.bashrc is now just a shim, and we automatically update our script with the actual code):

Screenshot 2026-06-10 at 5 03 43 PM

---

## Summary

A substantial pass over mngr's shell tab-completion, built on top of `mngr/list-schema` (its model-walking schema work is in the base). Highlights:

- **`-S`/`--setting` completion** — `mngr <cmd> -S <TAB>` now completes config `KEY`s and (where the value is constrained) `VALUE`s, reusing the same `config_keys` / `config_value_choices` data that backs `mngr config set`.
- **Builtin `agent_types.*` completion** — config keys now complete for builtin/registered agent types (`agent_types.claude.*`, `agent_types.codex.*`, …), derived from each type's config schema, not just the custom types defined in your config.
- **Segment-at-a-time drilling** — dotted keys complete one `.` segment at a time instead of dumping every fully-qualified key at once; a valued key completes to `KEY=` and defers its values to the next TAB.
- **Positional-counter fixes** — short value options (`-S`, `-m`, …) and the `--verbose` count option are now classified correctly, and bash's `=`-word-splitting of `KEY=VALUE` is handled, so completion of a later positional argument isn't wrongly suppressed.
- **Self-updating install** — completion installs as a small, stable rc *shim* that sources a mngr-managed file, which mngr regenerates automatically (on `mngr list`, and via `mngr extras completion`). Completion improvements reach users on upgrade with no manual rc edits. `mngr extras completion` migrates an old self-contained completion block to the shim (removing it only when it byte-matches a form mngr generated, never a hand-edited one), and an out-of-date install gets a throttled "refresh me" nudge (the deprecation warning shown above).

## Behavior

- **Drill-down.** `-S agent_t<TAB>` → `agent_types.` → `cl<TAB>` → `agent_types.claude.` → … Each branch (`section.`) or `KEY=` is inserted with no trailing space so you can keep typing; the next level / the values appear on the next TAB.
- **Key phase.** A key with a constrained value set completes to `KEY=` (values deferred); a free-form key completes to the bare `KEY`.
- **Value phase.** `-S logging.console_level=<TAB>` lists `TRACE`/`DEBUG`/…; `-S headless=<TAB>` lists `true`/`false`. `parent_type` completes to agent-type names; bool fields to `true`/`false`.
- Works on every command (it's a common option) and in **both zsh and bash** (which tokenize `KEY=VALUE` differently). Verified end-to-end interactively in real zsh and bash.

## Implementation

- `complete.py` — the lightweight per-TAB engine: `-S` candidate generation (`_get_setting_candidates`, handling the zsh/bash tokenization shapes), segment collapse (`_segment_keys`), and the positional counter (uniform long/short option classification + bash `=` coalescing).
- `completion_install.py` (new) — the completion shell-integration primitives: script/shim generation, managed-file writing, the version sentinel + stale-completion warning, and legacy-block migration. `extras.py` (install UX) and `list.py` (background refresh) import from it; the per-TAB path stays stdlib-only.
- `completion_writer.py` / `completion_cache.py` — record `-S`/`--setting`, every option's short and long forms, and schema-derived `agent_types.*` keys/values; new `setting_option_names` cache field.
- `utils/model_schema.py` — relocated from `cli/` so the config-layer writer can walk model schemas without importing the `cli` layer.

## Tests

- Unit + integration coverage for: `-S` key/value phases (both shells), segment drilling and `KEY=` value deferral, builtin agent-type keys, uniform option classification + the bash `=` residue, the managed shim / migration / stale-warning, and a guard test that the shim version is bumped whenever the generated completion function changes.
- Full `libs/mngr/imbue/mngr/cli` + `config` suites pass, type check passes, ratchets pass.


libs/mngr/imbue/mngr/cli/completion_install.py
```diff
@@ -0,0 +1,294 @@
...
+    for template in _LEGACY_COMPLETION_FUNCTION_TEMPLATES:
+        prefix, separator, suffix = template.partition("{python_path}")
+        if not separator:
+            continue
+        # The path is the only variable part of the line; everything else must match.
+        pattern = r"\n*" + re.escape(prefix) + r"[^\n]*" + re.escape(suffix) + r"\n*"
+        new_text, count = re.subn(pattern, "\n", rc_text, count=1)
+        if count:
+            return new_text, True
+    return rc_text, False
```
evgunter, line 221: slightly inelegant to check all of them instead of just the one for the shell in question, but this is a one-time thing so i think it's fine

joshalbrecht:
agreed that autocomplete is getting really fancy, but you sold me with the script upgrade :-P  

lgtm (will let you merge once tests pass / into non-standard base)

PR #2092, opened 2026-06-10
always clone local git build context before vps upload

evgunter:
will this break other providers? i'm confused about how this was working for vultr etc

---

## Problem

`mngr create -t aws` (and any VPS Docker backend: aws/vultr/ovh) fails during the post-build git seed when launched from a **primary checkout that keeps per-branch linked worktrees**:

```
remote: error: refusing to update checked out branch: refs/heads/mngr/aws
 ! [remote rejected] mngr/aws -> mngr/aws (branch is currently checked out)
error: failed to push some refs to 'ssh://.../code/mngr/.git'
```

### Root cause

The `aws` dogfood template uses build context `.` and `target_path = /code/mngr`. The remote `docker build` flow only cloned the context when it was a **linked worktree** (`.git` is a gitlink file) or `--git-depth` was set. From a primary checkout (`.git` is a directory), no clone happened, so the raw `.git` — **including `.git/worktrees/`** — was uploaded and `COPY . /code/mngr/`'d into the image. That admin dir registers the operator's *other* worktree branches as checked out, so the seed push (`git push +refs/heads/*` into `/code/mngr`) is refused for them. `git init --bare` on the target can't release a branch held by a linked worktree.

modal/docker never hit this because their context is the `make_tar_of_repo.sh` keyframe — a `git clone` + detached checkout, which carries no `.git/worktrees/`.

## Fix

Extend the existing clone-the-context behavior to **any local git checkout**, not just linked worktrees. A fresh `git clone` has a self-contained `.git` with no `worktrees/` admin — the same hygiene modal's keyframe gets. The operator's working tree (including uncommitted edits) is still overlaid on top via `rsync_worktree_over_clone`.

Extracted the clone+overlay into `_clone_build_context_for_self_contained_git` and added unit tests:
- a primary checkout with linked worktrees clones to a `.git` with no `worktrees/` admin, preserving uncommitted edits;
- a non-git context returns `None` (uploaded verbatim, no clone).

No template/settings change is needed — the existing `[create_templates.aws]` now works from a primary checkout.

## Tradeoff

Builds from a git-repo context now always clone locally first (previously only worktree contexts did). This matches modal's keyframe approach; the cost is a local `git clone file://` per create. The alternative — reusing `current.tar.gz` on VPS — needs the Dockerfile shipped inside the keyframe context (the VPS path rewrites `--file=` relative to the uploaded context), which is heavier; it remains a possible follow-up if shared build-layer caching is wanted.

## Testing

`just test-quick libs/mngr_vps_docker` (267 passed) + new `container_setup_test.py` cases (6 passed) + ratchets (56 passed). Full suite runs in CI.


joshalbrecht:
lgtm!

this only "worked" before because we never had worktrees in the repos we pointed at

PR #2093, opened 2026-06-11
azure

evgunter:
sometimes has these warnings, which are the azure package's fault. software
```
ev@evs-computer mngr % mngr c waiting-reason --branch mngr/unify-ts-plugins: -t azure
/Users/ev/.local/share/uv/tools/imbue-mngr/lib/python3.13/site-packages/azure/mgmt/compute/models/_models.py:12585: SyntaxWarning: invalid escape sequence '\W'
  <br>Has upper characters <br> Has a digit <br> Has a special character (Regex match [\W_])
/Users/ev/.local/share/uv/tools/imbue-mngr/lib/python3.13/site-packages/azure/mgmt/compute/models/_models.py:12695: SyntaxWarning: invalid escape sequence '\W'
  <br>Has upper characters <br> Has a digit <br> Has a special character (Regex match [\W_])
/Users/ev/.local/share/uv/tools/imbue-mngr/lib/python3.13/site-packages/azure/mgmt/compute/models/_models.py:23850: SyntaxWarning: invalid escape sequence '\W'
  <br>Has upper characters <br> Has a digit <br> Has a special character (Regex match [\W_])
Creating VPS Docker host curious-tireless-avocet (host-754f140a3e96486588370095db263fd4) ...
Creating VPS instance (region: westus, plan: Standard_D2s_v3)...
```

---

## Summary

Adds an `azure` provider backend for mngr — a thin adapter over the shared `VpsDockerProvider` (same pattern as `mngr_aws` / `mngr_gcp` / `mngr_vultr`). Agents run in a Docker container on an Azure Linux VM. Built on `mngr/gcp`.

**Live-verified end-to-end against a real Azure subscription**: `prepare` (registers resource providers + creates RG/vnet/subnet/NSG), `create` (VM + public IP + NIC + cloud-init + Docker + container + agent), `exec`, `destroy` (NIC/IP/disk cascade leaves zero orphans), discovery, and failed-create orphan reclaim all confirmed working.

## What it does

- **Auth** via `DefaultAzureCredential` (uses `az login` locally, service principal in CI); only `subscription_id` in config (falls back to `AZURE_SUBSCRIPTION_ID`). No credentials in config — matches the Modal/AWS/GCP convention.
- **`mngr azure prepare`** — one-time privileged setup: registers `Microsoft.Compute`/`Network`/`Storage` (fresh subscriptions start unregistered), creates the mngr-owned resource group + vnet + subnet + NSG (attached to the subnet, fail-closed on SSH CIDRs), tags the RG `managed-by=mngr`. The hot create path is lookup-only.
- **`mngr azure cleanup`** — safe inverse (mirrors `mngr aws cleanup`): refuses while any `mngr-provider`-tagged VM exists, else deletes the mngr-owned RG (cascading vnet/subnet/NSG); only deletes a group it owns.
- **Per-host create**: Standard public IP + NIC + VM, all with `delete_option=Delete`, so `destroy` is a single VM delete and NIC/IP/disk cascade. SSH key injected inline (in-memory map, like GCP). Ubuntu 24.04 LTS default; default size `Standard_B2s`.
- **Snapshots** = managed-disk snapshots. **Spot** = `--azure-spot` (priority=Spot, eviction=Delete, max_price=-1).
- **Failed-create cleanup**: a VM create that fails after the public IP + NIC are provisioned (e.g. `SkuNotAvailable`) would orphan them. They're cleaned up immediately when possible, or — because Azure reserves the NIC for 180s — reclaimed self-healingly at the start of the next create (age-gated so it never disturbs an in-flight concurrent create). The session-end test scanner reclaims them too.
- **Auto-shutdown**: best-effort cloud-init `shutdown -P` (matches Vultr; documented that on Azure this doesn't stop compute billing — the orphan scanner is the test cost backstop).
- Azure API payloads use **typed azure-mgmt SDK models** (the compute SDK doesn't remap snake_case dict bodies to the ARM wire format).

## Implementation notes

- Pure addition: only existing-file change is one line in `providers/registry.py` (`azure` in `_REMOTE_BACKEND_NAMES`), plus `pyproject.toml` cov, `uv.lock`, regenerated `create.md`, and changelogs.
- New deps: `azure-mgmt-compute`, `azure-mgmt-network`, `azure-mgmt-resource` (official Microsoft SDKs; `azure-identity`/`azure-core` were already vendored).
- Unit/integration tests pass; `ty` clean; ratchets clean. Release tests are triple-gated (`MNGR_AZURE_RELEASE_TESTS=1` + creds + subscription); the build-on-VM release test was run for real (see Follow-ups).
- Architecture verified and autofix clean.

## Follow-ups (this branch)

- **`-t azure` now builds the project Dockerfile on the VM** (gh + the full mngr toolchain), instead of coming up on a bare `debian:bookworm-slim`. Mirrors the `gcp` template: `build_arg = ["--azure-vm-size=...", "--file=libs/mngr/.../Dockerfile", "."]` — the worktree is the context, which the shared `mngr_vps_docker` flow clones (overlaying uncommitted changes) and resolves `--file` inside. Forwards `GH_TOKEN` + runs `gh auth setup-git`, sets `agent_args=--dangerously-skip-permissions` and `target_path=/code/mngr`, and builds on depot's cached remote builders (`[providers.azure] builder=DEPOT`; `DEPOT_TOKEN` exported at create time, `depot.json` supplies the project id). Covered by a new release test (`test_provider_create_builds_dockerfile_on_vm`) that builds a Dockerfile on a real Azure VM and asserts the container runs from the built image — **verified passing end-to-end** (no existing test exercised build-on-VM).
- **Unresolvable subscription/credential now raises `ProviderUnavailableError`, not `ProviderEmptyError`.** Azure being unreachable means its state is *unknown* (agents may exist), so `mngr list` now warns instead of silently dropping the provider and its agents.
- **Torn `azureProfile.json` reads are retried.** The az CLI rewrites that file non-atomically; a read racing a write could resolve to "no subscription" and (with the change above) make azure agents flicker out of `mngr list` and back. `read_az_cli_default_subscription` now retries before giving up; a genuinely absent file still short-circuits to `None`.

## Lifecycle fixes (this branch)

- **A cleanly-stopped agent is now `STOPPED` and reconnectable, not `CRASHED` + invisible.** A `mngr_vps_docker` host whose container is stopped but whose VPS is still reachable (the idle-watcher shutdown, a manual `mngr stop`, or a VPS reboot) was misreported as `CRASHED` and filtered out of `mngr conn`/`start` (`include_destroyed=False`), producing a confusing "Could not find agent" even though the agent stopped cleanly with its data intact. Discovery now tracks VPS reachability separately from container-running state: reachable + container-down → `STOPPED` (visible to conn/start); an *unreachable* VPS still derives to `CRASHED` (the genuine down case). Shared fix — benefits aws/gcp/vultr/ovh too.
- **Curated auth error.** An unresolvable subscription/credential previously surfaced the generic provider-unavailable help text ("ensure the provider backend is running (e.g. start Docker)") — wrong advice for cloud auth. It now gives Azure-specific guidance (`AZURE_SUBSCRIPTION_ID` / `az login` / `mngr azure prepare`), via an optional `user_help_text` on `ProviderUnavailableError`.

## Deferred (separate follow-up, not in this PR)

- **Deallocate-based stop/start (real cost savings) + idle auto-deallocate.** On Azure, neither `mngr stop` (container-level) nor the cloud-init `shutdown -P` actually stops compute billing — only `deallocate` or `destroy` does. True deallocate is **architecturally blocked** in the current `mngr_vps_docker` model: the host record lives *on the VM* and is read over SSH, so a deallocated (off) VM becomes unreadable and vanishes from `mngr list`/`conn`/`start`. Supporting it requires relocating host state off the VM (Azure tags or a durable store), mirroring how Modal decouples state onto a durable Volume. Scoped as a deliberate follow-up; until then, cost control is `destroy` + the orphan scanner.

Spec: `specs/azure-provider/concise.md`.


libs/mngr_azure/imbue/mngr_azure/backend.py
```diff
@@ -0,0 +1,321 @@
...
+           run cannot leak a billing VM. On Azure that net is two-layered --
+           cloud-init ``shutdown -P +N`` (from ``auto_shutdown_seconds``) powers
+           off the agent, and the conftest session-end orphan scanner
+           force-deletes any leaked VM tagged ``mngr-pytest-launched`` older than
+           the TTL (derived from the same ``auto_shutdown_seconds``). If it is
+           unset, fail closed here rather than silently leak a VM. NB: unlike
+           AWS/GCP, an OS ``shutdown -P`` on Azure leaves the VM "Stopped (not
+           deallocated)", which still bills for compute -- so on Azure the *cost*
+           guarantee comes from the orphan scanner, not from auto_shutdown. The
+           guard is still required so the TTL the scanner derives is well-defined.
```
evgunter, line 118: fixing this in a followup PR

libs/mngr_azure/imbue/mngr_azure/client.py
```diff
@@ -0,0 +1,926 @@
...
+    # lives only in per-VM os_profile config. This in-memory map bridges the base
+    # flow's upload_ssh_key -> create_instance(ssh_key_ids=[...]) handoff within a
+    # single process. A later fresh-process delete is a tolerant no-op.
+    _ssh_public_keys_by_id: dict[str, str] = PrivateAttr(default_factory=dict)
+    _cached_compute_client: Any = PrivateAttr(default=None)
+    _cached_network_client: Any = PrivateAttr(default=None)
+    _cached_resource_client: Any = PrivateAttr(default=None)
+
+    # =========================================================================
+    # Lazily-built management clients (overridden in tests to inject fakes)
```
evgunter: hmm

libs/mngr_azure/imbue/mngr_azure/client.py
```diff
@@ -0,0 +1,926 @@
...
+
+    def _reclaim_orphaned_network_resources(self) -> None:
+        """Best-effort delete unattached, aged mngr NICs / public IPs from failed creates.
+
+        ``create_instance`` provisions a public IP + NIC before the VM; a VM
+        create that fails (capacity / quota) leaves them orphaned, and Azure
+        reserves the NIC for the would-be VM for 180s so immediate cleanup is
+        blocked. Rather than block a failed create for ~3 minutes, the next create
+        reclaims them here -- self-healing on the next operation, the same pattern
+        discovery / GC use. Only resources older than the reservation window
```
evgunter: flagging

libs/mngr_azure/imbue/mngr_azure/client.py
evgunter: apparently aws and gcp do the same thing

joshalbrecht:
we chatted in person about:
- start / stop
- host directory
- how host data.json's are stored

libs/mngr_azure/imbue/mngr_azure/client.py
joshalbrecht: logically should be fixed in GC, not in create, if that makes sense?

joshalbrecht:
finally got a chance to review this whole thing, and lgtm except for my comments!

PR #2094, opened 2026-06-11
preserve session files for codex, opencode, pi-coding, and agy

evgunter:
Brings the four newer agent plugins (codex, opencode, pi-coding, antigravity/agy) to parity with the claude plugin's session preservation. On destroy (before the agent state dir is deleted), each agent's transcripts (raw + common), session/conversation-id history, **and native resumable session store** are copied to `<local_host_dir>/preserved/<name>--<id>/`, mirroring the state-dir layout and surviving remote host destruction. For remote agents the files are pulled to the local machine.

## Scope
- **Transcripts + id-history** — the raw + common transcript dirs, plus each plugin's root-session / conversation-id pointer file(s).
- **Native resumable session stores** — each agent's own store (what its `resume` path reads), targeted by its specific subpath so colocated credentials are **excluded** by construction:

  | Agent | Native store preserved | Credentials excluded (siblings) |
  |---|---|---|
  | codex | `plugin/codex/home/sessions` (rollout JSONLs) | `auth.json` symlink, `config.toml` |
  | opencode | `opencode.db` + its `-wal`/`-shm` WAL sidecars + `storage/` | `auth.json` symlink, `log/` |
  | pi-coding | `plugin/pi_coding/sessions` | `auth.json` (or none, with env-var auth) |
  | antigravity | `antigravity-cli/conversations/` (per-conversation SQLite) | oauth token, `settings.json`, keychain symlink |

- **Flag:** `preserve_on_destroy` (default True) on the four new plugins; claude keeps its existing `preserve_sessions_on_destroy`. A single flag preserves transcripts + native store, matching claude's one-flag model.

## Changes
- **`libs/mngr/api/preservation.py`** — `build_transcript_preserved_items(source)`, `preserve_agent_state(items, agent, host)` (online), and `preserve_host_agents_on_destroy(host, ctx, agent_type, items_for_agent)` (offline; skips non-readable hosts and filters discovered agents by agent type).
- **codex / opencode / pi-coding / antigravity** — `preserve_on_destroy` config field, `on_destroy`, `on_before_host_destroy` hookimpl, a `flag_gated_items` selector, and named path constant(s) + `PreservedItem`(s) for the native session store.
- **claude** — migrated onto the shared helpers (no behavior change); offline path now also filters by agent type.
- **Release-test harness (`agent_release_testing.py`)** — `run_agent_release_lifecycle` now *asserts* preservation: its destroy step verifies the transcripts (keyed on the seeded secret) and each profile's declared native session store (`native_session_preserved_relpaths`) actually landed in `preserved/<name>--<id>/`. Previously destroy was bare cleanup, so a swallowed preservation failure passed silently. The four lifecycle release tests gained `@pytest.mark.rsync` (the destroy step rsyncs the preserved directories on the local host).

Notable details:
- The **agent-type filter is load-bearing**: all four new plugins share the `preserve_on_destroy` key, so each plugin's `on_before_host_destroy` runs against every agent on the host; the filter ensures a plugin only preserves its own agent type.
- **opencode WAL sidecars** are load-bearing: the store is SQLite in WAL mode, so copying only `opencode.db` would drop the most recent (uncheckpointed) turns. The `-wal`/`-shm` sidecars are copied alongside it (skipped when absent, e.g. after a clean checkpoint).
- **antigravity macOS caveat** (documented in code + changelog): on macOS the `.db` is encrypted by the login-keychain "Antigravity Safe Storage" key, so a macOS-created store is readable on the same machine but not portable to a different machine/user. Linux uses a portable file-based store.
- A **FIXME** in the harness marks where this should grow into an actual resume-from-preserved-store check via `--adopt-session`, which is not yet implemented for these agents (only claude has it today). The arc currently asserts the bytes are preserved, not that they resume.

## Tests
- Unit/integration, core preservation, ratchets, `uv run ty check`, and `ruff`: clean across the touched projects (`mngr`, `mngr_codex`, `mngr_opencode`, `mngr_pi_coding`, `mngr_antigravity`).
- **Release tests do not run in CI.** Ran manually against the real binaries: the full-lifecycle release test for codex, opencode, pi-coding, and antigravity all pass (each now also asserting the native session store was preserved on destroy), plus the antigravity proto-schema release test.


libs/mngr/imbue/mngr/agents/agent_release_testing.py
```diff
@@ -228,11 +228,23 @@ def _assert_transcripts_preserved(
...
+      - codex: resumes by session id, but pops an interactive "Choose working directory
+        to resume this session" modal (recorded cwd != new worktree) that must be
+        auto-answered.
+      - opencode: the `session` row stores an absolute `directory` (the destroyed source
+        worktree); the recall POST returns 200 but silently no-ops against it. Must rebind
+        `session.directory` (and `project.worktree`) in opencode.db to the new work_dir.
+        Bytes incl. the `-wal`/`-shm` sidecars are sufficient once rebound.
+      - pi-coding: rewrite the `pi_session_file` pointer to the adopted jsonl's new path,
+        AND clear pi's blocking "cwd from session file does not exist" dialog (the cwd is
+        embedded in the session jsonl) -- ideally by rewriting that cwd before launch.
```
evgunter, line 249: working on this in a new pr

joshalbrecht:
lgtm

PR #2095, opened 2026-06-11
make --adopt-session work on preserved and live mngr claude sessions

evgunter:
made with pi-coding! (except autofix)

---

### What

`mngr create --adopt-session <session-id>` previously resolved a bare session ID only against the current Claude config dir (`$CLAUDE_CONFIG_DIR/projects/`) and the user-scope `~/.claude/projects/`. This PR extends that lookup to also search:

- every **live local mngr agent's** per-agent config dir (`<host_dir>/agents/<id>/plugin/claude/anthropic/projects/`), and
- **preserved sessions** from destroyed agents (`<host_dir>/preserved/<name>--<id>/plugin/claude/anthropic/projects/`, produced by `preserve_sessions_on_destroy`).

Passing a full path to a `.jsonl` file is unchanged.

### Why

A session ID is the natural handle users reach for, but until now you could only adopt by ID if the session lived in your shared Claude config. Sessions created by other mngr agents -- or left behind by an agent you've already destroyed -- required digging up the full `.jsonl` path. This makes those sessions adoptable by ID directly.

### How

- New helper `_mngr_session_projects_dirs(mngr_ctx)` enumerates the per-agent `projects/` dirs under the local host dir's `agents/` and `preserved/` subtrees.
- `_resolve_adopt_session` now takes `mngr_ctx` and searches, in priority order, the current config dir -> user-scope config dir -> live local mngr agents -> preserved agents (deduplicated by resolved path).
- Updated the `--adopt-session` help text and regenerated `create.md`.

### Scope / limitations

- Only the **local** host dir is scanned. Sessions on remote hosts (Modal/SSH/Docker) are not searched: enumerating them requires per-provider network discovery, and the existing copy step (`host.copy_directory(host, source_project_dir, ...)`) requires the source path to already exist on the destination host. This same constraint already applies to the pre-existing vanilla/user-config path, so `--adopt-session` remains a local-source feature. Cross-host adoption would be a larger, separate change.

### Testing

- Added unit tests covering adoption by ID from a preserved agent and from a live local mngr agent.
- Full `plugin_test.py` (208 tests) and `test_ratchets.py` pass; `ty check` is clean.
- The existing end-to-end `test_adopt_session.py` tests are release tests (not run in CI) and were not extended.


joshalbrecht:
lgtm

PR #2096, opened 2026-06-11
mngr usage for codex, opencode, and pi

evgunter:
didn't see how to get this info from antigravity yet

---

Extends `mngr usage` cost/usage tracking beyond Claude to **OpenCode, pi, and Codex** — all verified live end-to-end.

## Landed
- **Spec** `specs/agent-usage-plugins/spec.md`.
- **`mngr_usage` foundation** — `TokenSnapshot`; `pricing` (`compute_cost`). Anthropic prices mirror `apps/modal_litellm` (drift test); OpenAI/Codex prices mirror litellm directly (`litellm_pricing_test`). Unknown model → no estimate + WARNING, never `$0`.
- **Reader hook + 3 strategies** — `aggregate_usage_source` hookspec; `process_cumulative` (Claude), `session_cumulative` (Codex), `session_incremental` (OpenCode/pi). Cost preferred-reported / else token-estimated, flagged `REPORTED`/`ESTIMATED`.
- **Token/estimated surface** — `subscription_tokens`/`api_tokens`, `is_estimated` in cost blocks, per-session `tokens`/`model`/`cost_provenance`, human `(estimated)` marker.
- **`mngr_claude_usage`** — thin hookimpl claiming `claude` (unchanged output).
- **`mngr_opencode_usage` (new)** — in-process plugin, per-message reported cost. Live: `api_cost 0.066`.
- **pi** — writer in mngr_pi_coding's lifecycle extension (gated), `mngr_pi_coding_usage` (new) owns reader+gate. Live: `[pi-coding] 0.057`.
- **`mngr_codex_usage` (new)** — `codex_usage.sh` writer (rollout `token_count` → cost_snapshot; cache-exclusive input; 5h/7d windows), reader via session-cumulative + token-estimated cost. Live: windows render; tokens captured; cost correctly blank for `gpt-5.5` (newer than litellm) with a WARNING.

All four sources coexist in one `mngr usage` (claude / opencode / pi-coding / codex), each via its own reader hookimpl.

## Out of scope
Antigravity (no usage data in payload), claude-subagent-proxy (covered by mngr_claude_usage).

## Common-transcript converter parity

Alongside the usage work, this branch brings the common-transcript converters to parity across claude, antigravity, and codex (net change vs `main`):

- Each `common_transcript.sh`'s inline `python3` heredoc is extracted into a standalone `common_transcript_convert.py`, now type-checked, linted, and unit-tested directly. Event counting is unchanged from `main`'s model: the converter prints its appended-event count on stdout and the shell captures it.
- A shared `mngr_common_transcript_lib.sh` adds a convert lock that serializes the converter's read-modify-write against the turn-end `--single-pass` flush.
- codex reaches flush parity with claude/antigravity: its Stop / SubagentStop hooks run one synchronous flush when the agent goes WAITING, and its converter takes the convert lock.
- Converter errors go to the structured log only, never echoed to the agent's pane. Malformed / null lines are dropped silently, and input is decoded UTF-8 with `errors=replace` so a stray byte can't wedge the converter.

Plus a quality pass over the whole diff: dropped content-free codex token snapshots, resynced stale internal dependency pins (a failing `version_sync_test`), and minor docstring/spec corrections.


libs/mngr_pi_coding_usage/README.md
```diff
@@ -0,0 +1,31 @@
...
+(`<provider>/<model>`), and the session id (the pi session file's basename).
+`cost_mode` is `API_KEY` (pi bills a real provider key).
+
+## Why the writer is in the harness extension
+
+Unlike Claude (statusline shim) or OpenCode (auto-loaded `plugin/*.ts`), pi loads
+exactly one explicit extension owned by mngr_pi_coding. A standalone writer would
+require a harness change either way, so the writer is folded into the lifecycle
+extension (which already extracts the message usage), and this package stays the
+reader + gate.
```
evgunter, line 31: flagging

libs/mngr_codex/imbue/mngr_codex/resources/common_transcript_convert.py
```diff
@@ -0,0 +1,259 @@
...
+    return len(new_events)
+
+
+if __name__ == "__main__":
+    # Route routine per-line skip warnings to stdout. common_transcript.sh treats
+    # any stderr output as a "convert error", so a truncated trailing line caught
+    # mid-write must not land there; stdout is not captured (the shell reports the
+    # appended count by diffing the output file's line count). Genuine uncaught
+    # exceptions still hit stderr via the default traceback and surface as errors.
+    logging.basicConfig(stream=sys.stdout, level=logging.WARNING, format="%(message)s")
```
evgunter: [CLAUDE] Routing the converters' routine per-line skip warnings to stdout here so they don't trip `common_transcript.sh`'s "any stderr ⇒ convert error" check — a transcript caught mid-write (truncated trailing line) was producing a spurious `convert error` log each poll. stdout is not captured by the shell (the appended count comes from a `wc -l` diff of the output file), so this is harmless, but it does mean the module now writes to stdout, which the original extraction deliberately avoided. Same change is in the claude/antigravity converters. Flagging for a second opinion — no strong preference; alternatives would be a dedicated log file or suppressing only the per-line-skip path. (Added by an autofix pass; kept after review.)

evgunter: (not doing this anymore)

joshalbrecht:
lgtm once tests pass!

PR #2097, opened 2026-06-11
proper stop/start for aws

evgunter:
## Goal

Give AWS agents a Modal-like **idle-paused-but-resumable** lifecycle. Previously `mngr stop` / idle only stopped the inner Docker container while the EC2 instance and EBS volume kept billing; the GC backstop then *terminated* idle agents with no way back. This makes a paused AWS agent stop the EC2 instance (preserving the EBS volume) and be resumable by name.

Design: `specs/aws-ec2-stop-start-lifecycle/spec.md`. Decisions (agreed with the author): native EC2 stop/start (not Modal-style snapshot+recreate); **EBS snapshots dropped**; idle trigger = a **no-IAM self-stopping instance** (Phase 2; see below); backstop = GC stop-instead-of-destroy + ~7-day age-gated terminate (Phase 3); offline listing of paused hosts via **EC2 tags**; changes in `mngr_aws` plus one shared `mngr_vps_docker` resume fix (and a small gated `gc.py` hook in Phase 3).

## Status: Phases 1, 2 + 4 landed (idle auto-pause and manual stop/resume both work end-to-end)

**Done in this PR:**
- **Native EC2 stop/start.** `AwsVpsClient.stop_instance`/`start_instance`; `AwsProvider.stop_host`/`start_host` overrides. `mngr stop --stop-host` stops the instance (records `stop_reason=STOPPED`); `mngr start` finds the stopped instance by `mngr-host-id` tag, starts it, and rebinds `vps_ip` + known_hosts to the fresh public IP before restarting the container's sshd and reconnecting.
- **No-IAM self-stopping idle watcher (Phase 2).** An in-container activity watcher lets a sentinel file go stale; a host-side systemd `.path`/`.service` unit then runs `shutdown -P now`. EC2's `InstanceInitiatedShutdownBehavior` decides what that shutdown does, via the new `terminate_on_shutdown` config flag (default `false` -> `stop` = resumable idle-pause; `true` -> `terminate` = ephemeral, used by the release tests). This needs no `mngr-aws` IAM role / `iam:PassRole` / awscli, at the cost of the single-flag tradeoff (an instance is resumable-on-idle **or** self-terminating, not both).
- **Tag-based offline discovery** (required so a stopped host — which has no public IP and drops out of SSH discovery — still lists its agents and resolves by name): `add_tags`/`remove_tags`; the `persist_agent_data`/`remove_persisted_agent_data`/`list_persisted_agent_data_for_host` hooks compose the authoritative on-volume store with a per-field `mngr-agent-<id>-name`/`-type`/`-labels` tag mirror (one tag per field, so `labels` gets the full 256-char budget; the id is in the key); `discover_hosts_and_agents`/`to_offline_host` reconstruct stopped hosts from tags so they appear in `mngr list` (as STOPPED) and resolve for `mngr start`. A field value that still overflows 256 chars (realistically only `labels`) is dropped with a warning; a host with so many agents that mirroring another would exceed EC2's 50-tag limit raises a `NotImplementedError` (which the CLI turns into an "open an issue" prompt) rather than failing obscurely.
- **`prepare`/`cleanup` are security-group-only** (no IAM provisioning, since the self-stop path needs none). The operator's credentials need `ec2:StopInstances`, `ec2:StartInstances`, `ec2:CreateTags`, `ec2:DeleteTags` on top of the existing run/describe/terminate set; no `iam:*` unless you set the optional `iam_instance_profile`.
- **Shared resume fix in `mngr_vps_docker` (benefits every VPS-Docker provider).** Base `start_host` restarts the container's sshd after `docker start` (sshd is `docker exec`'d, not in the entrypoint, so it doesn't survive a stop/start), **relaunches the in-container activity watcher**, and **records fresh `BOOT` activity** on resume. Putting the watcher relaunch in the base (rather than AWS-only) also keeps Vultr/OVH auto-stopping after a resume. The activity refresh is what stops a resumed-but-idle host from being re-stopped within one watcher poll (its activity-file mtimes otherwise still predate the stop).

**Security-relevant design choices (the design was shaped this way deliberately; an adversarial audit of the branch confirmed the threat model — each issue would require an attacker holding account-level `ec2:CreateTags`):**
- **Host keys for the resume rebind come from mngr's locally-held keypairs, not EC2 tags.** The keypairs are generated locally and injected into the instance at create, so resume can pin `known_hosts` for the new IP from a source mngr controls. EC2 tags are account-writable, so sourcing host keys from them would let a tagger substitute their own key and MITM the resumed SSH — hence local keys only.
- **Offline discovery is robust to hostile tag data.** A malformed `mngr-host-id`/`Name` tag is skipped (with a warning) instead of aborting the whole `mngr list` / `mngr start` sweep, so one corrupt/forged tag can't DoS discovery.
- **Instance-by-tag resolution refuses ambiguity.** If more than one non-terminated instance carries the same `mngr-host-id`, `mngr start` refuses rather than acting on the first match (a duplicate tag could otherwise steer it onto an attacker-controlled instance).
- Unit-tested via botocore Stubber (stop/start, tag round-trip, stopped-host discovery, offline reconstruction, shutdown-behavior default vs `terminate`, composed-super agent-data hooks, local-key rebind, malformed-tag skip, duplicate-tag refusal).

**Release-test confirmation (real EC2, opt-in `MNGR_AWS_RELEASE_TESTS=1`, not in CI):**
- `test_provider_stop_host_stops_ec2_instance_and_start_resumes` **passed on real EC2** (2026-06-14, us-east-1): `create -> stop --stop-host` (asserted `stopped` via the EC2 API) `-> start` (asserted `running`) `-> exec` (post-resume command succeeded over the fresh IP) `-> destroy` (no leak).
- `test_provider_idle_watcher_auto_stops_then_resumes` **passed on real EC2** (2026-06-14): create with a short idle timeout, poll until the on-host watcher auto-stops the instance, then `mngr start` and confirm it stays up with a working post-resume command. This test surfaced the resume-re-stop bug and now confirms the fix.
- Note: the agent-tag mirror is forward-only — a *stopped* instance created before these commits lacks the `mngr-agent-*` tags, so it won't list its agents from tags (a freshly-created agent works). Resume SSH no longer depends on any tag, so it is not forward-only.

**Remaining (follow-up phases):**
- **Phase 3** — GC stop-instead-of-destroy + ~7-day age-gated terminate.
- **Phase 5** — S3-backed agent store for many-agent / large-label hosts (the per-field tag mirror covers the common few-agent case; a host that exceeds EC2's 50-tag limit raises the `NotImplementedError` pointing here).


libs/mngr_aws/imbue/mngr_aws/backend.py
```diff
@@ -223,6 +338,628 @@ def _list_provider_vps_hostnames(self) -> list[str]:
...
+        only EBS storage; the root volume (and all on-disk state) survives, so
+        ``start_host`` can resume it.
+
+        ``create_snapshot`` is intentionally ignored -- native EC2 stop preserves
+        the whole filesystem, so the base's docker-commit snapshot would be
+        redundant. The base container-stop + record-write is reused via
+        ``super()``; we then record ``stop_reason=STOPPED`` so the offline-state
+        derivation reports STOPPED (not CRASHED) while the instance is down, and
+        finally stop the instance. The ``stop_reason`` write must happen *before*
+        the instance stops, since the volume is unreachable once it does.
```
evgunter: i think we're dropping the underlying create_snapshot entirely in https://github.com/imbue-ai/mngr/pull/2105, so this is the correct norm going forward

libs/mngr_aws/imbue/mngr_aws/backend.py
```diff
@@ -223,6 +338,628 @@ def _list_provider_vps_hostnames(self) -> list[str]:
...
+        required ``id``/``name`` are missing, or even the id+name minimum exceeds
+        the 256-char tag-value limit (e.g. a very long agent name). Otherwise
+        prefers the richest representation that fits, dropping ``labels`` (user
+        metadata) first, then ``type``, before the id+name minimum. Including
+        ``labels`` lets an offline ``mngr label`` on a stopped host actually
+        round-trip; if a given agent's labels are too large to fit, they are
+        dropped and a warning is logged so the write does not silently no-op. An
+        agent whose id+name alone do not fit is skipped (with a warning) rather
+        than returning an over-limit value that AWS would reject.
+        """
```
evgunter: this is kind of silly, looking into alternatives

libs/mngr_aws/imbue/mngr_aws/client.py
```diff
@@ -94,6 +95,13 @@ class AwsVpsClient(VpsClientInterface):
     root_volume_size_gb: int = Field(default=30, description="Root EBS volume size in GB")
     root_volume_type: str = Field(default="gp3", description="Root EBS volume type")
     iam_instance_profile: str | None = Field(default=None, description="IAM instance profile name to attach")
+    terminate_on_shutdown: bool = Field(
+        default=False,
+        description=(
+            "Sets EC2 InstanceInitiatedShutdownBehavior: False -> 'stop' (resumable idle-pause), "
+            "True -> 'terminate' (ephemeral / self-cleaning). See AwsProviderConfig for details."
+        ),
+    )
```
evgunter, line 117: huh

libs/mngr_aws/imbue/mngr_aws/backend.py
evgunter: we're not dropping it but this is correct behavior

joshalbrecht:
> (an instance is resumable-on-idle or self-terminating, not both).

I *think* this is ok if it means what I think it means...  for "resumable on idle" hosts, they're still able to shut down when they become idle, right?

libs/mngr_aws/imbue/mngr_aws/backend.py
```diff
@@ -32,6 +61,97 @@
...
+# Per-agent metadata is mirrored onto the instance as up to three EC2 tags per
+# agent, keyed ``mngr-agent-<agent_id>-<field>`` (the agent id lives in the key,
+# not a value), so a *stopped* instance (no public IP, SSH unreachable) still
+# surfaces its agents in discovery and resolves by name. ``name``/``type`` are
+# stored raw; ``labels`` as compact JSON. One tag per field (rather than one
+# packed tag per agent) gives ``labels`` the full 256-char value budget; a field
+# whose value still overflows is dropped, not failed.
+AGENT_TAG_PREFIX: Final[str] = "mngr-agent-"
+_AGENT_TAG_FIELDS: Final[tuple[str, ...]] = ("name", "type", "labels")
+_MAX_TAG_VALUE_LEN: Final[int] = 256
```
joshalbrecht, line 73: I still think the whole "store the metadata in the tag" setup is a bit too strange -- is there a future PR changing that?

libs/mngr_aws/imbue/mngr_aws/client.py
joshalbrecht: that actually seemed kinda nice (if I'm understanding correctly)

joshalbrecht:
overall lgtm (except for my two comments!)

libs/mngr_aws/imbue/mngr_aws/backend.py
evgunter: yes: https://github.com/imbue-ai/mngr/pull/2159

evgunter:
yeah, the resumable on idle hosts are stopped when idle

PR #2098, opened 2026-06-11
implement waiting_reason for codex; update spec for the other agent types

evgunter:
## Summary

Documents the cross-plugin `waiting_reason` parity picture **and implements it for codex**, giving codex agents the same "why is this agent waiting?" listing field that `mngr_claude` already has.

### codex `waiting_reason` (new code)

Codex agents now report *why* they are waiting via a `waiting_reason` field in `mngr list` (`agent_field_generators` hook), reading marker files directly with no SSH/tmux:

- **`PERMISSIONS`** -- blocked on a tool-approval dialog. A new `PermissionRequest` hook touches a `permissions_waiting` marker (inline touch command); `PostToolUse` clears it once the approved tool runs, and the root `Stop` clears any stranded marker as a safety net (e.g. a cancelled/denied dialog). `CodexAgent.get_lifecycle_state` also promotes RUNNING -> WAITING while the marker is present, mirroring claude.
- **`END_OF_TURN`** -- idle, turn complete (the `active` marker is absent).

This only applies in supervised mode: with `auto_allow_permissions = true` (`approval_policy = "never"`) codex never prompts, so an agent never waits with reason `PERMISSIONS`.

The RUNNING -> WAITING promotion rule is factored into a pure `_resolve_lifecycle_state_for_permission` helper so it is unit-testable without a live tmux pane. Codex has **no** `PostToolUseFailure` event (unlike claude), verified against codex 0.139.0, so cleanup is `PostToolUse` + `Stop` only.

### Documentation / scoping

The agent-plugin-parity spec (section P) now classifies `waiting_reason` per agent type, distinguishing genuine gaps from non-goals:

- **claude, codex**: implemented.
- **opencode**: should be done, just unimplemented -- *not* blocked on upstream. opencode prompts interactively via its `ask` permission policy, and its event bus emits `permission.asked` / `permission.replied`, which the in-process extension already receives via its `event` hook. A `PERMISSIONS` marker on those events (the way codex's hooks work) needs no upstream change. The opencode README documents this.
- **antigravity**: should be done, but blocked on upstream -- agy prompts (its `ask` dialog blocks the agent) but fires no hook/event while blocked, so mngr has no signal to read.
- **pi-coding**: inapplicable -- pi has no tool-approval gate at all (runs tools, including shell, without a confirmation prompt by design), so there is no blocked-on-permission state to surface.
- **codex README**: documents the new `waiting_reason` field and its values in user-visible terms, plus the supervised-mode caveat.

## Test plan

- New `plugin_test.py` cases cover the `waiting_reason` field generator (`PERMISSIONS` / `END_OF_TURN` / actively-running) and the `_resolve_lifecycle_state_for_permission` promotion rule.
- New `clear_active_marker_test.py` cases cover the `Stop` safety-net clear of a stranded `permissions_waiting` marker, and that a nested (different-session) `Stop` leaves it alone.
- `codex_config_test.py` updated for the two new hooks.
- Verified live against codex 0.139.0 with the exact production inline hook commands.

Per-project changelog entries added for `dev`, `mngr_codex`, and `mngr_opencode`.


joshalbrecht:
lgtm

PR #2101, opened 2026-06-11
Remote Latchkey follow-ups

hynek-urban:
- Protect remote Latchkey Gateway with a password.
  - It's the same password that agents already use to access the primary gateway so it should "just work" from their perspective.
  - Reusing the gateway password feels relatively fine security-wise. It's just an opaque autogenerated string and the gateway is exposed only on localhost, anyway.
- When syncing encrypted credentials to the remote Latchkey gateway, only transfer those that the agent actually needs.
- Avoid leaking secrets in the logs by using temporary files instead of `env` (the tradeoffs seem a bit better).
- Attempt to fix a number of issues related to stale connections and remote setup race conditions.
- Move latchkey service catalog functionality to `mngr_latchkey` where it logically belongs (and remove the now redundant `/permissions/available/` root endpoint from the API).


joshalbrecht:
lgtm!

joshalbrecht:
I didn't merge this yet because I'm trying to get the remote minds created from main without having a mismatch between what was in the installer and what is currently on main

but once that is done I'm fine merging this

PR #2102, opened 2026-06-11
Route minds pool baking through justfile + minds pool create

joshalbrecht:
## Summary

Make `minds pool create` (the env-aware wrapper) the canonical way to bake minds pool hosts, surfaced through new justfile recipes, and fix the docs that still pointed at the low-level `mngr imbue_cloud admin pool create` flow.

### Justfile (the missing entries)
- `just bake-pool-host <attributes-json> <region> [workspace_dir] [count] [extra flags]` — wraps `minds pool create`
- `just list-pool-hosts` — wraps `minds pool list`
- private `_pool-dsn-args` helper — resolves the staging/production `host_pool` DSN from Vault (dev/ci auto-resolve downstream)

These let the wrapper derive the management SSH key + OVH AK/AS/CK from the activated tier's Vault entries automatically, so no secrets are exported by hand. Recipes are bash-3.2-safe (no `mapfile`; guarded empty-array expansion under `set -u`).

### Docs
- `apps/minds/docs/host-pool-setup.md` — steps 2, 5, dev workflow, cleanup
- `apps/minds/docs/staging-bringup.md` — step 7

Both now use `just bake-pool-host` / `minds pool create`, and clarify that the **baked version comes from the `--workspace-dir` checkout, not from `--attributes`** (which is only the lease-match label).

### Skill
- `.claude/skills/minds-justfile/SKILL.md` — routes any minds task (app, pool hosts, environments, deployments, tests) through the root justfile; directs adding a recipe when one is missing.

## Testing
- `just --list` parses; `just --show bake-pool-host` renders correctly.
- Smoke-tested `_pool-dsn-args`: no-activation → clean exit 2; dev → no DSN args; staging → correct `--database-url` + DSN from Vault.
- No Python changed.

- `cleanup-pool-hosts` still uses the pre-Vault `.minds/<env>/neon.sh` pattern; migrate it onto Vault-based DSN resolution.
- Consider moving the staging/production DSN resolution into `minds pool create` itself so the recipe and docs simplify further.

### DIFF
--- a/.claude/skills/minds-justfile/SKILL.md
+++ b/.claude/skills/minds-justfile/SKILL.md
@@ -27,16 +27,17 @@ a **minds deployment**, or **minds tests** --
    activated env) and usage.
 2. **Use the recipe.** Prefer `just <recipe> ...` over the underlying command.
 3. **If no recipe fits, ADD one.** Write a new, well-commented recipe that
-   wraps the canonical command (and reuses existing private helpers like
-   `_pool-dsn-args`), then use it. Do not paper over a missing recipe with a
-   one-off shell command -- the point is a named, auditable script that the
-   next person (or agent) can audit and re-run. Fix stale recipes you
-   encounter the same way.
+   wraps the canonical command, then use it. Keep the recipe thin -- push any
+   credential/secret resolution into the env-aware Python CLI rather than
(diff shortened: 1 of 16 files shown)


joshalbrecht:
lgtm

PR #2106, opened 2026-06-11
minds: reliably route stuck and stopped workspaces to the recovery page

gnguralnick:
## Problem

Opening a stopped or unresponsive Docker workspace could leave the app stuck on the plugin's "Loading workspace" loader, never advancing to the recovery page — even when the backend knew the workspace was unreachable. Two distinct gaps caused this:

1. **A stuck workspace's status could be lost after a chrome reload**, so the recovery auto-redirect never fired.
2. **Opening a workspace the landing page already knows is "Stopped"** still went through the normal loader and waited out the full HEALTHY→STUCK probe-failure window (~5s) before any restart was dispatched — and, combined with (1), could strand on the loader indefinitely.

## Root cause (1): lost one-shot status

The chrome shell auto-redirects to the recovery page when it holds a `stuck` `system_interface_status` for the currently-displayed workspace. That status reaches the renderer as a **one-shot** SSE event, but:

- `main.js` relayed it to the renderer yet did **not** include it in the replayable cache (`primeViewWithCachedChromeState` replayed workspaces/auth/requests/modal only); and
- the backend only (re)emitted `system_interface_status` on a state *transition* or a *brand-new* SSE connection.

So when the chrome renderer reloaded after a workspace went STUCK, its in-memory status was wiped and never restored: the tracker stayed STUCK (no new transition) and the single long-lived main-process SSE never re-pulled a snapshot. The renderer was permanently blind to the stuck state, so the redirect never fired.

This was traced against a live instance (tracker confirmed STUCK via the recovery endpoint; content view confirmed parked on the plugin loader; chrome shell observed reloading afterward) and reproduced/verified by driving the real `chrome.js` in a browser, then confirmed end-to-end with a live CDP A/B (fix ON → redirects in ~0.3s; fix OFF → stranded on the loader).

## Root cause (2): no fast path for a known-stopped mind

`landingRowClick` decided where a click went by looking only at the system-interface health (`stuck`/`restarting`/`restart_failed`). A freshly-stopped mind has no health status yet (the landing page even suppresses its health badge), so the click fell through to the normal `/goto/` loader and had to wait for the tracker to accumulate ~5s of continuous probe failures before a restart was dispatched — despite the landing page already knowing the container is "Stopped" (`mindLivenessByAgent`).

## Fix

- **Electron (`main.js`)**: `primeViewWithCachedChromeState` now replays the latest non-healthy `system_interface_status` for each agent to a freshly (re)loaded chrome/sidebar view, so a reloaded window re-learns stuck workspaces and redirects. The status map drops `healthy` entries so it stays scoped to agents that still need attention (mirroring `chrome.js`'s own semantics).
- **Backend (`app.py`)**: the `/_chrome/events` stream periodically re-asserts current non-healthy statuses (in addition to the connect-time snapshot and per-transition pushes), so a desynced renderer self-heals even if it missed the original event. Re-asserting `stuck` is idempotent client-side (the recovery-redirect lock prevents re-navigation). The re-assert runs on the events-loop wake, so the loop's idle wait is capped at the re-assert interval — making the self-heal cadence actually ~15s (matching the constant) rather than being stretched to the prior 30s idle floor.
- **Landing page (`Landing.jinja`)**: `landingRowClick` now routes a known-`STOPPED` mind straight to the recovery page with `intent=restart`, which confirms `host_offline` via its probe and cold-boots the host immediately. `intent=restart` is required because a never-probed stopped mind still reports HEALTHY to the tracker; without it the recovery handler would 302 the user right back to the loader.

## Testing

- `apps/minds/test/e2e/recovery-redirect.spec.js` (3 tests, all pass): drives the **real** `chrome.js` in a plain browser page (no Electron app, no Docker) and asserts the redirect happens iff a `stuck` status is held for the displayed workspace — baseline redirect; reloaded renderer re-primed *without* the status (reproduces the bug → no redirect); reloaded renderer with the status replayed (the fix → redirects).
- `apps/minds/test/e2e/landing-stopped-mind-restart.spec.js` (2 tests, all pass): renders the **real** landing page via `render_landing_page` and drives the real inline `landingRowClick` — a `STOPPED` mind routes to `/recovery?...&intent=restart`; a `RUNNING` mind navigates to its normal `/goto/` loader.
- The backend contract the fast path relies on (a HEALTHY-tracker mind + `intent=restart` renders/dispatches instead of 302-ing back) is already covered by `test_recovery_page_renders_for_healthy_agent_with_explicit_restart_intent` in the pytest suite (the CI gate).
- Note: the JS e2e specs are committed as fast, runnable regression artifacts but are **not** in the default CI gate (the JS e2e CI job targets `macos-launch.spec.js` only). The Python suite (including the recovery/`intent=restart` tests) runs in CI.
- `ruff check` clean on `app.py`; relevant minds unit/integration tests pass locally; full suite runs in CI.

### DIFF
--- a/apps/minds/imbue/minds/desktop_client/app.py
+++ b/apps/minds/imbue/minds/desktop_client/app.py
@@ -2173,15 +2173,23 @@ async def _handle_chrome_events(
             #
             # Clearing at the bottom of the loop instead would lose the
             # wakeup for any producer that fires between the drain and the
-            # bottom-of-loop clear, leaving the queued item idle for up to
-            # 30s -- a UX regression for health-state transitions like
-            # RESTARTING -> HEALTHY.
+            # bottom-of-loop clear, leaving the queued item idle until the
+            # next idle-wake timeout -- a UX regression for health-state
+            # transitions like RESTARTING -> HEALTHY.
             shutdown_event: threading.Event = request.app.state.shutdown_event
             connected = not await request.is_disconnected()
             while connected and not shutdown_event.is_set():
-                # Wait for a change signal or timeout (timeout for disconnect checks).
+                # Wait for a change signal or timeout. The timeout bounds both
+                # disconnect-detection latency and -- since the periodic status
+                # backstop below only runs on a loop wake -- the worst-case
+                # re-assert cadence on an otherwise-idle connection. Cap it at
+                # the re-assert interval so a steadily-stuck workspace really is
+                # re-asserted on that cadence rather than being stretched to a
+                # longer idle-wake period.
                 try:
-                    await asyncio.wait_for(change_event.wait(), timeout=30.0)
+                    await asyncio.wait_for(
+                        change_event.wait(), timeout=_SYSTEM_INTERFACE_STATUS_REASSERT_INTERVAL_SECONDS
+                    )
                 except TimeoutError:
                     pass
                 # Clear BEFORE draining so any producer firing between drain
--- a/apps/minds/imbue/minds/desktop_client/app.py
+++ b/apps/minds/imbue/minds/desktop_client/app.py
@@ -2052,6 +2052,18 @@ def _handle_dev_styleguide() -> Response:
     return HTMLResponse(content=render_dev_styleguide_page())
 
 
+# How often the chrome-events stream re-asserts the current non-HEALTHY
+# system-interface statuses, on top of the one-shot connect-time snapshot and
+# the per-transition pushes. This is a self-healing backstop: a chrome renderer
+# that lost its in-memory health state (e.g. a reloaded webview whose one-shot
+# ``system_interface_status`` was never replayed) re-learns a still-stuck
+# workspace within this interval and can finally redirect to the recovery page,
+# even though the tracker emitted no fresh transition. Re-asserting ``stuck`` is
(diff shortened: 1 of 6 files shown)


apps/minds/imbue/minds/desktop_client/app.py
```diff
@@ -2052,6 +2052,18 @@ def _handle_dev_styleguide() -> Response:
...
+# system-interface statuses, on top of the one-shot connect-time snapshot and
+# the per-transition pushes. This is a self-healing backstop: a chrome renderer
+# that lost its in-memory health state (e.g. a reloaded webview whose one-shot
+# ``system_interface_status`` was never replayed) re-learns a still-stuck
+# workspace within this interval and can finally redirect to the recovery page,
+# even though the tracker emitted no fresh transition. Re-asserting ``stuck`` is
+# idempotent client-side (the recovery-redirect lock prevents re-navigation), so
+# the only cost is one tiny event per non-healthy agent per interval.
+_SYSTEM_INTERFACE_STATUS_REASSERT_INTERVAL_SECONDS: Final[float] = 15.0
+
```
joshalbrecht, line 2065: I don't love this (feels smelly), but ok with it for now

joshalbrecht:
one minor comment worth potentially thinking about how to fix in the future, but otherwise lgtm

PR #2107, opened 2026-06-11
mngr: pause event --follow I/O for offline agents (persistent gated tail threads)

gnguralnick:
## Problem

A `mngr event <agent> --follow` whose target host is **offline** (stopped but not destroyed) was re-reading the agent's unchanging persisted event files once per second through the read-only volume. For the Docker provider each read is a separate `docker exec` into the shared state container, so a handful of stopped agents drove the Docker engine to a large wasted CPU load.

Observed live before the fix: ~30 `exec_create`/sec into a single `*-docker-state-*` container, pushing that container to ~90% CPU, with the bulk of followers polling agents whose containers weren't running. No actual agent work was happening.

## Fix

All in `libs/mngr/imbue/mngr/api/events.py`. Each per-source tail thread now lives for the whole follow session and **gates its own I/O on a shared online/offline signal** rather than being torn down and recreated on every transition:

- A shared `online_event` (`threading.Event`) is set while online and cleared while offline. While it is clear, every tail thread parks and does **no reads at all** (a stopped agent can't write events) — this is the wasted `docker exec` load, eliminated.
- A shared, swappable `target_holder` holds the current target. On a transition the consume loop just swaps `target_holder[0]` and flips the gate; threads read the current target on their next poll. They are never recreated.
- `_handle_online_offline_transition` collapses to: refresh the target, and on a net change swap the handle + set/clear the gate. The previous `set → join(timeout) → clear → restart` block is gone, which also removes its zombie-thread race (a tail thread that didn't honor the stop within the join timeout could survive the subsequent `stop_event.clear()`).
- The two old thread bodies (`_tail_source_thread_local` pygtail / `_tail_source_thread_remote` polling) are collapsed into stateless one-pass readers (`_read_local_source_once` / `_read_remote_source_once`) driven by a single persistent `_tail_source_thread`. A new `_resolve_read_plan` picks the mechanism (local pygtail vs. whole-file polling) and path from the current target; when those change across a transition the thread re-reads its source from the start, with the existing `emitted_event_ids` dedup in `_consume_event_queue` preventing any double emission.
- The periodic 10s source-directory rescan is still skipped while offline.

This affects every consumer of `mngr event --follow` (the CLI, `mngr forward`, and the minds desktop app), not just one provider.

### Why gate instead of start/stop?

The original CPU problem was the wasted *work* (the per-second `docker exec` reads), not the existence of threads. The earlier iteration of this branch fixed it by *not starting* threads while offline and tearing them down on a transition — but that start/stop churn is exactly what introduced teardown/restart races (the `stop_event.clear()`-after-timeout zombie, offset-file contention on a restart). Gating persistent threads removes the wasted work **and** that whole race class, with strictly less code.

## Tests

- `test_tail_source_thread_does_no_io_while_gate_closed_then_resumes` — the core guarantee: with the gate closed the thread issues zero reads (asserted against a volume-backed offline host), and resumes delivering once the gate opens.
- `test_tail_source_thread_follows_target_swap_without_recreation` — swapping `target_holder[0]` to a new path makes the same thread follow it (no recreation).
- The three `_handle_online_offline_transition` tests updated for the new gate semantics (online → sets gate + swaps target; offline → clears gate; no-op when unchanged).
- `just test-quick libs/mngr/imbue/mngr/api/events_test.py` → **84 passed, 0 failed**. Full `libs/mngr/imbue/mngr/api` dir → 789 passed.
- `ty check` clean, `ruff check` + `ruff format --check` clean, mngr ratchets pass.

## Behavior verification

The wasted-I/O elimination was measured live against minds-staging on the **earlier** (skip-start-threads) implementation of this branch:

| Metric | Before | After |
|---|---|---|
| `exec_create` into state container | ~30/sec (295 in 10s) | ~4/sec (56 in 14s) |
| Per-agent **events** polling (`cat …/events/requests/events.jsonl`, `ls …/events/services`) | dominated the load | **0 occurrences — eliminated** |
| State container CPU | **~90%** | **~0.01% steady** (brief ~7% spike every 10s) |

The remaining ~4/sec is the legitimate **discovery** poll (`mngr observe --discovery-only` reading `/mngr-state/host_state/host-*.json` every 10s), which the investigation identified as *not* the culprit.

The gated rewrite in this PR preserves the same no-reads-while-offline behavior — now locked in by the `…does_no_io_while_gate_closed…` unit test above — but I have **not** re-run the live staging measurement against the rewrite. (Equal-or-better is expected: same offline read-suppression, minus the thread churn.)

## Not yet done / notes

- Full `just test-offload` suite deferred to CI.
- **CI red is unrelated flakes, not this diff.** The failing tests (`agent_creator_test.py::test_start_creation_api_key_ai_does_not_mint_litellm_key` — already `@pytest.mark.flaky`; `mngr_modal/testing_provider_test.py::test_create_host_raises_on_ssh_setup_failure` — teardown error; `test_meta_ratchets.py::test_cli_docs_are_up_to_date` — timed out, passes locally with RC=0) live in other projects, don't import this module, and fail via timeout/teardown under runner contention. This diff is 3 files under `libs/mngr/imbue/mngr/api/`.


joshalbrecht:
I have two questions here:

1. Can events ever get dropped when we transition from online to offline? (I think almost certainly yes -- that's part of why we were reading the offline data. We probably *do* need to continue reading the offline data, but once we've read it a few times (and the thing went offline) then we can probably chill if it hasn't changed, if that makes sense?) Any kind of "we're going to dynamically start and stop these threads" makes me really suspicious about race conditions... better to just chill waiting for the host to come back online than start and stop the thread. Also consider cases like "it is rapidly going online and offline"
2. What remote event files do we even care about from minds? Like, why are we watching these at all? Logically, I'm not sure why we care at all...

joshalbrecht:
lgtm, thanks for the fixes

PR #2108, opened 2026-06-11
fix remote->local clone

evgunter:
was this ever working?

---

## Problem

`mngr clone <remote-agent> <name> --provider local` (e.g. moving an Azure agent to local) always failed with:

```
fatal: destination path '.../agent-....git' already exists and is not an empty directory.
```

## Root cause

In `Host._transfer_git_repo` (`libs/mngr/imbue/mngr/hosts/host.py`), the target `.git` bare repo is always created up front via `git init --bare` (this is idempotent and harmless for the push-based transfer branches). It then calls `_git_push_to_target`.

For a **remote source -> local target** transfer (`same_machine == False`, and the local target has no SSH info so `target_ssh_info is None`), `_git_push_to_target` took a `git clone --mirror <remote> <target>/.git` path. `git clone` requires an empty destination, but `<target>/.git` was just initialized, so the clone always aborted. Every other transfer direction uses `git push` into the pre-initialized bare repo and is unaffected — only this branch used `clone`.

## Fix

Replace the `git clone --mirror` with a mirror-style `git fetch` into the existing bare repo, reusing the same `GIT_MIRROR_PUSH_REFSPECS` the push branches use. This makes the remote-source branch symmetric with the rest of `_git_push_to_target`.

## Testing

Added `test_create_work_dir_git_mirror_from_remote_source_to_local_target`, an acceptance test that drives the remote-source -> local-target git-mirror path via a local sshd. Verified it fails on the pre-fix code with the exact original error and passes after the fix. All 17 `create_work_dir`/git-transfer tests in `test_host.py` pass.


joshalbrecht:
it probably never worked, that'd be cool if it did :)

lgtm

PR #2109, opened 2026-06-11
fix remote->remote clone

evgunter:
**Base:** this PR is stacked on `mngr/clone-fix` (#-base). It harmonizes the two overlapping fixes to `Host._git_push_to_target`.

## Problem

`mngr clone <agent> <new-agent> --provider aws` (and any agent creation transferring a git repo from one remote host to another) failed with:

```
Failed to push git repo from remote source: Warning: Identity file
/Users/.../providers/aws/aws/keys/container_ssh_key not accessible: No such file or directory.
No ED25519 host key is known for [host]:2222 and you have requested strict checking.
Host key verification failed.
```

## Root cause

The remote-source case ran `git push` **on the remote source host** while building `GIT_SSH_COMMAND` with the **target's** SSH key (`-i`) and known_hosts paths. Those live on the local orchestrator machine, not on the source host where the push runs, so ssh couldn't find the identity file or host key and `StrictHostKeyChecking=yes` aborted.

This only affected genuinely remote-to-remote transfers; local-source -> remote-target and remote-source -> local-target run their push/clone locally where the credential paths resolve.

## Fix (harmonized with `mngr/clone-fix`)

`mngr/clone-fix` independently fixed the **remote-source -> local-target** path (its `git clone --mirror` refused the pre-existing bare repo that `_transfer_git_repo` inits; it switched to a mirror-style `git fetch`). Both that fix and this one want the same low-level primitive, so they're unified:

- New `_git_mirror_pull_from_source(source_host, source_path, dest_git_dir)`: `git init --bare <dest>` (idempotent) followed by a mirror-style `git fetch --prune <url> <refspecs>`, run locally with the source's credentials. Works whether `dest` pre-exists (remote->local target) or is fresh (remote->remote relay's temp mirror).
- Remote-source -> remote-target now **relays** through a local temp bare mirror: pull from source with the source's creds, then push to target with the target's, both run locally where those files exist (mirrors the existing remote-to-remote rsync handling).
- The previously-buggy remote-source push-from-source branch is now unreachable and removed; an `assert source_host.is_local` documents the remaining invariant.

## Tests

- `test_git_transfer_remote_to_remote` (this PR): relay between two distinct SSH hosts backed by a local sshd; asserts the source commit and working tree land on the target.
- `test_create_work_dir_git_mirror_from_remote_source_to_local_target` (from `mngr/clone-fix`, retained): covers remote->local through the same primitive.

Together they cover same-machine, local->remote, remote->local, and remote->remote.


joshalbrecht:
lgtm!

(not merging bc non-standard base)

PR #2110, opened 2026-06-11
fix SSH keypair creation race + wrap paramiko error

evgunter:
Split off from \`mngr/aws\` for separate review (per request). The substantive change is the single commit \`mngr/vps_docker: fix racy SSH keypair creation + resilient host discovery\`; the other commits below it are just \`origin/main\` being merged in (can be rebased away for a cleaner diff if preferred).

## Why
Parallel host discovery intermittently aborted \`mngr create\`. Surfaced on the OVH release test (\`test_create_stop_start_destroy\`) and, via the same discovery path, suspected on the Vultr \`test_ssh_connectivity\` flake.

Root cause (two defects):

1. **Racy keypair creation.** \`save_ssh_keypair\` wrote the \`.pub\` non-atomically (truncate-then-write) and \`load_or_create_ssh_keypair\` had no lock. The discovery fan-out opens one SSH connection per VPS (OVH lists every VPS in the account) and each lazily creates the keypair on first use, so multiple threads raced to write the same \`.pub\`. A reader landing in the truncate window read a zero-byte file, and paramiko's per-connection certificate probe then raised \`ValueError: Not enough fields for public blob: []\`.

2. **Over-narrow discovery error handling.** \`_read_records_from_vps\` only caught \`MngrError\`, so that bare \`ValueError\` escaped, propagated as a \`ConcurrencyExceptionGroup\`, and aborted the whole \`create\`.

## Fix
- \`load_or_create_ssh_keypair\` serializes first-time creation behind an \`flock\`; \`save_ssh_keypair\` writes both key files atomically (temp file + \`os.replace\`). Adds a 16-thread concurrency regression test.
- \`_read_records_from_vps\` degrades any per-VPS failure (not just \`MngrError\`) to a warning and falls back to cached records — matching the best-effort pattern the destroy path already uses. Broad-except ratchet 7 -> 8.

## Verification
- New unit test \`test_load_or_create_ssh_keypair_concurrent_first_creation_is_consistent\` passes.
- OVH release suite re-run: 4/4 pass, including the previously-failing \`test_create_stop_start_destroy\`. Additional OVH/Vultr runs in progress.


joshalbrecht:
lgtm after conflicts addressed

PR #2112, opened 2026-06-12
mngr: apply host tmux config on existing servers; forward agent title to terminal tab

weishi-imbue:
## Problem

Agent tmux sessions received the mngr host tmux config only via `tmux -f <config> new-session`, but tmux reads a config file automatically only when it starts the server. Once any agent's server was running, every later session silently kept tmux defaults:

- `status-left-length` stayed 10, clipping `[mngr-<agent>]` in the status bar (the widening from #1716 never applied in practice).
- The `Ctrl-q` / `Ctrl-t` destroy/stop bindings were never bound.

Separately, the config never enabled `set-titles`, so tmux swallowed the title the agent program sets and the outer terminal tab fell back to e.g. `Basic(tmux)` in iTerm2.

A third issue surfaced in review: mngr's config re-sourced the user's `~/.tmux.conf` on every agent creation, which can corrupt a non-idempotent user config.

## Fix

- Run `tmux source-file <config>` right after `new-session`, so the config applies to running and fresh servers alike. The `-f` flag is dropped as redundant. The step is non-fatal (`|| true`): this config is cosmetic (status bar, titles, hotkeys) and must not block agent startup.
- Add `set -g set-titles on` + `set -g set-titles-string "#S  #T"` to the generated config, putting `<session-name>  <pane-title>` in the terminal tab.
- Stop sourcing the user's `~/.tmux.conf` from mngr's config (per @qi-imbue's review). tmux loads `~/.tmux.conf` itself, once, at server start (the agent server is no longer started with `-f`, so tmux's default config search applies); mngr's config now holds only mngr's own settings. This avoids re-running non-idempotent user config (`set -ag`, plugin `run-shell`) on every agent creation.

Verified empirically on an isolated tmux socket: `-f new-session` against a running server leaves `status-left-length` at 10 / `set-titles` off; `source-file` sets them to 20 / on, on both running and fresh servers.

## Tests

- `test_host_create_host_tmux_config_contains_mngr_settings`
- `test_host_create_host_tmux_config_does_not_source_user_config`
- `test_build_start_agent_shell_command_sources_config_after_new_session`

## Notes

- Sessions created before this change aren't re-sourced (the `has-session` early-exit skips them); a one-time `tmux source-file ~/.mngr/tmux.conf` fixes a live server.
- Coordinates with #2176 (mngr's own dedicated tmux socket): this PR will rebase on top of it. With the private server the sourcing fix is no longer a "mngr corrupts your tmux" bug, but it keeps mngr's own agent sessions rendering correctly.


libs/mngr/imbue/mngr/utils/test_ratchets.py
```diff
@@ -167,7 +167,9 @@ def test_prevent_num_prefix() -> None:
 
 
 def test_prevent_trailing_comments() -> None:
-    rc.check_trailing_comments(_DIR, snapshot(0))
+    # The 1 is a misfire: hosts/host.py's _TMUX_SET_TITLES_STRING contains a
+    # space-then-# inside a string literal (tmux format syntax, not a comment).
+    rc.check_trailing_comments(_DIR, snapshot(1))
```
weishi-imbue, line 172: Note for reviewers: this line bumps the `trailing comments` ratchet count to 1. The flagged "violation" is `_TMUX_SET_TITLES_STRING = "#S  #T"` in hosts/host.py — the space-then-`#` is inside a string literal (tmux format syntax for session name / pane title), not a trailing comment. The ratchet regex cannot distinguish string contents from comments, so this is a recorded misfire rather than a real violation.

qi-imbue:
Oh hey, I actually have an opinion on this since it's something I wanted to do but fell through the cracks :)

The user's `.tmux.conf` is not necessarily idempotent, and IMO we should not source it from within our `.tmux.conf`, and instead we should:

- Drop whatever code there is that points the initial tmux server at our custom `.tmux.conf` - i.e. a new tmux server should just source the user's `.tmux.conf`.
- Remove the line that sources the user's `.tmux.conf` from our custom `.tmux.conf`.
- Keep doing what this PR is doing - sourcing our `.tmux.conf` at agent creation.

---

There's another problem with our tmux configuration, which is that it will end up mutating the user's existing tmux session's config, and the user doesn't necessarily expect or appreciate that. For that I have always thought mngr's tmux socket should be separate from the system default, but the last time I tried to do this I ran into some weird issues and gave up.

joshalbrecht:
I agree with Qi's assessment!

As a separate improvement, making an entirely separate tmux server does sound like it might be a reasonable option to have (to keep things fully isolated), though I'm not entirely sure it should be the default (that would probably be fairly confusing to users)

@qi-imbue thoughts on the default after we make that other improvement?  (I'll add a ticket ticket for it)

qi-imbue (approved):

PR #2113, opened 2026-06-12
Minds: file-sharing fixes

hynek-urban:
(Accompanied by https://github.com/imbue-ai/forever-claude-template/pull/158.)

- Accept "~" in place of the home directory in permission requests.
  - The permission request backend still expands this so that the resulting permission is unambiguous.
  - The agent learns about the proper full path to use in actual webdav requests from the response.
- Properly deal with file names containing spaces or other characters that need to be encoded.


joshalbrecht:
lgtm

PR #2114, opened 2026-06-12
Always offer the `any` permission.

hynek-urban:
The implicit `any` permission was absent from the list of available permissions that agents got to see before sending permission requests. For services that didn't have any narrower permission definitions (like Linear), it resulted in an empty set of available permissions and the agent didn't know what to do. This PR fixes that.


joshalbrecht:
lgtm

PR #2115, opened 2026-06-12
Hide stale permission requests

hynek-urban:
When a Mind has been shut down, hide its old permission requests that were neither approved nor denied, if any. (Otherwise the permission dialogs don't work well anyway.)

Starting the Mind resurfaces the permission requests again.


joshalbrecht:
lgtm

PR #2121, opened 2026-06-12
Docker: tolerate 409 conflict when removing in-use build image during destroy/GC

weishi-imbue:
## Summary

`mngr destroy` (and any GC run) aborts with an uncaught `docker.errors.APIError: 409 Conflict` when the Docker provider tries to untag a per-host build image that a container still references:

```
409 Client Error ... Conflict ("conflict: unable to delete mngr-build-host-<id>:latest (must be forced) - container <c> is using its referenced image <img>")
```

### Root cause
`DockerProviderInstance._remove_build_image` (`libs/mngr/imbue/mngr/providers/docker/instance.py`) calls `images.remove(tag)` with no `force` and no exception handling. Docker returns 409 when the tag is the last reference to an image still used by a container. The raised `APIError` is not an `MngrError`, so it escapes the GC worker (`api/gc.py` only catches `MngrError`), re-raises at `future.result()`, and aborts the entire `mngr destroy` / GC run. This is reachable because:

- `delete_host` (the GC path) untags the build image but never removes the container, so a still-present container pins it.
- Per-host build images built from the identical default Dockerfile deduplicate to one image id with many tags, so a container from one host can pin the image another host's tag points to.
- Unlike snapshot-image and volume cleanup in the same `delete_host`, the build-image removal does not tolerate failures.

### Fix
Make the build-image untag best-effort: catch `docker.errors.DockerException` and log a warning, mirroring the snapshot/volume cleanup. `force=True` is intentionally **not** used (it would delete an image still referenced by another host's live container).

### Tests
Adds `test_remove_build_image_tolerates_container_still_referencing_it` (`docker_sdk`), which creates a container referencing a single-tagged build image and asserts `_remove_build_image` logs-and-continues (the tag survives). Verified locally that the test **fails (raises the 409 `APIError`) without the fix and passes with it**. The full acceptance/release suite runs in CI via offload rather than locally, per repo convention.

### Origin
Started as a diagnosis task (a GitHub issue was filed separately with the full root-cause write-up and reproduction); the fix, changelog, and regression test were added after the stop hook required a PR. Architecture verification, autofix, and conversation review gates have been run.


weishi-imbue:
closing in favor of #2209

PR #2123, opened 2026-06-12
validate --adopt-session up front

evgunter:
## Problem

Passing an unknown (or ambiguous) `--adopt-session` ID to `mngr create` crashed with a full "Unexpected error" traceback instead of a clean user-facing error:

```
ERROR: Unexpected error:
Traceback (most recent call last):
  ...
imbue.concurrency_group.concurrency_group.ConcurrencyExceptionGroup: Session ... not found in /Users/ev/.claude/projects. ...
```

## Root cause

The session ID was only validated in the claude agent's `on_after_provisioning`, which runs inside `provision_agent`'s `ConcurrencyGroup`. On exit, the group re-wraps the `UserInputError` in a `ConcurrencyExceptionGroup` (no longer a `ClickException`), so the top-level handler reports it as an unexpected error.

## Fix

Validate the `--adopt-session` ID up front in the claude plugin's `on_before_create` hook -- the same hook that already fail-fast-rejects non-claude agent types and the `--adopt-session` + `--from` combination. It runs before any host/worktree is created and outside the provisioning `ConcurrencyGroup`, so a bad ID surfaces as a clean, fail-fast `Error: ...` message. The check reuses the existing `_resolve_adopt_session` resolver; the session source is always local, so the early result matches the provision-time resolution.

This is deliberately narrow: only `--adopt-session` validation is affected. All other provisioning errors keep their existing behavior (a genuine bug still gets its full traceback). Earlier iterations that unwrapped *any* wrapped `MngrError` at the CLI handler / provision call site were dropped as too broad.

## Verification

- New unit test `test_on_before_create_rejects_unknown_adopt_session` asserts a clean `UserInputError` ("Session ... not found") from `on_before_create`.
- Two existing `on_before_create` "passes" tests updated to use a resolvable session path (the hook now actually resolves the ID).
- `plugin_test.py -k 'adopt or on_before_create or on_after_provisioning'`: 19 passed.

## Review gates

- Architecture verification: passed (re-verified for the targeted approach; endorsed as the right placement, reusing the existing fail-fast validation hook).
- Autofix: clean (0 issues).


joshalbrecht:
lgtm

PR #2125, opened 2026-06-12
minds: get launch-to-msg CI green (+ bundle mngr_aws)

weishi-imbue:
Fixes all three jobs of the `minds-launch-to-msg` workflow. Validated: full workflow green (run [27575511226](https://github.com/imbue-ai/mngr/actions/runs/27575511226)), FF-able from main.

- **build** — bundle `imbue-mngr-aws` in the four workspace-package lists (`build.js`, `env-setup.js`, `build_test.py`, electron `pyproject.toml`). #2145 added it as a minds dep but not to the bundle, so `uv lock` resolved it from PyPI and its day-old release hit the 14-day supply-chain cooldown — failing every build, including main's.
- **macos_launch** — assert against the content window (`pickContentWindow`) using stable hooks (`#skip-account-btn` / `a[href="/auth/login"]`) instead of the old `firstWindow()` + `Log in` button selectors that broke on the welcome-splash redesign.
- **launch_to_msg** — run the `mngr list` cross-check from `$HOME` like minds.app does, so it reads the minds host profile, not the mngr repo's own `.mngr/[providers.aws]` (which a bundled subset-mngr rejected).
- **launch_to_msg** — the W2-destroy poll now waits for actual teardown (`done`/404) instead of bailing on the transient `failed` status (pid-dead + host briefly still in discovery during slow lima teardown). Still fails a genuinely stuck host at the 240s deadline.


joshalbrecht:
lgtm

PR #2126, opened 2026-06-13
mngr_vps_docker: per-name btrfs snapshots to fix empty gVisor backups

joshalbrecht:
## Problem

On vps-docker (gVisor/`runsc`) hosts, the `host_backup` service captured data only on the **first** backup after boot; every subsequent hourly backup produced a 0-byte restic snapshot.

Root cause: the outer snapshot helper reused a single fixed btrfs subvolume path (`snapshots/current`), deleting and recreating it each cycle. The container reads that path through the gVisor gofer, which caches a handle to the first subvolume it opened. After the first delete+recreate, the container kept reading the stale (deleted) subvolume (`readdirent ... no such file or directory`, restic exit 3), so restic backed up nothing.

## Fix (this repo)

Rework `resources/snapshot_helper.sh` so each snapshot is created at a unique, caller-named path (`snapshots/<name>`) — never reusing a path, so the gofer always does a fresh lookup. The helper:
- creates a `snapshot` at `snapshots/<request_id>`, failing (not overwriting) on a name collision;
- deletes a `cleanup` target by name, validated to be a single path component (rejects `/` and `..`) so it can't escape the snapshots dir or touch the live subvolume; "already gone" is success.

Added functional helper tests (fake `btrfs`/`inotifywait`): per-name create, path-traversal rejection, idempotent cleanup, delete-by-name.

## Paired change (forever-claude-template)

The inner `host_backup` lib drives the new naming and garbage-collects old snapshots down to a retained count (`max_local_snapshots`, default 5), keeping the newest N each tick. That change lives in `forever-claude-template` and ships to hosts via `vendor/mngr`; it must land together with this one. New hosts only.

## Design

Full plan: `blueprint/host-backup-snapshot-rotation/plan-host-backup-snapshot-rotation.md`.

## Tests
- `mngr_vps_docker` fast suite: 276 passed (incl. ratchets); helper file: 6 passed.
- `ty` checks on changed files: clean.
- forever-claude-template `host_backup` full suite (with coverage): 68 passed.


joshalbrecht:
lgtm

PR #2127, opened 2026-06-13
flush transcript before switching to WAITING state

evgunter:
kinda ugly but seems correct

---

## Problem

Investigating the comment in Catalyst's `mngr_runner.py`:

> A short unconditional wait so the common-transcript converter (which writes on a ~5s timer) catches up with whatever the turn-end signal raced ahead of

The runner harvests a step's final answer from the **common transcript**, but learns the turn is over from a **separate** signal (`mngr wait --state WAITING`, i.e. the `active` marker being cleared). Those travel down independent, unordered pipelines:

```
session JSONL → stream_transcript.sh (1s) → raw → common_transcript.sh (5s) → common transcript → mngr event   [the ANSWER]
Stop hook → clear `active` marker → mngr wait --state WAITING                                                   [DONE]
```

DONE can beat the ANSWER by up to ~6s + conversion time. The old band-aid was a fixed 10s `time.sleep` on every step: slow, and still timing-dependent (no positive confirmation).

## Fix (mngr-side, the real fix)

Make the WAITING signal **causally downstream** of the converter. The converters already shipped an unused `--single-pass` flag; this wires it in:

- **claude** (`wait_for_stop_hook.sh`): synchronous `--single-pass` of stream then common, in `run_post_completion`, before `mark_inactive` clears the marker.
- **antigravity** (`statusline.sh`): same, on the busy→idle edge, before clearing the marker (edge-gated so it costs at most one pass per turn).

Both are best-effort (gated on script existence, `|| true`) so a flush failure can never strand the turn-end signal.

A concurrency wrinkle this surfaces: the on-demand flush runs alongside the 5s daemon, and both append to the same file with read-then-append dedup → possible duplicate lines. So the converter now takes a coarse **mkdir-based lock** (the codebase's portable idiom, same as codex's `codex_marker_lock`) around its read-modify-write. Stale locks (>1min) self-heal; timeout tunable via `MNGR_CONVERT_LOCK_TIMEOUT`.

## Catalyst side (separate `ai-scientist` repo, branch `mngr/catalyst-waiting`)

With the signal now ordered after the converter, the fixed 10s sleep becomes a bounded early-exit poll: parses on the first iteration in the common case (~zero added latency), with the grace budget kept as a fallback for the premature-`end_turn` case and for an mngr build without the flush.

## Scope

claude + antigravity only (the agent types Catalyst uses). codex is left alone (its root/subagent marker recompute is more delicate and unused here).

## Tests

- converter lock: held-lock skip, stale-lock break, concurrent-passes-no-duplicates (both plugins)
- statusline: flush-before-clear ordering + pipeline order; no flush off-edge or when busy
- claude `wait_for_stop_hook.sh` has no unit harness (pre-existing; needs `/proc`); its flush mechanism is verified manually + via the shared converter tests
- catalyst runner tests updated for the new constant/loop

## Shared library (de-duplication)

The convert lock and turn-end flush were initially duplicated across claude's and antigravity's `common_transcript.sh` / turn-end hooks. They are now extracted into a shared shell library `libs/mngr/imbue/mngr/resources/mngr_common_transcript_lib.sh`, provisioned to every agent's `commands/` dir alongside `mngr_log.sh` and `mngr_transcript_lib.sh` (via `Host._ensure_shared_shell_libs`) -- the same shared-shell-lib pattern already used for the raw-streamer primitives. Converters source it for the lock; turn-end hooks source it defensively and call the shared flush. codex's converter is left untouched (no flusher today; the lib is there for it to adopt later).

### DIFF
--- a/libs/mngr/changelog/mngr-catalyst-waiting.md
+++ b/libs/mngr/changelog/mngr-catalyst-waiting.md
@@ -1,0 +1,12 @@
+Added a shared shell library `mngr_common_transcript_lib.sh`, provisioned to every
+agent's `commands/` dir alongside `mngr_log.sh` and `mngr_transcript_lib.sh` (via
+`Host._ensure_shared_shell_libs`). It centralizes the common-transcript converter
+primitives shared across agent plugins:
+
+- the convert lock (a coarse mkdir-based mutex serializing the converter's
+read-modify-write so the background daemon and an on-demand `--single-pass` flush
+can't append duplicate events), and
+
+- the turn-end flush (one synchronous `--single-pass` of the raw streamer + common
+converter, in pipeline order), used by agent turn-end hooks so a WAITING-signal
+consumer can't outrun the converter.
--- a/libs/mngr/imbue/mngr/hosts/host.py
+++ b/libs/mngr/imbue/mngr/hosts/host.py
@@ -2269,7 +2269,11 @@ class Host(OuterHost, BaseHost, OnlineHostInterface):
             with log_span("Calling on_after_provisioning for agent {}", agent.name):
                 agent.on_after_provisioning(host=self, options=options, mngr_ctx=mngr_ctx)
 
-    _SHARED_SHELL_LIB_NAMES: ClassVar[tuple[str, ...]] = ("mngr_log.sh", "mngr_transcript_lib.sh")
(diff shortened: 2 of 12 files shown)


joshalbrecht:
agreed -- this is disgusting, but better than saying that we're waiting when, in fact, the common transcript is not yet up-to-date

lgtm

PR #2132, opened 2026-06-13
minds: restore checkout in clone_git_repo (fix local-worktree create)

joshalbrecht:
## Summary

Fixes agent creation from a **local git worktree on a target branch** (the `minds-start` dev flow for docker/lima), which currently fails with:

```
git checkout failed: error: The following untracked working tree files would be overwritten by checkout
```

### Root cause

A recent change (`02bb71b44`, "clone_git_repo + checkout_branch accept commit SHAs") rewrote `clone_git_repo` from `git clone --single-branch --branch <ref>` (which **checks out** a working tree) to `git init` + `git fetch` (which does **not**), in order to accept commit SHAs. The surrounding create flow was unchanged: for a local-worktree source it clones, then overlays the worktree with `rsync_worktree_over_clone`, then calls `checkout_branch`. With no checkout during the clone, the rsync'd files land **untracked**, and `checkout_branch` then aborts rather than clobber them. (`rsync_worktree_over_clone`'s docstring assumes it overlays a *checked-out* `git clone`.)

The ordering was never the problem — only the dropped checkout. This affects docker and lima local-worktree creates equally.

### Fix

Restore a final `git checkout --detach FETCH_HEAD` step in `clone_git_repo` so it produces a checked-out clone like `git clone` did. `checkout_branch` still runs afterward to rename the detached HEAD to a real local branch (so SHA/tag handling is unchanged).

### Tests

- `test_clone_git_repo_checks_out_working_tree` — asserts `clone_git_repo` leaves a populated, *tracked* (clean `git status`) working tree.
- `test_worktree_overlay_preserves_uncommitted_edits` — asserts the full clone → rsync-overlay → checkout flow succeeds and keeps the worktree's uncommitted edits.

Full `agent_creator_test.py` passes (44/44); the pre-existing `clone`/`checkout` tests still pass with the change.


joshalbrecht:
lgtm (and this came from local testing to ensure that this works)

hopefully it doesn't break the desktop client but i dont see any reason why it would

PR #2133, opened 2026-06-13
Stabilize flaky codex marker-lock concurrency smoke test

joshalbrecht:
## Problem

CI on the docker-restart merge failed on a single test:

`libs/mngr_codex/imbue/mngr_codex/resources/clear_active_marker_test.py::test_concurrent_root_stop_and_last_subagent_stop_clears_marker` — `Failed: Timeout (>10.0s) from pytest-timeout`.

The traceback shows the timeout firing during `subprocess.Popen` / `_fork_exec`, i.e. the OS fork itself, not a deadlock in the lock protocol. The test runs 8 iterations, each forking ~4 short-lived bash subprocesses (32 total). That is well under a second when the machine is idle (~1.3–2.6s locally), but on a heavily loaded CI runner (12.7k tests fanned across sandboxes) the fork/exec overhead alone can exceed the suite-wide 10s timeout. This failure is unrelated to the docker-restart change (the branch's diff vs `main` is empty; Gabriel's docker work is already in `main`).

## Fix

- `@pytest.mark.timeout(60)` — give this subprocess-heavy test realistic headroom over the global 10s budget. A genuine deadlock would still hang far longer than 60s, so this does not mask real bugs.
- `@pytest.mark.flaky` — let offload retry it automatically if it still occasionally exceeds under extreme load.

No production behavior changed; this is a test-only stabilization.

## Tests

- `clear_active_marker_test.py` (all 12 tests) and `mngr_codex/test_ratchets.py` pass locally.
- Full suite runs in CI.


joshalbrecht:
lgtm

PR #2134, opened 2026-06-13
prevent duplicate changelog apps + organize dev changelog by date

evgunter:
the prompt for the dev changelog is definitely more confusing now, unfortunately; i could do some silly thing where we also have it labeled by release even though that doesn't really make logical sense

---

## Why

Two issues surfaced from the nightly changelog-consolidation cron:

1. **Duplicate cron deployment.** Two `mngr/changelog-consolidation-*` branches were created within ~45 min on 2026-06-13. Investigation found **two Modal apps deployed** in the schedule's isolated environment: the tracked `mngr-schedule-changelog-consolidation` and an orphaned `mngr-changelog-schedule-modal` (deployed 2026-05-28 under a prior app-naming scheme). `mngr schedule remove` only stops the app matching the *current* computed name, so the old-named app was never stopped and kept firing nightly. (The orphan has already been stopped manually.)

2. **`dev/CHANGELOG.md` `[Unreleased]` grew unbounded.** `dev` is never released, so `release.py`'s finalize step never cuts its `[Unreleased]` into a dated section. It had accumulated ~100 lines since 2026-05-13, even though the file already had older `## <date>` sections.

## What

**Orphan-proofing the deploy:**
- `changelog_deploy.sh` now stops *every* app in the schedule's dedicated Modal environment before redeploy, via a new `--stop-all-apps` action in `changelog_schedule_utils.py`. Naming-scheme changes can no longer orphan apps.
- `modal app stop` now passes `--yes` in `remove_modal_schedule` (`mngr_schedule`), `modal_nuke.py`, and the test cleanup helper. Newer Modal CLIs prompt for confirmation and abort with "no interactive terminal detected" when run non-interactively — this was the deploy-time error.

**Date-organizing the dev changelog:**
- New `scripts/changelog_finalize_dev.py` + `cut_changelog_unreleased_to_date` in `changelog_release_utils.py` cut `dev`'s transient `[Unreleased]` into a `## <date>` section each consolidation run (mirroring `UNABRIDGED_CHANGELOG.md`). `dev` carries no standing `[Unreleased]` between runs.
- The consolidation prompt gains a step to run it, and step 4 now lets `dev` create its `[Unreleased]` on demand.
- The existing backlog was cut under `## 2026-06-13`.

## Note
The changelog cron must be **redeployed** (`./scripts/changelog_deploy.sh`) for the prompt/script changes to take effect — the cron snapshots the repo at deploy time.

## Tests
- New unit tests for the env sweep (`changelog_schedule_utils_test.py`) and the date-cut (`changelog_release_utils_test.py`); 62 changelog-script tests pass, `ty` clean. The env sweep was also dry-run against real Modal.


libs/mngr/imbue/mngr/cli/extras_test.py
```diff
@@ -360,6 +360,10 @@ def test_plugins_status_returns_string() -> None:
     assert len(status) > 0
 
 
+# Flaky on offload: the status code paths shell out (plugin/completion/claude
+# status) and can occasionally exceed the 10s pytest-timeout under sandbox load
+# (observed at 10.05s in CI; ~0.4s locally). Retry rather than fail the suite.
+@pytest.mark.flaky
```
evgunter: general question: it seems better to me to just bump this timeout rather than mark it flaky, but the guidance specifically says not to do this. why?

libs/mngr/imbue/mngr/cli/extras.py
```diff
@@ -525,8 +525,16 @@ def _install_default_agent_type(
...
-    """Print the status of all extras."""
+def _print_extras_status(
+    *,
+    claude_status_fn: Callable[[], tuple[bool, dict[str, bool]]] = _claude_plugin_status,
+) -> None:
+    """Print the status of all extras.
+
+    ``claude_status_fn`` is injectable (mirroring the ``status_fn`` seam on the
+    ``_install_*`` helpers) so tests can avoid shelling out to the ``claude``
+    CLI -- a Node process whose startup is the slow, variable part of this call.
```
evgunter, line 536: could replace all of these with monkeypatching; not sure which is better

libs/mngr/imbue/mngr/cli/extras_test.py
evgunter: (fixed this specific instance by i-can't-believe-it's-not-monkeypatching claude out since its presence was not actually material to the thing being tested)

joshalbrecht: what guidance says to mark as flaky instead of increasing the timeout?

I agree with you that that doesn't seem quite right...

libs/mngr/imbue/mngr/cli/extras.py
joshalbrecht: tbh I'm pretty confused what claude is doing in here at all? like, why is it special? I thought we moved away from it being special at all?

joshalbrecht:
lgtm other than comments

libs/mngr/imbue/mngr/cli/extras.py
evgunter: i'm not sure what you mean by "special", but we detect whether the user has `claude` to decide how prominently to offer the `mngr_claude` plugin to them

libs/mngr/imbue/mngr/cli/extras_test.py
evgunter: https://github.com/imbue-ai/code-guardian/blob/main/plugins/imbue-code-guardian/scripts/stop_hook_pr_and_ci.sh

```
        log_error "CI tests have failed for the PR!"
        log_error "Use the gh tool to inspect the remote test results for this branch and see what failed."
        log_error "Note that you MUST identify the issue and fix it locally before trying again!"
        log_error "NEVER just re-trigger the pipeline!"
        log_error "NEVER fix timeouts by increasing them! Instead, make things faster or increase parallelism."
        log_error "If it is impossible to fix the test, tell the user and say that you failed."
        log_error "Otherwise, once you have understood and fixed the issue, you can simply commit to try again."
```

evgunter: that plus `CLAUDE.md` saying to mark things with `@pytest.mark.flaky` seems to make agents reliably believe that they should mark things as flaky instead of bumping the timeouts (they are always telling me this)

libs/mngr/imbue/mngr/cli/extras.py
evgunter: it's actually for detecting whether we should offer `code-guardian` and `mngr-claude-skills`

PR #2140, opened 2026-06-14
improve help text and readme for mngr usage + add actual synopses for some commands

evgunter:
## Summary

Make `--since` (the way to bound which events `mngr usage` aggregates by age) discoverable on both surfaces it had been hidden on:

- **README** (`libs/mngr_usage/README.md`): adds a dedicated `## Filtering by event age` section naming `--since DURATION` as the event-age filter, clarifying that it only shapes the cost surface (rate-limit windows always use the freshest reading), and noting that `--max-age` is *not* a filter -- it's a stale-warning threshold for the snapshot file.
- **`mngr usage --help`**: moves the four group-level options out of the catch-all "Ungrouped" heading the click-option-group renderer was using. `--since` and `--preserved/--no-preserved` now render under the existing **Filtering** group (joined via `@add_agent_filter_options`), matching the sibling `mngr usage wait` subcommand where `--preserved/--no-preserved` already lives. `--max-age` and `--detail` render under a new **Display** group, matching the convention used in `mngr transcript` / `mngr events`.

No behavior change. Only the README, the help-text grouping, and the per-PR changelog entry move.

## Test plan

- [x] `uv run pytest libs/mngr_usage` -- 147 passed.
- [x] `uv run mngr usage --help` -- "Display" and "Filtering" groups render as expected; no "Ungrouped" section remains.
- [x] `uv run mngr usage wait --help` -- unchanged (the sibling already grouped correctly).
- [ ] CI on this PR.


libs/mngr_usage/README.md
```diff
@@ -42,6 +42,23 @@ agent's accumulated cost and rate-limit history still counts toward the totals.
...
+`api_cost.*`) computed off them. Default is 24h, configurable via the
+`since_seconds` option on the `usage` plugin config.
+
+`--since` only shapes the cost surface. Rate-limit windows always reflect the
+freshest reading across all agents -- they track an account-level counter, so
+there's no older reading to drop.
+
+`--max-age` is **not** an age filter: it's only the stale-warning threshold
+for the snapshot file (controls whether the human output prints a "snapshot
+last updated X ago" warning). It does not change which events are aggregated.
```
evgunter: possibly the current shape is kind of dumb and i should change it

joshalbrecht: minor, but yes, you'll probably want to change that flag name to make more sense (perhaps be longer, mention that it is just about warning, etc)

joshalbrecht:
lgtm other than the one comment

PR #2141, opened 2026-06-14
actually enforce changelog entries existing

evgunter:
i'm impressed that the pure CLAUDE.md adherence was so good i didn't notice this until now!

doesn't run in local pytest; depends on the PR's base branch, and it seems more annoying than helpful to require that the PR exists and has a base branch set / in-progress work otherwise correctly configures a base branch

---

## Problem

The per-PR changelog requirement was **unenforced**. The check ran as `test_pr_has_changelog_entry` (an `@pytest.mark.acceptance` test) inside the offload Modal sandbox, but that sandbox:

- does a fresh `git init` in `post-source-setup.sh` (so `main == HEAD`), and
- never fetches `origin` (sandboxes have networking disabled), and
- never receives `GITHUB_BASE_REF` (the justfile only forwards `GITHUB_HEAD_REF`/`GITHUB_REF_NAME`).

So `_resolve_diff_base()` fell through to `main`, which resolved to the `sandbox-init` commit (== HEAD), making `git diff main...HEAD` empty. No projects were ever flagged, and the test passed vacuously no matter what. PR #2139 demonstrated this: it bumped `libs/mngr_codex` with no changelog entry and CI went green.

## Fix

The check has exactly one input — the set of files the PR changed vs. its base — and that input only exists where a real base ref is present (a local checkout or the GitHub Actions runner), never in the sandbox. So enforcement moves out of the sandbox entirely:

- **`scripts/check_changelog_entries.py`** — a dependency-free (stdlib + the existing pure-stdlib `scripts/changelog_projects.py`) gate that resolves the diff base, computes the changed files, maps them to projects, and checks each touched project has its `<project_dir>/changelog/<branch>.md` entry.
- It **fails loudly** (exit 2) rather than passing when it cannot resolve a base distinct from HEAD — a vacuous green is impossible by construction now.
- **`check-changelog` CI job** (`.github/workflows/ci.yml`) runs it on the orchestrator on `pull_request`, after fetching the base branch. No `uv sync` needed, so it's fast.
- **`just check-changelog`** runs the same gate locally.
- The old sandbox-bound acceptance test (and its now-unused helpers) is removed from `test_meta_ratchets.py`.

## Tests

`scripts/check_changelog_entries_test.py` covers the mapping, the missing-entry detection, and crucially the regression: a single-commit repo where `main == HEAD` (the sandbox shape) must error, never pass.

## End-to-end verification

The gate was exercised live against this PR, not just unit-tested:

- Temporarily removed this PR's changelog entry (`dev/changelog/mngr-changelog-enforcement.md`) and pushed (commit `bd9687a01`). The **`check-changelog` CI job failed** with exit 1 on the real GitHub Actions runner:
  > changelog gate FAILED: missing changelog entries for branch 'mngr/changelog-enforcement' (diff base 'origin/main', via git diff --name-only origin/main...HEAD). This PR touches project(s) ['dev']; each needs its own entry file.

  Note the diff base resolved to `origin/main` (a real base distinct from HEAD) — exactly the diff the old in-sandbox test could never compute.
- Restored the entry (revert commit `c69977cf0`); the gate returns to passing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,36 @@ on:
     branches: ['**']
 
 jobs:
+  # Authoritative changelog gate. Runs on the orchestrator (not in an offload
+  # sandbox) because it diffs the PR branch against its real base branch, and
+  # the sandbox has no base ref -- doing the diff there passes vacuously. Pure
+  # stdlib, so no `uv sync`; just enough git history to compute the diff.
+  check-changelog:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
+    steps:
+      # Shallow checkout, then deepen ONLY HEAD's ancestry (same rationale as
+      # the offload jobs below: a bare `--unshallow` pulls every branch). The
+      # merge-base with the base branch is an ancestor of HEAD, so deepening
+      # HEAD plus fetching the base tip is enough for a three-dot diff.
+      - uses: actions/checkout@v6
+
+      - name: Fetch base branch and deepen HEAD for the changelog diff
+        run: |
+          if [ "$(git rev-parse --is-shallow-repository)" = "true" ]; then
+            git fetch --no-tags --unshallow origin "$(git rev-parse HEAD)"
+          fi
+          git fetch --no-tags origin \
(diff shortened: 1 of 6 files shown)


joshalbrecht:
lgtm

I've also noticed cases like this, thanks for finding the cause

PR #2143, opened 2026-06-14
make cli errors red

evgunter:
this has been bugging me for a while--is there a specific reason we wanted it to be plain text?

---

## Problem

A user-facing failure like the GCP "run \`mngr gcp prepare\` first" message (raised as a \`MngrError\`) was printed by click's default \`ClickException.show()\` as a plain \`Error: ...\` line in the **same color as normal output**, so it blended in. A new user could easily miss it.

## Fix

Override \`MngrError.show()\` to wrap its \`Error:\` line in the shared \`ERROR_COLOR\` (bold red) — the exact color the logging layer already uses for the \`ERROR:\` prefix on \`logger.error\`. This flags every \`MngrError\` subclass across the CLI (AWS, GCP, host/agent errors, etc.) in one place.

- **Visually flagged, not a crash:** exit semantics are unchanged (still a clean exit 1, never a traceback). The message just reads as an intentional, actionable error.
- **Respects the existing color policy:** gated on \`should_use_color\` (TTY + honors \`NO_COLOR\`), so piped/captured output stays plain.
- \`_should_use_color\` is promoted to public \`should_use_color\` in \`utils/logging.py\` so the foundational \`errors\` module can consult the same policy (no import cycle: \`logging\` only depends on \`primitives\`).

## Verification

- New unit tests in \`errors_test.py\` cover the TTY-colored, non-TTY-plain, \`NO_COLOR\`, and \`user_help_text\` paths.
- Manually verified end-to-end under a pty:
  - TTY: \`\x1b[1;38;5;196mError: Agent not found: ...\x1b[0m\`
  - \`NO_COLOR=1\`: plain \`Error: Agent not found: ...\`

Tests run: \`just test-quick "libs/mngr/imbue/mngr/errors_test.py libs/mngr/imbue/mngr/utils/test_ratchets.py libs/mngr/imbue/mngr/utils/logging_test.py libs/mngr/imbue/mngr/main_test.py"\` — 188 passed, 0 failed.


joshalbrecht:
lgtm

PR #2144, opened 2026-06-14
fix error message referencing removed flag

evgunter:
## Summary
The "branch already checked out" error from `mngr create` suggested `--in-place`, a flag that no longer exists — it was consolidated into `--transfer` in `36436099d` (`--in-place` -> `--transfer=none`). That consolidation never updated this string literal, so users were pointed at a non-existent flag.

This corrects the suggestion to `--transfer=none` and rewords the line to tell users to `cd` into the existing worktree before re-running (since `--transfer=none` only applies when target == source path).

## Test
`test_worktree_already_checked_out_gives_helpful_error` passes.


joshalbrecht:
lgtm

PR #2145, opened 2026-06-14
minds: add AWS compute provider (rename CLOUD launch mode to VULTR)

joshalbrecht:
## Summary

Adds **AWS** as a compute-provider option in the minds app, alongside Docker, Lima, Vultr, and Imbue Cloud. An AWS workspace runs in a runsc-hardened Docker container on an Amazon EC2 instance (the same outer-host/container model as the Vultr/OVH providers, so the secure latchkey gateway runs on the EC2 host outside the agent's container).

Implements the plan in `blueprint/aws-minds-compute-provider/`.

## What changed

- **`libs/mngr_aws`**: `mngr aws prepare` is now **read-only-first** — when the `mngr-aws` security group already exists with the required SSH ingress it issues no write call, so a re-run succeeds with a describe-only key. Privileged create/authorize only happen when something is actually missing.
- **`apps/minds`**:
  - New `LaunchMode.AWS`; the existing `CLOUD` mode was renamed to `VULTR` (clean rename, no alias).
  - Per-region `[providers.aws-<region>]` blocks written into the mngr profile settings at startup (only when AWS credentials are present, to avoid dead providers / discovery noise). The create address is `system-services@<host>.aws-<region>`.
  - Create form: AWS region selection (the regions with pinned default AMIs), geo-nearest default, and an inline note that AWS credentials come from the environment (`AWS_*` / `AWS_PROFILE` / `~/.aws`).
  - minds auto-runs `mngr aws prepare` for the chosen region before each AWS create.
  - The workspace listing shows a compute-provider label on every row; `aws-<region>` collapses to "AWS".
  - Depends on `imbue-mngr-aws`.
- **forever-claude-template** (committed in the external worktree on the same branch): new `[create_templates.aws]` block — runsc container, long-lived host (`idle_mode = "disabled"`, no `auto_shutdown_seconds`).

## Testing

- New unit tests: provider-label collapse, AWS create-command/address mapping, startup AWS-block writing (credential-gated), and the read-only-first prepare path.
- New manual release test (`apps/minds/test_aws_workspace_release.py`, gated on `MNGR_AWS_RELEASE_TESTS=1` + AWS creds) that drives a real AWS create through the minds command shape and verifies the runsc container on the EC2 outer host.
- Local: `mngr_aws` suite (131 passed), minds unit/integration (1345+ passed), ratchets, lint, and `ty` all clean.

Corresponding FCT PR: https://github.com/imbue-ai/forever-claude-template/pull/164


joshalbrecht:
lgtm

PR #2146, opened 2026-06-14
waiting_reason for opencode

evgunter:
Stacked on `mngr/waiting-reason` (the codex `waiting_reason` PR).

Adds the `waiting_reason` listing field to the **opencode** agent type, reaching parity with claude and codex. `mngr list` now reports *why* a WAITING opencode agent is blocked:

- **PERMISSIONS** -- a tool is blocked on an approval prompt (an `ask` permission policy).
- **END_OF_TURN** -- the agent is idle with its turn complete.

## How it works

opencode is a server with an event bus. The in-process lifecycle plugin (`mngr_opencode_plugin.ts`) already subscribes to it, so it now also:

- tracks the set of pending request ids from `permission.asked` (a tool blocked on approval) / `permission.replied` (answered), keeping a `permissions_waiting` marker present while any prompt is open. Tracking ids (rather than codex's single touch/remove flag) handles concurrent prompts from task-tool subagents.
- clears any stranded marker when the root session goes idle, as a safety net (codex uses the root `Stop`).

On the Python side, `OpenCodeAgent.get_lifecycle_state` promotes RUNNING -> WAITING while the marker is present (mirroring claude/codex), and `agent_field_generators` exposes `waiting_reason`, reading the marker files directly without SSH/tmux.

## Note on event names (sdk stubs vs. binary)

The `@opencode-ai/sdk` type stubs declare the events `permission.updated` / `permissionID`, but the **shipped opencode binary** (verified by inspecting the running 1.16.2 server) emits **`permission.asked`** with **`requestID`** on the reply. The binary is what actually flows through the plugin's `event` hook, so the parent branch's original `permission.asked` was correct about runtime. The plugin accepts **both** naming schemes defensively, since opencode self-upgrades. README and parity spec document the divergence.

## Tests

- Unit tests for the lifecycle promotion rule, the `get_lifecycle_state` override (live tmux), and the `waiting_reason` field generator (mirroring codex).
- Node-driven tests that replay the real runtime `permission.asked` / `permission.replied` / root-idle events through the actual `.ts` plugin and assert the marker is created/cleared, including the concurrent-prompt and stranded-marker cases, plus a case covering the sdk-stub aliases.

## Verification caveat

Event names and payload shapes were verified against the installed opencode 1.16.2 binary (not just the sdk type stubs) and the marker logic via node-driven tests over the real plugin code. A full end-to-end run against a live opencode agent triggering an actual approval prompt was not performed (matches the codex PR's scope; the shared release lifecycle test does not exercise permission prompts).


libs/mngr_opencode/imbue/mngr_opencode/resources/mngr_opencode_plugin.ts
```diff
@@ -294,20 +353,39 @@ export const MngrLifecyclePlugin: Plugin = async () => {
...
         }
         rebuildCommon()
         return
       }
 
+      // A tool is blocked on an approval prompt. The running opencode server
+      // (verified against the 1.16.2 binary) emits `permission.asked`; the
+      // `@opencode-ai/sdk` type stubs instead name it `permission.updated`. The two
+      // disagree, and opencode self-upgrades, so accept either -- both carry the
+      // request id in `properties.id`.
```
evgunter, line 374: sigh

joshalbrecht:
lgtm!

(not merged bc non-standard base)

PR #2148, opened 2026-06-14
Improve the discovery warning for unreachable VPS-backed hosts

evgunter:
## Problem

When a VPS-backed host is unreachable during discovery (e.g. the EC2 instance is up but `sshd` is wedged), its agents silently drop out of `mngr list`/`mngr conn`. The only signal was a stderr warning naming just the IP, so a transiently-unreachable host could be mistaken for a nonexistent agent.

This came out of a real recovery session where an AWS agent vanished from `mngr conn` because its host had OOM'd and `sshd` stopped answering.

## Changes

- The discovery warning in `VpsDockerProvider._read_records_from_vps` now names the **provider** (not just the IP) and states explicitly that **agents on the host are missing from listings** until it is reachable again.

That warning is logged at default verbosity (loguru console sink at `BUILD`/15; warnings are 30) on every command that runs discovery, including `mngr connect`, so the user already sees a clear, human-readable explanation right before any "agent not found" error.

## What this PR deliberately does *not* add

An earlier iteration also threaded an `unreachable_endpoints` property through `BaseProviderInstance` and appended a hint to the agent-not-found error in `find.py`. On review that was redundant with the always-logged warning above, so it was dropped to keep the change minimal. The net `libs/mngr` diff is therefore empty; only `mngr_vps_docker` changes.

## Tests

- `instance_discovery_test.py`: existing cache-fallback / not-ready / live-listing coverage continues to pass.

Unit tests for the touched files pass locally; `ty` clean.


evgunter:
not worth it

evgunter:
Reopening PR for continued work.

evgunter:
not actually useful

PR #2149, opened 2026-06-15
minds: destroy whole host reliably; fix silent-orphan teardown

joshalbrecht:
## Problem

Destroying a workspace from the minds app could report success in the UI while the underlying cloud instance kept running (and billing) — an invisible orphan. Observed live with an AWS workspace whose EC2 instance survived an in-app destroy.

## Root causes (all fixed here)

1. **Synchronous `mngr list` lookup, 10s-capped.** The destroy handler resolved the workspace's host id via `mngr list --include 'id == ...'` (an 8-region fan-out plus a per-host SSH `docker exec`). When that exceeded 10s it returned `None`.
2. **Unsafe single-agent fallback.** With no host id, destroy fell back to `mngr destroy <agent> -f`, which destroys only the workspace agent. Every minds host also runs a constant `system-services` agent, so the host (and its EC2/VPS) survived.
3. **Success decoupled from teardown.** The workspace was disassociated from the account synchronously on click, regardless of outcome, so a failed/partial destroy vanished from the UI. And "DONE" was derived from the workspace agent leaving discovery — never from the host actually being gone — so destroying one of two agents read as fully destroyed.

## Fix

- Resolve the host id from the **in-memory backend resolver** (it's immutable and already known for any workspace the user can see) — no `mngr list` subprocess. Refuse with a clear error (409) if the host can't be resolved, rather than partially destroying.
- Destroy **always fans out over the whole host** (workspace agent + `system-services`); the single-agent path is gone.
- A workspace stays **visible as destroying/failed** until its host is confirmed gone, and is disassociated only then — so a failed teardown is retryable instead of an invisible orphan.
- "DONE" now requires the **host** to be gone (workspace agent left discovery **and** host not still up), not just the agent.
- Trim the default AWS region set to the US datacenters (`us-east-1/2`, `us-west-1/2`); each region adds a provider `mngr list` queries every discovery cycle.

## Tests

- `destroying_test.py`: status machine is now host-based — FAILED when the wrapper exits but the host is still up (the exact bug); host id is recorded + read back; destroy command always fans out.
- `test_desktop_client.py`: landing finalize disassociates + deletes only when the host is gone (DONE); keeps + stays associated on FAILED.
- minds quick suite: 1352 passed; `ty` + `ruff` clean.

Follow-up (tracked separately by the author): orphan reconciler to catch any future "running host with no associated workspace".


joshalbrecht:
lgtm

PR #2152, opened 2026-06-15
fix deploy/lookup race for modal agents

evgunter:
## Problem

Acceptance tests creating Modal-backed agents intermittently failed with:

```
modal.exception.NotFoundError: Lookup failed for Function 'snapshot_and_shutdown' ...
```

This is a deploy-then-lookup eventual-consistency race: a freshly-deployed function may not yet be registered/propagated when we immediately look it up. The post-deploy lookup path (`DirectFunction.get_web_url`, which triggers Modal object hydration) had no retry, unlike the existing `deploy` (app-lock) and volume retries in `direct.py`.

## Fix

Add a retry-with-backoff on `modal.exception.NotFoundError` around `DirectFunction.get_web_url`, mirroring the existing `_VOLUME_RETRY` / `_DEPLOY_LOCK_RETRY` patterns (`_LOOKUP_RETRY`, 5 attempts, ~11s of headroom). `get_web_url` is a `@live_method` that lazily hydrates and leaves the object un-hydrated after a `NotFoundError`, so retrying re-attempts the network lookup. A genuinely-missing function still fails reasonably fast.

No tests are marked `@pytest.mark.flaky`: the root cause is fixed directly.

## Tests

- Added `test_get_web_url_retries_on_not_found_then_succeeds` and `test_get_web_url_does_not_retry_on_other_error` in `direct_test.py`.
- `just test-quick "libs/modal_proxy -m 'not acceptance and not release and not modal'"`: 101 passed.


joshalbrecht:
gross, but sure

lgtm

PR #2153, opened 2026-06-15
recommend aws/gcp/azure plugins when their CLI is detected

evgunter:
## Summary

The plugin install wizard now recommends and pre-selects the AWS, GCP, and Azure provider plugins when the corresponding CLI is detected on the user's system, mirroring how the `claude` and `modal` plugins are detected.

- Added `AwsSignalCheck` (`aws --version`), `GcloudSignalCheck` (`gcloud --version`), and `AzureSignalCheck` (`az --version`) to `plugin_catalog.py`.
- Marked the `aws`/`gcp`/`azure` catalog entries `is_recommended=True` and attached the new signals, so they appear in the wizard's recommended phase and are pre-checked when their CLI is present (an absent CLI just leaves them unchecked).
- `ovh` is left unchanged (no widely-installed standard CLI to detect against).

## Testing

- `just test-quick libs/mngr/imbue/mngr/plugin_catalog_test.py libs/mngr/imbue/mngr/cli/plugin_install_wizard_test.py` — 26 passed.
- Manually verified `check_signal` returns `True` for all three on a machine with the CLIs installed.


libs/mngr/imbue/mngr/plugin_catalog.py
```diff
@@ -207,6 +233,8 @@ class CatalogEntry(FrozenModel):
         package_name="imbue-mngr-azure",
         description="Azure Virtual Machines provider backend plugin for mngr",
         tier=PluginTier.INDEPENDENT,
+        signal=_AZURE_SIGNAL,
+        is_recommended=True,
```
evgunter, line 313: flagging (my thinking is these should be recommended even though they're experimental because we want people to know they exist)

joshalbrecht: sure, I think that's fine -- they should be working relatively well pretty soon

joshalbrecht:
lgtm!

joshalbrecht:
...but maybe we hold off on pressing merge until later in the week?

PR #2154, opened 2026-06-15
stop/start for gcp and azure

evgunter:
Brings the AWS stop/start (idle-pause + resume) lifecycle to the **GCP** and **Azure** providers: `mngr stop` halts live-instance compute billing (the disk and all state persist; storage still bills) and `mngr start` resumes the session faithfully. A stopped/deallocated host stays visible in `mngr list` and resolves by name (offline discovery), and idle agents self-stop. Design: `specs/gcp-azure-stop-start-lifecycle/spec.md`.

## Base branch

This PR targets the synthetic base `mngr/stop-gcp-azure-base` (= `mngr/azure` + `mngr/aws-stop` + `mngr/separate-snapshots`), so the diff is just the GCP/Azure stop/start work, not the merged provider branches.

## GCP

- `GcpVpsClient.stop_instance`/`start_instance` (poll status, return the fresh ephemeral IP) + instance-metadata helpers.
- `GcpProvider.stop_host`/`start_host`, label-based `_find_instance_for_host`, known_hosts rebind on the new IP.
- Idle self-stop with **no API call/IAM**: a guest poweroff lands a GCE VM in `TERMINATED` (stopped, disk preserved, no compute billing), so the in-container sentinel + host-side systemd watcher just powers it off.
- Offline discovery via instance **metadata** (GCE labels can't hold agent JSON).

## Azure

- `AzureVpsClient.deallocate_instance`/`start_instance` — **deallocate** (not OS shutdown, which keeps billing) halts compute billing; the static IP + host keys persist so resume needs no rebind. Tag add/remove via server-side `TagsOperations`; power state surfaced in the VM list via `expand=instanceView`.
- Idle self-deallocate via a **system-assigned managed identity**: the in-VM watcher calls the ARM `deallocate` API using its IMDS token. `mngr azure prepare` creates a least-privilege custom role and each VM gets a self-scoped role assignment. **Graceful fallback**: without the role-assignment privilege (Owner / User Access Admin), idle self-deallocate is disabled with a warning and `mngr stop`/`start` + a `shutdown -P now` circuit-breaker still work.
- Tag-based offline discovery (mirrors AWS). Adds the `azure-mgmt-authorization` dependency.

## Tests

Unit tests for both providers (client stop/start, offline-discovery reconstruction, idle-watcher units, role fallback). VM-level stop→start round-trips are covered by the existing per-provider release tests.

See `questions.local.md` (local, untracked) for the design decisions made while running unattended — I'll walk through them.


libs/mngr_azure/imbue/mngr_azure/backend.py
```diff
@@ -29,9 +56,112 @@
...
+        "Type=oneshot\n"
+        f"ExecStart={_DEALLOCATE_SCRIPT_PATH}\n"
+    )
+
+
+def _build_self_deallocate_script(sentinel_on_outer: str) -> str:
+    """Build the host-side self-deallocate script the idle ``.service`` runs.
+
+    Halts Azure compute billing from inside the guest: fetch the VM's
+    managed-identity token from IMDS (no az CLI needed -- plain curl), read this
```
evgunter, line 98: might be simpler to just install az, especially since it already uses azcopy elsewhere, but that would be a little cursed

libs/mngr_azure/imbue/mngr_azure/backend.py
```diff
@@ -203,6 +333,18 @@ def _list_provider_vps_hostnames(self) -> list[str]:
...
+        its IP (unlike a stopped GCE/EC2 instance, which loses its ephemeral IP and
+        is thus naturally excluded by the ``if main_ip`` check below). We
+        deliberately do NOT special-case deallocated VMs out here: the shared
+        discovery probe applies a bounded SSH connect timeout (pyinfra's
+        ``CONNECT_TIMEOUT``, 10s), so an unreachable VM fails fast and is surfaced
+        offline -- the same path a crashed-but-still-"running" host takes. The
+        deallocated host is then reconstructed from tags in
+        ``discover_hosts_and_agents``. This keeps discovery uniform (no
+        power-state-specific branch) at the cost of one bounded timeout when a
+        paused VM is present.
```
evgunter, line 331: flagging

libs/mngr_azure/imbue/mngr_azure/backend.py
```diff
@@ -212,6 +361,470 @@ def _list_provider_vps_hostnames(self) -> list[str]:
...
+            return json.dumps(labels, separators=(",", ":")) if labels else None
+        value = agent_data.get(field)
+        return None if value is None else str(value)
+
+    def _agent_field_tags(
+        self, agent_id: str, agent_data: Mapping[str, object], instance: Mapping[str, Any]
+    ) -> tuple[dict[str, str], list[str]]:
+        """Compute the ``mngr-agent-<id>-<field>`` tags to set, and stale ones to delete.
+
+        ``persist_agent_data`` is an upsert sometimes called with a partial record,
```
evgunter, line 484: this stuff is cleaned up in https://github.com/imbue-ai/mngr/pull/2159

libs/mngr_azure/imbue/mngr_azure/cli.py
```diff
@@ -297,6 +297,12 @@ def prepare(ctx: click.Context, **_kwargs: Any) -> None:
     # specific type. (allowed_ssh_cidrs is fail-open, so an empty list no longer
     # raises here -- it creates a no-ingress NSG with a warning.)
     result = client.ensure_network()
+    # Best-effort: create the least-privilege custom role that lets each VM's
+    # managed identity deallocate itself on idle (true cost parity). Returns None
+    # (after a clear warning) when the operator lacks roleDefinitions/write -- idle
+    # self-deallocate is then disabled but `mngr stop`/`start` still work, so this
+    # never fails prepare.
+    client.ensure_self_deallocate_role()
```
evgunter, line 305: flagging, but i think the mngr/volumes branch may have reworked this

libs/mngr_azure/imbue/mngr_azure/backend.py
joshalbrecht: seems weird to force you to install az

are you saying that the user must install azcopy? if so, can we remove that dependency? is it already part of the python package or something?

libs/mngr_azure/imbue/mngr_azure/backend.py
joshalbrecht: I don't really understand this / is hard to reason about how it will change in the volume based (non tag based) follow up

libs/mngr_azure/imbue/mngr_azure/backend.py
joshalbrecht: thanks!  is that one read for me to review?

libs/mngr_azure/imbue/mngr_azure/backend.py
joshalbrecht: I don't really understand this...

libs/mngr_azure/imbue/mngr_azure/client.py
```diff
@@ -811,6 +852,202 @@ def destroy_instance(self, instance_id: VpsInstanceId) -> None:
             return
         logger.info("Deleted Azure VM {} (NIC, public IP and OS disk cascade via delete_option)", instance_id)
 
+    def _await_lro(self, poller: LROPoller[None], timeout_seconds: float, description: str) -> None:
```
joshalbrecht: nit, don't use acronyms

libs/mngr_gcp/imbue/mngr_gcp/backend.py
```diff
@@ -38,6 +54,24 @@
...
 GCP_BACKEND_NAME: Final[ProviderBackendName] = ProviderBackendName("gcp")
 
+# Per-agent metadata is mirrored into instance metadata as up to three items per
+# agent, keyed ``mngr-agent-<agent_id>-<field>`` (the agent id lives in the key),
+# so a STOPPED instance (no external IP, SSH unreachable) still surfaces its
+# agents in discovery and resolves by name. GCE *labels* can't hold this (they
+# lowercase and restrict to ``[a-z0-9_-]``, 63 chars), so unlike AWS (EC2 tags)
+# the GCP mirror lives in instance *metadata*, whose values are large/permissive.
+AGENT_METADATA_PREFIX: Final[str] = "mngr-agent-"
+_AGENT_METADATA_FIELDS: Final[tuple[str, ...]] = ("name", "type", "labels")
```
joshalbrecht, line 64: I guess / hope this is changing in the later one?

libs/mngr_gcp/imbue/mngr_gcp/client.py
```diff
@@ -642,10 +755,76 @@ def list_instances(self, provider_tag: str | None = None) -> list[dict[str, Any]
...
+        concurrent metadata write -- rare, but possible when two agents persist
+        at once). No-op when there is nothing to change. Used to mirror
+        offline-discoverable metadata (host name, per-agent records) onto the
+        instance so a stopped VM still lists its agents and resolves by name.
+        AWS does the equivalent with EC2 tags; GCE labels are too restricted
+        (lowercase ``[a-z0-9_-]``, 63 chars) to hold arbitrary agent data.
+        """
+        if not updates and not delete_keys:
+            return
+        for attempt in range(2):
```
joshalbrecht, line 790: my only real complaint with this PR is the tags stuff

excited to check out the next PR that adjusts that!

joshalbrecht:
roughly lgtm except above comments -- my main one is definitely the tags stuff (vs volumes), so, excited for that PR

libs/mngr_azure/imbue/mngr_azure/backend.py
evgunter: this is on the remote host. aws, gcp, and azure are all sdk-only on the user's machine (except one optional gcloud call so that we can use gcloud defaults instead of having to explicitly ask the user; if gcloud is absent then it just asks the user)

libs/mngr_azure/imbue/mngr_azure/backend.py
evgunter: i was flagging more the IP keeping behavior. the volume based followup will change the tag reconstruction stuff

libs/mngr_azure/imbue/mngr_azure/backend.py
evgunter: it's really boring so i'm taking a long time to review it myself... so "you can if you want to but it's not up to my usual standards"

evgunter: (will post it when i am done reviewing it)

libs/mngr_gcp/imbue/mngr_gcp/backend.py
evgunter: which is changing?

libs/mngr_azure/imbue/mngr_azure/client.py
evgunter: `LROPoller` is azure's type name, but will switch to spelling it out elsewhere

PR #2155, opened 2026-06-15
Speed up `mngr message` in Minds.

hynek-urban:
As a user, I don't want to wait several seconds after clicking "Approve" or "Deny" on the permission request page.

---

After clicking one of the buttons, the underlying `mngr message` call took roughly 3.5s, resulting in a long delay in the UI. Of that:

- 1.5s was mngr's startup only (imports)
    - To make this part of the delay disapper, this PR introduces `MngrCaller`, a class that re-uses a pre-warmed forkserver instance where the imports are already resolved.
    - For now, only `mngr message` calls are routed through it. After this is merged, we can gradually migrate the other calls there, too.
    - (I also thought about simply using `mngr`'s Python API directly in-process but it seems the library is not ready for that because it modifies global state in various ways, e.g. when configuring the logging setup.)
- The remaining 2s is mngr's inherent runtime slowness
  - Apparently, as part of `mngr message`, the logic makes 11 (eleven) serial SSH calls. That's what takes most of those 2s.
  - 0.2s is also taken by parsing the events.jsonl file in search for a full snapshot (`find_latest_full_snapshot_offset`). Claude says that traversing the file back to front - as opposed to front to back - could be faster but I haven't verified this claim yet.
  - We get rid of this delay by simply running this in a background thread and not waiting for the result - it doesn't influence the permission dialog UI anyway.

The pre-warmed forkserver instance adds roughly 175MB memory to apps/minds. On the other hand, previously, each time we shelled out to `mngr`, those 175 MB would have also been consumed (and unlike now, in case of concurrent `mngr` calls, the memory would have been consumed multiple times).


apps/minds/imbue/minds/desktop_client/latchkey/handlers/messaging.py
joshalbrecht: soooo many "logger.warning" lines in this code

could you help us move to a more sensible error reporting world?

gabe is technically considering working on this I think but maybe you can coordinate with him to see if he's done anything yet.  And either way, it'd probably benefit from your experience + the things we learned about how to do error reporting well from when we did this in sculptor :)

hynek-urban: Do you mean adhering to the "do not log warnings" principle? That something is either an error (and then it should raise and be logged as such) or business as usual (and then it should be logged as such, too)?

joshalbrecht:
I spent a while looking at the core of this, and it actually looks pretty good!

lgtm except for the above comment, which I'll discuss on slack

joshalbrecht:
(I'm now also waiting to ship this until weishi's new installer is built, since it already has enough changes that I don't want to also roll this in there. This feels like it could have weird interactions via an installer, on osx, etc)

joshalbrecht:
ok, it works now! Adding so we can try it out

PR #2156, opened 2026-06-15
Improve the wording of selected system messages.

hynek-urban:
The messages sent to agents after approving or denying permission requests were needlessly referring to "blocked calls". In some situations, that was confusing to the agent (since there were situations when a permission request could have been sent without a call being blocked first).

The retry / don't retry follow-up should be clear to the agents from context (and I tested and validated this with opus-4-8 based agents).


joshalbrecht:
lgtm

PR #2157, opened 2026-06-15
clarify claude native plugin check; centralize claude presence check

evgunter:
## Summary

`_print_extras_status` (via `_claude_plugin_status`) checked for the Claude Code CLI with an inline `shutil.which("claude") is not None`, diverging from the canonical `CLAUDE.is_available()` system-dependency check in `deps.py`. This centralizes both that call site and the `is_claude_installed` test helper onto `CLAUDE.is_available()`, so the binary name and lookup logic live in one place.

Behavior is preserved today (`is_available()` is the same `shutil.which` call), but a future change to the canonical check now propagates everywhere.

## Changes

- `libs/mngr/imbue/mngr/cli/extras.py` — `_claude_plugin_status` uses `CLAUDE.is_available()`; dropped now-unused `import shutil`, added `CLAUDE` import.
- `libs/mngr/imbue/mngr/utils/testing.py` — `is_claude_installed()` defers to `CLAUDE.is_available()`.
- Added changelog entry.

## Testing

- `just test-quick libs/mngr/imbue/mngr/cli/extras_test.py` — 48 passed.
- `ty check` passes on the changed files.


joshalbrecht:
lgtm

PR #2158, opened 2026-06-15
don't mark timeouts as flaky

evgunter:
## Summary

Folds a short note into the existing flaky-test bullet in `CLAUDE.md`: if a test fails because of a timeout, do not mark it `@pytest.mark.flaky`; bump the timeout instead.

Adds a `dev/` changelog entry for the root-level file change.


CLAUDE.md
```diff
@@ -59,7 +59,7 @@ Only after doing all of the above should you begin writing code.
 - Use the shared fixtures (`temp_host_dir`, `temp_mngr_ctx`, `local_provider`, etc.) instead of creating your own.
 - Do NOT create tests for test utilities (e.g. never create `testing_test.py`). Code in `testing.py` and `conftest.py` is exercised by the tests that use it and does not need its own test file.
 - Do NOT create tests that code raises NotImplementedError.
-- If you see a flaky test, YOU MUST HIGHLIGHT THIS IN YOUR RESPONSE. Mark it with `@pytest.mark.flaky` so offload retries it automatically. Then try to fix the underlying flakiness in a separate commit.
+- If you see a flaky test, YOU MUST HIGHLIGHT THIS IN YOUR RESPONSE. Mark it with `@pytest.mark.flaky` so offload retries it automatically. Then try to fix the underlying flakiness in a separate commit. (If it was a timeout, do not mark it flaky; instead bump the timeout.)
```
joshalbrecht: "(If it was a timeout, do not mark it flaky; instead bump the timeout.)"
->
First investigate *why* the test is flaky. If it can be made more robust, do so. If the test is correct but it fundamentally requires additional time, increase the timeout for that individual test (but do NOT allow unreasonably long timeouts--in such a case, prefer to leave the test marked as flaky instead, eg, if this happened because of a weird infrastructure level fluke that sometimes makes an operation take *much* longer)

joshalbrecht:
lgtm after above

PR #2159, opened 2026-06-15
aws/azure buckets to store state accessible when the host is offline

evgunter:
not super happy with this but it does the thing i guess

---

## Provider state buckets (S3 / Azure Blob) for offline-readable mngr state

Gives the AWS and Azure providers a cloud object-storage bucket that holds mngr's control-plane state, so a stopped/offline instance's host record, agent metadata, and `host_dir` are readable without SSH and without the 256-char EC2/VM tag-value limit. GCP is intentionally untouched as a *feature* (its per-instance metadata allowance is sufficient) but shares the refactored offline-provider base. Design doc: `specs/provider-state-bucket/spec.md`.

### Known limitation — offline `host_dir` capture

Offline `host_dir` is captured **operator-driven** at `mngr stop`: mngr (already SSH-connected and holding the bucket credentials) rsyncs `host_dir` off the box (skipping sockets/other special files natively) into a local temp dir and uploads it to the bucket with your own creds. A host that **idle-self-poweroffs or crashes is _not_ captured** — its offline `host_dir` then reflects its last `mngr stop` (or is empty if it was never stopped that way). The state *records* are unaffected (always operator-written). Set `is_offline_host_dir_enabled = false` to disable capture entirely.

### Permissions / IAM

- **Offline host state** (the records that make `mngr list` / `start` / `event` work on a stopped host) is written by the operator with its own credentials. AWS: standard S3 access. Azure splits the control plane from the data plane, so `mngr azure prepare` also grants the prepare-runner's own principal **Storage Blob Data Contributor** on the state account (needs `Microsoft.Authorization/roleAssignments/write`) — without it the operator's `DefaultAzureCredential` can't read/write the state blobs (it hit `AuthorizationPermissionMismatch`). Only the prepare-runner is granted.
- **Offline `host_dir` needs no instance identity** — it's operator-driven, so there is no IAM instance profile (AWS) / managed identity (Azure), no `iam:PassRole` / `roleAssignments` for host_dir, and no on-box sync daemon. (The Azure VM still gets a system-assigned identity, used by idle self-deallocate.)
- `prepare` **fails loudly** when its permissions are missing — the bucket is required infrastructure: AWS needs SG + S3/STS for bucket create; Azure needs network + storage-account create + the operator blob-data grant.

### The bucket is required (no tag-mirror fallback)

The previous EC2/VM tag mirror is removed. With no bucket, mngr raises an actionable `mngr <cloud> prepare` error — on the `mngr create` / `mngr label` write path *and* on offline reads — rather than silently degrading or dropping a stopped host. `mngr create` fails fast (before launching the instance) when the bucket is absent. Bucket reads/writes propagate storage errors instead of swallowing them. `cleanup` refuses to delete a non-empty bucket unless passed `--force`.

### Why instances still carry lightweight identity tags (even though the bucket is required)

Removing the tag *mirror* does **not** remove tags entirely. Each instance still carries a small fixed set of `mngr-*` identity tags/labels — `mngr-host-id`, the cloud `Name`/host-name tag, `mngr-provider`, and `managed-by=mngr`. These are not the old mirror: they hold only *identity* (a handful of constant-size values), never the host record or agent records (that content now lives in the bucket).

They are still needed because **the bucket and the tags answer opposite lookup directions.** The bucket is keyed by `host_id` — it answers *`host_id` → state*. But a cloud listing (`DescribeInstances` / Azure VM list) returns *instances*, and the one thing you cannot get from the bucket is *instance → which host is this*: the bucket is keyed by the very id you're trying to recover. That instance→host correlation is required by two things that must work without SSH — listing stopped hosts (so a powered-off host still shows up by name in `mngr list`) **and** leak/orphan detection (finding a still-billing VM whose record is missing). Both are inherently infrastructure-side reconciliations of *live instances ⇄ known hosts*, and `mngr-host-id` is the cheapest key for them — it rides along free in the same listing call and keeps the live instance, not the mirror, as ground truth for what exists. No bucket re-keying removes this need; it would only relocate the correlation off the instance and add a denormalized index to keep in sync. (The old full-record-in-tags mirror — with its 256-char chunking and `TagLimitExceededError` — is gone, and this PR also removes the now-orphaned `add_tags`/`remove_tags` VPS-client methods it relied on.)

### What's in here

- **`S3StateBucket` / `BlobStateBucket`** holding the full host record + per-agent records, created by `mngr aws|azure prepare` and removed by `cleanup`.
- **Required-bucket model** as above (raises, no fallback, fail-fast create).
- **Offline `host_dir`** (on by default via `is_offline_host_dir_enabled`), operator-driven.
- **Destroy of a stopped host** now terminates its cloud instance instead of leaking it. The base teardown is SSH-based, so destroying a stopped (deallocated/powered-off) host used to raise `HostNotFoundError` and leave the still-billing instance + its mirrored state behind (often reporting success). `OfflineCapableVpsDockerProvider.destroy_host` now dispatches on the instance's actual power state (cheap tag/label lookup, no SSH): a stopped instance is terminated via the same `destroy_instance` primitive the online path uses and its external state is deleted, failing loudly (non-zero exit) if termination could not be carried out and idempotent when the instance is already gone. Fixes the start-then-destroy workaround.
- **GCP**: its offline instance-metadata store now holds the *full* host record (was a lossy field subset); no bucket needed.
- **Cross-provider unification** into shared `mngr_vps_docker` abstractions: `HostStateStore` / generic `BucketHostStateStore` over a `StateBucket` protocol, the `HostDirBackend` select-once strategy, `state_keys` (object-key layout), and `BaseStateBucket` (cloud-agnostic marshalling).

### Validation

- `ty` clean across `mngr` / `mngr_vps_docker` / `mngr_aws` / `mngr_azure` / `mngr_gcp`; unit tests green across the provider packages; ratchets pass.
- The bucket-backed **record** path's real-cloud release lifecycle (create → stop → list-while-stopped → start → exec → destroy) was validated on AWS / Azure / GCP earlier in the branch. The operator-driven **`host_dir`** capture (rsync pull) was validated end-to-end on **real AWS and Azure**: `mngr stop --stop-host` exits 0 even with a named pipe in `host_dir`, the captured tree (incl. a marker file, excl. the special file) lands in the bucket, and `mngr file get/list` serve it back offline. The Azure run also confirmed `mngr azure prepare`'s operator Blob-data-role grant (the operator creds read/write the state blobs). Shared cloud-agnostic capture (`BucketHostDirBackend` + `download_directory_from_outer`); only the bucket SDK differs per cloud. Also unit-tested.
- Full `just test-offload` runs in CI on push.


libs/mngr_aws/imbue/mngr_aws/backend.py
```diff
@@ -193,6 +274,51 @@ class AwsProvider(VpsDockerProvider):
     aws_client: AwsVpsClient = Field(frozen=True, description="EC2 API client")
     aws_config: AwsProviderConfig = Field(frozen=True, description="AWS-specific configuration")
 
+    # Lazily-resolved S3 state bucket. ``False`` means "not yet resolved"; once
+    # resolved it holds either an ``S3StateBucket`` or ``None`` (no bucket
+    # configured / account id unresolvable -- the legacy tag-mirror fallback).
+    _state_bucket_cache: S3StateBucket | None | Literal[False] = PrivateAttr(default=False)
```
evgunter: weird

libs/mngr_aws/imbue/mngr_aws/backend.py
```diff
@@ -323,8 +449,44 @@ def _create_vps_instance(
...
+                    identity.identity_name,
+                )
+                return None
+        except S3StateHostIdentityError as e:
+            logger.warning(
+                "Could not check the bucket-write IAM identity {}; launching without it: {}",
+                identity.identity_name,
+                e,
+            )
+            return None
```
evgunter: is this the behavior we want

libs/mngr_aws/imbue/mngr_aws/cli.py
```diff
@@ -54,8 +59,22 @@ class _AwsOperatorCliOptions(CommonCliOptions):
...
+# whether/how to provision the bucket-write IAM identity that lets an instance
+# push its host_dir to the bucket for offline reads.
+_HOST_DIR_IDENTITY_AUTO: Final[str] = "auto"
+_HOST_DIR_IDENTITY_REQUIRE: Final[str] = "require"
+_HOST_DIR_IDENTITY_SKIP: Final[str] = "skip"
+_HOST_DIR_IDENTITY_CHOICES: Final[tuple[str, ...]] = (
+    _HOST_DIR_IDENTITY_AUTO,
+    _HOST_DIR_IDENTITY_REQUIRE,
+    _HOST_DIR_IDENTITY_SKIP,
+)
```
evgunter: use enum

libs/mngr/imbue/mngr/cli/common_opts.py
```diff
@@ -128,6 +131,27 @@ def add_common_options(command: TDecorated) -> TDecorated:
...
+        "--use-offline-host-dir",
+        "use_offline_host_dir",
+        type=click.Choice(auto_toggle_choices()),
+        default=AutoToggle.AUTO.value.lower(),
+        show_default=True,
+        help=(
+            "Enable offline host_dir -- reading a stopped host's files via `mngr file` / "
+            "`mngr event` / `mngr transcript` while the host is powered off. " + AUTO_TOGGLE_HELP_SUFFIX
+        ),
+    )(command)
```
evgunter: clever, but probably we want to do this in a more normal way

libs/mngr/imbue/mngr/primitives.py
```diff
@@ -238,6 +238,33 @@ class PluginKind(UpperCaseStrEnum):
...
+
+    YES requires the capability and fails if it can't be enabled. NO disables
+    it. AUTO (the usual default) enables it when possible but degrades to a
+    warning instead of failing -- the right default for a capability that needs
+    privileges the operator may not have.
+    """
+
+    YES = auto()
+    AUTO = auto()
+    NO = auto()
```
evgunter: this is currently only used in one place but is the same shape as the always / auto / never pattern we have in a few places

libs/mngr_aws/imbue/mngr_aws/backend.py
```diff
@@ -281,81 +441,111 @@ def _idle_watcher_service_unit(self, sentinel_on_outer: str) -> str:
...
+        if bucket is None:
+            logger.debug("No S3 state bucket; skipping host_dir sync install for host {}", host_id)
+            return
+        host_dir_on_outer = self._host_dir_path_on_outer(host_id)
+        sync_target_uri = host_dir_sync_target_for(bucket.bucket_name, host_id)
+        service_unit = _build_host_dir_sync_service_unit(str(host_dir_on_outer), sync_target_uri)
+        timer_unit = _build_host_dir_sync_timer_unit(HOST_DIR_SYNC_INTERVAL_SECONDS)
+        with log_span("Installing AWS host_dir sync daemon"):
+            with self._make_outer_for_vps_ip(vps_ip) as outer:
+                outer.execute_idempotent_command(_build_awscli_install_command(), timeout_seconds=300.0)
```
evgunter: kind of weird to have to install awscli, but i guess it makes sense

libs/mngr_aws/imbue/mngr_aws/backend.py
```diff
@@ -367,147 +557,145 @@ def _is_instance_offline(self, instance: Mapping[str, Any]) -> bool:
...
+        host_id = host.id if isinstance(host, HostInterface) else host
+        try:
+            if not bucket.host_dir_prefix_has_objects(host_id):
+                self._warn_if_host_dir_identity_missing(host_id)
+                return None
+        except S3StateBucketError as e:
+            logger.warning(
+                "Could not probe host_dir prefix for host {}; treating volume as unavailable: {}", host_id, e
+            )
+            return None
```
evgunter: all the optional bucket fallback stuff in this branch is gross but i'm not sure what the clean solution is

evgunter:
[CLAUDE] Overnight cleanup summary (for your review)

While you were away I made this PR as clean as I could, then validated it against unit + real-cloud release tests. Here's everything notable, including judgment calls and a few flags.

## What I did (the arc)
1. **Verified + pushed the unified `_host_dir_backend` refactor** (the work in flight when you went to bed): `HostDirBackend`/`NullHostDirBackend` select-once strategy, sibling of `_state_store`; no provider call site re-tests `_state_bucket is None` / `is_offline_host_dir_enabled` outside the two cached-property selectors.
2. **Ran 2 rounds of design-simplification analysis** (5 read-only agents in round 1, 3 in round 2) and executed the worthwhile findings across 3 implementation forks + my own review fixes. Round 2's full-diff scan explicitly confirmed the branch is **converged** — no round 3 needed.
3. **Release-tested all three providers** on real cloud (create → stop → offline-read → start → destroy).
4. **Ran autofix** (focused on the volumes work since the last autofix point) — came back **clean, zero issues**.

## Simplifications landed
- **`BaseStateBucket`** extracted into `mngr_vps_docker` — the cloud-agnostic record marshalling + key layout now lives once; `S3StateBucket`/`BlobStateBucket` keep only the SDK client, raw object primitives, error translation, and bucket lifecycle (~160 lines of duplication removed).
- **Read-path fold**: tag fallback folded into `BucketHostStateStore.read_host_record`; deleted the `read_host_record_with_tag_fallback` wart. The `to_offline_host` bypass of `OfflineCapable` is load-bearing and kept.
- **host_dir lifted to `OfflineCapableVpsDockerProvider`** (`get_volume_for_host`/`get_volume_reference_for_host`/`_sync_host_dir_before_pause`/`_host_dir_backend`) — it's an offline-capability concern, not tag-mirror-specific. GCP inherits the no-op `NullHostDirBackend` default (behavior-preserving, None→None).
- **Partial `BucketHostDirBackend`** base sharing `volume`/`volume_reference`/`trigger_final_sync` (install_sync left per-provider — genuinely cloud-specific).
- **`emit_operator_result`** shared CLI dispatch (AWS/Azure/GCP prepare/cleanup); parametrized systemd unit builders; `_agent_dicts_from_tags` tag-listing dedup; `managed-by` tag via shared `state_keys.MANAGED_BY_TAG_*` (AWS); dropped dead code in AWS cli; moto fixtures moved to conftest; tightened a few weak test assertions; new `host_state_store_test` matrix.

(All of these also came out of your earlier review comments: `--purge-state`→`--force`, the `is_offline_host_dir_enabled` config field replacing the `--use-offline-host-dir` flag, the actionable `TagLimitExceededError`, the de-jargoned config/README wording, and the self-explanatory "Decision N" comments.)

## Judgment calls / deviations worth knowing
- **I caught and fixed a subtle behavior regression** one of the forks introduced: its read-path fold only fell back to tags when the bucket record was *absent*, silently dropping the prior fallback on a *malformed* or *errored* bucket record. I restored full parity (fallback on absent/error/malformed) and added the missing `host_state_store_test` matrix that pins it. Commit: "preserve tag fallback on malformed/errored bucket record".
- **Deferred (judged not worth the churn/risk now)**: (a) consolidating the AWS/Azure provider-building **test factories** into shared `testing.py` helpers — 3 separate forks independently judged it high-churn test-infra with no production benefit; (b) the `install_sync`/`trigger_final_sync` install *sequence* sharing (break-even on lines vs. hook ceremony, on a release-critical path). Both are clean follow-ups if you want them.
- **Left alone deliberately**: Azure's pre-existing `AZURE_MANAGED_BY_TAG_*` constants (they're on `main`, used by unrelated RG-management code; rewiring into `state_keys` would be a layering oddity); the AWS cleanup double-call of `_refuse_cleanup_if_instances_exist` (intentional ordering — the outer refuse must run before bucket/identity teardown).

## Flags
- **History is messy**: there are several `WIP: Fork ... mid-step snapshot` commits interleaved with the real ones. These are me satisfying the stop-hook's clean-tree gate while background forks were mid-step; content is captured correctly and the final tree is what matters. I did **not** squash them (repo rule: never rebase). Let me know if you'd like them cleaned up some other way.
- **One unrelated Azure VM**: a leak scan after the release tests found `mngr-outgoing-mellow-yak-6c0e2f...` in RG `MNGR`, but it's from **2026-06-11** (6 days old) with normal mngr agent-name styling — NOT from my release tests (those self-cleaned; all three exited 0 / leak-guard passed). I left it untouched since it isn't mine; flagging in case it's a real leak you want to reap.
- The fail-loud `--use-offline-host-dir yes` affordance is gone (provisioning is now always best-effort) — that was the explicit tradeoff of the config-unification you approved. A thin `--require-host-dir-identity` flag could restore it later if wanted.

## Verification
- `ty` clean across mngr / mngr_vps_docker / mngr_aws / mngr_azure / mngr_gcp.
- **5323 focused tests pass**, 1 skipped, 0 failed; ratchets pass (no count changes).
- **Real-cloud lifecycle release tests pass on all three**: AWS (us-east-1, 3m45s), Azure (4m12s), GCP (us-central1-a, 4m16s).
- autofix: clean, no fixes needed.

Next: handing off to the `bare-providers` agent to merge in this branch and resolve conflicts (per your request).

libs/mngr/imbue/mngr/cli/output_helpers.py
```diff
@@ -144,6 +144,32 @@ def emit_event(
...
+    match output_format:
+        case OutputFormat.JSON:
+            write_json_line(data)
+        case OutputFormat.JSONL:
+            emit_event(event_name, data, OutputFormat.JSONL)
+        case OutputFormat.HUMAN:
+            for line in human_lines:
+                write_human_line(line)
+        case _ as unreachable:
+            assert_never(unreachable)
```
evgunter, line 211: why would you do it this way lol

libs/mngr_aws/imbue/mngr_aws/backend.py
evgunter: (this is on the remote host, not the user's machine)

joshalbrecht:
> GCP is intentionally untouched as a feature 

Ok, but is host data available offline for GCP? Or was that just claude being lazy?

joshalbrecht:
also -- I'm totally down to say "the buckets are not optional -- you must have a bucket, get rid of all of the tags weirdness"

that would probably simplify some things?

evgunter:
it would simplify! i'm happy to do that if the additional permissions etc are acceptable. gcp is untouched because their metadata store is good enough that an additional bucket is not necessary for offline host data

libs/mngr_aws/imbue/mngr_aws/backend.py
```diff
@@ -48,29 +55,78 @@
...
-# agent, keyed ``mngr-agent-<agent_id>-<field>`` (the agent id lives in the key,
-# not a value), so a *stopped* instance (no public IP, SSH unreachable) still
-# surfaces its agents in discovery and resolves by name. ``name``/``type`` are
-# stored raw; ``labels`` as compact JSON. One tag per field (rather than one
-# packed tag per agent) gives ``labels`` the full 256-char value budget; a field
-# whose value still overflows is dropped, not failed.
-AGENT_TAG_PREFIX: Final[str] = "mngr-agent-"
-_AGENT_TAG_FIELDS: Final[tuple[str, ...]] = ("name", "type", "labels")
-_MAX_TAG_VALUE_LEN: Final[int] = 256
 # EC2 allows 50 (non-``aws:``) tags per resource. When a host has so many agents
```
joshalbrecht: is this comment just wrong? why does "number of agents" have anything to do with tags?

libs/mngr_aws/imbue/mngr_aws/backend.py
```diff
@@ -105,14 +161,87 @@ class ParsedAwsBuildOptions(ParsedVpsBuildOptions):
...
+            if not bucket.bucket_exists():
+                logger.debug(
+                    "S3 state bucket {} does not exist; using the EC2 tag mirror "
+                    "(run `mngr aws prepare` to create it)",
+                    bucket.bucket_name,
+                )
+                return None
+        except S3StateBucketError as e:
+            logger.warning("Could not check S3 state bucket {}; falling back to EC2 tags: {}", bucket.bucket_name, e)
+            return None
```
joshalbrecht: hella skeptical

libs/mngr_aws/imbue/mngr_aws/backend.py
```diff
@@ -281,81 +422,57 @@ def _idle_watcher_service_unit(self, sentinel_on_outer: str) -> str:
...
+        offline host_dir readability, so it is logged rather than failing create_host.
+        """
+        super()._on_host_finalized(host_id=host_id, vps_ip=vps_ip)
+        try:
+            self._host_dir_backend.install_sync(host_id=host_id, vps_ip=vps_ip)
+        except MngrError as e:
+            logger.warning(
+                "AWS host_dir sync install failed for host {} ({}); the stopped host's host_dir "
+                "will not be readable offline",
+                host_id,
```
joshalbrecht: warnings after exceptions always trigger me

libs/mngr_aws/imbue/mngr_aws/backend.py
```diff
@@ -367,147 +484,121 @@ def _is_instance_offline(self, instance: Mapping[str, Any]) -> bool:
...
-    def _offline_discovered_host_from_instance(self, instance: Mapping[str, Any]) -> DiscoveredHost | None:
-        """Build a STOPPED-state DiscoveredHost from an instance's tags, or None if not a mngr host."""
-        tags = self._tag_dict_from_normalized(instance)
-        host_id_str = tags.get("mngr-host-id")
-        if host_id_str is None:
+                return None
+        except S3StateHostIdentityError as e:
+            logger.warning(
+                "Could not check the bucket-write IAM identity {}; launching without it: {}",
+                identity.identity_name,
```
joshalbrecht: skeptical

IIRC we do all of this stuff roughly correctly in modal -- might want the agents to take a look and see if they can learn from that one and fix some of these edge cases?

libs/mngr_aws/imbue/mngr_aws/backend.py
```diff
@@ -367,147 +484,121 @@ def _is_instance_offline(self, instance: Mapping[str, Any]) -> bool:
...
+            profile_arn = self.provider.aws_client.get_instance_iam_profile_arn(VpsInstanceId(instance["id"]))
+        except MngrError as e:
+            logger.debug("Could not check IAM profile for host {} while diagnosing empty host_dir: {}", host_id, e)
+            return
+        if profile_arn is None:
+            logger.warning(
+                "Host {}'s instance has no attached IAM instance profile, so its host_dir was never "
+                "pushed to the bucket and is not readable offline. Re-run `mngr aws prepare` "
+                "with sufficient IAM, then recreate the host so it picks up the profile.",
+                host_id,
```
joshalbrecht: is it useful for me to keep commenting on these? :-P

evgunter: heh i commented on one of these at the top with like "all of these are gross, not sure what to do about them" and then piled on so many commits after it that it got buried

libs/mngr_aws/imbue/mngr_aws/backend.py
evgunter: multiple agents on one host

libs/mngr_aws/imbue/mngr_aws/backend.py
```diff
@@ -113,6 +164,77 @@ class AwsProvider(OfflineCapableVpsDockerProvider):
...
+
+        The bucket is the sole source of truth for agent records and the offline
+        host record. None means the bucket does not exist yet (no name
+        configured/derivable, or one whose name resolves but ``mngr aws prepare``
+        was never run); ``_state_store`` then raises an actionable error. A storage
+        error while probing existence propagates rather than masquerading as
+        "absent". The existence probe runs at most once per provider lifetime
+        (cached).
+        """
+        return self._resolve_existing_state_bucket()
```
evgunter, line 113: claude really wants this to be cached and optional instead of computed up front and required

joshalbrecht:
>  A host that idle-self-poweroffs or crashes is not captured — its offline host_dir then reflects its last mngr stop (or is empty if it was never stopped that way).

That doesn't seem great -- it might as well simply sync itself A) periodically, and B) when shutting down?

IIRC there's already something in the modal shutdown script that calls a "sync" command, which in these cases could be implemented with something like "aws s3 sync host_dir s3://wherever/" ?

joshalbrecht:
> GCP is intentionally untouched as a feature (its per-instance metadata allowance is sufficient) 

is the offline access to the host dir supported from here though? (my guess is no, but idk why that wouldn't also make sense on GCP)

evgunter:
my first attempt at this was having it sync itself periodically, but this requires the ability to give it its own iam role, which was not a permission i had on my account and so i figured it wouldn't be practical for users. (presumably there's some way to get around this, but i just reverted to having the user's machine do all the sync operations)

evgunter:
it is supported on gcp; gcp was the only one which worked properly before this PR, because its per-instance metadata store is big enough to store all the information we need without a bucket

joshalbrecht:
honestly, overall this looks pretty good!

re: GCP: hmm -- so it even has offline host data without syncing?

re: periodic syncing: I was originally thinking something simpler -- allowing the user to inject their credentials, and have a sync script running inside of the host (much like the idle-detection script), but that actually seems too insecure

doing it via the IAM role is probably the right way, but you're right that that will require higher permissions. I can make a follow up to do that (I have sufficient permissions for it)

still curious how the GCP thing would work without this!

evgunter:
won't our users probably have permissions more like mine in general?

evgunter:
sigh, my claude is now saying that the gcp one doesn't support host dir. and now i have questions for my release test runner claude which said everything was working...

will do in a followup PR

PR #2162, opened 2026-06-15
agent capability taxonomy + auto-install

evgunter:
this is a little silly in some ways, but i think having a granular organization like this will make it a lot easier to implement new agent types without accidentally missing important functionality, while also not imposing the burden of implementing everything up front (and hence encouraging placeholders)

---

## What

Makes each agent-type plugin's **capabilities** a fact the code carries, replacing the hand-maintained parity matrix that drifts from the code.

- **Capability registry** (`libs/mngr/imbue/mngr/agents/agent_capabilities.py`): one `AGENT_CAPABILITIES` list; presence detected via class-mixin (`issubclass`), `waiting_reason` field generator, plugin hookimpl, or sibling `mngr_<harness>_usage` plugin. A generated matrix (`libs/mngr/docs/concepts/agent_capabilities.md`) + drift-guard test replace the hand table (`just regenerate-agent-capabilities-doc`).
- **Contract-bearing mixins** (`interfaces/agent.py`): `HasStreamingSnapshotMixin`, `HasSessionPreservationMixin`, `HasUnattendedModeMixin`, `HasPermissionPolicyMixin`, `HasVersionManagementMixin`, `HasAutoInstallMixin` — each a small ABC whose abstract method the plugins implement; existing behavior is routed through it, so presence == implementation. The four existing transcript/headless mixins fold in.
- **Auto-install as a base capability**: shared `ensure_cli_installed` (`agents/installation.py`) called from each plugin's `provision()`, gated by consent locally / `is_remote_agent_installation_allowed` remotely. agy/opencode/codex gain install they lacked; claude keeps its version-pin flow. **Verified live on real Modal hosts** (which ship without the CLIs): opencode and agy both installed successfully during provision.
- pi implements `HasUnattendedModeMixin` degenerately (`auto_allow_permissions` pinned True, errors if set False) + a single-value `waiting_reason` generator — so the matrix has **no `n/a` cells** (every cell is Y/-, fully code-derived).
- **The matrix was corrected against the code** throughout (the drift this kills): per-resource policy = agy/opencode/codex (not claude — blanket hooks only); field generators = claude/codex/opencode (+ pi's degenerate); install split into auto-install vs version-management; the `code-guardian`/`fixme-fairy` skill variants are excluded (they reuse `ClaudeAgent` wholesale); `headless_claude` kept (genuinely distinct logic).

## Reviewed

Architecture-reviewed (approach validated; 4 findings addressed) and autofix-clean. Verified: ty across all touched files; affected unit/integration suites green; all ratchets pass.

## Notes for reviewers

- **agy/codex/opencode are NOT in the CI test images.** The new install-check (`command -v <binary>`) would fail provision *setup* tests (which run on a real local host without the binary), so those tests set `check_installation=False` — they test setup, not the binary. Install behavior is covered by `installation_test.py` + per-plugin install-path tests. If we later want the test images to carry these CLIs (so real `mngr create` / acceptance e2e exercise the install path), that's a deliberate image change to flag separately.
- **Deferred to a follow-up PR:** the registry-driven **release** harness that walks each agent's declared capabilities and runs a real `exercise_fn` against a live agent (the behavioral check detection can't give). The low-value prose-pointer coverage test was deliberately dropped, not shipped.
- **codex e2e auth (future):** when codex acceptance/e2e tests run the real agent, codex's custom `model_provider`/base-URL can route through the existing litellm gateway to claude/anthropic, avoiding a new OpenAI key. Part of the follow-up harness.

## Base

Targets `mngr/agent-mixins-base` — a purpose-built synthetic base = `main` + the agent-relevant feature branches from `ev/main` that are not yet merged to `main` (`agents-usage`, `agy-preserve`), excluding cloud/dev branches. (`waiting-reason`, `opencode-waiting-reason`, and `catalyst-waiting` have since merged to `main`.) See `dev/agent-mixins-synthetic-base.md` on the base branch.


libs/mngr/imbue/mngr/agents/default_plugins/command_agent.py
```diff
@@ -4,15 +4,27 @@
...
+class CommandAgent(BaseAgent[AgentTypeConfig], HasUnattendedModeMixin):
+    """Agent type that runs an arbitrary shell command (see ``register_agent_type``).
+
+    A bare command runner: it does not inherit ``CliBackedAgentMixin``, so the CLI-oriented
+    capabilities render ``n/a`` for it. It runs unattended by construction -- a plain command
+    has no in-run tool prompt to approve -- so ``is_unattended_enabled`` is always True.
+    """
+
+    def is_unattended_enabled(self) -> bool:
+        return True
```
evgunter: a little silly, but i think it's convenient given the taxonomy in general

libs/mngr/imbue/mngr/interfaces/agent.py
```diff
@@ -535,6 +535,17 @@ def get_common_transcript_scripts(self) -> Mapping[str, str]:
...
 
+class SupportsLiveOutputMixin:
+    """Marker for agents that publish a live, in-progress view of their output before a turn completes.
+
+    A bare marker with no contract of its own -- the concrete surface differs by agent kind:
+    a TUI agent exposes a streaming snapshot file (``HasStreamingSnapshotMixin``), while a
+    headless agent yields incremental stdout chunks (``StreamingHeadlessAgentMixin``). Both
+    inherit this so capability detection can treat "can stream live output" as one capability
+    regardless of how the agent surfaces it.
+    """
```
evgunter, line 542: i'm having an agent think about unifying this in a sensible way

libs/mngr/docs/concepts/agent_capabilities.md
```diff
@@ -0,0 +1,42 @@
+# Agent capabilities
```
evgunter, line 1: what's the right place for this file?

libs/mngr_antigravity/imbue/mngr_antigravity/plugin_test.py
```diff
@@ -159,11 +160,95 @@ def test_register_agent_aliases_maps_agy_to_antigravity() -> None:
...
 def _make_antigravity_agent(
     local_provider: LocalProviderInstance,
     tmp_path: Path,
     agent_config: AntigravityAgentConfig,
 ) -> AntigravityAgent:
+    # These setup tests run against a real local host where agy is not installed; the
+    # install check is irrelevant to provision setup (files/trust/config) and is covered
+    # separately, so skip it unless a caller opted in explicitly.
+    if "check_installation" not in agent_config.model_fields_set:
+        agent_config = agent_config.model_copy_update(to_update(agent_config.field_ref().check_installation, False))
```
evgunter, line 254: should probably just be in the test's config

libs/mngr/imbue/mngr/interfaces/agent.py
evgunter: https://github.com/imbue-ai/mngr/pull/2182

libs/mngr/imbue/mngr/agents/default_plugins/command_agent.py
joshalbrecht: this makes sense to me

libs/mngr/docs/concepts/agent_capabilities.md
joshalbrecht: that seems fine to me!

libs/mngr_antigravity/imbue/mngr_antigravity/plugin_test.py
joshalbrecht: agreed

joshalbrecht:
lgtm after tests pass, etc

fairly difficult to review this one super closely given the nature of the refactor, but it seems like a nice improvement

libs/mngr_antigravity/imbue/mngr_antigravity/plugin_test.py
evgunter: apparently nontrivial, FIXME'd it

PR #2164, opened 2026-06-15
minds: fix creating a mind from a remote git URL when no branch is specified

gnguralnick:
## Problem

Creating a mind from a remote git URL (e.g. a GitHub HTTPS URL) **without specifying a branch** failed during `mngr create`'s git-mirror push:

```
No refs in common and none specified; doing nothing.
Perhaps you should specify a branch.
fatal: the remote end hung up unexpectedly
```

## Root cause

`clone_git_repo` (`apps/minds/imbue/minds/desktop_client/agent_creator.py`) cloned via `git init` + `git fetch origin <ref>` + `git checkout --detach FETCH_HEAD`. The detached checkout is intentional: when a branch is given, the caller follows up with `checkout_branch` (`git checkout -B <name>`) to create a real local branch. When **no branch** is given, `checkout_branch` is never called, so the clone stayed on a detached HEAD with nothing under `refs/heads/*`. The downstream mirror push (`host.py::_git_push_to_target`) only pushes `+refs/heads/*:refs/heads/*` + `+refs/tags/*:refs/tags/*`, so it had nothing to push and failed.

## Fix

Split `clone_git_repo` into two explicit paths:

- **No branch** → a plain `git clone <url> <dir>`. Git resolves the remote's default branch natively (in one connection) and leaves a correctly-named local branch checked out — for any default (`main`, `master`, …). No `ls-remote` output parsing, no second round trip, and no dependency on HEAD being a symref into `refs/heads`.
- **Explicit branch/tag/SHA** → unchanged (`git init` + `git fetch origin <ref>` + detached checkout, then the caller's `checkout_branch`). `git fetch` is kept here because `git clone --branch` rejects commit SHAs.

The plain `git clone` does not populate `FETCH_HEAD`, which `checkout_branch` depends on. The local-worktree create call site now passes its branch through to `clone_git_repo` (mirroring the remote-URL call site), so the no-branch `git clone` path is never followed by `checkout_branch`; with no branch, the plain clone lands on the worktree's own branch and `checkout_branch` is skipped.

## Testing

- `test_clone_no_branch_lands_on_default_branch_and_is_mirror_pushable`: no-branch clone lands on a named branch and the mirror push succeeds.
- `test_clone_no_branch_uses_remotes_actual_default_branch_name`: a `master`-default repo clones onto a local `master` branch (guards the choice to resolve the real default rather than hardcode `main`).
- `test_worktree_overlay_preserves_uncommitted_edits` updated to pass `branch=` (mirrors the production call site).
- `agent_creator_test.py`: 50 passed. minds ratchets: 56 passed. `ty check`: clean.

## Follow-up (not in this PR)

The architecture review noted the no-branch vs explicit-branch asymmetry leaves an implicit `FETCH_HEAD` contract split between `clone_git_repo` and the caller's `if branch:` gate. A possible future refactor: have `clone_git_repo` always return on a named local branch (fold `checkout_branch`'s `-B` naming into the explicit-branch arm), retiring `checkout_branch` and the two-step. Larger than this bug warrants; left as a follow-up.


joshalbrecht:
lgtm

PR #2166, opened 2026-06-15
mngr_claude: skip response-streaming work while the agent is idle

gnguralnick:
## Summary

The Claude response-streaming watcher (`libs/mngr_claude/.../resources/stream_snapshot.py`, enabled by `streaming_snapshot_interval_seconds > 0`) does no work at all while the agent is idle: no transcript read, no tmux pane capture, no buffer write.

Previously the expensive pane capture was already gated on the `active` marker, but the poll loop still re-read the transcript and rewrote the stream buffer every interval while idle. Now `_run_one_poll` returns whether the agent was active, and `main()` threads that back in as `was_active` so the active→idle edge is detectable. On that edge we do exactly one clearing write -- empty the in-progress body and pin the id line to the last complete message, so a stale preview can't linger -- and every subsequent idle poll does nothing.

The rationale: streaming is a live preview of an actively-working agent. An idle agent has nothing to preview, and the durable transcript event already carries the finalized message, so the consumer (system interface / `mngr_robinhood`) loses nothing -- it already suppressed idle id-only changes and clears the provisional bubble on the finalized `assistant_message`.

## Changes

- `_run_one_poll` now returns `bool` (was active this poll) and short-circuits the idle path; the only idle-time work is the single active→idle clearing write.
- `main()` seeds `was_active = False` after the startup clear and threads the per-poll result forward.
- Two new unit tests: idle-from-idle does no work (buffer file never created); active→idle clears the body once and pins the id line.

## Testing

From the repo root:

```
uv run pytest libs/mngr_claude/imbue/mngr_claude/resources/stream_snapshot_test.py libs/mngr_claude/imbue/mngr_claude/test_ratchets.py
```

87 passed (includes the two new tests and the ratchets); `ruff format`/`check` and `ty` clean on the modified files.


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#4 from the same branch. This public repository is now a read-only mirror.

gnguralnick:
Continuing on imbue-ai/mngr-internal#120 — development has moved; this mirror is read-only, so PRs cannot merge here.

PR #2167, opened 2026-06-16
bump test_start_creation_imbue_cloud_ai_with_local_compute_mints_litellm_key timeout

evgunter:
## Summary

- Bumps the pytest-timeout cap on `test_start_creation_imbue_cloud_ai_with_local_compute_mints_litellm_key` from the implicit 10s to **30s wallclock**.
- Bumps the matching `_wait_until_finished` deadline from its default 10s to **20s**, leaving 10s of headroom under the outer cap so the helper's specific `AssertionError("creation <id> did not finish within 20.0s")` surfaces instead of racing pytest-timeout's generic `Failed: Timeout (>30.0s)` message.
- Same root cause as the file's already-`@pytest.mark.flaky` sibling `test_start_creation_api_key_ai_does_not_mint_litellm_key`: sync unit test, but setup spins up fresh `ConcurrencyGroup`s plus a recording http-server fixture that contend with other offload work.

## Test plan

- [x] `just test-quick "apps/minds/imbue/minds/desktop_client/agent_creator_test.py::test_start_creation_imbue_cloud_ai_with_local_compute_mints_litellm_key"` -- 1 passed in 6.69s
- [ ] CI

## Note

Codebase convention for occasionally-slow sync unit tests in this file is `@pytest.mark.flaky` (the sibling test) rather than a timeout bump. Going with a timeout bump as requested, but the flaky-mark may end up being the better fit if this still trips intermittently after the bump.


joshalbrecht:
lgtm

PR #2170, opened 2026-06-16
flush transcript converter correctly on macos and when interrupted

evgunter:
## What

Follow-up to the transcript-flush work (PR #2127, now merged). That PR added a synchronous `--single-pass` flush of the raw streamer + common-transcript converter at turn end, so a consumer that harvests the final assistant message from the common transcript on the WAITING transition (e.g. Catalyst's `mngr_runner`) can't outrun the converter. But it placed the flush at the start of `wait_for_stop_hook.sh`'s `run_post_completion`.

## The bug this fixes

`run_post_completion` is **skipped** on two turn-end paths:
- the **no-`/proc` fast path** -- `find_claude_pid` fails (no `/proc`), so the hook does `mark_inactive; exit` without `run_post_completion`. This is the path **macOS / local agents take** (verified live on macOS: STOP -> MARK_INACTIVE in ~27ms, no grace).
- the **SIGTERM/SIGINT handler** -- calls `mark_inactive` directly.

On those paths the `active` marker was cleared **without** flushing, so the converter race the flush was meant to close remained. Notably, Catalyst runs agents locally (`~/.mngr-catalyst`), which means the flush very likely wasn't running for it at all.

## Fix

Move the flush into `mark_inactive`, which **every** turn-end path calls right before clearing the marker -- so the flush is uniform across the grace, no-`/proc`, timeout, and signal paths.

A side benefit of the new placement: the flush now runs **immediately before** the `rm -f active` that clears the marker, rather than at the top of `run_post_completion` (before the potentially-slow autofix upload and `notify_user`). This tightens the window between "transcript flushed" and "marker cleared" to nearly nothing, which is exactly the invariant the WAITING consumer depends on.

Failure-safety is unchanged and layered: the call site is `command -v`-guarded (a missing/un-sourced lib is skipped), and `mngr_common_transcript_flush` itself wraps each converter pass in `|| true` and always returns 0, so neither a missing lib nor a flush failure can trip `set -e` and strand the turn-end signal.

## Verification

Created a real local (macOS) claude agent and confirmed via a temporary sentinel that the flush now executes on the no-`/proc` path before the marker is cleared, and the common transcript reflects the final message at that point. (`wait_for_stop_hook.sh` has no unit harness -- pre-existing; needs `/proc`.)


libs/mngr_claude/changelog/mngr-flush-on-all-stop-paths.md
```diff
@@ -0,0 +1,7 @@
+Fixed: the synchronous transcript flush at turn end (which keeps a WAITING-signal consumer
+from outrunning the common-transcript converter) now runs on *every* turn-end path.
+Previously it lived in `wait_for_stop_hook.sh`'s `run_post_completion`, which is skipped on
+the no-`/proc` fast path (macOS / local agents, where the Claude-ancestor PID lookup fails)
+and on the SIGTERM/SIGINT handler -- so on those paths the marker was cleared without
```
evgunter, line 5: is it a problem to slow down the SIGTERM/SIGINT handler with this?

joshalbrecht: I *think* it's ok, but how long does it end up waiting?

it would probably be bad if we were blocking SIGINT / SIGTERM for a non-trivial amount of time...

joshalbrecht:
lgtm other than our one comment, which we should discuss

libs/mngr_claude/changelog/mngr-flush-on-all-stop-paths.md
evgunter: updated to take a timeout. claude set it so the max value on SIGINT / SIGTERM is 4s (2s x 2). what should the actual max be?

joshalbrecht: that seems fine!

joshalbrecht:
lgtm after the last changes

PR #2171, opened 2026-06-16
somewhat more canonical common transcript format

evgunter:
Tracks the OpenTelemetry GenAI semantic conventions for the agent-agnostic common-transcript schema instead of bespoke field names. Design: `specs/common-transcript-standard/spec.md`.

**Base:** this PR targets `mngr/agent-mixins` (not `main`) because it builds on that branch's extraction of the common-transcript converters into real Python modules. Retarget to `main` once agent-mixins lands.

## Tier 1 — vocabulary alignment (all five emitters)
- Rename the assistant record's `stop_reason` -> `finish_reason` (the exact OTel term) in the schema and every emitter (claude/codex/antigravity Python converters; opencode/pi-coding TS emitters).
- Tool-call field names (`tool_call_id`/`tool_name`/`input_preview`) are deliberately left unchanged (see spec).

## Tier 2 — universal ordered `parts[]`
- Every assistant record now carries an ordered `parts[]` (text/tool_call segments, modelled on the OTel message parts) that preserves intra-turn text/tool interleaving — the canonical, agent-agnostic view the reader renders. Because it's universal, the reader has a single code path and **no per-agent fallback**.
- A `parts_ordered: bool` marks faithfulness: **true** for claude, pi-coding, opencode (all iterate their native ordered content) and trivially for codex (text-only assistant messages); **false** only for antigravity, whose native format doesn't record where tool calls sat relative to the text (best-effort text-then-tools).
- The flat `text` + `tool_calls` are kept as a convenience baseline; each emitter builds `parts[]` from the same source objects, so the two can't drift.

## Not in scope
- OTel as a transport/storage format (vocabulary + message-shape alignment only).
- A `reasoning` part type (deferred; latent sources noted in the spec: antigravity's discarded `PLANNER_THINKING`, claude's `thinking` blocks).
- Dropping the flat `text`/`tool_calls` for a single representation (a larger breaking change; tracked as a follow-up).

## Tests
Local (xdist): schema/reader/conformance-meta + all five emitters' converter/conformance tests pass (201 in the focused set; the node-driven pi/opencode tests ran for real); `ty` clean. Full suite via CI.

## Standards surveyed
There is **no published RFC or on-disk standard for an agent-transcript session file**, so the JSONL container stays ours; what we align to is field vocabulary and message shape. What was considered:

- **OpenTelemetry GenAI semantic conventions** (chosen). CNCF-backed; the only cross-vendor standard with real traction (Datadog, MLflow, Arize align to it). Its v1.37 *structured message model* represents a message as `role` + an ordered `parts[]` (`text`, `tool_call`, `tool_call_response`, `reasoning`) with `finish_reason` on outputs — which is what we borrow. It is a *telemetry* schema (spans + log-events to a collector), **not** a session-file format, so we adopt its vocabulary/shape, not its transport. Bonus: a future exporter from our records to OTel spans becomes a mechanical mapping.
- **OpenInference (Arize)**. OTel-based LLM-observability conventions, but messages are flattened into indexed span attributes (`llm.input_messages.0.message.role`) — span-only, even less suited to a standalone transcript file.
- **OpenAI / Anthropic message wire formats**. De-facto, but per-provider and divergent — i.e. exactly what the common transcript normalizes *away from*, so not a neutral target.
- **IETF**. No RFC exists; the active agent-related Internet-Drafts are about agent-to-agent comms, discovery, and security (e.g. `draft-ni-a2a-ai-agent-security-requirements`), not transcript serialization.

Refs: [OTel GenAI client spans](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/) · [GenAI attribute registry](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/) · [structured-message model discussion](https://github.com/open-telemetry/semantic-conventions/issues/1913) · [OpenInference conventions](https://github.com/Arize-ai/openinference/blob/main/spec/semantic_conventions.md)


libs/mngr/imbue/mngr/cli/transcript.py
```diff
@@ -147,16 +147,31 @@ def _format_event_human(event: dict[str, Any]) -> str:
...
+                    if part.get("type") == "text":
+                        content = part.get("content", "")
+                        if content:
+                            lines.append(content)
+                    elif part.get("type") == "tool_call":
+                        tool_name = part.get("tool_name", "unknown")
+                        preview = part.get("input_preview", "")
+                        lines.append(f"  -> {tool_name}({preview})")
+            else:
+                # Fall back to the flat text + tool_calls baseline.
```
evgunter: hm

joshalbrecht:
lgtm after tests pass!

PR #2174, opened 2026-06-16
Bump Latchkey, improve wording.

hynek-urban:
- Bump Latchkey version to 2.16.1 to get the latest fixes.
- Display the wildcard permission as "all" instead of "any" and, when checked, disable the other checkboxes to make the semantics clear to the user.


joshalbrecht:
lgtm

PR #2178, opened 2026-06-16
Bake bare-metal slices: minds pool create --backend slice, golden-image Docker pre-bake, throttled parallel bakes

joshalbrecht:
## Summary

Makes baking bare-metal **slices** (lima VMs carved on a pre-registered OVH bare-metal box) a first-class, fast, parallel-safe operation through `minds pool create` and the `just` recipes.

### `minds pool create --backend {ovh_vps,slice}`
- Adds a `--backend` option (default `ovh_vps`, unchanged). With `--backend slice` it bakes a slice instead of ordering an OVH VPS, resolving the tier's `pool-ssh` key (and host_pool DSN for shared tiers) from Vault — same ergonomics as the OVH path.
- New thin `just bake-slice-{dev,prod}` recipes, mirroring `bake-pool-host-*`.

### Golden-image Docker/inotify pre-bake
- `admin server prep` now `virt-customize`s the staged guest image to pre-install the **pinned** Docker Engine (same version `mngr_vps_docker` pins) + inotify-tools. Each slice VM's first-boot guards are presence-only, so they now skip the per-VM `get.docker.com` install — removing a per-slice network dependency and speeding up bakes.

### Throttled, self-healing parallel bakes
- `--max-concurrency` (default 4): bakes at most K slices at once, queues the rest, and logs progress as each completes. Bounds box contention so each `mngr create` stays under its (now 45-min, per-create) timeout — one slice timing out never aborts the others.
- After the bakes, reconciles the box's lima VMs against the DB and **reaps orphans** (a carved VM with no `pool_hosts` row, e.g. left by a create killed by its own timeout before the insert). Only untracked slice-prefixed VMs are removed.

## Validation (live, against the production box)
- Golden image verified to contain `docker-ce 29.5.1` + inotifywait; bakes show **0** `get.docker.com` installs.
- `--count 4`: 4/4 succeeded (~27 min).
- `--count 8` unthrottled: 0/8 (all hit the 30-min create budget under contention; left 8 orphan VMs — the motivating bug).
- `--count 8 --max-concurrency 4`: **8/8 succeeded** (~55 min, 2 waves), orphan reap found 0, box VM count == DB slice count.

## Tests
Unit tests added/updated for the backend split, `--max-concurrency` forwarding, and the pure orphan-computation; `ty` clean; ratchets green. Full suite runs in CI.


joshalbrecht:
lgtm

PR #2181, opened 2026-06-16
adopt existing session for agy, codex, opencode, pi + fixes for misc bugs found

evgunter:
## Per-agent session adoption

Adds **session adoption** for the four non-claude TUI agents (antigravity, codex, opencode, pi), so a freshly created agent can resume an existing conversation — matching what claude already does. Adoption is exposed through the shared `--adopt`/`--adopt-session` create option and a first-class `CreateAgentOptions.adopt_session` field, gated to agent types that implement `HasSessionAdoptionMixin`.

A session can be adopted from any of three local sources, resolved by session id (or an explicit path to the native store):

- a **preserved** session from a destroyed agent,
- a **live** mngr agent's store, or
- the **user-native** (plain-CLI) store.

### Per-agent mechanics

Each agent resolves the adopt argument across its stores, copies its native session store into the new agent, rewrites the recorded working directory to the new agent's work dir (so the CLI resumes instead of prompting to pick a directory), and writes its resume pointer:

- **codex** — rollout JSONL under `sessions/YYYY/MM/DD/`; rewrites `payload.cwd` in the adopted rollout.
- **opencode** — `opencode.db` (SQLite); rebinds the stored `session.directory`, `project.worktree`, and `project_directory` rows via the stdlib `sqlite3` module on a local copy of the db (no host `sqlite3` CLI dependency).
- **antigravity** — per-conversation `<id>.db` store; resumes by conversation id (directory-agnostic).
- **pi** — session JSONL; rewrites the embedded `cwd` in the first record.

### Combining `--adopt` and `--from`, and error behavior

- `--adopt` is **repeatable**: every named session is copied into the new agent (each coexists in the agent's session switcher) and the last one is resumed.
- `--adopt` may be **combined with `--from <agent>`**: every named session plus the clone's source conversation is made available, and the clone's conversation is the one resumed.
- A `--from` clone copies the workspace but not the state dir, so the source agent's native session store is transferred separately and rebound (`transfer_cloned_agent_session_store`). Previously only claude carried the source conversation forward on `--from`; all five agents now do.
- **Warn vs. raise:** an explicit `--adopt` of an unusable/ambiguous id is a hard error caught up front (before any host or worktree is created). A `--from` clone whose source has no resumable session is a **warning**, not an error — `--from` is fundamentally a workspace clone, so carrying the conversation forward is a bonus; the agent still starts (falling back to the last `--adopt`, or fresh).
- For claude, `--adopt A --from X` now **content-merges** the adopted session into the destination's encoded project dir instead of refusing on a whole-directory clobber; a genuine per-file collision (same session id in both) still raises.

### Shared core

- `HasSessionAdoptionMixin` (read-side counterpart to `HasSessionPreservationMixin`) and the agent-agnostic `--adopt` validation in `create()`.
- Helpers in `api/preservation.py` shared across the adoption plugins: `adopt_sessions` (the orchestrator: copy every `--adopt` session, then the `--from` clone, then resume the clone-or-last), `iter_agent_session_paths` (scan live + preserved agents), `transfer_cloned_agent_session_store` (the `--from` carry-forward), `dedupe_by_resolved_path`, `run_adopt_session_preflight` (the per-plugin `on_before_create` fail-fast, tied to the core validation gate), and `require_unique_match` (the uniform not-found/ambiguous resolver tail).

## `mngr create --yes` dismisses first-run dialogs

`--yes` (auto-approve) now dismisses claude's first-run **dialogs** (onboarding, effort callout, work-dir trust) in the per-agent config — previously only auto-dismissed for a remote/unattended agent or the explicit `auto_dismiss_dialogs` config. `--yes` deliberately does **not** accept bypass-permissions mode: it auto-approves prompts, it does not silently widen tool permissions (those stay governed by `auto_allow_permissions`/unattended).

## Common transcript: tool calls on the assistant turn

The codex common-transcript converter now records a tool invocation as a nested assistant `tool_calls` entry (sharing the id of the paired `tool_result`). codex models a tool call as a standalone rollout item, so previously the assistant turn carried no call — diverging from claude/antigravity/opencode/pi, which all nest it. The canonical envelope is now consistent across agents, and each emitter has a test pinning it.

## TUI readiness

`TUI_READY_INDICATOR` (the readiness signal `mngr` polls the pane for before sending keystrokes) is now a `str | re.Pattern` union: a plain string is an exact substring match (as before), a compiled `re.Pattern` is `re.search`. The matching mode is chosen by the value's **type**, not its contents. This fixes antigravity readiness for **agy 1.0.9**, which dropped the `? for shortcuts` footer hint `mngr` keyed on; agy now matches the input box itself (a rule, the `>` prompt, a rule) via regex, which survives both a fresh start and a resume.

## Release-test harness

The shared agent release-test harness drives every agent through one lifecycle arc — create → WAITING → message → stop → start → recall a pre-restart secret → destroy → adopt the preserved session into a fresh agent → recall again — proving the store actually *resumes*, not just that its bytes survive. Session adoption is now exercised unconditionally for every agent that preserves on destroy.

Observing the RUNNING marker is now required of every agent (was a per-profile opt-in). Forcing a bash tool call (asserting the nested assistant `tool_call` plus its `tool_result`) is enabled for claude/codex/opencode/pi; it stays gated off for antigravity, which makes the call but produces no in-turn `tool_result` (it runs the command async and ends the turn before the result settles, and its transcript decoder does not decode tool output) — tagged `FIXME(agy-forces-tool-call)` with the root cause. Per-agent adoption logic is additionally covered by unit tests.


libs/mngr/imbue/mngr/agents/agent_release_testing.py
```diff
@@ -96,14 +98,17 @@ class AgentReleaseProfile(abc.ABC):
     asserts_usage: bool = False
     # Whether destroy preserves this agent's transcripts to the local preserved/ dir
     # (the preserve-on-destroy feature, default-on for every current agent type). When
-    # set, the arc asserts the transcripts actually landed there after destroy. A
-    # profile whose agent disables preservation sets this False to skip that check.
+    # set, the arc asserts the transcripts landed there after destroy AND adopts the
+    # preserved session into a fresh agent (new worktree) to prove it resumes -- so every
+    # preserving agent must implement ``adopt_session_arg``. A profile whose agent disables
+    # preservation sets this False to skip both checks.
```
evgunter: need to fix, it shouldn't be able to skip the checks

libs/mngr_antigravity/imbue/mngr_antigravity/plugin.py
```diff
@@ -559,6 +672,133 @@ def on_destroy(self, host: OnlineHostInterface) -> None:
...
+          its root conversation.
+
+        Either way ``assemble_command`` then resumes the recorded id via ``agy --conversation``.
+        The explicit-vs-clone dispatch lives in the shared ``dispatch_session_adoption``.
+        """
+        dispatch_session_adoption(
+            options.adopt_session,
+            options.source_agent_state_location,
+            on_explicit=lambda args: self._adopt_session(host, args[-1]),
+            on_clone=lambda location: self._adopt_cloned_session(host, location),
```
evgunter: hmm

libs/mngr_antigravity/imbue/mngr_antigravity/plugin.py
```diff
@@ -559,6 +672,133 @@ def on_destroy(self, host: OnlineHostInterface) -> None:
...
+                "Clone adopt: source agent {} has no agy conversation store; cloned agent will start a fresh session",
+                source_location.path,
+            )
+            return
+        conversation_id = self._pick_cloned_conversation_id(host, source_location)
+        if conversation_id is None:
+            logger.warning(
+                "Clone adopt: transferred agy store from {} has no resumable conversation; "
+                "cloned agent will start a fresh session",
+                source_location.path,
```
evgunter, line 769: hmm

libs/mngr_claude/imbue/mngr_claude/test_claude_agent_e2e.py
```diff
@@ -0,0 +1,146 @@
...
+* Post-``--`` args. ``--dangerously-skip-permissions`` lets the forced bash tool call
+  run without a permission pause, ``--pass-env ANTHROPIC_API_KEY`` carries the key to the
+  agent, and ``--model haiku`` pins the cheapest tier (the seed/recall turns don't need
+  more).
+
+* Adoption resolves by the preserved session JSONL's absolute path. claude has no
+  root-session-id sidecar file (unlike codex); the preserved native store is the
+  per-agent ``projects/<encoded-work-dir>/<session-id>.jsonl`` tree, and
+  ``_resolve_adopt_session`` accepts a ``.jsonl`` path directly, so the path is both
+  unambiguous and independent of the encoded-cwd subdir name.
```
evgunter: the other agents should also do things like this

libs/mngr_claude/imbue/mngr_claude/test_claude_agent_e2e.py
```diff
@@ -0,0 +1,146 @@
...
+            return "Release test requires ANTHROPIC_API_KEY in the environment and `claude` on PATH."
+        return None
+
+    def setup(self, tmp_path: Path) -> AgentReleaseContext:
+        # Seed ~/.claude.json (in the autouse fixture's temp HOME) with the dialog
+        # dismissals + ANTHROPIC_API_KEY approval claude needs to start non-interactively.
+        # Per-work-dir trust is added automatically by ``mngr create --yes``. This also
+        # returns a subprocess env carrying the redirected HOME and the isolated
+        # MNGR_HOST_DIR / tmux server from the autouse fixture.
+        env = setup_claude_trust_config_for_subprocess(trusted_paths=[], root_name="mngr-claude-release-test")
```
evgunter: what is this doing that `--yes` doesn't

libs/mngr_codex/imbue/mngr_codex/plugin.py
```diff
@@ -880,6 +891,298 @@ def assemble_command(
...
+        """
+        transferred = transfer_cloned_agent_session_store(
+            host, self._get_agent_dir(), source_location, _AGENT_SESSIONS_RELPATH
+        )
+        if not transferred:
+            logger.warning(
+                "Clone adopt: source agent {} has no codex session store; cloned agent will start a fresh session",
+                source_location.path,
+            )
+            return
```
evgunter: also should probably error

evgunter: (depending on its interaction with --from)

libs/mngr_opencode/imbue/mngr_opencode/opencode_config.py
```diff
@@ -272,3 +279,155 @@ def build_opencode_config(
...
+    quoted_session = _sqlite_quote(session_id)
+    quoted_dir = _sqlite_quote(str(new_directory))
+    return (
+        f"UPDATE session SET directory = {quoted_dir} WHERE id = {quoted_session};"
+        f"UPDATE project SET worktree = {quoted_dir} "
+        f"WHERE id = (SELECT project_id FROM session WHERE id = {quoted_session});"
+        f"INSERT INTO project_directory (project_id, directory, time_created) "
+        f"SELECT project_id, {quoted_dir}, CAST(strftime('%s','now') AS INTEGER) * 1000 "
+        f"FROM session WHERE id = {quoted_session} "
+        f"ON CONFLICT (project_id, directory) DO NOTHING;"
```
evgunter, line 421: sigh

libs/mngr_claude/imbue/mngr_claude/plugin.py
```diff
@@ -334,17 +334,27 @@ class ProvisioningContext(FrozenModel):
     """Runtime context derived from host type and transfer mode."""
 
     is_unattended: bool = Field(description="Agent runs without user interaction (remote/deploy)")
+    is_auto_approve: bool = Field(
+        default=False,
+        description="The human auto-approved mngr's prompts (`mngr create --yes`). Dismisses first-run "
+        "*dialogs* (onboarding/effort/trust), but -- unlike `is_unattended` -- does not accept "
+        "bypass-permissions mode (tool auto-allow is governed by `auto_allow_permissions`).",
+    )
```
evgunter, line 342: kinda janky but i think this is ~correct behavior

libs/mngr_claude/imbue/mngr_claude/test_claude_agent_e2e.py
evgunter: and now they do!

joshalbrecht:
I'm not opposed to these changes, but they are slightly beyond my ken :)

lgtm provisional on you feeling good about them

joshalbrecht:
(and assuming tests pass, obviously)

evgunter:
there might be a few rough edges, but i think overall they're a big improvement!

PR #2183, opened 2026-06-17
better tui ready check for resumed claude

evgunter:
## Problem

When a Claude Code session is **resumed** (not freshly created), sending a message times out. Users hit `Timeout waiting for message submission signal (waited 90.0s)`.

Root cause: on resume, mngr treats the agent as ready as soon as Claude's `SessionStart` hook fires (which flips the agent to `WAITING` and writes `session_started`). But at that moment Claude's input box is still replaying the resumed transcript. mngr pastes the message and presses Enter into a not-yet-interactive TUI; the Enter is swallowed, `UserPromptSubmit` never fires, no enqueue marker appears, and `send_message` times out.

The fresh-create path avoided this only by accident: `InteractiveTuiAgent.wait_for_ready_signal` waits for the `"Claude Code"` welcome banner (gated on `is_creating`). That banner never renders on `claude --resume`, and the two resume entry points (`mngr start` resume message; `mngr message --start` → `ensure_agent_started(is_creating=False)`) never ran any TUI-ready wait at all.

## Fix

- `send_message` now waits for TUI readiness before pasting, covering **all** send paths (create-initial-message, `mngr start` resume message, `mngr message --start`). It is a fast no-op when the TUI is already ready.
- Claude's `TUI_READY_INDICATOR` changes from `"Claude Code"` to the input-prompt glyph `❯`, which renders on both fresh start and resume.
- `wait_for_tui_ready` gains a testable `timeout_seconds` param.
- The indicator still comes from the per-agent `get_tui_ready_indicator()`, so Codex/Antigravity keep their own indicators; OpenCode/Pi use their own delivery and are unaffected.

## Verification

Diagnosed, then replicated with a local stress loop (stop → `mngr message --start` against a real local Claude TUI with a growing transcript), run both with and without the fix:

| Run | Fix live? | Indicator | Result |
|---|---|---|---|
| Baseline | No | `Claude Code` | 4 of 6 resumes timed out (~95s each); messages never submitted (session size flat) |
| Fixed | Yes | `❯` | 0 of 12 resumes failed (~4.5–7s each); every message submitted (session grew 17KB→96KB) |

Unit tests: `tui_agent_test.py` + `tui_utils_test.py` (23 passed) and the Claude plugin suite green.

## Notes / residual risk

- Codex/Antigravity now also gate on their footer indicator at send time; fine unless those indicators can scroll off mid-turn (only Claude was live-tested here).


joshalbrecht:
lgtm

not the prettiest, but I think this will work

PR #2188, opened 2026-06-17
Self-healing permissions

hynek-urban:
Bowei [reported a weird case](https://imbue-ai.slack.com/archives/C0B1JACQ9HT/p1781646709485929) where the agents host permission directory didn't exist.

This could happen when someone deletes the data or when a failure at agent creation time happens.

This PR makes this "self-healing": missing symlinks and missing directories are re-created, missing permission files are replaced by default permissions. We still log this as an error to be further investigated (once we have good error reporting in place.)


joshalbrecht:
lgtm

PR #2189, opened 2026-06-17
Hynek/adjust latchkey discovery logging

hynek-urban:
- Do not use `logger.warning()`.
- Replace some `logger.debug()` calls with `logger.trace()` (those are relatively low-value entries that unnecessarily comprised the bulk of mngr_latchkey's events.jsonl file).


joshalbrecht:
lgtm

PR #2190, opened 2026-06-17
Bare-metal slices: explicit order options + server-id, prep cache-ownership fix, disk reap

joshalbrecht:
## Summary

Operator-facing improvements to the imbue_cloud bare-metal / slice provisioning flow, made while standing up a new production bare-metal box in US-WEST (OVH `hil`, Hillsboro) and baking `minds-v0.3.1` slices onto it.

## Changes

- **`admin server order` requires explicit option selection.** Plans whose mandatory OVH eco option families (e.g. `bandwidth`, `vrack`) offer more than one choice previously failed to order (`expected exactly one X option to auto-pick`). Added a repeatable `--option <planCode>` flag; single-offer families are still auto-selected, and the error for an unspecified multi-offer family lists each offer with its monthly price so you can re-run. (An interim commit auto-picked the cheapest; that was replaced with the explicit flag per review.)

- **Slice baking requires an explicit `--server-id`.** `admin pool create --backend slice` (and the `minds pool create` wrapper) no longer auto-selects the box with the most free slots — you name the target box (from `admin server list`), validated ready + with enough free slots.

- **Fixed a box-prep bug that made every slice bake fail.** The prep script runs as root and staged the slice base image under the lima user's `~/.cache`, leaving `~/.cache` itself root-owned, so `limactl` (run as the lima user) couldn't create `~/.cache/lima` (`permission denied`). Prep now creates + chowns the cache-dir chain to the lima user and repairs an already-root-owned `~/.cache` on re-run.

- **Orphan reap now reclaims leaked data disks, not just VMs.** A failed carve whose rollback `limactl delete` can't unlock the disk leaves the disk behind (VM gone, disk still holding the box slot). The end-of-bake reap now reconciles the box's lima disks against the pool DB and force-deletes (unlocking first) any slice disk with no row.

## Testing

- Targeted unit suites green (ordering, bare_metal, bare_metal_db, bare_metal_prep, lima_slice_client, cli/admin, cli/server, minds cli/pool); `ty` type-check clean; CLI docs regenerated.
- Exercised end-to-end against real production OVH: ordered + delivered + set up the US-WEST box and baked 8 `minds-v0.3.1` slices (8/14 slots) successfully after these fixes.
- Full `just test-offload` not yet run locally — relying on CI here.

Changelog entries added for `mngr_imbue_cloud`, `minds`, and `mngr`.

### DIFF
--- a/apps/minds/changelog/mngr-new-production-west.md
+++ b/apps/minds/changelog/mngr-new-production-west.md
@@ -1,0 +1,4 @@
+`minds pool create --backend slice` now requires `--server-id` (the bare-metal
+box to bake the slices onto, from `mngr imbue_cloud admin server list`), forwarded
+to the underlying `mngr imbue_cloud admin pool create`. Slice baking now targets
+an explicitly-chosen, ready server rather than auto-selecting one.
--- a/apps/minds/imbue/minds/cli/pool.py
+++ b/apps/minds/imbue/minds/cli/pool.py
@@ -129,6 +129,7 @@ def build_create_admin_args(
     is_recycle_enabled: bool,
     is_dry_run: bool,
     is_deferred_install_wait_skipped: bool,
+    server_id: str | None = None,
     max_concurrency: int | None = None,
 ) -> list[str]:
     """Compose the ``mngr imbue_cloud admin pool create`` argv from minds-side inputs.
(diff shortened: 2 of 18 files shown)


joshalbrecht:
lgtm

ran these live and confirmed improvements

PR #2191, opened 2026-06-18
Fix imbue_cloud restart leaving leased minds broken

gnguralnick:
## Problem

Restarting a stopped `imbue_cloud` (leased pool) mind left it dead and UI-unrecoverable, even though its data survived on the volume. This is **bug C** from the remote-mind-recovery breakdown: a transient blip funneled users into the recovery flow, the recovery action stopped the container, and it could not be brought back ("Start step of host restart failed").

There were **two** root causes, and the second is load-bearing:

1. **Routing (`get_host`).** `mngr start` only re-starts a host when `get_host` reports it offline: `ensure_host_started` dispatches on the static type — `case Host()` returns as-is (start_host skipped), only a non-`Host` `HostInterface` (e.g. `OfflineHost`) calls `start_host`. But `ImbueCloudProvider.get_host` returned an online `ImbueCloudHost` **unconditionally**, with no container-state probe. So the start command skipped `start_host` entirely and SSHed straight into the dead container. (This is why the bug bit even with the re-bootstrap below in place.)

2. **Re-bootstrap (`start_host`).** Even once reached, a bare `docker start` + `_build_host_object` is not enough: the in-container sshd is launched via `docker exec` (the container CMD is just a sleep), so it does **not** survive a stop/start, and the per-host authorized key and host key may not persist either.

Unlike the local docker and vps-docker providers (whose `get_host` already returns an offline host for a stopped container, and which re-bootstrap sshd on restart), `imbue_cloud` had neither — so the earlier vps-docker fix did not cover it.

## Fix

**1. `get_host` routes a stopped container through `start_host`.** It now probes the inner container's running state over the outer root SSH (`docker_inspect_running`) and returns `to_offline_host(host_id)` when the container is stopped — mirroring `VpsDockerProvider.get_host`. `get_connector` is decoupled to build the host object directly from the lease, since a connector is needed regardless of running state and `get_host` can now return a connector-less `OfflineHost`.

**2. `start_host` re-bootstraps the container's SSH** over the outer root SSH (which works independently of the container's sshd):

- relaunch sshd (`start_container_sshd`)
- re-seed the per-host authorized key — covers `/root` not persisting (reuses the shared helper, see 3)
- wait for the container sshd to accept connections
- re-scan and re-record the served host key — reconciles host-key drift (detail D) so strict host-key checking succeeds

**3. The shared `build_add_authorized_keys_command` is now idempotent.** The restart re-seed originally hand-rolled its own `grep -qxF`-guarded command because the shared seeding helper (used by every provider at lease/create time) did a blind append. That helper now guards each append with `grep -qxF` (grouped in a subshell so the `||` doesn't break the surrounding `&&` chain), so it never accumulates duplicate `authorized_keys` lines on a re-run. `_reseed_container_authorized_key` now delegates to it, removing the duplication; the first-time callers (docker, lima, modal, vps-docker) are unaffected in practice but are now safe against accidental double-application.

This reuses the shared vps-docker / ssh-host-setup building blocks rather than adding new machinery.

Scope is **C only**. A (minds funneling transient remote blips into recovery) and B (the restart guard) are intentionally untouched. No activity-watcher relaunch is needed here because imbue_cloud minds run with `idle_mode = disabled`.

> Note: `get_host` now opens an outer SSH connection and can raise if the leased VPS is unreachable (it was connection-free before). This matches `VpsDockerProvider.get_host` and is intended — you cannot start a host whose VPS is down.

## Test plan

- [x] `test_get_host_returns_offline_host_when_container_stopped` / `..._online_when_running` — assert the routing fix (stopped container resolves to a non-`Host` offline host so `ensure_host_started` reaches `start_host`; running container stays online).
- [x] `test_start_host_rebootstraps_container_ssh` drives the real `start_host` against a recording stub outer and asserts the full re-bootstrap sequence (container start -> sshd relaunch -> authorized-key re-seed -> wait -> host-key re-scan).
- [x] `test_build_add_authorized_keys_command_is_idempotent` runs the generated shell twice against a temp-redirected path and asserts each key appears exactly once (validates the guard + subshell grouping).
- [x] Regression tests refactored onto the shared stub-outer pattern (`_StubOuter` via `cast`), replacing the hand-rolled `OuterHostInterface` subclass.
- [x] `just test-quick libs/mngr_imbue_cloud` — 378 passed; consuming providers (`libs/mngr` providers, vps-docker, lima, modal, non-slow) — 1150 passed; `ty` (type check) clean; `ruff check` / `ruff format` clean.
- [ ] Live-lease validation: a real stop/start on a throwaway lease is the definitive end-to-end check (needs connector creds + a paid lease; not run in CI).

> Unrelated flaky test seen locally: `ssh_utils_test.py::test_load_or_create_ssh_keypair_concurrent_first_creation_is_consistent` times out under heavy xdist contention (passes in isolation; concerns SSH **keypair** generation, not authorized_keys). Out of scope for this branch.


libs/mngr_imbue_cloud/imbue/mngr_imbue_cloud/providers/instance.py
```diff
@@ -1535,8 +1584,50 @@ def start_host(
...
+            start_container_sshd(outer, container_id)
+            # Re-seed the per-host authorized key in case the container's ``/root``
+            # did not persist across the stop/start. The connector installs it at
+            # lease time, but we cannot rely on that surviving a restart.
+            self._reseed_container_authorized_key(outer, container_id, host_id)
+            self._wait_for_container_sshd(leased)
+            # The restarted container may serve a different host key than the one
+            # recorded at lease time; re-scan and update known_hosts so the
+            # ImbueCloudHost's strict host-key checking succeeds.
+            self._scan_and_record_container_host_key(host_id, leased.vps_address, leased.container_ssh_port)
```
gnguralnick, line 1601: i added this logic when i was testing on the ovh vps (rather than the bare-metal slice) since restarting that did cause keys to be changed and need to be reseeded and rerecorded.

on the new slice hosts i think this is unnecessary - since we're just restarting a container within the host, keys won't be changed. though i don't think it's harmful to have, in case the overall machine does end up being restarted for some reason?

joshalbrecht: we cannot / should not have this -- it's a security issue, sorry :(  

otherwise you are more exposed to MITM attacks

joshalbrecht: I'm going to merge this PR, remove these bits, and then check that the remote instances are robust to restarts

joshalbrecht:
will fix this up after merging

PR #2192, opened 2026-06-18
Supervisord migration: blueprint + minds-dev-workflow Vault docs

joshalbrecht:
The supervisord migration itself lands in the **forever-claude-template** repo (it's all template/bootstrap/skill changes): https://github.com/imbue-ai/forever-claude-template/pull/176

This monorepo PR carries only the pieces that live here:

- **`blueprint/supervisord-services/plan-supervisord-services.md`** — the implementation plan for replacing the custom bootstrap service manager with supervisord (and making `system-services` a real background Claude agent).
- **`.claude/skills/minds-dev-workflow/SKILL.md`** + **`dev/changelog/mngr-supervisord.md`** — a "Vault (for pool / slice bakes)" note added while testing the change end-to-end: bakes need an interactive `vault login -method=oidc`; the minds wrappers auto-apply the imbue HCP `VAULT_ADDR`/`VAULT_NAMESPACE` defaults (from `vault_reader.py`), so they work with just the token; raw `vault`/`mngr imbue_cloud admin` commands need those two exported (a `127.0.0.1:8200` "connection refused" is a missing address, not a logged-out session).

## Verification

The FCT change was verified live across all three providers (see FCT PR #176 for detail): Docker, Lima (incl. the supervisor-systemd-conflict fix + the git-2.39 `--orphan` fix), and an imbue_cloud slice (leased fast-path) — supervisord runs all 8 programs, `system-services` runs a real idle Claude, and runtime-backup commits.


joshalbrecht:
lgtm

PR #2193, opened 2026-06-18
Make pool destroy slice-aware (tear down the lima VM, not just the row)

joshalbrecht:
## Summary

`admin pool destroy` now branches on the row's `backend_kind`, mirroring how `pool create` branches on `--backend`. Previously destroy only knew the OVH-VPS teardown path (cancel the OVH VPS, then drop the row), so destroying a bare-metal **slice** pool host:

- 404'd trying to cancel a non-existent OVH VPS at the box's address, and
- with `--skip-vps-cancel`, dropped the `pool_hosts` row while leaving the slice's lima VM running on the box — stranding a slot indefinitely (no cron reaps slice-VM orphans; only a subsequent bake does).

Now the teardown follows the backend:
- **slice** → destroy the lima VM (and data disk) on the bare-metal box, freeing the slot, then drop the row.
- **ovh_vps / legacy-null** → cancel the OVH VPS, as before.
- **`--skip-vps-cancel`** → drop the row only, any backend.

Either teardown runs *before* the row delete, so a failure keeps the row and the operation stays retryable — never a stranded VPS or slice VM.

## Changes

- `libs/mngr_imbue_cloud/cli/admin.py`: `PoolHostUnderlyingTeardown` enum + pure `resolve_underlying_teardown(...)` planner; `_destroy_slice_pool_host_vm(...)` helper; `pool_destroy` rewired to read `backend_kind`/`bare_metal_server_id`/`lima_instance_name` and dispatch via `match`/`assert_never`; updated `emit_json`, flag help, and docstring.
- `libs/mngr_imbue_cloud/cli/server.py`: new `destroy_slice_vm(server, lima_instance_name)` — the targeted counterpart of the orphan reaper (reuses `LimaSliceVpsClient.destroy_instance` + `_pool_private_key_path`).
- `apps/minds/cli/pool.py`: `pool destroy` wrapper injects both OVH creds and `POOL_SSH_PRIVATE_KEY` from Vault when not skipping, so `just destroy-pool-host <id>` works for a slice with no flags. (The wrapper can't cheaply know the row's backend, and every pool-bearing tier has both Vault entries — documented inline.)
- `justfile`: updated `destroy-pool-host` recipe comment.
- Changelog entries for `libs/mngr_imbue_cloud`, `apps/minds`, and `dev`.

## Tests

- 4 new unit tests for `resolve_underlying_teardown` (slice→VM, ovh→VPS, legacy-null→VPS, skip→none) — the slice→VM case locks the previously-broken behavior.
- Ran locally: affected unit tests (admin/server/minds-pool) pass, `ty check` clean, ratchets pass, changelog gate ok, and `autofix` found no issues.
- Following existing convention, the live slice-destroy path has no end-to-end integration test (the OVH cancel path likewise is only covered via its pure helper, since the live paths need a real DB/box/OVH).


joshalbrecht:
lgtm

PR #2194, opened 2026-06-18
just the skills changes from mngr/examine-errors-skill

evgunter:
Extracts only the skill-instruction and verified-accurate doc changes from `mngr/examine-errors-skill`, leaving out all logic code changes.

- The four existing `identify-*` skills now take a `target_path` argument (a whole library or any subdirectory within one) instead of a bare library name, resolving the scan scope and its containing library explicitly.
- New `identify-suspicious-edge-cases` skill that flags over-broad exception catches, fallback `else` branches, defensive guards, and unnecessary `| None` types.
- `libs/mngr/non_issues.md`: clarified that the `is_`-prefix and missing-`Field()` exemptions apply specifically to CLI command functions and CLI-options data classes (e.g. `ListCliOptions`) that mirror user-facing click options; internal boolean fields on non-CLI data classes still use `is_`. Verified against `ListCliOptions` and style_guide.md line 677.

Deliberately excluded the `list.md` `host.ssh.key_path` doc edit: it describes the `Path | None` behavior that only exists alongside the `primitives.py` logic change, which is not part of this branch. On this branch `key_path` is non-optional, so that edit would be inaccurate here.


joshalbrecht:
lgtm!

PR #2196, opened 2026-06-18
unified provider release test harness + fixing some bugs it caught

evgunter:
## Summary

Unifies the per-provider end-to-end release tests into instantiations of **one shared harness**, mirroring the agent release-test harness (`agent_release_testing.py`). Implements the **trips** from `specs/provider-release-tests.md` (vendored into this PR) — single-boot, multi-assertion "trips" amortized across one provisioned host — so each provider's release suite collapses to a thin profile. Along the way it **surfaced and fixed three real provider error-UX bugs** (AWS/GCP/Modal).

Covers **AWS, GCP, Azure, Modal** across container + bare isolation.

> Based on `mngr/bare-providers` (merged in) — the spec targets that end-state (`mngr_vps`, realizer seam, `IsolationMode`, capability flags). Retarget the PR base to `main` once `bare-providers` lands.

## The harness

- **`imbue.mngr.providers.provider_release_testing`** (core): the `ProviderReleaseProfile` contract + `run_provider_release_trip{1,2,3,4}` drivers. Isolation-agnostic — speaks only through capability booleans + profile-supplied probes, never imports `mngr_vps`; delegates the CLI to the shared `run_mngr_subprocess`.
- **`VpsCloudReleaseProfile`** (`mngr_vps/testing.py`): shared plumbing for the VPS-family clouds (AWS/GCP/Azure) — probes via the common `VpsClientInterface`.
- **Thin per-provider profiles**: AWS/GCP/Azure subclass the VPS base + parametrize over `CONTAINER`/`NONE`; Modal implements the ABC directly.

## Trips (real provisioning; release tier, not in CI)

**Trip 1 — full lifecycle** (create → tagged-resource probe → exec marker → plain stop → `--stop-host` real-stop / loud refusal → start → persistence → snapshot → sketchy-kill → CRASHED → gc → backend-clean):

| | Container | Bare |
|---|---|---|
| AWS | 5:34 ✅ | 4:08 ✅ |
| GCP | 6:13 ✅ | 5:11 ✅ |
| Azure | ✅ | ✅ |
| Modal | 2:30 ✅ | n/a |

**Trip 2 — idle auto-shutdown** (create idle → poll cloud API until compute genuinely stops → resume + marker survived; Modal asserts sandbox-terminated, no resume): AWS 4:36 ✅, GCP 6:47 ✅, Azure 6:48 ✅, Modal 2:49 ✅ (container; bare via equivalence). Uses `--idle-timeout` (resumable stop) over the terminate-on-cap path.

**Trip 3 — snapshot survives destroy** (snapshot → destroy → portability probe): Modal 2:49 ✅ (portable: `create --snapshot` restores the marker); AWS 3:40 ✅ (non-portable docker-commit → asserts the snapshot is gone after destroy); GCP/Azure container same non-portable branch; bare skips (`supports_snapshots=False`).

**Trip 4 — error classification** (no-boot, ~35s): unresolvable creds → contract `ProviderUnavailableError` + provider-correct curated help; `--vps-*` build arg → migration hint. **All four honest** after the fixes below.

| Provider | Trip 4 |
|---|---|
| AWS | ✅ contract error + `aws configure` help |
| GCP | ✅ contract error + `gcloud auth application-default login` help |
| Azure | ✅ (already curated: `az login`) |
| Modal | ✅ contract error + `uvx modal token set` help |

## Provider error-UX bugs fixed in this PR (surfaced by Trip 4)

- **AWS** (`mngr_aws/backend.py`): missing-cred `ProviderUnavailableError` now carries curated `user_help_text` pointing at `aws configure` (was the generic "start Docker" default).
- **GCP** (`mngr_gcp/backend.py`): same — now points at `gcloud auth application-default login` + project/ADC setup.
- **Modal** (`mngr_modal/backend.py`): the auth-failure path now raises the **contract `ProviderUnavailableError`** (was a plain `MngrError`), with curated `uvx modal token set` help — so `mngr list`/`gc` keep it visible like the other clouds.
- Plus a shared test-helper fix: `delete_modal_apps_in_environment` passes `--yes` to `modal app stop` (non-interactive runs).

## Notes
- **Offline host_dir read** is folded into **Trip 1** as an opt-in step (env `MNGR_RELEASE_TEST_OFFLINE_HOST_DIR=1`, off by default so the happy-path trip stays lean): in the `--stop-host` window it asserts a stopped host's host_dir marker is served from the offline mirror (`mngr file get … --relative-to host`) without starting the host. AWS and Azure both validated end-to-end (Blob/S3 state-bucket host_dir capture); GCP (`NullHostDirBackend`) and Modal skip. Replaces the need for a separate offline-host_dir trip.
- Two dependency bugs the trips surfaced were fixed upstream + merged in: Azure operator Blob-RBAC (`mngr/volumes`) and GCP-bare SSH-resume host-key wait (`mngr/bare-providers`).
- Existing per-step lifecycle tests are retained for now; they'll be retired once the trips are stable (a follow-up, which also collapses the remaining `_run_mngr` duplication).
- Still owed (docstring backlog): Trip 1b (N agents per host).

## Test plan
Release tier — run manually with creds + opt-in, e.g.:
```
MNGR_AWS_RELEASE_TESTS=1 uv run pytest -m release -p no:xdist --no-cov \
  libs/mngr_aws/imbue/mngr_aws/test_release_aws.py
```


joshalbrecht:
lgtm!

I wasn't able to to audit every detail, but this does seem like an improvement over what we have right nwo

joshalbrecht:
merged to main

PR #2198, opened 2026-06-18
detect agent subtypes properly for transcript

evgunter:
## Problem

`mngr transcript <agent>` errored with `Unknown agent type 'X'` for an agent whose type is a config-defined subtype -- a custom `[agent_types.X]` with `parent_type = "claude"` (no class of its own; it inherits `ClaudeAgent`).

The command's precheck resolved the type via a flat `get_agent_class()` lookup, which only knows plugin-registered classes (`claude`, `codex`, ...). Subtypes aren't in that registry, so it raised. `mngr transcript` was the only production caller doing a flat lookup; every other command resolves via `resolve_agent_type()`, which walks the `parent_type` chain (and aliases).

## Fix

- `_assert_agent_type_supports_transcripts` now resolves through `resolve_agent_type()`, so subtypes and aliases map to their parent's class.
- A type that does not resolve at all still **blocks** with the resolver's clear error -- the precheck exists to refuse types we don't know how to read, so it must not fall through to transcript discovery.
- Documented `get_agent_class()` as a low-level primitive that does NOT resolve subtypes/aliases, pointing callers at `resolve_agent_type()`.
- Tests: a custom `coder` (`parent_type = claude`) agent's transcript resolves and renders; an unresolvable type is blocked.

## Testing

`just test-quick` over `transcript_test.py`, `agent_config_registry_test.py`, `agent_registry_test.py`: passing. `ty check` clean on the changed files.


joshalbrecht:
lgtm

PR #2199, opened 2026-06-18
fix `mngr plugin list` mislabeling blocked claude_subagent_proxy as enabled

evgunter:
claude_subagent_proxy keeps causing trouble by not being disabled in weird edge cases...

(here it was actually disabled it just didn't report that correctly)

---

## Problem

`mngr plugin list --fields name,enabled` reported `claude_subagent_proxy` (an opt-in plugin) as `enabled=true` even though it is actually blocked: after `create_plugin_manager()`, `pm.is_blocked("claude_subagent_proxy")` is `True` and its hooks never fire.

## Root cause

There were two divergent representations of "disabled":
- `read_disabled_plugins()` (the pre-reader) is opt-in-aware and is what `create_plugin_manager` uses to actually `set_blocked`.
- `config.disabled_plugins` is rebuilt by `_apply_plugin_overrides` from only the `[plugins.*]` enabled flags, so it **drops** the opt-in-derived set.

Separately, pluggy's `set_blocked` leaves the blocked name in `list_name_plugin()` with a `None` plugin object, so `_gather_plugin_info` still emitted it and `_is_plugin_enabled` (config-only) reported it enabled. Because blocking is strictly one-way (there is no `set_unblocked`), the field could diverge from `pm` in both directions.

## Fix

Two parts:

1. **Root fix (`loader.py`):** `load_config` now folds the opt-in-derived disabled set into `config.disabled_plugins`, so the field is faithful for *all* consumers (provider filtering, config validation, completion, the plugin list). Gated on `MNGR_LOAD_ALL_PLUGINS` exactly like `create_plugin_manager`, so doc/tooling runs still load every plugin. The CLI/`[plugins.*]` block pass is left untouched (opt-in plugins are already blocked at startup), preserving the strict `--disable-plugin` typo check.

2. **Display backstop (`plugin.py`):** `_gather_plugin_info` ANDs the config check with `not pm.is_blocked(name)`. `pm` is the runtime ground truth, and this catches blocks that never reach `config.disabled_plugins` — e.g. command-default / create-template `--disable-plugin`, applied after `load_config` in `setup_command_context`.

## Behavior

- Default (no explicit enable): `claude_subagent_proxy` → `enabled=false`; `pm.is_blocked()` is `True`. Agrees.
- `[plugins.claude_subagent_proxy] enabled = true`: → `enabled=true`, and actually loads. Agrees.
- `MNGR_LOAD_ALL_PLUGINS=1`: → `enabled=true` (loaded for tooling); not marked disabled.

All three verified end-to-end against a throwaway `MNGR_HOST_DIR` in a non-git cwd.

## Tests

- `loader_test.py`: opt-in plugin included in `disabled_plugins` by default; excluded when explicitly enabled; excluded under `MNGR_LOAD_ALL_PLUGINS`.
- `plugin_test.py`: blocked plugin reports disabled (fails without the display fix); unblocked complement reports enabled.

### DIFF
--- a/libs/mngr/changelog/mngr-subagent-proxy-begone.md
+++ b/libs/mngr/changelog/mngr-subagent-proxy-begone.md
@@ -1,1 +1,5 @@
-Fixed `mngr plugin list` mislabeling opt-in plugins (e.g. `claude_subagent_proxy`) as `enabled=true` when they are actually blocked. The reported `enabled` state now reflects the plugin's real block state: opt-in plugins that were not explicitly enabled show `enabled=false`, while a plugin enabled via `[plugins.<name>] enabled = true` still shows `enabled=true`.
+Fixed `mngr plugin list` mislabeling opt-in plugins (e.g. `claude_subagent_proxy`) as `enabled=true` when they are actually blocked.
+
+The reported `enabled` state now reflects the plugin's real block state: opt-in plugins that were not explicitly enabled show `enabled=false`, while a plugin enabled via `[plugins.<name>] enabled = true` still shows `enabled=true`.
+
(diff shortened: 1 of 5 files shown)


joshalbrecht:
lgtm

PR #2200, opened 2026-06-18
add version pinning and auto-update control to all agent types except antigravity

evgunter:
only lightly reviewed

---

## Summary

Gives every agent type (claude, antigravity, codex, pi, opencode) the ability to (a) **pin the version that gets auto-installed** and (b) **govern the CLI's self-updater** via a uniform `update_policy`.

## Current behavior (after this PR)

Each agent gets an `update_policy` field (`AUTO` / `ASK` / `NEVER`):
- **`NEVER`** — block the CLI's self-updater so the installed version is preserved (the concrete lever differs per agent; see table).
- **`AUTO`** — leave the CLI's own auto-updater enabled.
- **`ASK`** — prompt at provision time. Only **codex** implements a prompt flow; on the other agents `ASK` behaves like `AUTO`.

**Default when unset:** `NEVER` (block self-update), so a managed agent stays on its installed version — **except** an *attended* agent that has an interactive update flow, which defaults to `ASK`. Unattended (remote/deploy) agents always default to `NEVER`. In practice: claude/antigravity/pi/opencode default to **self-updater disabled**; codex defaults to **ASK** (prompt on an attended local run, just notify when unattended). Set `update_policy = "AUTO"` to opt back into a CLI's auto-updater.

| Agent | Version pin | Self-updater disable (`NEVER`) |
|---|---|---|
| claude | `version` (already existed) | `DISABLE_AUTOUPDATER=1` in the agent env |
| codex | `version` → `npm i -g @openai/codex@<v>` + reconcile | already disabled (`check_for_update_on_startup=false`); a pin also skips the provision-time update check |
| antigravity | none — Google's installer can't pin (latest-only; no version arg/env var, build-id'd URLs) | `AGY_CLI_DISABLE_AUTO_UPDATE=true` (confirmed in the `agy` binary) |
| pi | `version` → npm `@<v>` + reconcile | `PI_SKIP_VERSION_CHECK=1` |
| opencode | `version` → installer `VERSION=<v>` + reconcile | `"autoupdate": false` in the per-agent `opencode.json` |

## How this relates to claude's preexisting behavior

**Before this PR, mngr did _not_ disable Claude Code's auto-updater on local agents.** The per-agent `.claude.json` is seeded from the user's `~/.claude.json` (`sync_claude_json` defaults `True`), and mngr never set `autoUpdates` or `DISABLE_AUTOUPDATER`. So:
- If `~/.claude.json` set `autoUpdates` → that value was inherited.
- If it omitted `autoUpdates` → Claude Code's native-install default (auto-update **on**) applied.
- Only a brand-new user with **no** `~/.claude.json` got the generated default `autoUpdates: false`.

Net: **local claude agents typically auto-updated.** Remote agents were unaffected too (no env var was set).

**This PR changes that:** claude now disables the auto-updater by default (`update_policy` unset → `NEVER` → `DISABLE_AUTOUPDATER=1`), on both local and remote agents. The disable is keyed off the `DISABLE_AUTOUPDATER` env var — the documented, authoritative lever — so it works regardless of whether your `~/.claude.json` sets `autoUpdates` to true/false or omits it. It is **skipped in `use_env_config_dir` (shared) mode**, where mngr deliberately leaves your claude environment alone. Set `update_policy = "AUTO"` to restore the old auto-updating behavior.

(The other agents' self-updaters were likewise not previously disabled by mngr, so they too now default to disabled.)

## Shared infrastructure (`libs/mngr`)
- `AgentUpdatePolicy` enum + `resolve_update_policy` (explicit value wins; else `NEVER`, or `ASK` for an attended ask-capable agent).
- `extract_cli_semver` / `verify_pinned_cli_version` — verify an installed CLI matches a pin; lenient when the `--version` output is unparseable, error only on a confirmed mismatch.

## Notes / decisions
- **Antigravity version pinning is impossible, not merely omitted:** Google's `install.sh` always installs the latest build from a manifest (no version argument/env var), and per-version download URLs embed an undiscoverable build-id (both confirmed against the real installer). So there is no `version` field; use `update_policy = "NEVER"` (the default) to freeze whatever build was installed.
- `ASK` is only implemented for codex (its existing prompt + `codex update` flow); other agents treat `ASK` as `AUTO`. Reuses the existing `modify_env_vars` agent hook rather than adding a new callback.
- Mechanisms were verified against vendor docs/installers; `AGY_CLI_DISABLE_AUTO_UPDATE` was confirmed directly in the `agy` binary's strings.

## Testing
- New unit tests for the shared enum/resolver, the version helpers, and each agent's install-command / env-var / config wiring (including the default-disabled behavior).
- Ran the touched packages' unit + ratchet suites locally (all green); full suite runs in CI.


joshalbrecht:
> Antigravity version pinning is impossible, not merely omitted:

lol, come on, that is just trash. How far google has fallen...

joshalbrecht:
lgtm

leaving for @evgunter to merge after doing a slightly closer review if you want (per your comment on the PR)

evgunter:
lgtm

PR #2201, opened 2026-06-18
Harden minds workspace recovery logic against provider unavailability

gnguralnick:
## Summary

Extends minds' workspace-recovery flow to handle remote (Imbue Cloud) minds. Recovery was designed and tested only against local providers (docker/lima) before the imbue_cloud provider was finalized; on a flaky connection it would funnel a healthy remote mind onto the recovery page and offer a destructive host restart that routes through the same unreachable backend and cannot help.

Core idea: **gate recovery on provider reachability, and never offer a destructive action that can't help.**

## Changes

**libs/mngr_imbue_cloud**
- `_list_leased_hosts_cached` now raises a typed `ProviderUnavailableError` when the connector is unreachable at the transport level (`httpx.HTTPError`: connection refused / DNS / timeout), with curated help text that does not tell a cloud user to "start Docker". Auth (`ImbueCloudAuthError`) and account-config failures keep their own types. The error surfaces in `mngr list`'s `errors[]` as `exception_type == "ProviderUnavailableError"`, which is the typed signal the recovery probe keys on.

**apps/minds**
- The recovery host-health probe now reads `mngr list`'s JSON body (its `errors[]`) even on a non-zero exit, and adds two classification tiers above all host/interface tiers: `PROVIDER_UNAVAILABLE` (connector/daemon unreachable -> Retry + auto-reconnect poll, no restart) and `WORKSPACE_UNREACHABLE` (auth/account problem -> show reason, no restart). Both are attributed to this workspace's own provider.
- The probe's `mngr list` now runs under a much shorter (30s) timeout, and a timeout is itself classified as `PROVIDER_UNAVAILABLE` -- the full-network-outage signature. Previously a dropped connection left the diagnostic on the "Loading workspace" loader for the full ~2-minute restart timeout and then offered a restart that couldn't help. The in-container exec probe is skipped when the listing timed out (it would only add another doomed round-trip over the same dead network).
- The recovery page renders a "Can't connect to ..." page (Retry + a geometrically backed-off background poll that auto-returns once the provider is reachable) and a generic "Can't reach your workspace -- `<reason>`" page. Neither offers nor auto-dispatches a restart.
- The Retry button now shares the prominent full-width primary-button styling with the Restart button (previously it fell back to an unstyled native browser button).
- Internal cleanup: the recovery probe's nonzero-exit-tolerant `mngr` runner (`_run_mngr_capturing`) is now the single subprocess-launch primitive, and the existing raise-on-nonzero `_run_mngr` delegates to it instead of duplicating the launch/timeout handling.

**docs**
- Adds the design docs: `blueprint/remote-mind-recovery/plan-remote-mind-recovery.md` and a root-level diagnosis note `remote-mind-recovery-bugs.md`.

## Not included
- Provider-side re-bootstrap of a stopped imbue_cloud container (the documented **Bug C**: `ImbueCloudProvider.start_host` does a bare `docker start` and doesn't restore in-container sshd / `authorized_keys` / the host key). That is the root cause of the original incident and is out of scope here; this PR is the minds-side safety net so a transient blip can't reach the destructive path in the first place.
- A pre-stop interface re-confirmation on the destructive `--stop-host` path (re-probe just before the stop, abort if recovered): explored on this branch but dropped as non-essential -- the provider-reachability gating plus the recovery page's own auto-return already cover the recovered-on-its-own case.
- A separate longer remote STUCK-entry threshold: an earlier commit added one and it was reverted (`dd262625b`) in favor of gating at the recovery page itself, so the `SystemInterfaceHealthTracker` is unchanged.

## Testing
- New/extended unit tests: provider-unavailable raising + auth-error preservation (`mngr_imbue_cloud`); `errors[]` extraction, the two new tiers, and the timed-out-listing classification (recovery probe); recovery-page branches + the provider-unavailable Retry page (templates).
- Full `just test-offload` runs in CI.


joshalbrecht:
@gnguralnick I'm going to merge this because it is better than what we have, but FYI it is doing it the wrong way -- we just need to be a lot more careful about the states that we get back from mngr

It fundamentally doesn't make sense to poll provider health separately from host health. Instead, we need to have a single source of truth (mngr list) that appropriately separates between "this host cannot be reached due to network connectivity" and other states. Anything else risks race conditions and madness.

Does that make sense?

PR #2204, opened 2026-06-18
offline host_dir for gcp

evgunter:
medium level of actually reviewed

---

## Summary

- Mirrors the AWS S3 / Azure Blob shape for offline `host_dir`: a stopped GCE instance's `host_dir` is now readable via `mngr event` / `mngr file` etc., captured operator-side at `mngr stop` and uploaded to a Google Cloud Storage bucket.
- Host + agent records still live in GCE instance metadata (they fit comfortably and need no `prepare` step); only `host_dir` (whose size can exceed metadata limits) goes to GCS.
- `mngr gcp prepare` now creates the GCS bucket (`mngr-state-<project_id>` by default, configurable via `[providers.gcp] state_bucket_name`); `mngr gcp cleanup` deletes it (with a new `--force` flag matching AWS).
- The shared provider release harness's Trip 1 offline-host_dir step now runs against GCP too, asserting `mngr file get` serves the marker from the offline mirror without resuming the host. This step now runs **by default** for providers that support it (GCP/AWS/Azure), instead of only when `MNGR_RELEASE_TEST_OFFLINE_HOST_DIR=1`; set the env var to `0` to skip it.

New config fields on `GcpProviderConfig`: `state_bucket_name`, `is_offline_host_dir_enabled` (default on).

## Test plan

- [x] 959 cloud-provider unit tests pass (`libs/mngr_gcp` + `libs/mngr_aws` + `libs/mngr_azure` + `libs/mngr_vps`).
- [x] 14 new `state_bucket_test.py` unit tests for `GcsStateBucket` / `GcsVolume` (in-memory fake GCS via test-only subclasses, no `monkeypatch.setattr`).
- [x] 9 new `state_bucket_backend_test.py` tests for the `GcpProvider`'s `_state_bucket` / `_host_dir_backend` selection paths.
- [x] Updated cli/config tests for the new bucket fields and prepare/cleanup output shape.
- [x] Release test passes against real GCP: `MNGR_GCP_RELEASE_TESTS=1 MNGR_RELEASE_TEST_OFFLINE_HOST_DIR=1 uv run pytest --no-cov -n 0 -m release libs/mngr_gcp/imbue/mngr_gcp/test_release_gcp.py::test_provider_release_trip1` — `2 passed in 716.58s` (both `CONTAINER` and `NONE` isolation), exercising the offline `host_dir` read from the GCS mirror while the host stays stopped.


joshalbrecht:
lgtm

joshalbrecht:
merged to main

PR #2205, opened 2026-06-18
pool list: emit every pool_hosts column (fix missing region/backend_kind)

joshalbrecht:
## What

`mngr imbue_cloud admin pool list` (and the `minds pool list` wrapper) was emitting a hand-maintained 10-column subset of the `pool_hosts` table. It omitted:

- `region` (so every row showed `null` — the reported bug)
- `backend_kind` (couldn't tell a slice from an OVH VPS without `psql`)
- the slice identifiers `bare_metal_server_id`, `lima_instance_name`, `lima_disk_name`
- `host_name`, `vps_instance_id`, `ssh_user`, `ssh_port`, `container_ssh_port`

A baked slice host therefore looked like a region-less OVH VPS.

## Fix

Drive both the `SELECT` and the emitted JSON keys from a single `_POOL_HOST_LIST_COLUMNS` tuple (all 20 columns), so the two can never drift again. `emit_json`'s `default=str` already serialises the UUID/datetime values, so no per-column coercion is needed.

Added `test_pool_list_covers_every_pool_host_column`, mirroring the existing `test_pool_hosts_insert_has_required_columns` regression guard: it asserts the column tuple equals the full schema, so any future `pool_hosts` migration column must be added here too.

## Verification

- Unit tests: full `admin_test.py` + project ratchets — 80 passed.
- Live `minds pool list` against staging now shows all 20 columns, including `region: "US-WEST-OR"` and `backend_kind: "slice"`.


joshalbrecht:
lgtm

PR #2206, opened 2026-06-18
clean up provider args

evgunter:
## AWS
Removes the configurable `default_ami_by_region` field. `default_ami_id` now defaults to `None`; when unset, the pinned per-region default (`DEFAULT_AMI_BY_REGION`) is consulted for the chosen region. End behavior is identical -- only the configuration surface is simpler (one knob instead of two).

## GCP
`project_id` config field defaults to `None` instead of `""`, making "unset" explicit and matching the other optional identifier fields (`default_region`, `default_zone`, `service_account_email`). Resolution behavior is unchanged: an unset `project_id` still falls back to the project ADC resolves from the environment.

Docs, help text, and the VPS migration hint updated accordingly. Changelog entries added for `mngr_aws`, `mngr_gcp`, `mngr`, and `mngr_vps`.


joshalbrecht:
lgtm

joshalbrecht:
merged to main

PR #2208, opened 2026-06-18
original-claude-compatible __extend/__assign

evgunter:
i know i just reworked this, but i came up with this better solution to make it be a lighter change to your normal claude config

(the implementation is a bit messy but i think the core is good)

---

## Problem

`settings_overrides` is folded into a file the external AI CLI also reads (Claude Code's / antigravity's `settings.json`). mngr's internal merge operators are key **suffixes** (`key__extend` / `key__assign`), but the external CLI does not understand them and would treat `permissions__extend` as a junk literal key. So a user mixing mngr and vanilla Claude gets inconsistent behavior, and the operator surface mngr teaches doesn't transfer to real Claude files.

## Approach

On the `settings_overrides` path only, replace the suffix surface with a single top-level **`__mngr_merge`** map (dotted key path -> `"extend"` | `"assign"`) that vanilla Claude/agy silently ignore. mngr's own (non-`settings_overrides`) config is unchanged and still uses the suffixes.

```toml
[agent_types.claude.settings_overrides.permissions]
allow = ["Bash(npm *)"]
[agent_types.claude.settings_overrides.__mngr_merge]
"permissions.allow" = "extend"   # or "assign"
```

- **Desugar at config-load** (`key_resolver`): `__mngr_merge` rewrites to the internal suffix form (leaf takes the operator suffix; every ancestor takes `__extend` so the recursive merge reaches it), so the rest of the algebra is unchanged. Raw suffix keys / nested `__mngr_merge` are a hard error; directive paths/ops are validated.
- **Default stays strict**: a bare key assigns with the narrowing guard. The narrowing error now prints the **exact `__mngr_merge` patch** to get native-CLI (merge-across-layers) behavior.
- **Shared fold** (`overlay_merge.apply_settings_patch`): used by both `mngr_claude` and `mngr_antigravity`. Antigravity now gets parity (principled merge + narrowing guard instead of a wholesale `settings.update`). A stray `__mngr_merge` in the base settings file is stripped (no-op on the floor; the external CLI ignores it).
- **`mngr config extend`** on a `settings_overrides` path writes the bare value + a `__mngr_merge` `extend` directive instead of a `__extend` suffix.

Within-layer reset-then-add is intentionally dropped on this path (inexpressible in the clean JSON an external CLI reads).

## Scope / decisions (confirmed with @evan.gunter)

- `__mngr_merge` only for the Claude/antigravity `settings_overrides` path; general mngr config keeps the suffixes.
- Hard replace (no suffixes accepted on this path) — safe because the overlay system is unreleased.
- Full parity extended to antigravity.

## Tests

998 passing across `libs/mngr/imbue/mngr/config`, `cli/test_config.py`, the ratchets, and both plugins' suites; `ty` clean. Docs updated (config README + both plugin READMEs); changelog entries added for all three projects.


joshalbrecht:
I don't really love this, but I guess it works and is handy, so I'm willing to merge it

joshalbrecht:
merged to main

PR #2209, opened 2026-06-19
Fix mngr destroy/gc crash on docker build-image removal conflict

weishi-imbue:
## Problem

`mngr destroy <agent>` started failing with an uncaught Docker `409 Conflict` traceback (exit 1), **even though the requested agent was destroyed successfully**:

```
docker.errors.APIError: 409 ... conflict: unable to delete mngr-build-host-<id>:latest
  (must be forced) - container <id> is using its referenced image
```

`mngr destroy` runs: destroy agent -> destroy its host -> post-destroy GC sweep, and that sweep eventually calls `provider.delete_host` on aged-out, already-destroyed hosts. A single bad-state docker host (a leaked container still referencing the per-host build image) made `delete_host`'s build-image removal raise a raw `409`, which—not being an `MngrError`—bypassed GC's continue/aggregate path and aborted the whole sweep.

## Root cause

`delete_host` removed the build image while a leftover container still referenced it. Removing the image's only tag (`mngr-build-<host_id>`) deletes the underlying image, which Docker refuses with `409 (must be forced)` while a container references it. The bad state arises from an earlier `destroy_host` whose container removal failed, or a crashed/failed host, or hosts left over from testing.

## Fix

**`delete_host` ensures no trace of the host before forgetting it.** It re-runs `destroy_host` (idempotent) up front to clear any leftover container, then purges records. GC calls `delete_host` standalone on aged-out / `FAILED` / `CRASHED` hosts (it does *not* call `destroy_host` immediately before), so this is the last opportunity to clear runtime resources. Because the container is removed first, the build-image removal never hits the `409`. `_remove_build_image` force-removes the tag: mngr always relinquishes its `mngr-build-<host>` tag—deleting the image when nothing else references it, or just untagging it (leaving the image) when something does (e.g. a snapshot committed from the host's container, which shares its layers; verified against a live daemon that snapshot restore is unaffected).

**A failed cleanup keeps the host record for retry instead of orphaning the resource.** `delete_host` deletes the record (forgets the host) only when cleanup fully succeeds; if a resource is left behind it raises a `CleanupFailedGroup` and keeps the record, so the next sweep retries (a non-retryable failure stays sticky, but *visible* — surfaced as a leak with a cause-specific exit code — rather than a silent orphan). GC records the leak and continues, and counts the host as deleted / emits its `destroyed` event **only on a clean delete**.

This removes the previous exception-driven ladder (`_is_image_in_use_conflict`, a `force` parameter, nested try/except) in favor of the single `destroy_host` delegation.

## Review comments addressed

- **`delete_host` shouldn't forget a host whose cleanup failed** — the record deletion now lives outside the `collecting_cleanup_failures()` block, so a `CleanupFailedGroup` raised on exit skips it and the record is kept for retry.
- **GC shouldn't report a leaked host as destroyed/deleted** — `emit_host_destroyed` and `result.machines_deleted` are gated on a clean delete (the `else`); a leak goes only to `result.failures`.
- **Over-broad host-volume catch** — `remove_directory` is idempotent on a missing path (`DockerVolume`'s `rm -rf`, `LocalVolume`'s `is_dir()` guard), so an exception there never means "no volume"; it's recorded as a `HOST_RESOURCE_REMAINS` leak (the `OSError` / `MngrError` / `DockerException` cases — permission, `rm -rf` non-zero, exec-into-state-container failure — each verified against a live daemon).
- **Test teardown swallowing exceptions** — `make_docker_provider_with_cleanup` collects per-host `CleanupFailedGroup` leaks and `pytest.fail`s loudly after best-effort cleanup; "couldn't even attempt cleanup" conditions (daemon unavailable / resource-guard-blocked SDK) are tolerated.
- **Dead `except NotFound` catches** — removed the ones around `_remove_build_image` (its `images.list` guard makes "absent" a no-op); kept the ones around direct `container.remove` / `images.remove(snap.id)` / `stop_host` where `NotFound` is reachable or the no-`NotFound` behavior isn't contractual.

## Tests

- `_remove_build_image` force/no-op; `delete_host` clean-delete vs leak (stuck container, volume-removal failure) now asserting the record is **kept** on failure; GC offline-delete leak path asserts the host is **not** counted as deleted.
- Underlying Docker semantics (409 on a referenced single-tag image, force-untag leaving the image, unforced delete once the container is gone, snapshot independence, `remove_directory` idempotency, the three volume-error types) verified against a live daemon.


libs/mngr/imbue/mngr/providers/docker/instance.py
```diff
@@ -1544,33 +1552,63 @@ def destroy_host(self, host: HostInterface | HostId) -> None:
...
+                        )
 
-        if self.config.is_host_volume_created:
-            volume_id = self._volume_id_for_host(host_id)
+            if self.config.is_host_volume_created:
+                volume_id = self._volume_id_for_host(host_id)
+                try:
+                    self._state_volume.remove_directory(f"volumes/{volume_id}")
+                except (FileNotFoundError, OSError, MngrError) as e:
+                    logger.trace("No host volume to clean up for {}: {}", host_id, e)
```
joshalbrecht: skeptical -- that's way too broad of a set of errors to simply say "guess there was no volume"

there are lots of other things that can happen! And a lot of them don't mean "there is no volume"

libs/mngr/imbue/mngr/providers/docker/testing.py
```diff
@@ -210,13 +211,13 @@ def make_docker_provider_with_cleanup(
...
+            except (MngrError, CleanupFailedGroup, docker.errors.DockerException, OSError):
                 pass
             try:
                 provider.delete_host(provider.get_host(host.host_id))
-            except (MngrError, docker.errors.DockerException, OSError):
+            except (MngrError, CleanupFailedGroup, docker.errors.DockerException, OSError):
                 pass
-    except (MngrError, docker.errors.DockerException, OSError):
+    except (MngrError, CleanupFailedGroup, docker.errors.DockerException, OSError):
         pass
```
joshalbrecht: these are very suspicious! Why are we silently ignoring so many exceptions?

at the very least we ought to be logging warnings here, but really, this failure to clean up should probably fail the test suite very loudly with a "failed to properly clean up" message (*after* we've done our best to clean everything up)

libs/mngr/imbue/mngr/providers/docker/instance.py
```diff
@@ -802,12 +802,20 @@ def _remove_build_image(self, host_id: HostId) -> None:
...
+        its referenced image"). Forcing drops just the tag (all mngr needs);
+        the image lingers as a dangling layer, reclaimed once that container
+        is gone.
         """
         tag = self._build_image_tag(host_id)
         if not self._docker_client.images.list(name=tag):
             logger.trace("No build image to remove for host {}", host_id)
             return
-        self._docker_client.images.remove(tag)
+        self._docker_client.images.remove(tag, force=True)
```
joshalbrecht, line 880: this is half of the right solution

Instead, we need to update delete_host to:
1. call _remove_build_image normally
2. catch the exception you observed. If we see that exception, then have delete_host call destroy_host to ensure that the container is actually removed
3. then after ensuring the host was actually destroyed, delete_host can call this function again in a way that ends up calling remove with force=True

Does that make sense?  The docker provider's delete_host needs to ensure that, in fact, all traces of the host are gone

joshalbrecht:
lgtm after comments are addressed!

thanks for tracking down and fixing :)

libs/mngr/imbue/mngr/api/gc.py
```diff
@@ -362,7 +362,16 @@ def _gc_single_host(
...
+                        # records were deleted but a resource was left behind. Record
+                        # the leak and continue the sweep (the host record is gone, so
+                        # it still counts as deleted) rather than letting one host's
+                        # leak abort GC. Mirrors the online-destroy path below.
+                        try:
+                            provider.delete_host(host)
+                        except CleanupFailedGroup as group:
+                            with results_lock:
+                                result.failures.extend(group.failures)
                         emit_host_destroyed(
```
joshalbrecht: this needs to be in an "else", no?

we wouldn't want to "emit_host_destroyed" if, in fact, there was an exception!

libs/mngr/imbue/mngr/providers/docker/instance.py
```diff
@@ -1552,33 +1554,75 @@ def destroy_host(self, host: HostInterface | HostId) -> None:
...
+                        CleanupFailure(
+                            category=CleanupFailureCategory.HOST_RESOURCE_REMAINS,
+                            message=f"failed to remove host volume for host {host_id}: {e}",
+                            host_id=host_id,
+                        )
+                    )
+
+            self._host_store.delete_host_record(host_id)
+            self._container_cache_by_id.pop(host_id, None)
+            self._evict_cached_host(host_id)
```
joshalbrecht: these are supposed to happen *if we have successfully deleted the host*

but if there were failures... it's not clear that we should remove *our knowledge* of this host...  otherwise it cannot be retried in the case of a transient failure, misconfiguration, etc.

I think this block needs to be guarded by "if no errors AND if not forced"

(ie, we'll need a way to forcibly delete and bypass things that seem like they are stuck... otherwise )

whereas if there *are* errors, we probably should raise those errors... and actually, it doesn't look like we're doing anything with the "failures" var?

weishi-imbue: >  the "failures" var

`with collecting_cleanup_failures() as failures:` collects failure and then raise `CleanupFailedGroup` on exit. 
I've moved these 3 lines (delete_host_record etc.) to outside of the with block, so if there're failures, they're skipped.

>  these are supposed to happen if we have successfully deleted the host
I tend to agree. Just to note this would change existing behavior. 

Current pre-pr behavior: best effort clean up, try to delete whatever it can. (mngr thinks it's clean next run but it risks of leaking resources) 

I've updated the PR, now it changes to: treat any unknown error as failure, and abort on failure. i.e. keep the host record so we can retry next time. (now a non-retry-able failure would be sticky, but i think it's what we want.)

>  "if no errors AND if not forced"

gc already hardcoded the `--on-error continue` behavior. it's unclear what "force" should mean during cleanup. in my opinion, we should always force remove those tags that's tagged by mngr. and we should not "force" remove host record if error occurred destroying the underlying resources.

joshalbrecht: that all sounds right except for the very last bit -- the idea is that "force" would mean "delete our knowledge of this resource even when some resources were unable to be cleaned up"

this is important because otherwise you have no way to proceed (except manually) if there is a bug where, for example, you've already taken care of those resources, but mngr doesn't realize it

it's quite a corner case, and we don't need to change it right now -- I was just explaining why it might be needed

weishi-imbue: sync'ed offline, will need a tag (--force) to guard this behavior.

PR #2212, opened 2026-06-19
Avoid keychain dialogs popping up on Macs.

hynek-urban:
Make sure the encryption_key env var is set for all latchkey calls.


joshalbrecht:
lgtm

PR #2214, opened 2026-06-19
Do not use forks without exec

hynek-urban:
Using forkserver to get pre-warmed mngr processes didn't work (see [here](https://imbue-ai.slack.com/archives/C0ASLEFFD62/p1781744713531079)).

The next best thing is probably keeping a pre-warmed mgnr process and communicate with it over a socket which is what this PR does.

(It also adjusts the core process-related interfaces a little.)


joshalbrecht:
lgtm

PR #2215, opened 2026-06-19
imbue_cloud: realign restart tests with simplified start_host

joshalbrecht:
## Problem

CI on `main` started failing after a batch of recently merged provider PRs. Two `mngr_imbue_cloud` unit tests fail (in both `test-offload` and `test-docker`):

- `test_get_host_returns_online_host_when_container_running`
- `test_start_host_rebootstraps_container_ssh`

Both are cases of a merged PR changing production code while a test (added on another branch) was left describing the old behavior.

## Root causes & fixes

**1. Stale `docker inspect` stub.** `docker_inspect_running` now reads `{{.State.Status}}` and routes through the shared `is_running_container_state` rule (`== "running"`), but the test's `_StubOuter` still returned a boolean (`"true"`/`"false"`, the old `{{.State.Running}}` protocol), so a running container was misread as stopped and `get_host` returned an `OfflineHost`. The stub now returns a container status string.

**2. `start_host` no longer re-seeds keys / re-scans the host key.** Commit `ec6f9cf6` ("Remove changes that should be unnecessary") removed the authorized-keys re-seed and host-key re-scan from `start_host` -- correct, because a `docker stop`/`docker start` preserves the container filesystem, so only the sshd *process* (launched via `docker exec`, not the entrypoint) needs relaunching. But the test and `start_host`'s docstring still required the removed steps. Updated both to match.

## Verification

- `libs/mngr_imbue_cloud` package unit/integration tests: 379 passed.
- `instance_test.py`: 11 passed.
- `ty` type check + ruff: passed (pre-push hooks).


joshalbrecht:
lgtm

PR #2218, opened 2026-06-20
Auto-recover sshd + system-services after container/VM restart

joshalbrecht:
## Summary

Makes sshd (and the minds `system-services` agent) recover automatically after an out-of-band container/VM restart, so workspaces come back without a manual `mngr start` and even when the minds desktop app is closed.

Implements the plan in `blueprint/sshd-restart-robustness/`.

### mngr core / providers (this repo)
- **Self-healing sshd entrypoint** (`ssh_host_setup.py`, docker `instance.py`/`volume.py`, vps `container_setup.py`): the host-container PID-1 entrypoint now (re)starts sshd on every boot once a host key has been provisioned, so `docker restart`, a daemon restart, or a host reboot brings ssh back on its own. mngr's explicit sshd start is now idempotent via a procps-free `/proc` guard.
- **Concurrency-safe `mngr start`** (`hosts/host.py`, `interfaces/host.py`, `cli/start.py`): `Host.lock_for_starting()` holds a dedicated cross-actor `flock` (local `fcntl` + remote held-over-SSH paramiko channel, block-forever) so a desktop-driven start and an in-host boot-hook start can't race. Separate from the idle-suppression `host_lock`.
- **`post_host_create_outer_command` hook** (`interfaces/host.py`, `api/create.py`, `cli/create.py`, `config/data_types.py`): runs commands once on the host's *outer* machine (VM/daemon host) after create, via `host.outer_host()`; skips with a warning when there's no outer. Lets FCT config install a VM-level systemd unit declaratively.

### forever-claude-template (separate repo)
Committed in `.external_worktrees/forever-claude-template` on the matching branch (not part of this PR):
- `--restart=unless-stopped` on docker/pool_host/imbue_cloud agent containers.
- lima: in-VM systemd unit running `mngr start system-services` on boot.
- pool_host bake: outer-VM systemd unit (via the new hook) that starts the container and relaunches system-services inside it on boot.

## Testing
- `ty check` clean; ratchets pass; full `libs/mngr` + `libs/mngr_vps` non-infra suites pass locally (4552 tests). New unit tests cover the entrypoint, guarded sshd start, the remote start-lock command, `lock_for_starting` mutual exclusion, and the outer-command skip path.
- Infra/acceptance tests run in CI via offload.

## Caveats / needs verification on real infra
- The FCT boot hook's `mngr start system-services` (inside the VM/container) sources `/mngr/env` for `MNGR_*` context; this end-to-end path (esp. lima env resolution) needs verification on real lima/pool hosts, which aren't available in this environment.
- `vultr`/`aws` modes were intentionally left out of the boot-hook/restart-policy scope (agreed scope was docker + lima + imbue_cloud).


libs/mngr/imbue/mngr/api/create.py
```diff
@@ -433,14 +462,32 @@ def _run_post_host_create_commands(
...
+    """
+    if not commands:
+        return
+    with host.outer_host() as outer:
+        if outer is None:
+            logger.warning(
+                "Skipping {} post-host-create outer command(s): this provider exposes no outer host",
+                len(commands),
+            )
+            return
```
joshalbrecht: this should raise an error

libs/mngr/imbue/mngr/hosts/host.py
```diff
@@ -762,6 +804,64 @@ def lock_cooperatively(self, timeout_seconds: float = 300.0) -> Iterator[None]:
...
+        try:
+            with log_span("acquiring start lock at {} (over SSH)", lock_file_path):
+                channel.exec_command(_build_remote_start_lock_command(lock_file_path))
+                _wait_for_lock_acquired_marker(channel)
+            yield
+        finally:
+            try:
+                channel.shutdown_write()
+            except (OSError, EOFError):
+                pass
```
joshalbrecht: we probably want at least a logger.debug here

joshalbrecht:
lgtm

PR #2219, opened 2026-06-20
Migrate minds desktop client from FastAPI/asyncio to Flask

joshalbrecht:
Replaces the minds desktop client's FastAPI/uvicorn/asyncio web stack with a synchronous Flask app served by a graceful Werkzeug WSGI server. Strict-parity framework swap: same routes, status codes, headers, redirects, and SSE wire format.

## What changed
- **New modules**: `desktop_client/state.py` (typed `app.state` replacement + `get_state`), `responses.py` (Flask response factories), `server.py` (graceful Werkzeug server + `desktop_client_runtime` with one ordered shutdown sequence).
- **Converted off FastAPI/asyncio**: `app.py`, `api_v1`/`supertokens_routes` (Flask blueprints), `webdav` (direct WSGI mount, drops `a2wsgi`), `api_key_auth` (decorator), `request_handler` + latchkey grant/deny handlers (sync), `sharing_handler`, `cli/run.py` (serves via the new runtime).
- **SSE**: sync generators wrapped in `stream_with_context`; drop `is_disconnected` (rely on write-failure + shutdown flag).
- **Shutdown**: SIGINT/SIGTERM flips the shutdown flag and wakes SSE streams before the server drains, then closes the HTTP client, stops consumers, and drains the root concurrency group.
- **Deps**: remove `fastapi`, `uvicorn`, `a2wsgi`, `python-multipart`, `websockets`; add `flask`. Regenerate the electron `uv.lock`; delete the stale `demo_desktop_client.py`.
- **Tests**: all desktop-client/cli/telegram tests migrated to Flask's test client; new `server_test.py` for the runtime + SSE-shutdown path.

## Verification
- `just test-quick "apps/minds -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'"` -> 1393 passed, 6 skipped.
- `ty check imbue` -> clean. Ratchets pass.


joshalbrecht:
roughly lgtm

apps/minds/imbue/minds/desktop_client/webdav.py
```diff
@@ -14,87 +14,62 @@
...
+    the app without populating the state still see 401s rather than 500s.
     """
 
-    async def app(scope: Scope, receive: Receive, send: Send) -> None:
-        if scope["type"] != "http":
-            await _send_unauthorized(send)
-            return
-        authorization = _extract_authorization_header(scope)
+    def app(environ: WSGIEnvironment, start_response: StartResponse) -> Iterable[bytes]:
+        authorization = environ.get("HTTP_AUTHORIZATION")
```
hynek-urban, line 60: Where does this come from? I don't think we read anything like this from env vars previously; makes me a little suspicious of the whole refactor :) Originally, the `_AUTHORIZATION_HEADER` constant was used which seems to be gone now.

apps/minds/electron/main.js
```diff
@@ -2067,12 +2067,19 @@ async function promptMindShutdown() {
   return { proceed: true, stop: true, running };
 }
 
-function fetchInitialChromeState(timeoutMs = 4000) {
+function fetchInitialChromeState(timeoutMs = 10000) {
```
hynek-urban, line 2070: If the refactor comes with _"no user-visible behavior change"_ as claimed, why do we need longer timeouts? Are we sure this isn't a symptom of some sort of degradation?

joshalbrecht: this one was actually a very intentional bump from me to temporarily alleviate a race condition that I've known about for a while, and which I encountered during testing (and thus was able to root cause)

roughly, the issue is that if login state takes too long to come back, you would appear to be logged out. This is annoying.

I didn't bother fixing correctly though, because Gleb's new onboarding designs put that login usage later in the process.

apps/minds/imbue/minds/desktop_client/webdav.py
joshalbrecht: I double checked, and this is doing exactly the same thing.

The previous code accessed the value via:

```
_AUTHORIZATION_HEADER: Final[bytes] = b"authorization"

...

def _extract_authorization_header(scope: Scope) -> str | None:
    """Return the raw ``Authorization`` header value or ``None`` if absent."""
    for name, value in scope.get("headers", ()):
        if name == _AUTHORIZATION_HEADER:
```

The code you highlighted is just how that same thing is done in WSGI

hynek-urban: I presume we don't set this env var explicitly anywhere, though so it means it depends on some implicit default now?

hynek-urban: I guess it's fine but I just still find it strange :)

hynek-urban: Actually, ignore me, I get it now.

PR #2222, opened 2026-06-21
Share a bare-metal slice box across developer environments

joshalbrecht:
## Summary

Makes it safe for multiple developer environments (each with its own per-env Neon DB) to bake and run lima "slices" on the **same** bare-metal box. Previously, sharing a box was destructive: the per-env orphan reaper deleted any `mngr-slice-*` VM it didn't track, so envs sharing a box would delete each other's live slices; envs also miscounted capacity (overcommit) and could collide on host ports.

Implements the plan in `blueprint/share-bare-metal-across-dev-envs/`.

## What changed

- **Env-stamped slice names** (`mngr-slice-<env>-<host-hex>`): the owning env is baked into each slice's lima instance + disk names. Legacy un-stamped slices keep working and are never touched. (`--slice-env-name` on `admin pool create`, forwarded by `minds pool create`.)
- **Box-derived occupancy**: slice baking computes free slots from the box's real occupancy (every env's slices + legacy) instead of the per-env DB, so independent envs can't over-subscribe a box.
- **Brief-lock split carve**: each carve reserves its slot + host ports under a box-wide `flock` (creates the instance without booting via `limactl create`), then boots it after releasing the lock. Replaces the old per-invocation port-window partitioning, so concurrent cross-env bakes never collide on capacity or ports.
- **Env-scoped reaper**: the post-bake orphan reaper only deletes the active env's own stamped slices.
- **`minds env destroy` slice teardown**: tears down the env's unleased pool slices on their boxes before deleting the per-env DB (new `admin pool teardown-slices`). Leased slices still go through their agent's release path.
- **Connector reconcile**: the hourly cleanup cron now reconciles each box's slices against the DB, scoped to the deployment's own env (`MINDS_ENV_NAME`) — reaping its own untracked slices and alerting on rows whose VM vanished.

## Testing

- `libs/mngr_imbue_cloud`: 392 passed (unit + integration).
- `apps/minds` cli + `apps/remote_service_connector` app_test: 289 passed.
- Ratchets for all three projects: 168 passed (bumped the `mngr_imbue_cloud` lima-YAML ratchet 31→38 for the new, legitimate lima-YAML handling).
- Full offload suite runs in CI.


apps/minds/imbue/minds/cli/pool.py
```diff
@@ -198,6 +203,44 @@ def build_create_admin_args(
...
+    """
+    try:
+        pool_private_key = read_pool_private_key_from_vault(env_name)
+    except VaultReadError as exc:
+        logger.warning(
+            "No pool SSH key in Vault for env '{}' ({}); skipping slice teardown (env likely never baked slices).",
+            env_name,
+            exc,
+        )
+        return
```
joshalbrecht: this seems like a bad idea -- not having an SSH key shouldn't mean "we can't clean up", it should be an error that is raised and alerts us about a bad state

apps/remote_service_connector/imbue/remote_service_connector/app.py
```diff
@@ -2173,6 +2173,131 @@ def clean_up_slice_on_box(
...
+    divergence_count = 0
+    for server_id, public_address, lima_service_user in servers:
+        if not public_address:
+            continue
+        user = lima_service_user or "root"
+        try:
+            box_instances = _list_box_lima_names(public_address, user, management_key_pem, "list --json")
+        except (PoolHostCleanupError, paramiko.SSHException, OSError) as exc:
+            logger.warning("Slice reconcile skipped box %s (could not list lima resources): %s", public_address, exc)
+            continue
```
joshalbrecht: this doesn't make sense -- if we cannot list the lima resources, we should fail -- we certainly didn't reconcile anything!

apps/remote_service_connector/imbue/remote_service_connector/app.py
```diff
@@ -4106,7 +4231,26 @@ def cleanup_removing_pool_hosts() -> dict[str, int]:
...
+        # Audit this env's slices on every box against the DB (alert-only: it never
+        # auto-deletes, to avoid racing an in-flight bake). Scoped to MINDS_ENV_NAME so
+        # it is safe on a box shared by multiple dev envs. Best-effort: never fail the sweep.
+        try:
+            divergence_count = reconcile_slice_boxes(conn, _current_minds_env_name())
+        except (psycopg2.Error, paramiko.SSHException, OSError, KeyError) as exc:
+            # KeyError: POOL_SSH_PRIVATE_KEY unset on a deployment that does have
+            # boxes; still degrade to a warning rather than failing the whole sweep.
+            logger.warning("Slice box reconcile failed (continuing): %s", exc)
+            divergence_count = 0
```
joshalbrecht: what? No, this is also stupid. Failures should be failures, not silently swallowed!!!!

joshalbrecht:
lgtm

this was required to even come close to unbreaking the ability for developers to make instances in their own dev envs (while sharing our bare metal servers)

PR #2224, opened 2026-06-21
Handle unauthenticated providers consistently

joshalbrecht:
## Summary

Makes `mngr` handle unauthenticated/unavailable providers the same way everywhere, replacing the previous grab-bag (AWS/GCP/Azure raised verbose errors and exited 1, Vultr printed an ad-hoc `WARNING` and exited 0, OVH silently returned empty, and a full `mngr list` surfaced only whichever provider failed first).

### Error taxonomy
- `ProviderNotAuthorizedError` is now a subclass of `ProviderUnavailableError`, carrying structured `short_reason` / `short_remediation` fields plus verbose `user_help_text`. Every credentialed backend raises it for missing/invalid credentials.

### `mngr list`
- `--on-error abort` (default): fails immediately on the first provider error (now exits with the granular provider-inaccessible code `6` for auth/unavailable failures).
- `--on-error continue`: runs the rest of the listing, then reports every failed provider in one consistent block — a single line per provider on stderr for human/template output (`<provider>: <reason> — <remediation> (disable: ...)`), and structured entries (`exception_type`, `help_text`, `is_provider_inaccessible`) in the `errors` array for `--format json` / `jsonl`. Exits `6` when all errors are provider-inaccessible, else `1`.

### Per-provider behavior changes
- **Vultr / OVH**: now raise `ProviderNotAuthorizedError` when unauthenticated instead of silently returning zero agents (exit 0). The bespoke Vultr warning print is removed.
- **AWS / GCP / Azure**: raise the standardized error; credential/metadata (IMDS) resolution is bounded by a configurable `credential_timeout_seconds` (default 10s, on the shared cloud VPS config) so an unauthenticated provider fails fast instead of hanging. Azure now validates its credential eagerly (it previously checked only the subscription id).
- **Modal**: `ModalAuthError` becomes a `ProviderNotAuthorizedError` subclass (message preserved), so Modal auth failures are categorized consistently.
- Fixed OVH being omitted from the internal remote-backends list, so it is skipped in local-only contexts (e.g. tests) like the other cloud backends.

## Testing
- New unit tests: error categorization / structured fields (`ProviderErrorInfo`), exit-code selection, the concise error-line renderer, the `call_with_timeout` helper, and an end-to-end continue-mode listing with an unauthenticated provider.
- Ran the affected unit/integration suites locally (api/list, cli/list, errors, timeouts, registry, and all touched provider backends + their ratchets) — all green. The full suite runs in CI.
- Verified manually: `mngr list` now reports all five cloud providers consistently and exits 6; timing confirms providers still discover in parallel (full list ≈ slowest single provider, not the sum) with no stalls.

## Notes
- Release tests (not run in CI) were not executed (they require real cloud credentials); their existing assertions on the contract message and curated help commands remain satisfied by the preserved messages.


joshalbrecht:
lgtm

PR #2232, opened 2026-06-22
Pin all imbue_cloud SSH host keys (remove TOFU)

joshalbrecht:
## Summary

Removes trust-on-first-use (TOFU) across the imbue_cloud pool flow by pinning SSH host keys end-to-end. Implements `blueprint/pin-imbue-cloud-host-keys/plan-pin-imbue-cloud-host-keys.md`.

- **Plumb slice/VPS keys**: the bake surfaces the VM-root + container sshd host public keys via `mngr create --format json` (new `get_ssh_host_public_keys` provider method) and persists them on the connector's `pool_hosts` row. `/hosts/lease` returns both; the client pins them (fast path) / pins the locally-known rebuilt key (slow path) instead of scanning. `_scan_and_record_container_host_key` and `_scan_ssh_host_key` are gone.
- **OVH box host key**: generated by us and injected during OS reinstall via an inline base64 `postInstallationScript` (secret stays in the authenticated request body; ed25519-only), recorded on `bare_metal_servers`.
- **Pin everywhere**: connector lease/teardown SSH (no more `AutoAddPolicy`), admin box SSH (no more `StrictHostKeyChecking=no`), and the lima slice client (no more `accept-new`) all pin the recorded key.
- **Migration**: new nullable key columns (migration 011) + a one-shot `mngr imbue_cloud admin pool backfill-host-keys` keyscan command (the single sanctioned, migration-only TOFU). After it runs, leasing fails closed on any row missing a pinned key.

## Test plan
- Unit/integration: connector lease (keys returned, fail-closed on null), pin anti-clobber, OVH post-install script + reinstall injection, insert/select round-trips, lima client strict pinning. All green locally; `ty check` clean; ratchets pass.
- Full offload suite + live deploy/bake verification on a dev env in progress.


libs/mngr_imbue_cloud/imbue/mngr_imbue_cloud/cli/server.py
```diff
@@ -175,28 +176,56 @@ def _derive_public_key(private_key_path: Path) -> str:
...
+    """Yield ssh ``-o`` options pinning the box's host key when it is known.
+
+    The provision flow injects the box's host key at OS reinstall, so it pins
+    strictly here (no trust-on-first-use). The standalone ``prep`` command on a box
+    we did not reinstall has no key to pin yet and falls back to accept-new
+    (operator-driven first contact; the key is later captured by the keyscan
+    backfill or a re-provision).
+    """
+    if not box_host_public_key:
+        yield ["-o", "StrictHostKeyChecking=accept-new"]
```
joshalbrecht: hmmm... why is this necessary?

joshalbrecht: Good catch — it was the one remaining trust-on-first-use, and only the standalone `admin server prep` command hit it (it took `--server-address` and had no recorded key to pin; the main `server setup` flow already pins the injected key strictly). Rather than keep a TOFU fallback, I removed it: `prep` now takes `--server-id`, resolves the box's recorded sshd host key from `bare_metal_servers`, and pins it strictly — failing closed (pointing at `server setup` or the one-time `admin pool backfill-host-keys` keyscan) if no key is recorded. The `accept-new` branch is gone entirely, so no box SSH can fall back to TOFU. Done in 90ccb7667.

joshalbrecht:
lgtm

PR #2233, opened 2026-06-22
Force runc runtime for the local docker provider on macOS in minds

gnguralnick:
## Summary

On macOS, the minds desktop app now forces the local-docker provider's container runtime back to the Docker default (`runc`) when it spawns the Python backend.

The bundled forever-claude-template `[providers.docker]` block sets `docker_runtime = "runsc"` to harden agents with gVisor. runsc is Linux-only and is not present in Docker Desktop's VM on macOS, so `mngr create` failed immediately with `unknown or invalid runtime name: runsc`.

`apps/minds/electron/backend.js` now injects `MNGR__PROVIDERS__DOCKER__DOCKER_RUNTIME=runc` into the spawned backend's environment when `process.platform === 'darwin'`. The `mngr create` subprocess inherits this env. The override is skipped if the user has already set the variable themselves.

This is the same escape hatch that the local-docker e2e test (`test_desktop_client_e2e.py`) already sets for CI runners that lack runsc.

## Testing

- No JS lint/unit-test harness exists for the electron layer, so there is no automated coverage for `backend.js` (consistent with the existing pattern). The change is a guarded, platform-conditional env injection.


gnguralnick:
closing in favor of splitting into separate providers (docker and gvisor)

PR #2234, opened 2026-06-22
Sentry

hynek-urban:
Copy the Sentry setup from the old Sculptor and adjust for our purposes.

- There are three new Sentry projects (`minds-production`, `minds-staging` and `minds-dev`)
- Here's a test error that proves the integration works: https://generally-intelligent-e3.sentry.io/issues/7570133472
- Includes uploading logs to the existing `traceback-uploads-*` buckets (only for staging and production).
- Sentry events contain both the `git_sha` and `release`.
- For now, the Sentry integration is actually disabled by default. It needs to be enabled by setting these two env vars:
  - `MINDS_SENTRY_ENABLED=1`
  - `MINDS_SENTRY_S3_UPLOADS=1`
  - (This is a temporary solution - I expect Gabe to override this with a user-facing dialog / settings.)


apps/minds/imbue/minds/utils/sentry/core.py
```diff
@@ -0,0 +1,831 @@
...
+        # See this PR for more: https://gitlab.com/generally-intelligent/generally_intelligent/-/merge_requests/5789
+        #
+        # Questions to the above:
+        # - why are we not relying on the Sentry's logger for this?
+        # - won't the call to `logger.exception` itself try to send something to Sentry causing recursion?
+        # - the following message will likely hit an error inside Loguru handler because it is not allowed
+        #   to call emit from inside emit (that's what we're in here).
+        logger.opt(exception=e).error("Failure when processing event in before_send hook: {}", e)
+        # NOTE: this re-raise will get suppressed by Sentry and treated as if `before_send` returned `None`
+        raise
```
joshalbrecht, line 485: maybe worth slightly closer review?

apps/minds/imbue/minds/utils/sentry/core.py
```diff
@@ -0,0 +1,831 @@
...
+    # the number of initial reports to allow before starting to apply rate limiting
+    initial_reports_without_rate_limiting: int = 2
+    # the time (in seconds) that must pass since the last report of a given exception before allowing
+    # another report it is multiplied by the number of times the exception has been passed-thru since
+    # the app start after the first throttling event
+    timeout_factor: float = 60.0
+    # maximum number of different exceptions to track for rate limiting
+    # once this number is exceeded, all events will be passed through unfiltered
+    max_tracked_rate_limited_exceptions: int = 10_000
+
```
joshalbrecht, line 219: that's a little strange?

apps/minds/imbue/minds/utils/sentry/loguru_handler.py
```diff
@@ -0,0 +1,359 @@
...
+            "type": "log",
+            "level": self._logging_to_event_level(record),
+            "category": record.name,
+            "message": record.message,
+            "timestamp": datetime.fromtimestamp(record.created, timezone.utc),
+            "data": self._extra_from_record(record) if not self.strip_extra else {},
+        }
+
+
+def log_error_inside_sentry(
```
joshalbrecht, line 326: is this maybe what we were supposed to call from that other method I commented on above?

joshalbrecht:
lgtm!

comments can be fixed in post

apps/minds/imbue/minds/utils/sentry/loguru_handler.py
hynek-urban: That's probably half the solution (the other half is making sure the error appears in the log, too).

apps/minds/imbue/minds/utils/sentry/core.py
hynek-urban: What do you find strange about it? It seems like a protection against excessive memory usage of the exception history tracker.

apps/minds/imbue/minds/utils/sentry/loguru_handler.py
hynek-urban: Edit: turns out this function calls before_send hooks again (as part of the capture_event() call) so I ended up moving the protections here.

apps/minds/imbue/minds/utils/sentry/core.py
joshalbrecht: >     # once this number is exceeded, all events will be passed through unfiltered

I mean, wouldn't the normal way be to have a queue or something of the "most recent X errors"?  Maybe the comment is just misleading, but it made me think that nothing would be matched one that number was hit

PR #2236, opened 2026-06-22
minds: drive workspace recovery from passive discovery instead of a second `mngr list` sampler

gnguralnick:
## Summary

Workspace recovery used to fire its own synchronous `mngr list` on every recovery-page load -- a second, independently-clocked sampler of host lifecycle and provider reachability, separate from the passive discovery that already feeds the workspace list and providers panel. Two samplers of the same state can disagree, so the recovery page could contradict the sidebar. This collapses that state onto the single passive sampler and revises the post-outage handling so the recovery page only ever classifies from trustworthy state. It also closes a websocket-only path that could strand a workspace whose backend died, and lifts an `mngr` constant so the freshness threshold can't drift from the discovery cadence.

### minds: recovery reads from passive discovery

- The recovery host-health probe no longer runs `mngr list`. Host lifecycle state, the system-services agent, and provider reachability are read from `backend_resolver` (`get_host_state` / `get_system_services_agent_id` / `get_provider_errors`) plus discovery freshness. The host-health endpoint is now effectively instant (a resolver read, no network round-trip), and a dropped connection can no longer strand the diagnostic on a slow probe.
- The in-container `mngr exec` probe (Layer B -- *why* the inner interface isn't answering) is retained but fires **only when Layer A is healthy** (provider reachable + host RUNNING), with an explicit 30s cap, so an outage never pays a doomed provider round-trip.

### minds: gate the recovery redirect on a post-outage snapshot

Rather than the earlier transient recovery tier plus client-side convergence loop, the redirect to the recovery page is now gated on a discovery snapshot taken *after* the outage began. In the brief window right after a drop, the latest snapshot can still predate the outage and report a just-stopped container as RUNNING; acting on it would misclassify the tier (e.g. asking the user to confirm a restart for a container that is already stopped). Minds keeps you on the auto-refreshing "Loading workspace" loader until a post-outage snapshot lands (typically within one discovery poll), then routes straight to the right place: a stopped container's restart is auto-dispatched with nothing to confirm, a genuinely-wedged-but-running workspace still asks before a disruptive host restart, and a down provider routes to the "Can't connect to ..." page (which reconnects automatically once it recovers).

### minds: one "Can't connect to ..." page

The two nearly identical backend-unreachable tiers/pages (`PROVIDER_UNAVAILABLE` / `WORKSPACE_UNREACHABLE`) collapse into a single `BACKEND_UNREACHABLE` tier and page, covering a down provider connector, a stopped/paused local Docker daemon, or an expired login. The old split could route the same "Docker isn't available" situation to either page and left one of them unable to reconnect on its own. The page surfaces the provider's own error verbatim (e.g. Docker's "Docker Desktop is manually paused. Unpause it through the Whale menu or Dashboard.") rather than a minds-authored per-provider message, offers a Retry, and always returns you to the workspace once the backend recovers.

### mngr_forward: emit the backend-failure envelope on websocket failures

The websocket forward path now emits a `system_interface_backend_failure` envelope when the backend connection fails (unresolved target, SSH-tunnel setup failure, refused host-loopback dial, or a connect-time backend-websocket failure), matching what the HTTP/SSE paths already did. Previously only HTTP failures emitted it, so an already-loaded SPA whose backend died over its only live (websocket) channel would silently retry forever -- never enrolling the agent as a recovery-probe suspect, so the recovery redirect never fired and the user was stranded on a frozen workspace.

### mngr: public discovery-cadence constant

Exposed `DISCOVERY_STREAM_POLL_INTERVAL_SECONDS` (in `imbue.mngr.api.discovery_events`), previously a private `_`-prefixed value. Minds derives its discovery-freshness threshold from the cadence, so sharing one constant keeps the two from silently drifting apart.

`get_freshness_timestamps` is added to the `BackendResolverInterface` ABC (None default) so the resolver-sourced read is uniform with the other Layer-A accessors. Removed as dead code: the `mngr list` JSON-parsing/argv helpers and the old per-provider provider-error message builders.

Design doc: `blueprint/remote-mind-recovery/plan-remote-mind-recovery.md`.

## Test plan

Relevant suites to run (gated by the full offload suite on CI):

- [ ] `apps/minds/imbue/minds/desktop_client/recovery_probe_test.py`
- [ ] `apps/minds/imbue/minds/desktop_client/test_desktop_client.py`
- [ ] `apps/minds/imbue/minds/desktop_client/templates_test.py`
- [ ] `apps/minds/imbue/minds/desktop_client/forward_cli_test.py`
- [ ] `apps/minds/imbue/minds/desktop_client/system_interface_health_test.py`
- [ ] `libs/mngr_forward/imbue/mngr_forward/server_test.py`
- [ ] Full offload suite green on CI


apps/minds/imbue/minds/desktop_client/app.py
```diff
@@ -2526,13 +2527,23 @@ def _build_requests_payload(
...
-# "can't reach the provider" in tens of seconds instead of stranding the user on
-# the "Loading workspace" loader for the full restart timeout. A timeout here is
-# itself classified as provider-unavailable (see _run_host_health_probe).
+# Hard timeout for the recovery host-health probe's in-container ``mngr exec``.
+# Far shorter than the restart ceiling: this is a *diagnostic* that gates the
+# recovery UI. The exec touches the provider (``get_host`` -> the connector's
+# ~30s httpx) before reaching the container, so it must carry its own 30s-class
+# cap rather than inheriting the 120s restart default -- otherwise a wedged
+# host could gate the recovery UI for the full restart timeout. The probe is
+# only fired when Layer A is already healthy (provider reachable + host
```
gnguralnick: "Layer A" is meaningless here

apps/minds/imbue/minds/desktop_client/app.py
```diff
@@ -2526,13 +2527,23 @@ def _build_requests_payload(
...
+# RUNNING), so it never stacks a doomed round-trip during an outage.
 _HOST_HEALTH_PROBE_TIMEOUT_SECONDS: Final[float] = 30.0
+# How recent the last full discovery snapshot must be for the resolver's
+# Layer-A view to positively establish provider reachability. Discovery polls
+# every ~10s; a healthy poll emits a snapshot each cycle, so a snapshot older
+# than this means discovery has stalled (the post-outage window) and the host
+# state it last reported can no longer be trusted to offer a destructive
+# restart. Comfortably above the normal inter-snapshot interval to avoid
```
gnguralnick: is there  a way to just derive this based on whatever constant controls how often discovery polls?

apps/minds/imbue/minds/desktop_client/app.py
```diff
@@ -2710,8 +2684,12 @@ def _run_mngr(concurrency_group: ConcurrencyGroup, argv: list[str], env: dict[st
     ``except MngrCommandError`` callers are unaffected), a nonzero exit as a bare
     ``MngrCommandError`` (discarding stdout, unlike ``_run_mngr_capturing``), and
     a launch failure as a bare ``MngrCommandError``.
+
+    ``timeout_seconds`` defaults to the generous restart ceiling; the recovery
+    host-health exec probe overrides it with a much shorter cap since it must
```
gnguralnick: if the two uses of this function are restart and host-health probe  and they use different values maybe the parameter just shouldn't have a default value?

apps/minds/imbue/minds/desktop_client/app.py
```diff
@@ -3219,15 +3197,53 @@ def _handle_host_health_probe_api(
...
     concurrency_group: ConcurrencyGroup,
 ) -> HostHealthResponse:
-    """Run the batched ``mngr exec`` probe + ``mngr list`` lookup, return the response.
-
-    Composes the response from three independent inputs: ``mngr list`` for
-    host state / services-agent state, the batched in-container
-    ``mngr exec`` probe, and the plugin's resolver-snapshot mirror.
+    """Compose the host-health response from the passive resolver + an in-container probe.
+
+    Layer A (provider reachability + host lifecycle) is read from the
```
gnguralnick: just say "provider reachability and host lifecycle" - no need to say "Layer A". simlar for where you mention layer b

apps/minds/imbue/minds/desktop_client/recovery_probe.py
```diff
@@ -711,55 +611,57 @@ def _classify_dispatch_tier(probes: tuple[Probe, ...], provider_error: ProviderP
...
+    ``provider_error`` is this workspace's provider-level error read from the
+    resolver's ``get_provider_errors()``; when present it drives the
+    PROVIDER_UNAVAILABLE / WORKSPACE_UNREACHABLE tiers and its message is carried
+    on the response as ``unreachable_reason``. ``provider_label`` is the friendly
+    provider name for that page's title.
+
+    ``reachability_confirmed`` is whether the resolver's Layer-A view is fresh
+    enough (no provider error + recent discovery snapshot) to positively
+    establish provider reachability. When false, the would-be destructive
+    HOST_UNRESPONSIVE classification falls to REACHABILITY_UNCONFIRMED instead.
```
gnguralnick: i think there might not actually be any need to have REACHABILITY_UNCONFIRMED be its own state. if we aren't sure we can reach the provider we should just say PROVIDER_UNAVAILABLE. it seems the functional impact of REACHABILITY_UNCONFIRMED on the template-side is to start looping to await updated discovery state - we should just do this for PROVIDER_UNAVAILABLE as well in case we make it through in the next round of polling.

joshalbrecht:
roughly lgtm, but it does feel a little brittle still for some reason, hard to put my finger on exactly...

PR #2237, opened 2026-06-22
minds: choose Docker container runtime (runc/runsc) per create instead of hardcoding gVisor

gnguralnick:
## What

The minds Docker compute provider always ran under gVisor (`runsc`), which fails on macOS where gVisor is unavailable -- forcing the `MNGR__PROVIDERS__DOCKER__DOCKER_RUNTIME=runc` workaround. This makes the container runtime a per-create choice, defaulted per platform:

- `DockerRuntime` enum (RUNC/RUNSC) + `default_docker_runtime()` (runc on macOS, runsc on Linux) in `apps/minds/imbue/minds/primitives.py`.
- An advanced "Container runtime" select on the create form (`Create.jinja` + `templates.py`), shown only for the Docker compute provider and pre-selected to the platform default.
- `app.py`: the two create handlers (form POST + JSON API) parse/validate the `runtime` field, defaulting to the platform value.
- `agent_creator.py`: threads `docker_runtime` through the create chain (`start_creation` -> `_create_agent_background` -> `_MngrCreateAttemptParams` -> `_attempt_mngr_create` -> `run_mngr_create` -> `_build_mngr_create_command`); the DOCKER case stacks `--template docker_runsc` only when RUNSC is chosen (RUNC is the docker template's default, so it adds nothing).

macOS users no longer need the env-var workaround; Linux keeps the hardened runsc default and either can be overridden per workspace.

## Pairs with / depends on

**imbue-ai/forever-claude-template#196** (FCT branch `gabriel/macos-gvisor-issue`), which adds the `[create_templates.docker_runsc]` overlay and removes the hardcoded `docker_runtime = "runsc"` from `[providers.docker]`. This change is inert -- in fact broken on the default Linux path -- without that template.

**This is a hard release dependency, not just a "land together":** the create form and `test_desktop_client_e2e.py` resolve the workspace template against the pinned FCT tag (`FALLBACK_BRANCH = "minds-v0.3.3"` in `templates.py`), which does **not** contain `docker_runsc`. On Linux, `default_docker_runtime()` returns RUNSC, so every default create runs `mngr create ... --template docker --template docker_runsc` and fails with `Error: Template 'docker_runsc' not found`. Merging FCT#196 alone does not fix this -- a new FCT tag containing the overlay must be cut and `FALLBACK_BRANCH` bumped to it.

Until that happens, the `test-docker-electron` CI job (and any real Linux create on the default runtime) will fail with the missing-template error. This is a pre-existing condition on the branch, not introduced by the latest `origin/main` merge.

## Test plan

- [x] `agent_creator_test.py` + `templates_test.py`: 219 passed (`just test-quick`), incl. new tests for runsc-overlay stacking/ordering, the runc no-overlay case, non-docker modes ignoring the runtime, and form rendering/defaulting.
- [x] `test_desktop_client.py`: 151 passed. minds ratchets: 56 passed.
- [x] `ty` type check: clean on all changed files.
- [x] Manually rendered the create form: runtime select with runc/runsc, platform default pre-selected, show/hide JS wired for Docker-only.
- [ ] Full suite green in CI. **Blocked**: `test-docker-electron` fails with `Template 'docker_runsc' not found` until FCT#196 ships in a tagged release and `FALLBACK_BRANCH` is bumped (see "Pairs with / depends on" above). The ruff-format check on `templates_test.py` is fixed in this branch.


joshalbrecht:
lgtm, will merge together

joshalbrecht:
actually, I had to run before I got to this -- could you fix the conflicts and then merge this plus the paired one?

gnguralnick:
merged FCT - it should get tagged when weishi makes the next one, after which i'll merge this one

PR #2239, opened 2026-06-22
just minds-start: add optional `fct` arg to override the FCT worktree

weishi-imbue:
## What

Adds an optional positional `fct` argument to the `just minds-start` recipe.

`minds-start` previously hardcoded the forever-claude-template worktree at `.external_worktrees/forever-claude-template`. This lets a launch target a different FCT worktree/branch — the directory that gets synced into `vendor/mngr/` and exported as `MINDS_WORKSPACE_GIT_URL`:

```
just minds-start fct=.external_worktrees/my-fct-worktree
just minds-start fct=/abs/path/to/fct
```

An absolute path is used as-is; a relative one is resolved against the mngr root. Omitting `fct` preserves the previous default, so existing invocations are unchanged.

## Notes

- `dev/`-only change (root-level `justfile`), with a `dev/changelog` entry.
- Split out of the docker GC-crash branch (#2209) to keep that PR focused.


joshalbrecht:
lgtm

PR #2244, opened 2026-06-23
Avoid logger.warning() in Minds.

hynek-urban:
- Convert all `logger.warning()` calls in `apps/minds` either to `logger.info()` or to `logger.error()`
- Introduce a ratchet to prevent further `logger.warning()` calls.

The judgement for the individual calls (whether they should be info() or error()) was done by Claude. I went through the changes and they seemed more or less reasonable to me. I think if any of them is misclassified, we'll eventually find out and will change it when that happens.


apps/minds/imbue/minds/cli/env.py
```diff
@@ -415,7 +415,7 @@ def _load_dev_credentials_from_vault(vault_prefix: str, *, cg: ConcurrencyGroup)
     try:
         ovh_secret = read_vault_kv(VaultPath(f"{vault_prefix}/ovh"), parent_concurrency_group=cg)
     except VaultSecretNotFoundError as exc:
-        logger.warning("No ovh Vault entry at {}/ovh ({}); proceeding with empty OVH credentials.", vault_prefix, exc)
+        logger.info("No ovh Vault entry at {}/ovh ({}); proceeding with empty OVH credentials.", vault_prefix, exc)
```
joshalbrecht, line 418: should be an exception (is part of dev CLI commands, better to crash than fail silently)

same deal as the below code, which raises errors when keys are missing

hynek-urban: But the comment above clearly says that this entry is optional and the exception should be caught?

hynek-urban:
Closing as obsolete - we'll do it again, possibly a bit differently.

PR #2245, opened 2026-06-23
Add just minds-start-cloud recipe for imbue_cloud slice testing

joshalbrecht:
## Summary

Adds a `just minds-start-cloud` recipe to launch the minds desktop client (Electron) in dev mode for testing the `imbue_cloud` provider against pre-baked pool slices.

Unlike `just minds-start` (which targets local Lima/Docker dev and forces `MINDS_USE_LOCAL_WORKSPACE_DEFAULTS=1` + `MINDS_WORKSPACE_*`), `minds-start-cloud` leaves the create form's shipped fallbacks in place -- the canonical `forever-claude-template` remote plus `FALLBACK_BRANCH` (the `minds-v<version>` release tag baked into production slices). Those shipped defaults are exactly the identity stamped into a slice by `just bake-slice-prod <region> minds-v<version>`, so an `imbue_cloud` create gets a fast-path lease of that slice instead of falling back to the slow rebuild path.

It also skips the live-mngr -> `vendor/mngr/` rsync that `minds-start` does on every launch, since a leased pool slice already carries the mngr snapshot baked at its tag.

## Context

Came out of an operational task: deploy `dev-josh-1`, bake a `minds-v0.3.2` slice, then launch the app to test the slice via the `imbue_cloud` provider. The existing `minds-start` recipe couldn't be used as-is because its local-worktree defaults break the imbue_cloud fast-path match.


joshalbrecht:
lgtm

PR #2246, opened 2026-06-23
Add Latchkey 2.18.0 & Google OAuth to Minds

pseay-imbue:
## Summary

Adds a **Minds-owned Google OAuth fallback**: when a mind needs a Google API and its credentials aren't valid, the user authenticates against the Minds consent screen first, and only sees the old "create your own Google project" self-setup flow if that fails. Also bumps the bundled/installed Latchkey CLI to **2.18.0** (which introduces the `auth prepare` subcommand this depends on) and raises the compatibility floor to match.

Verified end-to-end in a local dev env: Gmail authenticates via the Minds consent screen; a service the Minds client can't cover falls through to self-setup.

## The Google OAuth fallback

For a `google-` service whose credentials are **not valid** (and that supports browser sign-in), `grant()` runs a three-step chain (`apps/minds/.../latchkey/handlers/predefined.py`, via `_authenticate_google`):

1. **Current credentials** — if valid, grant immediately (unchanged).
2. **Minds Google OAuth** — if no client is registered, register the Minds OAuth client (`latchkey auth prepare`) and open the **Minds consent screen** (`latchkey auth browser`). If a client is already registered, reuse it.
3. **Self-setup fallback** — on *any* failure of step 2, **clear the registered client** (`latchkey auth clear`) and fall through to `auth browser-prepare` + `auth browser` (the user provisions their own Google project).

- Scope is gated to `google-` service names; **every other service keeps today's behavior**.
- New thin `Latchkey` primitives in `mngr_latchkey/core.py`: `auth_list`, `auth_prepare`, `auth_browser_login`, `auth_clear` (and `auth_browser` refactored to reuse the bare login).
- The dialog is unchanged (single "authenticating… opening a browser" notice).

### Important fix: don't get stuck on the Minds client

`auth prepare` *persists* the Minds client, and `auth browser` only triggers `browser-prepare` (self-setup) when **no** client is registered. So a failed Minds attempt would otherwise leave the Minds client registered and the fallback would keep re-using it (failing) instead of letting the user self-set-up. Step 3 clears the registered client first, so self-setup actually engages — and this self-heals a client left stuck by a prior failed attempt.

## Latchkey 2.18.0

- Bundled in the Electron app (`package.json` + `pnpm-lock.yaml`).
- Installed on remote VPS gateways (`remote_gateway.LATCHKEY_VERSION`).
- `LATCHKEY_MIN_VERSION` raised `2.17.1` → `2.18.0` (first release with `auth prepare`).

## Credentials & follow-ups (infra, not code)

- The Minds OAuth client id/secret are read from the `MINDS_GOOGLE_OAUTH_CLIENT_ID` / `MINDS_GOOGLE_OAUTH_CLIENT_SECRET` **environment variables** (kept out of source). **Production wiring is still required**: these must be injected into the packaged/deployed app's launch environment, or the production app reads empty creds and falls through to self-setup.
- The real Minds Google Cloud OAuth client must have these **7 APIs** enabled + matching consent-screen scopes: Gmail, Calendar, Drive, Docs, Sheets, People, and Analytics (Data + Admin). `google-directions` (Routes API) is API-key-based, not OAuth, so it isn't on the client and correctly never hits the Minds consent flow.

## Testing

- Unit tests for the new `Latchkey` primitives (`core_test.py`, fake `latchkey` binaries).
- Unit tests for the `grant()` 1→2→3 ordering across all branches via an extended `FakeLatchkey` (records the call sequence), including a regression test asserting `auth_clear` runs before the self-setup `auth_browser`.
- One acceptance test (`test_google_oauth_fallback.py`) driving the three paths end-to-end, asserting call order and the on-disk `latchkey_permissions.json`.
- Local: `mngr_latchkey` 353 passed, minds latchkey handlers 58 passed, ratchets pass, `ty`/`ruff` clean. Full offload suite runs in CI.

## Notes

- A blueprint plan for the feature is included at `blueprint/minds-google-oauth-fallback/`.
- The Minds-specific constants/prefix live in `mngr_latchkey/core.py` (reviewed layering choice); a catalog-driven `managed_oauth_client` field is the natural refactor if a second managed provider is ever added.
- `run.py` / `app.py` deltas in this diff are main's own fixes that rode in on a merge of `main`, not part of this feature.


apps/minds/package.json
```diff
@@ -18,7 +18,7 @@
   },
   "dependencies": {
     "@todesktop/runtime": "^1.6.0",
-    "latchkey": "^2.17.2"
+    "latchkey": "^2.18.0"
```
hynek-urban: In the meantime, I bumped this to 2.19.x (already merged to main) :)

apps/minds/imbue/minds/desktop_client/latchkey/handlers/predefined.py
```diff
@@ -404,6 +410,61 @@ def grant(
...
+        it (failing) instead of letting the user self-set-up. The clear also
+        self-heals a client left stuck by a prior failed attempt, and is a
+        harmless no-op when nothing is registered.
+
+        ``auth_list`` read failures degrade to "not registered" (see
+        :meth:`Latchkey.auth_list`), so the Minds attempt is still made.
+
+        Returns ``(is_success, detail)`` like :meth:`Latchkey.auth_browser`.
+        """
+        is_already_registered = service_name in self.latchkey.auth_list()
```
hynek-urban: I understand why it's used here (there's no obvious way of telling whether a client ID is already present in the credentials store) but `latchkey auth list` is actually a bit problematic. It sends requests to all services where credentials exist and can often be very slow (because some of the services, e.g. Ramp :), are slow to respond).

Instead, could we do the same thing that was in place for the "normal" `auth browser` call, i.e. optimistically trying to `auth browser` and only if it fails with the error of "run latchkey browser-prepare first", we will know we still need to inject the client details? The logic for that already exists in `mngr_latchkey/core.py`.

Relatedly, would it make sense to encapsulate the logic (including the _authenticate_google() part) inside `auth_browser()` in mngr_latchkey? That way we wouldn't need to define the flows in two different modules.

apps/minds/imbue/minds/desktop_client/latchkey/handlers/predefined.py
```diff
@@ -404,6 +410,61 @@ def grant(
...
+            is_success, detail = (False, "")
+            if is_prepared:
+                is_success, detail = self.latchkey.auth_browser_login(service_name)
+        if is_success:
+            return True, detail
+        # The fast attempt failed. Clear any client registered for the service
+        # (the Minds client we just prepared, or a stale one) so the self-setup
+        # ``auth_browser`` triggers ``browser-prepare`` from a clean slate
+        # instead of re-using the failing client.
+        self.latchkey.auth_clear(service_name)
```
hynek-urban: I'm a bit afraid that we would clear self-created OAuth clients which is not desired because there's a limited number of those that can be created per project.

Could we only clear() right after we injected the pre-registered OAuth client details and failed the subsequent `auth browser` call? (That way we would only clear the details that are easy to re-insert because we know them.)

hynek-urban (commented):
Looks great! I guess we'll eventually want to have a different way of detecting whether we want to enter the fallback but I think this is fine as the first version :)

I just left you a few comments alongside the changes.

apps/minds/imbue/minds/desktop_client/latchkey/handlers/predefined.py
pseay-imbue: Good idea, I'll move it there!

apps/minds/imbue/minds/desktop_client/latchkey/handlers/predefined.py
pseay-imbue: Yes, good catch.

pseay-imbue:
Thanks for the feedback @hynek-urban! I've moved the logic into mngr_latchkey so it's all together and then I don't use list anymore - I just try the browser auth and input the Minds client upon failure (i.e., indication there is no setup client). I also made it only clear the Minds client credentials, not those of another path.

I just manually tested it and it works as expected :)

I suppose one issue now is what the user should do if there is a client they don't want to use, but that probably won't happen (if they have preregistered a client, it should be good; and if they haven't, they'll be setup with Minds or one created on their google account) and isn't highly relevant to this PR.

hynek-urban:
@pseay-imbue Looks and works great!

As for changing the client, we want to add more controls for the user to be able to see his current logins and log out in general; I think this will be connected.

Great work :)

libs/mngr_latchkey/imbue/mngr_latchkey/core.py
```diff
@@ -171,6 +173,33 @@ class CredentialStatus(UpperCaseStrEnum):
...
+)
+
+# Minds-provided Google OAuth client, registered for a ``google-*`` service via
+# ``latchkey auth prepare`` so the user signs in against the Minds consent
+# screen instead of self-provisioning their own Google Cloud project. A single
+# pair is reused for every google service. This is an installed/desktop-app
+# OAuth client, so the "secret" is not truly confidential -- it ships inside
+# the distributed client.
+MINDS_GOOGLE_OAUTH_CLIENT_ID: Final[str] = "991889009876-ms5ln5jnvqmsrgpmi2nipkv7atmoaks8.apps.googleusercontent.com"
+MINDS_GOOGLE_OAUTH_CLIENT_SECRET: Final[str] = "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"
```
weishi-imbue, line 201: is this intended? this conflicts with the PR description of 

> The Minds OAuth client id/secret are read from the MINDS_GOOGLE_OAUTH_CLIENT_ID / MINDS_GOOGLE_OAUTH_CLIENT_SECRET environment variables (kept out of source).

hynek-urban: Yes, it is. The PR description is obsolete.

weishi-imbue: sync'ed offline with Hynek. PR description is stale.

PR #2247, opened 2026-06-23
docker: report a running container with a dead inner sshd as UNAUTHENTICATED, not CRASHED

gnguralnick:
## What

When a Docker-backed workspace's container is still `running` but its inner sshd has died (e.g. you `pkill sshd` inside it, surfacing as "Error reading SSH protocol banner"), `mngr list` / `mngr forward` reported `host.state = CRASHED`.

The chain: agent enumeration SSHes into the host and raises `HostConnectionError`, so `get_host_and_agent_details` falls back to the offline view. The generic offline-state derivation treats a shutdown-capable provider with no recorded clean stop as `CRASHED` — even though the Docker daemon (the ground truth for container lifecycle, reachable without inner SSH) still reports the container as `running`.

Downstream, minds' recovery page reads this as fully offline (`dispatch_tier=host_offline`), auto-dispatches `restart-host?host_already_stopped=1`, skips the stop step, and the start step's `mngr start` then fails because it SSHes into a still-running-but-sshd-dead container.

An identical sshd-kill on an `imbue_cloud` workspace already classifies correctly (`host_unresponsive`), because imbue_cloud reads container state out-of-band and reports `UNAUTHENTICATED` for "container running but inner SSH unreachable".

## Fix

Mirror the imbue_cloud behavior for Docker, at the lowest-coupling seam:

- Add a concrete, default-`None` provider hook `get_connection_error_fallback_state(host_id)` on `ProviderInstanceInterface`. The `HostConnectionError` fallback in `get_host_and_agent_details` consults it (only for non-authentication failures) and, when it returns a state, applies it to `host_details.state` before building the agents' details — so each agent's nested `host.state` inherits it too.
- Override the hook in `DockerProviderInstance` to return `HostState.UNAUTHENTICATED` when the Docker daemon reports the container as `running`, else `None`.

The generic `derive_offline_host_state` and every other provider's behavior are unchanged (the hook defaults to `None`), and `discover_hosts` (the GC-critical path that must never swallow failures into an empty list) is untouched. With this, the same Docker sshd-kill classifies as `HOST_UNRESPONSIVE` in minds: the user gets the manual restart button (consent), whose stop step is not skipped, so the container is actually stopped and then started successfully.

## Note on the `UNAUTHENTICATED` state name

We are intentionally reusing the existing `HostState.UNAUTHENTICATED` here rather than introducing a new state. `UNAUTHENTICATED`'s literal meaning is "we reached sshd but auth failed", so for "container running but inner sshd dead" it is a semantic stretch. This is a *pre-existing convention*: `mngr_imbue_cloud` already maps a running-but-inner-SSH-unreachable container to `UNAUTHENTICATED` (introduced 2026-05-10 in `5e1b6bfd1`, "Use outer SSH to determine real container state for unreachable hosts"; later moved into `providers/listing.py`). This PR follows that convention so the two providers report the same state for the same condition. A more accurately-named state (e.g. `UNREACHABLE`) would be cleaner but is a broader vocabulary change spanning imbue_cloud, `mngr_wait.TERMINAL_HOST_STATES`, the `mngr wait` CLI help, and docs — out of scope here. Flagging for a follow-up discussion with Josh.

## Verification

TDD: a failing test was written first (asserting the fallback reports a non-offline state, never `CRASHED`), confirmed red (`assert <HostState.CRASHED> == <HostState.UNAUTHENTICATED>`), then made green by the fix.

- New tests:
  - `provider_instance_test.py`: the connection-error fallback applies a provider's state override to both the host and every agent's nested host; without an override it preserves the default offline-derived `CRASHED`.
  - `docker/instance_test.py`: `get_connection_error_fallback_state` returns `UNAUTHENTICATED` for a running container (fake daemon), `None` for a stopped/absent one.
- Locally green: the affected files — `provider_instance_test.py`, `docker/instance_test.py`, `mock_provider_test.py` (124), `mngr_imbue_cloud` instance/listing (27), and `mngr` ratchets (56) — plus `ruff` and `ty` on the changed files.
- Both imbue-code-guardian gates run clean (no CRITICAL/MAJOR): autofix (1 sub-threshold MINOR, see below) and verify-conversation.
- Full suite is deferred to CI (not run locally).

Known sub-threshold note (left as-is): the Docker override re-touches the daemon in the offline-fallback path; if the daemon dies in the narrow window after `get_host` already succeeded, that one host degrades to a per-host error entry in `mngr list` (contained by `_process_host_with_error_handling`) rather than an offline row. Swallowing the daemon error here would itself be a silent-failure anti-pattern, so it's surfaced.


joshalbrecht:
I guess that's technically true, but this is a little confusing

lgtm for now

PR #2249, opened 2026-06-23
Fix custom parent_type agent types losing inherited settings through multi-layer config loads

gnguralnick:
## Problem

A custom agent type that declares only `parent_type` (e.g. `[agent_types.worker]` with
`parent_type = "claude"`) is supposed to inherit all of its parent type's non-default
settings. This worked when the config was resolved straight after a single merge, but once
the real config load applied a second settings layer, every inherited value silently reset
to the child config class's defaults -- so a `worker` built on `claude` came out with
`use_env_config_dir`, `auto_dismiss_dialogs`, `auto_allow_permissions`, and `cli_args` all
back at their defaults.

## Root cause

`merge_models_via_overlay` reparses each merged result with `model_validate`, which marks
every present field as explicitly set. The base side of a whole-config merge is serialized
with a full `model_dump()`, so a registry entry carried through unchanged loses its sparse
`model_fields_set` (it becomes "all fields"). The later `parent_type` inheritance merge uses
`exclude_unset` semantics on the child, so once the child entry is non-sparse its defaulted
fields look explicitly set and clobber the parent's values. (A single merge looked fine
because a child-only-on-the-override entry's merged dict still held only the keys it set; the
second merge is what re-marked the carried-through base entry as fully set.)

This restores the same `model_fields_set` invariant already fixed for
`ProviderInstanceConfig.merge_with` and `RecursivePluginConfig.merge_with` (#1982): the
remaining gap was that the overlay reparse did not preserve per-entry `model_fields_set`.

## Fix

After the reparse, restore each merged model's `model_fields_set` -- recursively, including
nested sub-models and every `RegistryField` entry -- to the union of the base's and the
override's actually-set fields, rather than "all fields". A new `_restore_model_fields_set`
helper rebuilds each node via `model_construct(_fields_set=...)` (no revalidation, so
already-validated concrete subclass entries are carried through untouched). The fix lives in
the merge, not in `agent_config_registry.py` or any settings file.

## Tests

Two regression tests in `agent_config_registry_test.py` exercise a >=2-layer
`MngrConfig.merge_with` followed by `resolve_agent_type`, asserting a `parent_type` child
inherits a parent's non-default scalar and that a child override still wins. Verified to fail
on `main` without the fix. Full config suite: 458 passed.

---


joshalbrecht:
lgtm

PR #2250, opened 2026-06-23
minds: don't compile CSS in postinstall under ToDesktop --prod install

weishi-imbue:
## Problem

The scheduled `minds launch-to-first-message` health check (and the underlying ToDesktop build) has failed on every mngr main commit since the Tailwind v4 migration (`gleb/ux-cleanup`). The `build` job dies at 6% with:

```
Error: Command failed with exit code 1: pnpm recursive install --prod
```

## Root cause

ToDesktop's cloud builder installs the app with `pnpm recursive install --prod`. `--prod` omits devDependencies but **still runs lifecycle scripts**. The migration changed `postinstall` from `bash scripts/fetch_tailwind.sh` (only needs `curl`) to `pnpm run build:css`, which shells out to the `tailwindcss` CLI — a **devDependency**. Under `--prod` that binary isn't installed, so `postinstall` exits 1 (`tailwindcss: command not found`) and the whole install (and build) aborts before any platform is packaged. The local mac-runner steps use a full `pnpm install --frozen-lockfile`, so they never hit this.

## Fix

Move the postinstall into `scripts/build-css.js`, which no-ops when the Tailwind CLI isn't installed. The packaged app already ships the compiled `app.min.css` inside the minds wheel (built locally by `build.js` before upload), so the cloud install never needs to compile CSS. A full dev `pnpm install` still pre-builds the stylesheet as before.

## Verification

Reproduced ToDesktop's environment locally with a minimal pnpm project:
- old `postinstall` under `--prod` → `tailwindcss: command not found`, install exits 1 (matches CI);
- new `postinstall` under `--prod` → skips, install succeeds;
- new `postinstall` under full install → builds `app.min.css`.

Empirical CI verification (dispatch of launch-to-first-message against this branch) in progress.


joshalbrecht:
roughly lgtm

gdenisov:
I wonder if we need the postinstall step at all — theoretically in the dev workflow we call watch:css with `just minds-start` and in the world of build.js the `build:css` is also called explicitly

weishi-imbue:
close in favor of #2257

PR #2251, opened 2026-06-23
Add minds pool backfill-host-keys wrapper + recipe (staging host-key backfill)

joshalbrecht:
## Summary

Adds an env-aware `minds pool backfill-host-keys` command and a `just backfill-pool-host-keys` recipe wrapping the existing `mngr imbue_cloud admin pool backfill-host-keys`. The wrapper resolves the staging/production `host_pool` DSN from the tier's `<vault_prefix>/neon.DATABASE_URL` Vault entry (exactly like `minds pool list` / `destroy`), so the one-time SSH host-key backfill is a named, auditable recipe instead of a hand-rolled Vault/DSN lookup in bash.

This is the migration path called for by the host-key-pinning change (`94458ad`/`90ccb76`): rows + bare-metal boxes baked before the host-key columns existed need their sshd host public keys captured once so leasing/prep stop failing closed.

## Operational steps performed (staging)

- Re-deployed the host-key-pinning connector to staging: `minds env deploy --yes-i-mean-staging` (applied migration `011_host_public_keys.sql`; both apps healthy).
- Ran the one-time backfill via the new recipe: 1 pool host + 1 bare-metal box backfilled, 0 skipped. Re-run is idempotent (0/0); verified all three key columns are now populated.

## Testing

- `apps/minds/imbue/minds/cli/pool_test.py` — 42 passed (incl. new `test_build_backfill_host_keys_admin_args_forwards_dsn_when_present`)
- `apps/minds/imbue/minds/test_ratchets.py` — 56 passed
- `uv run ty check` (apps/minds) — clean


joshalbrecht:
lgtm, deployed

PR #2252, opened 2026-06-23
minds: discovery-pipeline health watchdog

gnguralnick:
## Summary

Minds' workspace-discovery pipeline (the background `mngr forward` consumer + the detached latchkey-forward supervisor producer) keeps the workspace list, the per-workspace liveness dots, and the recovery flow up to date. Previously, if it silently stalled (the producer stopped emitting updates) nothing noticed, and if the consumer crashed outright you only got a CRITICAL "Forwarding subprocess died -- restart minds" notification you could easily miss.

This adds an app-global **discovery-health watchdog** that detects a stalled or dead pipeline and self-heals it before surfacing a terminal, app-level recovery screen.

This is the discovery-watchdog work only. It is rebuilt cleanly on top of the now-merged recovery-redundancy work (#2236) -- the branch had carried a stale, pre-#2236 copy of that recovery code, which has been dropped in favor of main's version, leaving just the net-new watchdog here.

### `DiscoveryHealthWatchdog` (`discovery_health.py`)

A `HEALTHY` / `RECONNECTING` / `BLOCKED` state machine driven by two signals:

- **Producer stall** -- the resolver's snapshot-freshness age. When discovery snapshots go stale past threshold, the watchdog runs a cheap-to-heavy producer re-kick ladder: `LatchkeyForwardSupervisor.bounce()` (SIGHUP the observe child; gateway + reverse tunnels untouched) -> `restart()` (full supervisor restart) -> give up (`BLOCKED`). A cold start that never produces a first snapshot escalates into the same `BLOCKED` tier.

- **Consumer death** -- the forward consumer's unexpected-exit signal transitions straight to `BLOCKED`.

While it heals in the background nothing interrupts the user: an already-open workspace keeps working, with no banner -- only the existing "time since last discovery" indicator in the providers panel.

### Pipeline wiring (`forward_cli.py`, `run.py`, `app.py`, `state.py`)

- `forward_cli.py`: the lifecycle watcher now reports an unexpected consumer exit to registered `on_unexpected_exit` callbacks (-> the watchdog) instead of dispatching the CRITICAL "Forwarding subprocess died" notification, which is removed. The consumer no longer takes a `notification_dispatcher`.
- `run.py`: constructs the watchdog (wired to a `SupervisorProducerRemediator`), registers the consumer-death callback before the consumer starts, and launches the background poll loop.
- `app.py`: a background loop polls resolver freshness and drives the watchdog; the chrome SSE stream surfaces only the terminal `BLOCKED` transition via a new `discovery_health` event.

### Electron recovery screen (`main.js`, `shell.html`)

The `discovery_health` `blocked` event drives the existing `shell.html` error takeover with a "lost track of your workspaces" message and a "Restart Minds" button wired to the existing backend-restart path (which respawns the discovery producer + consumer). The takeover is guarded so an SSE reconnect can't re-drive it, and is reset by the retry handler so a re-block after a restart surfaces again.

### Included fix

`run.py` constructed `SupervisorProducerRemediator` with a positional argument, but it is a pydantic `MutableModel` whose `supervisor` field is keyword-only -- so `minds run` would have raised at startup once the watchdog was wired. Fixed to pass `supervisor=...`. (The watchdog tests never exercised the full `run()` entrypoint, so it had slipped through.)

Design doc: `blueprint/discovery-health-watchdog/plan-discovery-health-watchdog.md`.

## Test plan

- [x] `apps/minds/imbue/minds/desktop_client/discovery_health_test.py`, `forward_cli_test.py`, `test_desktop_client.py` -- 195 passed locally
- [x] `apps/minds/imbue/minds/test_ratchets.py` -- 56 passed locally
- [x] `ty check` on all changed `.py` files -- clean
- [x] `ruff check` on all changed `.py` files -- clean
- [x] changelog gate (`minds` + `dev`) -- ok
- [ ] Full offload suite green on CI


apps/minds/electron/main.js
```diff
@@ -1547,6 +1554,21 @@ function handleChromeSSEEvent(evt) {
...
+    // App-global discovery-pipeline health. Only the terminal `blocked` state is
+    // ever sent (the reconnecting tier heals silently in the background). The
+    // watchdog has exhausted its self-healing -- or the consumer died -- so agent
+    // forwarding is down / the app is unusable. Take over every window with the
+    // error screen; its Retry button runs the existing restart path (shut down +
+    // restart the backend, respawning the discovery producer + consumer).
+    if (evt.state === 'blocked' && !discoveryBlockedShown) {
+      discoveryBlockedShown = true;
+      showErrorInAllWindows(
+        "Minds has lost track of your workspaces and can't reconnect on its own. Restart Minds to recover.",
```
gnguralnick: "Minds has disconnected from your workspaces and can't automatically reconnect. Restart the app to recover. Your data has not been lost."

apps/minds/imbue/minds/desktop_client/discovery_health.py
```diff
@@ -0,0 +1,316 @@
...
+        try:
+            self.supervisor.bounce()
+        except LatchkeyError as e:
+            logger.warning("Discovery watchdog: producer bounce failed: {}", e)
+
+    def restart(self) -> None:
+        try:
+            self.supervisor.restart()
+        except LatchkeyError as e:
+            logger.warning("Discovery watchdog: producer restart failed: {}", e)
```
gnguralnick: seems not good to swallow this - if we can't restart, that means there's a big failure

apps/minds/imbue/minds/desktop_client/discovery_health.py
```diff
@@ -0,0 +1,316 @@
...
+
+    _lock: threading.Lock = PrivateAttr(default_factory=threading.Lock)
+    _health: DiscoveryHealth = PrivateAttr(default=DiscoveryHealth.HEALTHY)
+    # When the watchdog first observed state, used as the freshness baseline for
+    # the cold-start case (no snapshot has ever arrived, so there is no
+    # ``last_full_snapshot_at`` to age). Set on the first ``evaluate`` call.
+    _started_at: datetime | None = PrivateAttr(default=None)
+    # How many ladder rungs have run in the current stall: 0 = none, 1 =
+    # bounced, 2 = restarted. Reset to 0 on recovery.
+    _rungs_attempted: int = PrivateAttr(default=0)
```
gnguralnick: this seems like a silly framing. there are only two rungs; we just need to know whether we've done each. i'm not sure this ladder metaphor really accomplishes much

gnguralnick: if we can't restart, and we're going to go to the fully block state at the next poll time, we probably could just go to that state now

apps/minds/imbue/minds/desktop_client/discovery_health.py
```diff
@@ -0,0 +1,331 @@
...
+from pydantic import ConfigDict
+from pydantic import Field
+from pydantic import PrivateAttr
+
+from imbue.imbue_common.mutable_model import MutableModel
+from imbue.mngr_latchkey.core import LatchkeyError
+from imbue.mngr_latchkey.forward_supervisor import LatchkeyForwardSupervisor
+
+# How stale the last full discovery snapshot must be before the watchdog treats
+# the pipeline as stalled. Discovery polls every ~10s, so this is ~3-4 missed
```
joshalbrecht, line 69: discovery polls every 10 seconds!? that seems fairly aggressive...

all of this logic seems kind of complicated and brittle...

gnguralnick: followed-up on slack - improving/simplifying/hardening this and maybe reducing discovery cadence are potential follow-ups to this work

PR #2253, opened 2026-06-23
Add Modal compute provider (DIRECT) + workspace lifecycle in minds

minhtrinh-imbue:
## What

Adds **Modal** as a first-class compute provider for minds workspaces (DIRECT mode — the local machine authenticates with its own Modal token), and wires the full workspace lifecycle (state display, resume, destroy) into the minds UI so a Modal workspace behaves like any other.

## Why

Enables running minds workspaces on Modal cloud sandboxes (2 CPU / 4 GB, 24h), selectable from the Create form, without standing up per-user cloud backends.

## Highlights

**Provider + create flow**
- "Modal (1-day ephemeral)" option in the Create form; bootstrap registers a persistent `[providers.modal]` block (`is_persistent=true` is load-bearing — an ephemeral app would die the instant the `mngr create` subprocess exits).
- Sandboxes sized 2 CPU / 4 GB with a 24h timeout (matches lima/docker); config validates `default_sandbox_timeout + buffer <= 86400` (Modal's hard cap).

**Lifecycle in the UI**
- Projects list shows correct **Running / Stopped** state for Modal (previously blank). A paused/stopped host gets a **Resume** control; a running/unknown one keeps a recovery **Restart**, so a row is never control-less. No Stop button and excluded from the quit prompt (Modal can't stop host compute).
- `start_host` auto-restart skips the bare "initial" snapshot, so resuming a dead host no longer spawns an agent-less billing orphan; an initial-only host steers the user to recreate.
- Recovery no longer auto-runs `mngr start` on page open for Modal — resume is an explicit click (docker/lima auto-dispatch unchanged).

**Teardown / destroy hardening**
- Destroy passes `MNGR_HOST_DIR` to the detached `mngr destroy` (it was the only minds→mngr shell-out that omitted it, so teardown targeted the wrong project and silently no-op'd).
- Stop-host is guarded server-side (clear 409 for resume-only providers instead of a 500); recovery Restart skips the unsupported stop step for Modal.
- `stop_host` surfaces a `terminate()` failure instead of swallowing it; the "Destroying…" page reveals a Dismiss escape hatch instead of spinning forever.
- **Running-host resolution fix:** `mngr destroy`/`start` of a *running* Modal host read agents only from the state volume (which can lag the live host) → "agent not found" → eternal spinner. The resolver now re-reads a still-missing running host's agents live (the same source `mngr list` uses) on a lookup miss. **Scoped to Modal** via a new `discovers_agents_from_live_host` provider flag (default True; only Modal returns False) — every other provider is untouched, and there's zero extra work on the common path.

**mngr core**
- `derive_offline_host_state` guards a malformed/untrusted `stop_reason` (falls back to `CRASHED` instead of raising out of discovery).

**Scope note**
- `ModalMode.PROXIED` is a `NotImplementedError` stub: Modal is imbue-internal and its sandboxes are ephemeral, so it is DIRECT-only.

## Diff
36 files, +1850 / −125 (production + tests + per-project changelogs for `apps/minds`, `libs/mngr`, `libs/mngr_modal`).

## Testing
Local: `ty` clean on changed files; minds desktop_client + ratchets, `libs/mngr` api/find/discover + ratchets, and `libs/mngr_modal` suites all green. Real-Modal acceptance tests run in CI offload. Manually verified end-to-end on Modal DIRECT: create → correct state → resume → destroy a running workspace from the UI.


libs/mngr/imbue/mngr/api/discover.py
```diff
@@ -147,6 +151,76 @@ def _all_identifiers_found(
     return not remaining
 
 
+def _reconcile_running_hosts_with_live_agents(
```
joshalbrecht: I'm kind of skeptical of the changes in this file...

what was the precise bug that happened here and necessitated this? I'm fairly certain that the modal provider's discovery of hosts and agents should already be consistent

libs/mngr_modal/imbue/mngr_modal/instance.py
```diff
@@ -434,6 +461,90 @@ def supports_volumes(self) -> bool:
...
+                    # --branches --tags selects exactly refs/heads/* and refs/tags/*
+                    # (no remote-tracking refs), matching GIT_MIRROR_PUSH_REFSPECS semantics.
+                    self.mngr_ctx.concurrency_group.run_process_to_completion(
+                        ["git", "-C", str(source_path), "bundle", "create", bundle_path, "--branches", "--tags"],
+                    )
+                except ProcessError as e:
+                    raise MngrError(f"Failed to build git bundle of work-dir repo: {e}") from e
+
+            with log_span("Uploading git bundle to sandbox over control plane", remote=remote_bundle):
+                sandbox.copy_from_local(Path(bundle_path), remote_bundle)
```
joshalbrecht: this is no more robust than what we were already doing -- it can still have network failures, etc

In fact, this is strictly worse -- it now can also leave unclean threads, sockets, and async tasks that make retrying considerably harder than just retrying the git operation

why would't we just retry the flaky git operation directly with a simply tenacity decorator?

minhtrinh-imbue:
Superseded by #2310 (clean minimal Modal provider, no liveness layer).

PR #2255, opened 2026-06-23
minds: don't let a malformed permission-request agent_id crash the requests consumer

pseay-imbue:
## Problem
A latchkey permission request whose `agent_id` is **not** a valid `agent-…` id crashes the `latchkey-permission-requests-consumer` thread. `_maybe_recover_host_permissions` does `AgentId(event.agent_id)` with no guard, and `AgentId` requires the `agent-` prefix:

```
File ".../minds/cli/run.py", in _maybe_recover_host_permissions
    agent_id = AgentId(event.agent_id)
imbue.imbue_common.ids.InvalidRandomIdError: AgentId must start with 'agent-', got 'ENV_AGENT'
```

Once that thread dies, **no permission request — for any agent or service — is ever surfaced to the UI again**, and it re-crashes on the same persisted record on every restart. This happens when an agent hand-crafts a `POST /permission-requests` body with a placeholder id (`ENV_AGENT`); the gateway accepts it because `permission_requests.mjs` only checks the field is a non-empty string. `_displayable_pending_requests` has the same unguarded `AgentId(req.agent_id)`.

## Fix
Guard both spots that build an `AgentId` from a streamed/stored request:
- `run.py` `_maybe_recover_host_permissions`: skip recovery (with a warning) on a malformed `agent_id` instead of killing the consumer thread.
- `app.py` `_displayable_pending_requests`: skip such requests rather than raising and taking down the whole request panel.

A malformed/hand-crafted request now degrades gracefully (it just isn't surfaced) instead of wedging the entire permission system.

(Supersedes #2254, which GitHub auto-closed when the branch was renamed.)


joshalbrecht:
see slack

pseay-imbue:
Moved fix to https://github.com/imbue-ai/mngr/pull/2258

PR #2257, opened 2026-06-24
minds: build Tailwind CSS at consumption points, not in postinstall

gdenisov:
## Problem

The scheduled `minds launch-to-first-message` health check (and the underlying ToDesktop build) has failed on every mngr `main` commit since the Tailwind v4 migration (`gleb/ux-cleanup`). The `build` job dies at 6% with:

```
Error: Command failed with exit code 1: pnpm recursive install --prod
```

## Root cause

ToDesktop's cloud builder installs the app with `pnpm recursive install --prod`. `--prod` omits devDependencies but **still runs lifecycle scripts**. The migration changed `postinstall` to `pnpm run build:css`, which shells out to the `tailwindcss` CLI -- a **devDependency**. Under `--prod` that binary isn't installed, so `postinstall` exits 1 (`tailwindcss: command not found`) and the whole install aborts before any platform is packaged.

## Why not just guard the hook

Compiling CSS is a build-artifact step, not an install step. The cloud `--prod` install never needs to compile CSS at all: the packaged app already ships the compiled `app.min.css` inside the minds wheel (force-included via `[tool.hatch.build] artifacts` and built locally by `scripts/build.js` before upload). Guarding `postinstall` to no-op when Tailwind is absent (the approach in #2250) keeps a misplaced step that now silently skips in one environment and still redundantly recompiles on every full install.

## Fix

Remove the `postinstall` hook entirely. CSS is now built only at its genuine consumption points, all of which already existed:

- **Packaging** -- `scripts/build.js` -> `buildCss()` (this is what ships in the wheel).
- **Dev run** -- `pnpm start` -> `prestart` compiles once, then `watch:css` keeps it fresh.
- **The `minds_electron` e2e test** -- the only path that relied on `postinstall`, because it launches `electron main.js` directly. Wired in explicitly: a "Build Tailwind CSS for the e2e app" step in the CI job (`.github/workflows/ci.yml`) and a `minds-css` dependency on the `just minds-test-electron` recipe.

The app already degrades gracefully if `app.min.css` is missing (`app.py` logs a warning and serves unstyled), and the e2e test asserts structure, not styling -- so this is belt-and-suspenders for representative screenshots, not a correctness fix for that test.

## Verification

- `pnpm install --frozen-lockfile` then `pnpm run build:css` run end-to-end: install succeeds and (with `postinstall` gone) no longer builds CSS; `build:css` then compiles a 35KB minified Tailwind v4 `app.min.css`.
- `just -n minds-test-electron` confirms `minds-css` now runs before the pytest.
- CI YAML validated; the new step lands in the `test-docker-electron` job.

Alternative to #2250 (guard approach). Close that in favor of this if the relocation is preferred.


gdenisov:
## CI status: `test-docker-electron` red — pre-existing `main` regression, not this PR

The single failing check (`test-docker-electron`) is **not caused by this PR's CSS change**. It is a regression already present on `main`:

- `main` @ `a34b04d05` (the #2252 `gabriel/discovery-watchdog` merge) **failed `test-docker-electron`**; the 7 prior `main` runs were green. This PR inherits the failure only because the stop hook merged `a34b04d05` into the branch.
- This PR's new "Build Tailwind CSS for the e2e app" step **passed**; Electron launched and the test drove the create form before failing. `test-offload` and `test-offload-acceptance` (full suite) passed.
- **Root cause:** the new app-global discovery watchdog escalates to `restart()`, which "fully restart[s] the supervisor (bounces the gateway + reverse tunnels)". In the auth-less e2e environment discovery always stalls (repeated `Failed to list imbue_cloud accounts`), so the watchdog tears down the `mngr forward` supervisor that serves the page the Playwright test drives over CDP → `playwright._impl._errors.TargetClosedError: Page.wait_for_timeout: Target page, context or browser has been closed`.

The fix belongs at the source (#2252) — e.g. the watchdog should not restart the supervisor in the no-auth e2e scenario, or the e2e test should opt out of the watchdog. This PR will go green once `main` is fixed and re-merged.

PR #2258, opened 2026-06-24
minds: reject malformed permission-request agent_id at the gateway

pseay-imbue:
## Summary

An agent can `POST /permission-requests` to the latchkey gateway with an `agent_id` that is not a valid `agent-...` id (e.g. a hand-crafted body with a placeholder like `ENV_AGENT`). The gateway only checked that the field was a non-empty string, persisted it, and returned 201. The desktop client's consumer later built `AgentId(event.agent_id)`, which raised `InvalidRandomIdError`, killed the `latchkey-permission-requests-consumer` thread, and silently stopped every subsequent permission request — for any agent or service — from reaching the UI (re-crashing on the same record on each restart).

## The proper-layer fix (two parts)

**1. Reject at the gateway — so the agent is made aware.** The agent's tool call is a raw HTTP POST, so that's where the error must surface. `permission_requests.mjs` `parsePermissionRequestBody` now validates `agent_id` against `VALID_AGENT_ID_PATTERN` (`^agent-[0-9a-fA-F]{32}$`) and throws `InvalidRequestBodyError` (HTTP 400) **before** the request is persisted. The malformed record never gets created, and the agent learns its request failed at the point it made it. The JS pattern mirrors Python's canonical `imbue.imbue_common.ids.RandomId._validate`; a cross-language test posts a freshly generated `AgentId` to guard against drift.

**2. Catch ANY error in the consumer loop — so a bad record can never wedge the consumer.** `PermissionRequestsConsumer._run` now wraps the per-request parse + dispatch in a broad `try/except` that logs the failure with `logger.opt(exception=e).error(...)` (full traceback) and skips just that request. This replaces the previous narrow `except InvalidRandomIdError` in `run.py` (now removed, along with its unused import): the consumer thread is the inbox's source of truth, so it must survive *any* unprocessable request, not just one specific error.

The gateway is the primary defense; the consumer loop catch is the defense-in-depth backstop for legacy/already-persisted records. The desktop client also still hides from the list any request whose agent can't be resolved.

## Tests

- Gateway: updated fixtures from the now-invalid `agent-1` to a conforming id; added `test_post_rejects_malformed_agent_id` (parametrized; asserts 400, names `agent_id`, and asserts no file is persisted) and `test_post_accepts_generated_agent_id` (drift guard).
- Consumer: added `test_consumer_survives_an_on_request_error_and_keeps_processing` (a callback that raises on one record still lets the next one through).
- Bumped the `test_prevent_broad_exception_catch` ratchet 16 → 17 for the new, logged, justified catch.

## Design

A blueprint plan documenting the decision lives in `blueprint/gateway-agent-id-validation/`.


hynek-urban:
LGTM!

PR #2264, opened 2026-06-24
Report frontend errors to Sentry

hynek-urban:
Report more errors via Sentry:
  - Report JS errors in pages served from the local Minds backend.
  - Report errors in the Electron process.

Uses three newly created Sentry projects (`minds-frontend-production`, `minds-frontend-staging`, `minds-frontend-dev`) because apparently, it's discouraged to mix languages within a single Sentry project.


joshalbrecht:
lgtm once conflicts are resolved!

PR #2266, opened 2026-06-24
Log errors that happened while handling errors

hynek-urban:
Addresses the feedback from https://github.com/imbue-ai/mngr/pull/2234.

- Use `log_error_inside_sentry()` when an error happens in one of the `before_send` hooks.
- Harden `log_error_inside_sentry()`:
  - Make sure it cannot spiral into recursion.
  - Also log the errors in the application log, not just to Sentry.


joshalbrecht:
lgtm, thanks for fixing!  I remember this being such a pain before...

PR #2267, opened 2026-06-24
Mutate shared Claude config to dismiss startup dialogs (not permissions)

qi-imbue:
## Summary

For Claude shared config mode (`isolate_local_config_dir=False`), mngr now mutates the user's global Claude config to dismiss the cosmetic startup dialogs, instead of leaving it untouched.

Previously, shared mode made no writes to the user's config, so a fresh `~/.claude.json` (or a stub) re-triggered the trust/onboarding/effort screens on every agent and could intercept tmux input.

### What changed

- Shared mode dismisses the cosmetic startup dialogs (trust, onboarding, effort callout, cost threshold) directly in the config file claude actually reads: `$CLAUDE_CONFIG_DIR/.claude.json`, or `~/.claude.json` when the var is unset. New `find_shared_claude_config()` resolver mirrors `modify_env_vars`' `CLAUDE_CONFIG_DIR` resolution; new `_dialog_dismissal_config_path()` picks the right file per mode.
- `_dismiss_start_dialogs`, the `on_before_provisioning` validation, and the cost-threshold ack now run for local shared agents (still skipped for remote, which are always isolated).
- It honors `auto_dismiss_dialogs` in shared mode.
- It **never** accepts bypass-permissions mode (`bypassPermissionsModeAccepted`) through the global config — that remains governed by `settings.json` (`skipDangerousModePermissionPrompt`), per the existing `auto_dismiss_claude_dialogs` contract.
- Fixed the hooks write **and** load gates (`provision` and `assemble_command`) to key off the resolved predicate `not _is_isolated_config_dir()` rather than the deprecated `use_env_config_dir` alias, so mngr's managed `--settings` hooks are both written and loaded when shared mode is set via the current `isolate_local_config_dir=False` flag.

Updated the field description and README accordingly.

## Testing

- `just test-quick libs/mngr_claude` (offload-excluded markers): 612 passed.
- New tests: `find_shared_claude_config` resolution, shared-mode provision dismisses dialogs in the global config but not the permission flag, shared-mode `on_before_provisioning` validation against the shared/default config, and managed `--settings` injection via the `isolate_local_config_dir=False` flag.
- vet (agentic) reports no issues.

### DIFF
--- a/libs/mngr_claude/imbue/mngr_claude/plugin.py
+++ b/libs/mngr_claude/imbue/mngr_claude/plugin.py
@@ -108,8 +108,8 @@ from imbue.mngr_claude.claude_config import complete_onboarding
 from imbue.mngr_claude.claude_config import dismiss_effort_callout
 from imbue.mngr_claude.claude_config import encode_claude_project_dir_name
 from imbue.mngr_claude.claude_config import find_project_config
-from imbue.mngr_claude.claude_config import find_user_config_in_unisolated_mode
 from imbue.mngr_claude.claude_config import find_user_config_in_isolated_mode
+from imbue.mngr_claude.claude_config import find_user_config_in_unisolated_mode
 from imbue.mngr_claude.claude_config import fold_hook_configs
 from imbue.mngr_claude.claude_config import get_agent_claude_config_dir
 from imbue.mngr_claude.claude_config import get_agent_claude_plugin_dir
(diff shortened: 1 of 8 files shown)


libs/mngr_claude/imbue/mngr_claude/plugin.py
```diff
@@ -1579,6 +1580,23 @@ def _is_isolated_config_dir(self) -> bool:
         """
         return self.agent_config.resolve_isolate_local_config_dir() or not self.host.is_local
 
+    def _dialog_dismissal_config_path(self) -> Path:
+        """Return the global ``.claude.json`` whose startup-dialog state this agent dismisses.
+
+        Both modes record dialog dismissals (trust, onboarding, effort callout, cost
```
qi-imbue, line 1586: I'm actually slightly surprised by this and then I found out we have a bunch of `click.confirm` that asks the user before mutating the global Claude config.

Is this really necessary for the isolated config mode?

joshalbrecht: it's probably slightly less complicated to have "should it use click.confirm" be configured independently from "should the config dir be shared"

so no, it's not necessary, but it's probably a little less complex to keep them orthogonal?

joshalbrecht:
lgtm once tests pass!

PR #2268, opened 2026-06-24
minds: fix window restore reopening to landing page + creating dead-end

gnguralnick:
## Problem

Bugs in how minds windows reopen on relaunch:

1. Windows restored on app relaunch reopened to the **main landing page** instead of the mind they were showing.
2. A reopened window sometimes showed a **black screen with bare text "Unknown agent creation"** (e.g. after the app was closed following a failed creation) and stayed stuck there.
3. (Found mid-review) With the restore fix in place, two open minds could collapse to a **single landing-page window** on relaunch, and — relatedly — the landing page is missing some minds for a few seconds after launch.

## Root cause

**Window restore.** A workspace window's live content URL is on the agent subdomain (`agent-<id>.localhost:<mngr_forward_port>/...`). `saveSessionState` ran that through `toRelativeBackendUrl`, which does `pathname + search + hash` — **stripping the host entirely**. The agent identity collapsed to a bare path (usually `/`), and on restore `toAbsoluteUrl('/')` resolved against the *minds backend* origin: the landing page.

**Cold-start discovery gap (bug 3 + the 2→1 collapse).** `fetchInitialChromeState` resolved on the **first** `/_chrome/events` `workspaces` snapshot, which at cold start is the empty/partial state before discovery has polled all providers. That partial list both painted an incomplete landing page and — once the restore fix made `filterRestorableUrls` able to parse the agent id out of the persisted `/goto/<agent>/` URL — caused **every** restored workspace window to be dropped (absence from the partial list misread as "destroyed"), collapsing to one landing-page window.

**Creating dead-end.** `GET /creating/<id>` returned a bare `404 "Unknown agent creation"` when the in-memory creation was gone (app restart / cleaned-up failed creation), and a restored/reopened window navigated to it.

## Fix

`apps/minds/electron/main.js`:
- `toPersistedContentUrl` persists workspace windows as the port-independent `/goto/<agent>/` bridge path (carries the agent id; survives the changing `mngr_forward` port). `toRestoredContentUrl` rebuilds the live URL on launch via `workspaceUrlForAgent`. Both it and `filterRestorableUrls` resolve the persisted relative URL to absolute before `parseWorkspaceId` (`new URL()` throws on a bare path). Restoring to the mind's root is intentional — the mind persists its dockview layout server-side and restores it on a fresh root load.
- `fetchInitialChromeState` now holds the loading screen until it sees `discovery_complete: true` (or a 25s timeout), so restored windows **and** the landing page see the complete mind list. The restore path only filters windows against the workspace list when discovery is complete; on the timeout fallback it keeps **every** saved window — absence from an incomplete snapshot is not evidence of destruction (matching the live-path principle, which only navigates a window away on positive destroy evidence).

`apps/minds/imbue/minds/desktop_client/app.py`:
- New `discovery_complete` flag on the `workspaces` SSE payload, via `_is_discovery_complete` (wraps the existing `_is_discovery_fresh`: true once a **fresh** full snapshot has landed; a cached last-good snapshot carries an old timestamp so it reads not-complete; non-mngr resolvers are always complete). Emitted on connect and on the diff-update, and pushed when it flips even if the list is unchanged.
- `_handle_creating_page` redirects to `/` (303) on unknown creation instead of the 404 page. The status / onboarding / logs endpoints keep their 404s — they are XHR/SSE callers, not navigations.

## Tests
- `app_discovery_complete_test.py`: `_is_discovery_complete` is false before any snapshot, true after a fresh one, false for a stale/cached one, and true for non-mngr resolvers.
- `test_desktop_client.py::test_creating_page_redirects_to_landing_for_unknown` updated (was `..._returns_404_...`) to assert the 303 → `/` fallback.
- `ruff` + `pyright` clean. Full unit + integration suite verified green on CI offload (the only failure was the stale 404 assertion above, now fixed).
- `main.js` has no JS unit-test harness in this repo (only heavyweight Playwright e2e; repo rules forbid test-only exports), so its persist/restore + wait-for-snapshot logic was verified by a throwaway script replicating the exact functions/regexes (round-trip, dead-workspace filtering, timeout-fallback keep-all, origin-fallback all pass). **The end-to-end cold-start behavior still needs verification in the running app** (the original regression lived in cold-start integration timing a replica can't model).


libs/mngr/imbue/mngr/providers/docker/instance.py
```diff
@@ -1657,6 +1683,21 @@ def discover_hosts(
...
+            raise ProviderUnavailableError(self.name, f"Cannot list Docker hosts: {e}") from e
+
+        # A present-but-stopped state container means host records are
+        # unreachable: every read exec's into a stopped container, and
+        # ``list_all_host_records`` swallows that into an empty list -- which
+        # discovery would publish as "zero hosts", dropping every host (and, for
+        # a long-lived cached provider instance, the read never restarts it).
+        # Surface it as ProviderUnavailableError so an empty list still means
+        # "genuinely zero hosts" and the retain-on-error path keeps last-known
+        # hosts instead. Read-only discovery must not start it; the owning app
```
gnguralnick, line 1696: or maybe read-only discovery _should_ start it? not sure

apps/minds/imbue/minds/cli/run.py
```diff
@@ -250,6 +252,27 @@ def run(
...
+    # Restart this env's mngr docker *state* container before the discovery
+    # producer spawns. The quit flow stops it (``stop_active_env_state_container``)
+    # so no stray container is left running; but if it's still stopped when
+    # discovery polls, the docker provider can't read host records and reports
+    # zero hosts -- which blanks the restored windows and the landing page on the
+    # first snapshot. Bring it back up first. Best-effort: a no-op without docker,
+    # and a start failure must not block startup (discovery then degrades to
+    # ProviderUnavailable + retain-last-known rather than a misleading empty).
+    try:
+        start_active_env_state_container(
```
gnguralnick, line 269: because of the mngr docker changes in this PR, without this the app will crash on startup. but if we wanted the mngr-side to include starting the state container for reads we could get rid of this step

libs/mngr/imbue/mngr/providers/docker/instance.py
joshalbrecht: better for discovery to error -- discovery should not change state

joshalbrecht:
lgtm

PR #2269, opened 2026-06-24
Convert minds Vault secrets to split per-key leaf layout

joshalbrecht:
## What

Switches all minds Vault access from the old **flat** layout (one KV entry per service holding many fields) to the **split** layout, where each key is its own single-`value` leaf:

```
# old:  secrets/minds/<tier>/<service>   -> { "A": "...", "B": "..." }
# new:  secrets/minds/<tier>/<service>/A -> { "value": "..." }
#       secrets/minds/<tier>/<service>/B -> { "value": "..." }
```

This matches the canonical convention in the `~/project/vault` repo (path basename = env var name, actual value under `value`).

## Changes

- `vault_reader.py`: `read_vault_kv` now lists the service directory and reads each leaf's `value` to reconstruct the `{key: value}` dict callers expect; `write_vault_kv` writes one leaf per key; `delete_vault_kv` removes every leaf. **No fallback to the old flat layout.**
- `generation.py`: catches the dedicated `VaultSecretNotFoundError` for a missing entry.
- Updated callers' docstrings/error messages, operator help text (`neon_db`, `cleanup_released_hosts`), `scripts/push_vault_from_file.py`, and the vault-setup / host-pool-setup / staging-bringup docs.
- Rewrote `vault_reader` + `generation` unit tests for the two-call (list + per-leaf get) model.
- Added `scripts/remove_old_flat_vault_secrets.py`: a guarded, dry-run-by-default cleanup tool that retires the old flat entries for a tier once they're verified fully mirrored.

## Out of scope

`scripts/changelog_deploy.sh` reads `secrets/mngr/dev/*`, which was **not** mirrored to the split layout (only `secrets/minds/*` was), so it is intentionally left unchanged.

## Still pending

- Run `remove_old_flat_vault_secrets.py dev --yes` (after review) to retire the old dev paths.
- Deploy to `dev-josh-1` and verify the deploy reads only the new locations.


joshalbrecht:
roughly lgtm

joshalbrecht:
new changes lgtm too

PR #2272, opened 2026-06-24
Bump offload from 0.9.7 to 0.9.9 (latest)

DanverImbue:
Bumps the pinned [offload](https://crates.io/crates/offload) version from `0.9.7` to `0.9.9` (the latest release).

Offload is a Rust crate pinned in two places, both updated here:

- `.github/workflows/ci.yml` — cargo cache key, the `offload --version` check, and the `cargo install offload@...` invocation (in both jobs).
- `libs/mngr/imbue/mngr/resources/Dockerfile` — `OFFLOAD_VERSION`, the binary baked into the Modal sandbox image (which is documented to track the CI pin).

Historical `CHANGELOG.md` entries and the feature-availability comments in `scripts/snapshot_minds_e2e_state.py` ("offload v0.9.7+") are left untouched, as they document when a feature was introduced rather than pinning a version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,12 +102,12 @@ jobs:
             ~/.cargo/registry
             ~/.cargo/git
             ~/.cargo/bin/offload
-          key: cargo-offload-0.9.7-${{ runner.os }}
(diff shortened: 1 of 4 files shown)


joshalbrecht:
lgtm, thanks danver!

PR #2273, opened 2026-06-24
minds: user-controlled error reporting + get-help / report-a-bug (phases 1-2)

gnguralnick:
First two phases of the minds error-reporting & "get help" work (the blueprint plan covers all four; phases 3-4 — in-workspace and out-of-workspace agent help — will be stacked as PRs on top of this one).

## Phase 1 — consent, settings, and live Sentry gating
- `MindsConfig` persists three per-machine settings: `error_reporting_consent_given`, `report_unexpected_errors`, `include_error_logs` (all default off).
- First-launch **consent screen** (`Consent.jinja`) shown ahead of welcome/login until answered; the same two toggles live on the **account settings page** (with a FIXME noting they should move to a dedicated app-settings page once one exists).
- Sentry now **always initializes**, but what it sends is gated **live** by these settings: a `before_send` gate drops automatic events while reporting is off (manually-submitted reports always pass via a `manually_submitted` tag), and log/traceback attachment is gated by `include_error_logs`. The `MINDS_SENTRY_ENABLED` / `MINDS_SENTRY_S3_UPLOADS` env vars no longer gate anything.

## Phase 2 — get-help button + report-a-bug
- A permanent question-mark **"get help" button** in the top bar beside the inbox. It opens a modal — an **Electron overlay** mirroring the inbox machinery (new `toggle-help` IPC + `main.js` helpers + preload method; browser fallback navigates the content frame to `/help`).
- The modal offers **"have an agent help"** (intentionally **disabled** for now — wired up in a later phase) and **"report a bug to Imbue"**. The form has a description, a conditional include-logs checkbox, and an Advanced section (app diagnostics, in-workspace details, a recorded remote-access flag).
- A **shared collector** (`report_collector.py`) plus `core.submit_manual_bug_report` back both the local form and a new authenticated **`POST /api/v1/agents/<id>/report`**, so a future in-workspace agent and the local form produce identical reports. All Sentry submission stays in the outer app.

## Intentional scope cuts (deferred to later phases)
- Agent-help flow is UI-disabled.
- Heavy/sensitive collection (full project code, full chat-history content, deep in-container/recovery probes) is not collected; the form only exposes controls for what is actually gathered.
- Resource usage covers cpu/load/disk via stdlib only (no `psutil`); system memory is not collected.

## Testing
- Ran locally and passing: the relevant unit + integration tests (minds config, sentry core, report collector, desktop-client routes, auth templates), the minds ratchet suite, `uv run ty check`, and `ruff` lint/format.
- The full Modal `just test-offload` suite could not run in the dev workspace (no Modal tokens) — relying on CI here.
- Reviewed via validate-diff, autofix, and verify-conversation gates (no actionable findings).


apps/minds/electron/sentry.js
```diff
@@ -71,29 +86,235 @@ function fixupReleaseId(releaseId) {
...
+    // the Electron process's stderr and outlives the app on quit. Under
+    // `just minds-start` that stderr is a socketpair `concurrently` reads to
+    // prefix `[electron]` output; because the orphaned handler never closes it,
+    // `concurrently` never sees EOF, its kill-others never fires, and the dev
+    // launcher hangs after every quit. Packaged builds aren't wrapped in
+    // `concurrently`, so they keep native-crash minidump reporting.
+    integrations: (defaults) =>
+      paths.isDev()
+        ? defaults.filter((integration) => integration.name !== 'SentryMinidump')
+        : defaults,
```
joshalbrecht: hmmm... I'm always suspicious when I see "isDev" style flags -- they're pretty much just asking for bugs

Why exactly do we need this? Is there no way to make it work the same way in dev?

gnguralnick: agreed, i'm having a claude see if there's another way to do this. this was just the most direct fix to get rid of the thing that was causing concurrently not to quit.

apps/minds/electron/sentry.js
```diff
@@ -71,29 +107,249 @@ function fixupReleaseId(releaseId) {
...
   Sentry.setTag('git_sha', gitSha);
-  console.log(`[sentry] Initialized (environment=${environment}, release=${fixupReleaseId(releaseId)}).`);
+  console.log(
+    `[sentry] Initialized (environment=${environment}, release=${fixupReleaseId(releaseId)}); ` +
+      'automatic reporting gated live by the report_unexpected_errors user setting.'
+  );
+}
+
+// -- Backend-down report enrichment (logs + host basics) --
+//
```
joshalbrecht, line 165: isn't most of this stuff already done by python functions? (ex: the inclusion of logs, etc)

seems weird to have it duplicated...

gnguralnick: messaged in slack too, but this is so that we can still get logs for reports of the entire backend crashing

PR #2281, opened 2026-06-25
Serialize same-repo Docker mind creations on the clone path

gnguralnick:
## Problem

Creating a Docker mind intermittently fails during the work-dir transfer with:

```
rsync: [Receiver] getcwd(): No such file or directory (2)
rsync error: errors selecting input/output files, dirs (code 3) at util1.c(1126) [Receiver=3.4.2]
```

It only reproduces when two minds are created for the **same repo** at roughly the same time.

## Root cause

The failing rsync is the **local** one on the user's Mac (Homebrew rsync 3.4.2), spawned by `mngr create`. rsync 3.4.2 calls `getcwd()` at startup and hard-aborts (exit 3) if its cwd has been deleted out from under it (rsync bug 6422; older rsync tolerated this).

In `agent_creator.py`, minds clones the template repo to a **fixed**, repo-name-keyed path: `clone_target = tempdir/"minds-clone-<repo>"` (stable on purpose, so Docker layer caching hits). At the start of every creation it does `if clone_target.exists(): shutil.rmtree(clone_target)` and re-clones, then runs `mngr create` with `cwd=clone_target`; mngr's `rsync` subprocesses inherit that cwd.

When a second creation of the same repo overlaps the first, its `rmtree` deletes the directory that the first creation's still-running `mngr create` (and its rsync) has as its cwd. The first creation's rsync then starts, calls `getcwd()`, finds the cwd gone, and aborts. The path is keyed only on repo name, so any two overlapping creations of the same repo collide — it is a pre-existing latent race that rsync 3.4.2 makes fatal.

## Fix

Add an in-process per-clone-path lock (`ClonePathSerializer`) held across the entire span a creation owns the shared clone dir — the `rmtree` + clone (+ worktree overlay) and branch checkout through the **completion** of `mngr create` (which uses the dir as its cwd and git-mirror source). A second creation of the same repo now waits until the first finishes that step, so no creation ever `rmtree`s a clone dir another creation is still using.

- The lock is keyed **per clone path** (per repo), so creations of different repos still run concurrently.
- A plain local-directory source (used in place — no clone, no `rmtree`) passes `clone_target=None` and is unserialized.
- Trade-off (intentional): two concurrent same-repo creations now run sequentially. We chose this over making the clone path unique, which would lose the Docker-layer-cache reuse the stable path exists for.

To keep the lock correct and testable, the lockable region was extracted into `AgentCreator._clone_and_create`, and source resolution into `_resolve_workspace_source` / `_materialize_workspace_dir`, so the `clone_target` the lock keys on is computed once and the flow can be driven with injected fakes (no test-only hooks in production).

## Tests

- New regression test `test_concurrent_same_repo_creations_serialize_on_clone_path`: pins one creation inside the locked region (via a blocking fake `imbue_cloud_cli`, an existing DI seam) and asserts a second same-repo creation cannot enter the clone+create region until the first releases. Verified it **fails without the lock** (both creations enter the region) and passes with it.
- New `ClonePathSerializer` unit tests: same-path serialization (max-concurrency 1), distinct paths run concurrently, `None` is never serialized.
- New `_resolve_workspace_source` unit tests: remote URL / local worktree / plain local dir, the same-repo-name → same `clone_target` collision property, and the missing-path error.
- `agent_creator_test.py`: 59 passed. `test_ratchets.py`: 56 passed. `ty check` + `ruff` clean. (6 local restic/backup failures are environmental — no `restic` binary installed locally — and don't touch this change.)
- Full `just test-offload` suite runs here in CI.


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#5 from the same branch. This public repository is now a read-only mirror.

gnguralnick:
Continuing on imbue-ai/mngr-internal#119 — development has moved; this mirror is read-only, so PRs cannot merge here.

PR #2282, opened 2026-06-25
Discovery watchdog: retry forever instead of dead-ending; harden latchkey bounce-watcher; fix empty error details

gnguralnick:
## Motivation

A user hit the terminal "Minds has disconnected from your workspaces and can't automatically reconnect" screen, and "Restart Minds" didn't recover it. Diagnosis: the discovery-pipeline health watchdog detected a *producer* stall (the `mngr observe` grandchild of the `mngr latchkey forward` supervisor stopped emitting), ran one bounce + one restart, and — when freshness didn't return in time — gave up to the terminal `BLOCKED` takeover. Because the underlying supervisor kept re-wedging, a manual restart spawned another supervisor that re-stalled the same way, so the dead-end persisted. The takeover's "Show details" box was also empty (the blocked path passed `null` for details).

This reworks the watchdog so a producer stall is treated as *degraded, not fatal*.

## What changed

- **Event-based stall signal.** The stall signal is now the age of the resolver's last discovery *event* (`last_event_at`) rather than its last full *snapshot*. Since a full snapshot bumps both timestamps, `last_event_at >= last_full_snapshot_at`; keying off events means a producer that is alive and still emitting incremental updates — just slow to complete a full re-poll (e.g. one provider is down) — reads as healthy and is left alone instead of being needlessly restarted. A *dead* supervisor manifests the same way (its `mngr observe` grandchild stops emitting), so freshness covers that case too — no separate liveness probe is needed.
- **Backoff-forever, no give-up.** The producer path no longer reaches `BLOCKED`. On a stall it bounces once, then issues repeated supervisor restarts on a capped exponential backoff (15s → 30s → 60s → 120s → … capped at 5 min), retrying indefinitely. A failed restart is just another "did not help" (log + keep backing off), not terminal. The backoff is capped on purpose: a full restart re-provisions every managed host, so hammering a merely-slow producer would make things worse. (The backoff exponent is clamped before doubling so the unbounded retry counter can't `OverflowError` and silently kill the watchdog thread.)
- **Takeover reserved for consumer death.** `BLOCKED` (the full-screen takeover) is now reached *only* when the forwarding subprocess — which is also the HTTP traffic proxy — dies, i.e. the genuinely-unusable case. A producer stall stays in the silent background `reconnecting` state (the open workspace still works); the providers panel's "time since last discovery" counter remains the only passive signal.
- **Empty details fix.** When the takeover does appear, its "Show details" panel now shows the tail of the log instead of an empty box (`main.js` passed `null`; now passes `readLastLogLines(50)`, matching the other takeover sites).

The watchdog rework is backend-only on the `apps/minds` side (no Electron behavior change beyond the details fix: the SSE only ever emitted `discovery_health` for `BLOCKED`, so making producer stalls never reach `BLOCKED` is what removes the takeover for that case). This PR **also** carries a `libs/mngr_latchkey` change — see "Also in this PR" below.

### Considered and dropped: a supervisor-liveness trigger

An earlier revision added a `LatchkeyForwardSupervisor.is_running()` probe so the watchdog could remediate a dead supervisor immediately instead of waiting out the stall timer. It was removed: it only sped up detection of a *fully-exited* supervisor by ~30s (a state that's silent-and-still-usable anyway), the freshness signal already subsumes it (a dead supervisor stops emitting events), and it did nothing for the actual incident (a wedged-but-alive supervisor). It also cost a `_supervisor_seen_alive` latch, a startup race against the background supervisor-init thread, and a per-poll on-disk/psutil probe — complexity not worth a marginal speedup. Net result: simpler and covers the same failure modes.

## Also in this PR: bounce-watcher robustness (`mngr_latchkey`)

Originally deferred as follow-up (a); now included, because it is the backend reason the watchdog's cheap `bounce()` remediation was a no-op in the incident.

The `mngr observe` discovery child of the `mngr latchkey forward` supervisor was tracked as a *checked* concurrency-group strand. Every bounce (and every shutdown) terminates it via SIGTERM → exit `-15`, which the group then re-checks as a failure and raises as a `ConcurrencyExceptionGroup` — both at the bounce respawn and at supervisor exit. That group escaped `_run_sighup_bounce_watcher`'s narrow `(OSError, RuntimeError)` guard and **permanently killed the watcher thread**, so every later SIGHUP bounce (a provider toggle, or the watchdog's own `bounce()`) became a silent no-op for the rest of that supervisor's life. This is exactly why the cheap `bounce()` remediation was useless and the watchdog had to lean on the heavier `restart()`.

Fixes:

- Spawn the observe child with `is_checked_by_group=False` (in both `start()` and the bounce respawn): a deliberately-terminated stream's SIGTERM exit is expected, not a failure. This removes the `ConcurrencyExceptionGroup` at the bounce respawn *and* at supervisor exit.

- The SIGHUP bounce watcher now catches `Exception` (log + continue), so a long-lived daemon survives any single bounce's error instead of dying on it.

- `bounce_observe()`'s terminate/respawn guards and the shutdown `stop()` guard are widened to the concurrency-group failure types they can actually raise (including `subprocess.TimeoutExpired` on a force-kill overrun), so a slow-to-die child can no longer abort the rest of teardown.

- A broad-exception ratchet ticks 0 → 1 for the watcher's deliberate daemon safety net (documented in `test_ratchets.py`).

Real-world confirmation: a live `latchkey_forward.log` showed this exact `ConcurrencyExceptionGroup … exit code -15 … mngr observe` firing ~22 times (once per supervisor teardown); on this branch those are gone. (Distinct from the sleep/wake death-screen incident, which the watchdog rework above fixes — that one was verified by hand on this branch.)

## Diagnosis confirmed by the supervisor log

The user's `latchkey_forward.log` confirmed the failure shape. The root cause was **a full disk** (`[Errno 28] No space left on device` while the `mngr observe` producer wrote a provider temp file; logging itself was failing too). That produced a **wedged-but-alive** supervisor: the process stayed up, but its SIGHUP bounce-watcher thread died on an unhandled exception, making every `bounce()` a silent no-op. The **event-based stall signal + backoff-forever** handles exactly this: the watchdog sees the dead events, escalates to a full `restart()` (fresh supervisor), and keeps retrying instead of dead-ending. (Caveat: that log was a supervisor instance from 06-16, not provably the exact 06-24 screenshot session, since the supervisor log is per-instance — but the mechanism is the same and ENOSPC persists until disk is freed.)

## Follow-ups (intentionally NOT in this PR)

- **(b) Disk-full surfacing (`minds`).** Since backoff-forever never recovers a full disk and the producer-stall state is intentionally silent, the user gets no signal to free space. The actual ENOSPC originates three process layers down (in the detached `mngr observe` grandchild) and is only logged, never threaded up — and the disk-full condition breaks the very file-write channel that would carry it. The robust approach is a **direct `shutil.disk_usage(data_dir)` check in minds, gated on the watchdog being stalled**, surfacing a soft "disk almost full — free up space" notice (not the terminal takeover).

(Follow-up (a), bounce-watcher robustness, is now implemented in this PR — see above.)

## Test plan

- [x] `discovery_health_test.py` — fresh-event healthy; fresh-event-healthy-despite-stale-snapshot; stall → reconnecting + bounce; bounce → restarts on growing backoff, never blocks; failed restart keeps retrying; recovery resets; consumer death → terminal BLOCKED; cold-start grace then stall; cold-start never-recovers keeps retrying; backoff holds at cap without OverflowError past 1024 restarts. (13 passed)
- [x] `test_desktop_client.py -k discovery_health` SSE tests. (2 passed)
- [x] `apps/minds` ratchets pass (trailing-comment ratchet at baseline); `ty` clean on changed files. (56 passed)
- [x] `mngr_latchkey` `discovery_stream_test.py` + `cli_test.py` — a real bounce terminates and respawns the observe child without surfacing the `-15` strand as a failure; the SIGHUP watcher survives an unexpected `ConcurrencyExceptionGroup` from one bounce and still services the next. Both fail against the pre-fix code. (31 passed)
- [x] Full `mngr_latchkey` suite (354 passed); `mngr_latchkey` ratchets (56 passed; broad-exception at 1); `ty` clean on changed files.
- [ ] Full suite via CI.


joshalbrecht:
lgtm I guess, but I still kinda dislike how complicated and ad-hoc all of this restart logic is

Would be worth implementing a more coherent strategy for this stuff later (ie, a better way of understanding if any of our required background processes died, and if so, why, and being able to surface such failures to us (via sentry) and the user (so they know what failed) in a sensible way...)

PR #2285, opened 2026-06-25
mngr_latchkey: reap duplicate latchkey-forward discovery producers per directory

gnguralnick:
## Problem

Two Sentry bug reports (minds-production, 0.3.3): a running docker mind disappeared from the app ("0 minds" → "create a mind" redirect), and providers the user had disabled kept erroring in the terminal.

Root cause (traced from the real on-disk discovery-events file): multiple `mngr latchkey forward` supervisors were running against the same shared `.minds` latchkey directory — left by prior/concurrent app instances (e.g. several sculptor workspaces + `utilities/mngr`). Each runs its own `mngr observe --discovery-only` child, so several producers wrote the one shared discovery-events file with **divergent, frozen-at-startup provider configs**. Orphans kept polling (and erroring on) providers the user had since disabled, and their conflicting snapshots made the consumer flap agents in and out.

`LatchkeyForwardSupervisor.ensure_running()` only reconciled the PID in `latchkey_forward.json`; when that record was missing/stale it spawned a fresh supervisor without noticing the already-running orphans.

## Fix

`ensure_running()` now enforces one forward per latchkey directory: it reaps every `mngr latchkey forward` bound to the same resolved `--latchkey-directory` — **along with all of that forward's descendant processes** (its `mngr observe` producer, the `latchkey gateway`, and reverse `ssh` tunnels) — except the one matching the live on-disk record. An unrecorded orphan is replaced, not adopted (it may run stale code/config).

Safety:
- Scoped by **resolved `--latchkey-directory` equality**, so a supervisor for one profile never signals a sibling profile's forward (`.minds` vs `.minds-staging`/`.minds-dev`).
- Descendants are captured as live `psutil.Process` handles (snapshotting `(pid, create_time)`) and the forward PID is re-verified immediately before terminating — so a PID recycled in the scan→kill window is never signalled.

minds invokes this via `restart()` on every startup.

## Scope / caveats

- **Fully fixes** the "disabled providers keep erroring" symptom and removes cross-producer snapshot divergence.
- **Does not by itself fully fix** a mind disappearing: a *single* producer can still momentarily report a provider healthy while omitting an agent (Docker intermittently returning no containers without raising), and the consumer currently drops the agent on the first such snapshot. Hardening consumer-side retention (hysteresis) and the Docker provider's empty/error behavior is separate follow-up work.

## Test plan

- [x] `forward_supervisor_test.py`: cmdline parsing, descendant enumeration, reap-on-same-dir, adopt-recorded-and-reap-other, do-not-reap-different-dir safety boundary, the PID re-verify guard, and an **end-to-end test that a reaped forward's child process is terminated**.
- [x] Fixed a pre-existing flaky `_spawn_test.py` reporter (atomic write) surfaced by the new process-spawning tests.
- [x] `just test-quick "libs/mngr_latchkey -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'"` → 359 passed, 0 failed (run repeatedly).
- [ ] **Manual cold-start re-test recommended** (unit tests use fake binaries): quit/reopen the minds app and confirm exactly one `mngr latchkey forward` per `.minds`, the docker mind stays visible, and disabled providers stop erroring.


gnguralnick:
## Additional production scenario (minds-production 0.3.4): new imbue_cloud mind stuck "waiting for workspace to be ready"

Hit this live today and traced it; it's the same root cause this PR targets, with a third symptom worth recording, plus one scope observation about the consumer.

### Symptom
Creating a new imbue_cloud mind (`taxonomizing`) hung on "waiting for workspace to be ready", then fell through to the recovery page with `dispatch_tier=host_unresponsive`. Production log:

```
WARNING: Workspace readiness probe for agent-192b7cbc... timed out after 300s (last status=503); publishing redirect anyway
Layer-2 host-state probe for agent-192b7cbc...: dispatch_tier=host_unresponsive
```

The host itself was completely healthy the whole time: SSH in fine, `system_interface` RUNNING under supervisord, `curl localhost:8000/` -> 200 with the correct hostname/agent-id, cloudflared connected, and the host-side `LATCHKEY_GATEWAY` reverse-tunnel port (1989) listening. So the readiness 503 and the empty recovery snapshot (no host state, no plugin-resolver services, `mngr exec` returns no stdout) were entirely client-side.

### Same root cause as this PR
The process list showed exactly the duplicate-forward condition this PR fixes — multiple `mngr latchkey forward` on the shared `~/.minds/latchkey`:
- `9363` — recorded/active forward (production `Minds.app`, bundled binary; matches `latchkey_forward.json`)
- `18586` and `21221` — left over from a **sculptor dev-workspace** minds build (`.../sculptor/workspaces/84230ac7.../node_modules/.bin/latchkey`), pointed at the **production** `~/.minds/latchkey`, ~1 day old, each with its own `mngr observe --discovery-only` child writing the shared events file.

So multiple producers were writing divergent snapshots into the one shared discovery-events file, and the consumer was flapping the new agent's host-state / `system_interface` registration in and out -> proxy never had a stable backend to resolve -> 503 for the full readiness window. `restart()`-on-startup with this PR's reaping would have removed `18586`/`21221` at the 11:45 app launch and left a single coherent producer. The user's manual "kill all `mngr forward` processes + relaunch" is effectively what the reap does, and it fixed it immediately (post-restart state: exactly one supervisor + one producer + one consumer).

This is a distinct presentation from the two Sentry reports in the PR description ("mind disappears" / "disabled providers keep erroring"), but the same mechanism. Worth adding to the test-plan / scenario list as the "new mind never goes ready (503/host_unresponsive)" face of the bug.

### Scope observation: the `mngr forward --observe-via-file` consumer is not covered by the reap
The reap targets `mngr latchkey forward` processes (cmdline must contain both `latchkey` and `forward`) and their `children(recursive=True)`. The discovery **consumer** (`mngr forward --observe-via-file`, the traffic proxy) is neither: it's a **sibling** of the supervisor, not a descendant. Confirmed against the live tree:

```
minds run (backend)
├── mngr forward --observe-via-file     <- consumer (sibling)
└── mngr latchkey forward               <- supervisor (reaped if duplicate)
    └── mngr observe --discovery-only    <- producer (descendant, reaped with it)
```

The consumer is spawned by `minds run` (`start_mngr_forward` -> `subprocess.Popen`), and its only cleanup is `EnvelopeStreamConsumer.terminate()` on graceful lifespan shutdown. A force-quit/crash orphans it, and nothing reaps orphans on startup. In practice they accumulate: this machine also had a consumer from **~2 days earlier** (`84657`) still running alongside two from today.

I don't have evidence that an orphan consumer *independently* causes the 503 — each consumer keeps its own isolated resolver, binds the default `8421` then falls back to an ephemeral port if taken, and the app routes to the port its *own* freshly-spawned consumer reports via `wait_for_listening`. So orphan consumers read the shared events file read-only and don't corrupt the live consumer's view; they look like co-leftovers / a resource leak (squatting `8421`, redundant per-host SSH tunnels, redundant file tailing) rather than a root cause here. But the lifecycle asymmetry is real — this PR makes supervisors self-heal to exactly-one on startup, while consumers have no equivalent reaping — so it's worth deciding whether an analogous "one consumer per `minds run`, reap pre-existing orphans" pass belongs here or as a follow-up.


^ written by claude

joshalbrecht:
yes, we probably don't want spurious mngr forward instances running

however, mngr forward *should* be able to be configured (via some flag) to kill itself when its parent dies

this is *different* from mngr latchkey -- that process needs to *continue* running even when the minds backend exits, IIRC?

does that make sense?

joshalbrecht:
lgtm, will merge because this seems important to fix

PR #2286, opened 2026-06-25
Fix AWS workspaces wedging at startup: mount /run as tmpfs for supervisord

minhtrinh-imbue:
Mounts `/run` as a tmpfs in AWS workspace containers so supervisord can install its control socket — on AWS the container's gVisor gofer-backed rootfs rejects the hard link supervisord uses to place that socket (`os.link` → `EOPNOTSUPP`), which left every AWS workspace stuck "unresponsive" with no services started.

Essentially one file: `apps/minds/imbue/minds/bootstrap.py` adds `default_start_args = ["--tmpfs", "/run"]` to the per-region `[providers.aws-<region>]` blocks; AWS-scoped, since ovh/vultr/imbue_cloud already get a tmpfs `/run`. (Plus a one-line test assertion and the required changelog.)


joshalbrecht:
lgtm, thanks for the fix!

PR #2290, opened 2026-06-26
ci(launch-to-msg): green the launch_to_msg + macos_launch pair (release gate)

weishi-imbue:
## What

Iterates the `minds-launch-to-msg.yml` mac-runner CI until **both** jobs pass as a pair (the release gate). The branch is `main` + the commits below.

**Verified green pair:** run [28222028718](https://github.com/imbue-ai/mngr/actions/runs/28222028718) — `launch_to_msg ✓` + `macos_launch ✓` (binary `commit_sha=e38ad54`, FCT `wz/vendor-sync-main`).

### `launch_to_msg` (self-hosted, Python e2e)

- Scroll the chat transcript to the live tail before the canned-body assertion (the transcript virtualizes off-screen rows, so the canned reply can be mounted but scrolled out of view).

- Treat create-agent auto-navigation to the workspace URL (`agent-*.localhost`) as the DONE signal, instead of polling `/status` against the workspace origin (which returns HTML and reads as empty).

- Dismiss the post-login "Help improve Minds" error-reporting consent screen (`#consent-continue`) that otherwise blocks the home page.

### `macos_launch` (GitHub-hosted, Playwright Electron smoke test)

- Bound the fixture teardown so a cold CI mac can't wedge it for the full test timeout: SIGKILL the app, `unref()` the worker's stdout/stderr **first** (detached helpers — minds backend, `mngr latchkey forward`, crashpad — inherit and hold the worker's stdio, so the worker never exits otherwise), then a 10s-bounded `pkill` of those helpers and an 8s-capped `app.close()`. Previously an unbounded `execSync('pkill')` ran before the `unref`, hung for the entire 600s test timeout, and the `unref` never executed.

- **Check out the trigger ref (not `commit_sha`) for the e2e harness**, matching `launch_to_msg`. `macos_launch` previously pinned both the binary *and* the playwright spec/fixtures to `commit_sha`, so harness fixes on the dispatched branch were silently never run by this job — the two jobs tested different harness code. The binary under test is still pinned to `commit_sha` via the build artifact.

### Infra / docs

- `mac-runner-reset.sh`: drop `set -e` (a best-effort cleanup must not abort on the first failure and leak lima VMs) and fail loudly on the app-install curl/unzip/mv.

- `release.md`: when iterating CI, always dispatch a `template_ref` whose `vendor/mngr` is synced to the same mngr SHA being built — a stale FCT vendor silently rejects a renamed agent-config field, so the in-VM agent never starts and the e2e wedges at "Waiting for initial chat agent…" (reads as a frontend hang, is really vendor skew).

## Why

The launch-to-msg pair is the release gate for the minds app. These are the harness and workflow fixes needed to make it green on current `main`; merging them lets the release dispatch (`--ref main`) inherit them.


joshalbrecht:
lgtm, thanks for fixing these!

PR #2292, opened 2026-06-26
minds: (for caretaker) serve the user's timezone to workspaces via GET /api/v1/timezone

pseay-imbue:
The minds-desktop side of the nightly **Caretaker** feature: the desktop client now **serves the user's timezone on demand** (`GET /api/v1/timezone`) instead of pushing it into each workspace at create time. Pairs with the forever-claude-template PR (imbue-ai/forever-claude-template#212: cron/anacron scheduling, the Caretaker, and the in-workspace blinking tab — its bootstrap pulls this endpoint at boot to set the container clock).

## What's in here
- **`GET /api/v1/timezone`** (`apps/minds`) — returns the IANA timezone of the machine the desktop client runs on, e.g. `{"timezone": "America/Los_Angeles"}` (empty string when undeterminable). Read via `tzlocal` (new direct dependency, already pinned transitively in the lock) and validated against the tz database, so the response is always a real IANA name or an explicit unknown. New `host_timezone` module carries `read_host_timezone()` + the generalized `validate_iana_timezone()` (moved out of `create_helpers`); the route is registered in the API schema registry so the published document stays in sync.
- **Workspace reachability** (`libs/mngr_latchkey`) — every agent's baseline gateway permissions gain `minds-api-timezone-read`, a read-only grant for `GET /minds-api-proxy/api/v1/timezone` mirroring the existing API-schema read grant. Workspace agents (and the FCT bootstrap at boot) can fetch the user's current timezone with no per-agent permission dialog — e.g. to re-check it whenever the user schedules something.
- **Push path removed** — the `timezone` field on the create request, the create-route threading, and `agent_creator`'s `build_timezone_write_script` / `_write_scheduler_timezone` mngr-exec machinery are gone; nothing writes `runtime/scheduler/timezone` anymore. (The old PR description's "create form captures the browser timezone" never actually existed — the push was backend-only.)
- **Design docs** — `blueprint/caretaker-scheduler/` gains superseded-banners recording the revisions since the original plan: this pull model, and the FCT side's replacement of the custom scheduler with cron/anacron.

## Testing
- `PYTEST_MAX_DURATION_SECONDS=540 just test-quick apps/minds` — **1687 passed, 0 failed** (3 skipped).
- `PYTEST_MAX_DURATION_SECONDS=300 just test-quick libs/mngr_latchkey` — **395 passed, 0 failed**.
- `uv run ruff check apps/minds libs/mngr_latchkey` — clean.
- New tests: endpoint returns 200 + a ZoneInfo-loadable name (or "") and 401 without auth; `host_timezone` validator accept/reject cases; latchkey baseline assertions extended for the new grant.

Existing workspaces keep their old baseline permissions file, so their agents will 403 on the new route until recreated — the FCT side treats that as "timezone unknown, keep current clock", so nothing breaks.

Draft — open for review alongside the FCT PR.


joshalbrecht:
lgtm when the other PR is ready

weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#6 from the same branch. This public repository is now a read-only mirror.

pseay-imbue:
Continuing on imbue-ai/mngr-internal#96 -- development has moved; this mirror is read-only, so PRs cannot merge here.

PR #2296, opened 2026-06-26
Add Sentry error reporting to `mngr latchkey forward`

hynek-urban:
`mngr latchkey forward` is a daemon process with an independent lifecycle. As such, it should take care of its own error reporting; see the discussion [here](https://imbue-ai.slack.com/archives/C0ASLEFFD62/p1782314508043949?thread_ts=1782230596.736299&cid=C0ASLEFFD62).

- Move the core Sentry logic to `imbue_common`
- Use the shared Sentry logic both from `apps/minds` and from `libs/mngr_latchkey`
- The Minds app is responsible for communicating the right Sentry settings to `mngr latchkey forward` via env vars.
- Changes in user consent on the Minds app's side are propagated to a json file on the disk which is read every time an event is about to be sent (similarly to how the Minds app does it).

Here's a test event that shows that it works: https://generally-intelligent-e3.sentry.io/issues/7577344748/


joshalbrecht:
lgtm, will merge after the API branch

joshalbrecht:
tests passed here: https://github.com/imbue-ai/mngr/pull/2371

PR #2297, opened 2026-06-26
Workspace permissions

hynek-urban:
- Introduce a new type of permission request specifically for the `workspaces` API.
- Accepted and validated by the `permission-requests.mjs` extension that agents communicate with.
- Render a dialog in the UI, allowing users to override the extent of the given permission.

An alternative was to keep using the "predefined" permission requests which would seemingly lead to less code and duplication. The downside was that it wasn't possible to have workspace-specific permission request validation. Specifying "targets" or rendering the permission dialog in a specific way would have also been rather hacky.

Here's how a permission request of the new type can be created by an agent:

```
   latchkey curl -X POST http://latchkey-self.invalid/permission-requests \
     -H "Content-Type: application/json" \
     -d "{
       \"agent_id\": \"$MNGR_AGENT_ID\",
       \"rationale\": \"Testing the workspace permission request flow.\",
       \"type\": \"workspace\",
       \"payload\": {
         \"permissions\": [\"minds-workspaces-destroy\"],
         \"target_workspace_id\": \"agent-REPLACE_WITH_TARGET_32HEX\"
       }
     }"
```

See `.agents/skills/latchkey` or `.agents/skills/file-sharing` for examples of previous skills that communicated this to agents.


joshalbrecht:
merged as part of https://github.com/imbue-ai/mngr/pull/2315

please take a look at the additional latchkey related changes that I made there in case I messed anything up!

PR #2302, opened 2026-06-26
OVH reinstall transient-error retry + vault soft-delete skip

joshalbrecht:
Two small operational fixes split out of the (now-merged) split-secrets work, made while baking production hosts and cleaning up dev/staging/prod vault secrets.

## `mngr imbue_cloud admin server setup`: retry OVH reinstall on transient compatibility-lookup error
The OS reinstall kickoff (`POST /dedicated/server/{s}/reinstall`) intermittently fails OVH's OS-compatibility lookup for a perfectly valid template (`"Error while retrieving compatibility details for operating system debian12_64"`), then succeeds unchanged on a retry seconds later -- observed live ordering a production box. Previously this hard-failed the whole `server setup`.

The reinstall POST is now wrapped in a `poll_for_value` retry (5 min budget, 15s interval) that retries **only** on that specific transient marker and propagates any other error immediately. The ed25519 host keypair is generated once and reused across attempts, so the pinned host key stays stable regardless of retry count.

## `scripts/remove_old_flat_vault_secrets.py`: skip soft-deleted secrets
`vault kv get` on a secret whose latest version was soft-deleted returns exit 0 with a null `data.data` (the metadata still carries a `deletion_time`), not the exit-2 not-found the script assumed. A leftover soft-deleted `minds/staging/test` entry crashed the staging run. Such paths are now treated as absent and skipped.

## Tests
Three new unit tests on `_ReinstallStartAttempt` (success wraps the task; transient marker -> retry; non-transient -> propagates). `ordering_test.py` (26), `mngr_imbue_cloud` ratchets, and the full `slices/` dir (131) all pass locally.


joshalbrecht:
lgtm

PR #2303, opened 2026-06-26
ci(launch-to-msg): scroll chat to tail on every reply read (fix v0.3.4 flake)

weishi-imbue:
## What

Fixes the flaky `launch-to-msg` timeout that blocked the **minds-v0.3.4** release run (run 28228005836): the W1 cross-workspace follow-up waited 8 min for the agent's reply token and timed out though the agent had replied.

## Why

minds-v0.3.4 only auto-scrolls the chat to a new message when the transcript is **already pinned to the bottom** (verified manually). After the e2e navigates away and back (cross-workspace follow-up) or reloads, the view stays scrolled up, and the virtualized transcript leaves the new prompt + reply in unmounted rows absent from `document.body.innerText` — so the reply-count wait never sees them.

## Fix

Route every chat-transcript read (first-message reply, cross-workspace follow-ups, post-reload history) through a shared `wait_for_function` helper that scrolls `.app-content` to its tail on each poll. Previously only the Slack canned-body check did this.

Validated against the v0.3.4 anchor (binary `minds-v0.3.4` / FCT `minds-v0.3.4`).


joshalbrecht:
lgtm

PR #2304, opened 2026-06-26
Anchor TMR test scope on docstrings; run TMR over all of mngr's release tests

qi-imbue:
## Summary

Re-anchors the scope contract that TMR (test map-reduce) enforces from each
test's *tutorial block* to its *docstring*, generalizing TMR beyond the e2e
tutorial tests, migrates every mngr release test to the new scheme, and points
the TMR workflow at all of mngr's release tests. See
`specs/docstring-anchored-tmr.md`.

Previously TMR was specific to the mngr e2e tutorial tests: the prompts talked
about `mega_tutorial.sh`, `e2e.write_tutorial_block`, keeping tests "1:1 with the
tutorial", etc. None of that applied to the other release tests. Now each test's
docstring is the contract: the mapper agent makes the test verify exactly what
its docstring says and treats the docstring as read-only (the one exception being
an outdated `Tutorial block:` section, which it may correct along with
`mega_tutorial.sh` as a `FIX_TUTORIAL`).

## What changed

**Scheme / infrastructure**
- `specs/docstring-anchored-tmr.md` — the design.
- Docstring format: tutorial-anchored tests carry the verbatim block under a
  `Tutorial block:` section followed by a `Scope:` section; other tests carry a
  plain scope docstring.
- `scripts/tutorial_matcher.py` extracts the block from the docstring's
  `Tutorial block:` section (supports raw `r"""` docstrings for blocks with
  backslashes) instead of from `write_tutorial_block(...)`.
- The e2e fixture captures each test's docstring into `docstring.txt`; the detail
  renderer and `serve_test_output` show it under a "Docstring" heading.
  `write_tutorial_block` is removed.

**TMR generalized**
- `mapper.j2` / `reducer.j2` are docstring-anchored and drop the tutorial-1:1 /
  e2e-only framing; `FIX_TUTORIAL` is kept as the single docstring-mutation
  exception. The recipe injects the e2e-only `--mngr-e2e-run-name` flag only for
  e2e tests, so non-e2e release tests no longer error.
- The `sync-tutorial-to-e2e-tests` skill emits the new format and crystallizes
  each block's implicit requirements into its scope.

**Migration of all mngr release tests**
- All 32 e2e tutorial files migrated; `tutorial_matcher.py` confirms 100%
  tutorial↔test correspondence (and the migration caught a real tutorial drift:
  the `mngr --help` block listed `push, pull` vs `git`).
- Non-tutorial libs/mngr tests (e2e, cli, install/upgrade, docker) and the
  per-provider / per-agent packages (`mngr_aws`, `mngr_gcp`, `mngr_azure`,
  `mngr_modal`, `mngr_claude`, `mngr_codex`, `mngr_schedule`, …) got scope
  docstrings. Docstring-only; no test logic changed.

**TMR configuration**
- `.github/workflows/tmr.yml` (manual + daily scheduled) now defaults to all of
  mngr's release tests: `test_paths: libs/mngr`,
  `pytest_args: -m "release and not docker and not docker_sdk"` (343 tests).
  Docker-marked release tests are excluded because they need a real Docker daemon
  and run on a GitHub runner in `release-tests.yml`, not on TMR's Modal hosts.

## Notes
- Trivial/redundant tests surfaced during the migration were flagged, not
  deleted (deletions in the non-tutorial tier aren't matcher-verifiable).
- Release tests don't run in CI; the migration is docstring-only (plus a few
  stale-block corrections), so runtime behavior is preserved by construction.


joshalbrecht:
lgtm

PR #2308, opened 2026-06-26
minds: agent-escalated bug reports open a human-reviewed pre-filled modal, plus get-help "have an agent help"

gnguralnick:
## Overview

This is the "continued" phase of minds **get-help / error reporting**. The theme: **a human gates every Sentry send**. An in-workspace agent that hits a problem can escalate it, but instead of submitting a report itself, it opens the desktop app's report-a-bug modal *pre-filled* for a person to review and send. The same get-help modal also gains a "have an agent help" option that spawns an `/assist` chat inside the workspace.

### Agent-escalated bug reports → pre-filled modal (human-reviewed)
- The authenticated per-agent report route (`POST /api/v1/agents/<id>/report`) no longer submits to Sentry. It publishes an `OpenHelpRequest(description, workspace_agent_id)` through a new `HelpModalRequestBroker`, which fans the request out over the chrome SSE stream as an `open_help` event; the Electron shell opens `/help?workspace=<id>&description=<diagnosis>` in the window showing that workspace. The user reviews what to attach and submits through the normal `/help/report` collector. The route returns `{"ok": true}` (no `event_id` — nothing was transmitted).
- A pre-filled `description` now forces the modal into **report mode**. Previously the modal defaulted to agent-help whenever a workspace was set, so an escalation (which carries both a workspace *and* a description) opened in agent-help mode and would have spawned another `/assist` chat instead of the report form.

### "Have an agent help" in the get-help modal
- When opened from a loaded workspace, the modal offers (and defaults to) "Have an agent help fix the problem". Choosing it posts to `POST /help/assist`, which spawns a new chat inside that workspace's container (`mngr exec … mngr create --template chat`) seeded with `/assist <description>`; the system interface auto-opens its tab.

### Recovery-page "Report a problem" gating
- The link no longer shows on the transient "Loading workspace" spinner; it appears only on terminal states that offer an action: workspace-unresponsive (restart), the restart-dispatch error (restart), and provider-unreachable / "Can't connect to …" (retry).

### Latchkey
- A baseline rule allows any in-workspace agent to `POST …/agents/<id>/report` without a prior per-agent grant, ordered ahead of the unauthorized gate (detent stops at the first matching scope). The route's bearer-key auth still applies. Interim allowance pending the broader minds-API-surface latchkey work.

## Base / merge

`main` was merged into this branch (the repo uses merge, not rebase), so the diff is up to date with main's `/api/v1` spectree consolidation. The report route was reconciled onto that framework: `@require_api_or_cookie_auth` + `BugReportRequest` validation, returning `OkResponse`, and the published `GET /api/schema` document was updated to match (no longer advertising the dropped `event_id`).

## Testing

- Unit tests cover: the broker fan-out / subscribe / unsubscribe, the report route opening the modal (and *not* submitting to Sentry), `/help/assist` validation + dispatch arg-building (including hostile-input shell-quoting), the help-page `description` pre-fill, the report-mode default when a description is present, and the recovery-page link gating.
- Targeted suites pass locally — `test_desktop_client.py`, `assist_chat_test.py`, `help_modal_requests_test.py`, and latchkey `agent_setup_test.py`; `ruff` and `ty` clean on the touched files. CI runs the full offload suite.

## Changelog

Per-project entries are included for every touched project: `apps/minds`, `libs/mngr_latchkey`, and `dev`.


apps/minds/imbue/minds/desktop_client/help_modal_requests.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class HelpModalRequestBroker(MutableModel):
+    """Fans ``OpenHelpRequest``s out to every subscribed chrome-events SSE connection.
+
+    Each subscriber is a ``(queue, wake_event)`` pair owned by one SSE
+    connection: ``request_open`` enqueues the request and sets the event so the
+    connection's loop drains it. Subscriptions are added on connect and removed
+    in the loop's ``finally``.
+    """
```
joshalbrecht, line 47: huh -- what SHOULD happen if an error report request is made by an agent in one workspace and you have multiple workspace windows?

if I'm reading this correctly, it would open in all windows? That seems confusing. But I could be reading it wrong...

joshalbrecht:
lgtm after my one comment

apps/minds/imbue/minds/desktop_client/help_modal_requests.py
gnguralnick: sent in slack - it doesn't show in all windows, but it also doesn't necessarily show up in the correct window. will update

gnguralnick: (resolved) - this code does correctly only open the modal in the requesting workspace's window, falling back to the last focused window if the requesting workspace doesn't have a window open. there was a bug on FCT side where agents were giving the wrong ID; this is now fixed.

PR #2309, opened 2026-06-26
Pin changelog-consolidation schedule user_id to a committed constant

qi-imbue:
## Problem

The nightly `changelog-consolidation` schedule is deployed to a Modal environment named `mngr-changelog-schedule-<user_id>`. That `user_id` was a **random per-profile UUID** (`uuid4().hex`), generated and persisted lazily on first use by mngr's config layer. As a result, the deployed environment's identity depended entirely on whichever machine first deployed the schedule.

Redeploying from a fresh checkout (a different machine, or a reset profile) rolled a *different* random `user_id`, so `mngr schedule remove`/`add` operated on a brand-new, empty environment while the live schedule kept running in the original one — silently forking instead of replacing. This bit us during a recent redeploy: the deploy's `--stop-all-apps` sweep stopped the live cron apps, but `schedule add` then failed because the (different) target environment didn't exist, leaving the job down.

## Fix

Pin `user_id` to a committed constant so the environment name is machine-independent:

- Add `USER_ID = "920da0b3…"` to `scripts/changelog_schedule_utils.py` (the existing source-of-truth module), exposed via a new `--print-user-id` CLI flag. This is the value of the live production environment.
- `changelog_deploy.sh` and the `changelog-trigger` justfile recipe now `export MNGR_USER_ID` from that flag, so every deploy and on-demand trigger targets the same environment regardless of host.

`MNGR_USER_ID` is mngr's supported override for the otherwise-random `user_id`. The `--stop-all-apps` sweep is unchanged — it remains identity-agnostic (sweeps every `mngr-changelog-schedule-*` env by prefix).

Note: if a machine already has a stale changelog profile holding a *different* `user_id`, mngr fails loudly on the mismatch; the fix is to delete `~/.mngr-changelog-schedule` and let it regenerate against the pinned value. This is documented in the deploy script.

## Tests

- Added `test_user_id_is_pinned_lowercase_hex` and `test_cli_print_user_id_matches_constant` to `scripts/changelog_schedule_utils_test.py`.
- `just test-quick scripts/changelog_schedule_utils_test.py`: 13 passed, 0 failed.


joshalbrecht:
lgtm

PR #2310, opened 2026-06-26
Add Modal as a cloud compute provider (create / destroy / reconnect)

minhtrinh-imbue:
[human written]

In this PR, we made the mngr_modal plugin supporting mngr verbs:
- `create` boots a Modal sandbox 
- `list` discovers hosts/agents (read live off running sandboxes, from the state volume for offline ones)
- `destroy` terminates the sandbox and clears its records
- `start` reconnects to a live sandbox OR recreates it from the latest resumable snapshot
-`stop` snapshots then terminates.

I tested this branch in the Minds UI for all of the following cases:
- making a new workspace that was modal-backed
- starting an old workspace that was modal-backed, as long as the sandbox was still online
- destroying a minds workspace that was modal-backed. If its sandbox was still online, it is terminated too. If the sandbox was already gone (timed out), there is nothing to terminate, so destroy just removes the agent records and marks the host DESTROYED, and the workspace disappears from the list cleanly.
- Clicking the restart button from Minds home.
  - If the sandbox was still online, `mngr start` reconnects to the live sandbox (fast, no new sandbox).
  - If the sandbox was gone but has a resumable snapshot (it was stopped cleanly, or hit its graceful idle/max-age timeout), `mngr start` boots a brand-new sandbox from that snapshot (the slow cold-boot described below).
  - If the sandbox was gone and only its bare create-time ("initial") snapshot exists (e.g. it was hard-killed or crashed before it could snapshot), `mngr start` refuses with "no resumable workspace state", and Minds shows "Workspace cannot be recovered -- recreate the workspace" instead of looping on a restart that can never succeed.

One edge case on the minds side: restarting a Modal workspace cold-boots a brand-new sandbox from its snapshot, which takes a couple of minutes -- but minds' default "Restart workspace" timeout (30s, tuned for local containers) would call that a failure while the sandbox was still healthily booting, so a longer restart timeout is wired in, Modal-only.

My motivation for this last piece will make it very helpful for evals: I can run a Minds evals on fanned-out Modals, let it finish, maybe time out, then later reopen that exact workspace, hit restart, and have it come right back from the snapshot so I can see the chat & try out any web artifacts -- so revisiting Modal evals is clean and doable directly in Minds, instead of something I lose when the sandbox goes away or something I must export from Modal and import to another provider.


apps/minds/imbue/minds/desktop_client/app.py
```diff
@@ -3026,6 +3053,18 @@ def __call__(self, exc: BaseException) -> None:
...
+def _is_modal_workspace(backend_resolver: BackendResolverInterface, agent_id: AgentId) -> bool:
+    """Whether the workspace's host runs on the Modal provider.
+
+    Modal host restarts cold-boot a new sandbox from a snapshot, so they need a
+    longer restart budget than other providers. Resolved from the discovery
+    snapshot's provider name; unknown agents (no display info) are treated as
+    non-Modal so they keep the fast default budget.
+    """
+    info = backend_resolver.get_agent_display_info(agent_id)
+    return info is not None and info.provider_name == _MODAL_PROVIDER_NAME
```
joshalbrecht: this, and the above constants, and all of the changes in this file, make me a little skeptical

it seems like we'd more generically want to allow hosts to take as long as they need to start up? Certainly I would expect eg AWS hosts to be sometimes fairly slow to start as well

this gets in to a larger change -- rather than timing out and trying to automatically declare that something failed, our operations should generally be allowed to take ~forever, but we should tell the user when something takes longer than expected and thus could indicate a problem

in other words, the right behavior is probably something like:
- user or app tries to start any host / agent
- we have per-provider rough estimates of how long it should be until we *warn* the user (in the UI that they're looking at) that it's taking longer than normal to start
- we always make it possible for users to see the logs / what is happening during the start (in case it got stuck somewhere). This is eventually particularly useful for debugging agents as well
- only after a ridiculously long period of time (like, 15 minutes or something) do we say "it looks like this failed to start", and we just fail and stop there (we don't retry, etc).
- The *user* can then manually click something like "retry" to try the whole process again from the beginning

This general approach is:
- more robust
- simpler
- more transparent
- more consistent

It's probably worth working with @gnguralnick  on getting our starting / restarting to work a bit more simply / as I laid out above

For now, could we simply raise all of the default limits so that we don't need anything modal-specific here?

apps/minds/imbue/minds/desktop_client/templates.py
```diff
@@ -1071,6 +1071,22 @@ def render_inbox_unavailable_fragment(message: str = "") -> str:
         // but unreachable (bouncing it would interrupt user agents, so we want
         // explicit consent before doing so).
         function renderUnresponsive() {
+          // If the last restart failed because the host has no resumable state (e.g. a
+          // Modal sandbox terminated without a clean stop -- only its create-time snapshot
+          // exists), a restart can never succeed. Guide the user to recreate instead of
+          // looping on a Restart button that cannot work. Keyed off the error text, so it
+          // applies to any provider that surfaces the same "no resumable" failure.
+          var errText = errorEl ? (errorEl.textContent || '') : '';
+          if (errText.indexOf('no resumable workspace state') !== -1) {
```
joshalbrecht: it's fairly weird and brittle to key off of the string of the error here

I'm also not fully convinced that we should need special casing for modal -- it can always resume, it's just that sometimes, it might be resuming from an earlier state that doesn't include your previous state because it wasn't shut down cleanly.

we could warn, when starting, if that was the case, but I don't see why it should be implicated here for a special case

apps/minds/imbue/minds/primitives.py
```diff
@@ -67,6 +67,13 @@ class LaunchMode(UpperCaseStrEnum):
     LIMA = auto()
     IMBUE_CLOUD = auto()
     AWS = auto()
+    # Runs the agent in a Modal sandbox using the local machine's own Modal token
+    # (``modal token new``) -- resolves the ``modal`` provider instance (DIRECT).
+    # Modal sandboxes are ephemeral (~1 day max), so it is surfaced as "Modal
+    # (1-day ephemeral)" and is testing-only. (A keyless/connector-proxied variant
+    # is intentionally not offered: Modal is imbue-internal, so users authenticate
+    # to it directly.)
+    MODAL_DIRECT = auto()
```
joshalbrecht: the naming of "MODAL_DIRECT" is a little weird

(same comment applies not just to this constant, but the name of all other vars in this PR -- we don't have any other way of accessing modal from minds, do we?)

libs/mngr_modal/imbue/mngr_modal/config.py
```diff
@@ -58,8 +66,13 @@ class ModalProviderConfig(ProviderInstanceConfig):
         description="Base directory for mngr data on the sandbox (defaults to /mngr)",
     )
     default_sandbox_timeout: int = Field(
-        default=900,
-        description="Default sandbox timeout in seconds",
+        default=86_310,
```
joshalbrecht: it's very weird to change these constants in this PR -- we should just be setting them for minds in the create template for the forever-claude-template repo!

(this comment applies to all of the default changes here)

libs/mngr_modal/imbue/mngr_modal/instance.py
```diff
@@ -152,6 +154,29 @@
 # Fixed namespace for deterministic VolumeId derivation from Modal volume names.
 _MODAL_VOLUME_ID_NAMESPACE: Final[uuid.UUID] = uuid.UUID("c8f1a2b3-d4e5-6789-abcd-ef0123456789")
 
+# Snapshot names used by the modal provider. The "initial" snapshot is captured
+# right after provisioning (before the agent runs), so it carries no workspace
+# state; "stop" is captured on graceful stop/pause and carries the user's work.
+# Auto-restart must skip the bare "initial" snapshot or it would spin a sandbox
+# with no agent -- an agent-less, billing orphan.
+INITIAL_SNAPSHOT_NAME: Final[str] = "initial"
+STOP_SNAPSHOT_NAME: Final[str] = "stop"
```
joshalbrecht: I don't love these changes -- they:
1. are backwards incompatible
2. collapse the names of the snapshot into a smaller namespace where some get overwritten
3. are mixing our concerns from minds into the mngr provider (and they don't belong there)

yes, it doesn't make sense to resume from an initial snapshot (because you end up without the agent having been created), but we could have fixed that bug separately in a more minimal way I suspect (also, I wonder why we even bother with the initial snapshot if it cannot be resumed from there...)

apps/minds/imbue/minds/desktop_client/app.py
MT-GoCode: duly noted... changes to this reverted and I'll relay to gabe sometime

libs/mngr_modal/imbue/mngr_modal/config.py
MT-GoCode: fixed

apps/minds/imbue/minds/desktop_client/app.py
gnguralnick: that makes sense. the timeout constants for the restart operations were not super formally chosen; just ones that seemed reasonable. and at the time i was testing there were still only local (docker/lima) minds, so worth updating now that remote is fleshed out

apps/minds/imbue/minds/desktop_client/workspace_recovery.py
```diff
@@ -69,6 +69,12 @@
 _HOST_RESTART_STARTUP_WAIT_SECONDS: Final[float] = 30.0
 # Poll cadence while waiting for the system interface to come back post-restart.
 _RESTART_PROBE_INTERVAL_SECONDS: Final[float] = 1.0
+# Substring of HostShutdownNotSupportedError's message (raised by ``mngr stop --stop-host``
+# when a provider's ``supports_shutdown_hosts`` is False, e.g. Modal). A host-restart on
+# such a provider hits this on the stop step; it is expected, not a failure -- ``mngr start``
+# performs the restart on its own (reconnect-if-alive, else recreate-from-snapshot), so the
+# restart worker treats it as "skip the stop" and proceeds rather than aborting.
+_HOST_SHUTDOWN_NOT_SUPPORTED_SIGNAL: Final[str] = "does not support stopping hosts"
```
joshalbrecht: minor: it'd be better to do this in a more robust way, eg, via exit code or an explicitly shared string (ex: some imported constant)

libs/mngr_modal/imbue/mngr_modal/instance.py
```diff
@@ -1988,12 +1997,22 @@ def start_host(
...
-                host_record.certified_host_data.snapshots, key=lambda s: s.created_at, reverse=True
+            # Auto-select the most recent snapshot, skipping the bare "initial" one
+            # (captured right after provisioning, before the agent ran): auto-restoring
+            # it would spin a sandbox with no agent. An explicit snapshot_id can still
+            # name the initial snapshot.
+            resumable_snapshots = sorted(
+                (snap for snap in host_record.certified_host_data.snapshots if snap.name != "initial"),
+                key=lambda snap: snap.created_at,
+                reverse=True,
             )
```
joshalbrecht: I suspect that we should probably, as a separate earlier PR, change this behavior so that we simply avoid that initial snapshot in the first place? 

If it cannot be resumed, what is the point?

(then we wouldn't need this logic here)

libs/mngr_modal/imbue/mngr_modal/instance.py
```diff
@@ -2063,14 +2082,23 @@ def start_host(
...
         self._uncache_host(host_id)
 
+        # Clear stop_reason so a now-running host is no longer derived as PAUSED/STOPPED.
+        # Otherwise a host resumed here and then hard-killed would be mis-derived as
+        # PAUSED/STOPPED (never GC'd, wrongly shows a Resume control) instead of CRASHED.
+        # The setup helper below persists this cleared data to both the host record and
+        # the host's data.json, mirroring what mngr_lima does on restart.
+        restored_host_data = host_record.certified_host_data.model_copy_update(
+            to_update(host_record.certified_host_data.field_ref().stop_reason, None),
+        )
```
joshalbrecht: I don't fully understand why this is needed -- from the rest of this, it looks like our modal sandboxes cannot even be restarted or resumed? So how is restoration playing into this?

libs/mngr_modal/imbue/mngr_modal/instance.py
```diff
@@ -2539,16 +2569,55 @@ def discover_hosts_and_agents(
...
-                ref = validate_and_create_discovered_agent(agent_data, host_id, self.name)
-                if ref is not None:
-                    agent_refs.append(ref)
-
-            result[host_ref] = agent_refs
+            # A RUNNING host's agents are read LIVE off the sandbox; only offline
+            # hosts fall back to the state volume. See _discover_agent_refs_for_host.
+            result[host_ref] = self._discover_agent_refs_for_host(
+                host_id, host_record, is_running, agent_data_by_host_id.get(host_id, [])
+            )
```
joshalbrecht, line 2595: This is actually an interesting larger design question that I'd like @qi-imbue 's thoughts on:

**Should "local" agents on a "remote" host be seen by a remote version of mngr or not?**

- If "yes", it's a little bit strange -- what is the name of the provider for those agents? It's not "local" (that'd be confusing"), and it's not quite <provider-name>, because they're not really such agents...
- If "no", there can be this de-sync between what the remote state volume says, and what the local mngr would see

I *think* I'm leaning towards "no" (ie, revert these changes), but curious if there's something I'm not seeing?

apps/minds/imbue/minds/desktop_client/workspace_recovery.py
MT-GoCode: fixed

libs/mngr_modal/imbue/mngr_modal/instance.py
minhtrinh-imbue: addressed - discussion on slack

libs/mngr_modal/imbue/mngr_modal/instance.py
minhtrinh-imbue: addressed - discussion on slack

libs/mngr_modal/imbue/mngr_modal/instance.py
minhtrinh-imbue: Should "local" agents on a "remote" host be seen by a remote mngr?

They already are, for every other cloud provider — AWS, Vultr, and OVH (all three share mngr_vps's VpsProvider, which does a per-provider live read). Modal was the only one missing it.

On naming: when mngr list returns, each agent is stamped with the host's provider — DiscoveredAgent.provider_name = self.provider_instance.name, i.e. "modal" for everything on a modal host. So both system-services and all other agents spawned in the modal sandbox get named as "modal." 

Should they be visible? Yes. Without it there's a de-sync (as you said) with a concrete cost: mngr destroy sees and terminates system-services (which is on the state volume) but misses the workspace agent (created by the in-sandbox mngr, so never written to the volume). Then the cleanup sweep, which checks the host live via host.get_agents(), still sees that workspace agent, concludes the sandbox is in use, and skips shutting it down. Net: the Modal sandbox keeps running and billing. This happened during testing, which is why I added this.

So this part of the commit isn't modal special-casing, it brings Modal up to spec with AWS/Vultr/OVH, which do this... I'd suggest no reversion.

joshalbrecht: ok, I think that makes sense!  Will wait for Qi to weigh in as well

qi-imbue: It's a bit weird but it seems to be a natural conclusion of mngr's design decisions.

I suppose the alternative here is to segregate local agents created from within the Modal sandbox (e.g. use a different host dir so that they can't be discovered together) but that may also feel a bit weird. Overall LGTM.

PR #2312, opened 2026-06-26
minds: stop cold-start warm-up from falsely restarting a healthy workspace

gnguralnick:
## Problem

On a cold start (or wake from sleep), opening a *healthy* imbue-cloud workspace would sit on "Loading workspace…" for tens of seconds and then needlessly restart the workspace's system services. Root cause, traced from a Sentry bug report + uploaded production logs: the freshly-spawned `mngr forward` has not resolved the workspace's route yet, so probes fail with `UNRESOLVED` / 503 for the ~10s it takes discovery to warm up. That warm-up was mistaken for an outage — the workspace was marked STUCK and the recovery flow auto-restarted it, turning an instant load into tens of seconds plus an unnecessary restart.

## Fixes

**A — ignore `UNRESOLVED` for suspect enrollment.** `UNRESOLVED` means the forward has no route for the agent at all, so a restart (which routes through the forward) cannot help regardless. In practice it is either a cold-start warm-up (self-resolves the moment the forward catches up) or a genuinely-gone agent (a restart cannot revive). A workspace that is present but unreachable does *not* land here: discovery retains its (stale) route, so the dial failure surfaces as `CONNECT_ERROR` / a 5xx, which still enrolls and still drives recovery.

**B — do not auto-restart an already-recovered workspace.** The recovery page's host-health probe runs a slow in-container exec, so the background probe loop can flip the workspace back to HEALTHY while it is in flight. The recovery page's auto-dispatched restarts now carry `auto_dispatched=1`, and `_dispatch_restart` skips the restart (returns 202) when auto-dispatched and the tracker is already HEALTHY. Manual restarts are unaffected and always proceed.

Also in this PR: diagnostic logging across the system-interface health / recovery path (so this is traceable from a future bug report), and removal of a stale recovery-gate FIXME now that the discovery-health watchdog backstops a persistently-stalled pipeline with its `BLOCKED` app-takeover.

## Verification

- **Fix B** confirmed firing in production logs: after a false STUCK, the workspace self-recovered and the guard skipped the restart (`Skipping auto-dispatched system-interface restart … already recovered to HEALTHY`).
- **Fix A**: an earlier cut gated `UNRESOLVED` on `has_completed_initial_discovery()`, but that signal is decoupled from the forward's route-readiness and was disproven in real-app testing (the suspect still enrolled). Corrected to ignore `UNRESOLVED` unconditionally. Worth one more real-app cold-start pass to confirm no enrollment occurs.
- Tests: enrollment-policy matrix + a new auto-dispatched-restart-guard test. Targeted suites green locally; lint / format / `ty` clean. Full suite runs in CI.


joshalbrecht:
lgtm, but in general, see my DM about this class of errors

PR #2313, opened 2026-06-26
minds: fix the in-app bug-report and agent-assist get-help flows on the error/loading takeover screens

gnguralnick:
This branch hardens the in-app get-help / bug-report flow, which surfaces on the full-window "takeover" screens where the workspace content view is replaced (the error takeover, and the "Loading workspace" / recovery screens). It fixes three separate problems.

### Motivation

On the full-app error takeover (the screen users hit when the discovery pipeline goes to the terminal `blocked` state, where the backend is still up), clicking "Report a bug" did nothing. The button is wired to open the in-app `/help` report modal (`canUseBackendReport=true` -> `toggleHelp`), but it was unreachable for two layered reasons:

1. **Invisible:** the error-state window layout in `updateBundleBounds` collapsed every non-chrome view -- including `modalView` -- to `0x0`, so the modal opened with zero size.
2. **Unclickable:** once made visible, the takeover screen (`shell.html`) is a full-window `-webkit-app-region: drag` region (so the frameless window can be dragged by it). macOS unions drag regions across all stacked views, so the OS treated the overlaid modal's pixels as a draggable titlebar and swallowed every click -- the modal looked visible but nothing inside it responded.

### What changed

- `apps/minds/electron/view-layout.js` (new): the per-view bounds math is extracted from `updateBundleBounds` into a pure `computeBundleViewBounds` helper -- no `electron` imports, so it is unit-testable under plain node (mirrors the existing `startup-routing.js` pattern). The visibility fix lives here: in an error state with an open modal, the modal overlays the full window instead of collapsing to `0x0`. Gated on `isErrorState && modalVisible` (not `modalVisible` alone) so a quit -- which leaves `modalVisible` true while hiding the modal via `setVisible(false)` -- still collapses it.
- `apps/minds/electron/main.js`: `updateBundleBounds` now delegates to the helper. Behavior-preserving except the one intended fix.
- `apps/minds/electron/shell.html`: drops its window-drag region while a modal is open (a `body.modal-open { -webkit-app-region: no-drag }` rule toggled by a `modal-state-changed` listener), fixing the clickability bug. This mirrors how the normal chrome titlebar (`chrome.js`) already suppresses its drag region for the same reason; the `modal-state-changed` IPC is already broadcast to the chrome view by `openModal`/`closeModal`, and `main.js` replays the current state on `did-finish-load` so a modal open across a takeover-screen reload isn't missed.
- `apps/minds/test/unit/view-layout.test.js` (new): node unit tests covering the regime decisions (normal inset layout; error+modal -> overlay; error-without-modal -> collapsed; quitting-with-modalVisible -> collapsed; loading -> collapsed).

### Motivation

When the bug-report (`/help`) modal is opened over a "Loading workspace" screen, its textarea lost focus once per second and was impossible to type into. The text already entered stayed put (the modal view is never reloaded), but the focus kept getting cleared.

Root cause: the loading screen re-attempts the workspace by **full-reloading itself every second** -- the `mngr_forward` proxy loader via a `<meta http-equiv="refresh" content="1">`, and the workspace-recovery page's "restarting" state via `window.location.assign`. In Electron, a full reload of one `WebContentsView` steals OS focus from any sibling view layered over it (the report modal sits in a separate `modalView`), so the textarea was blurred on every tick. Electron has no per-`WebContentsView` focus-on-navigation / focusable control to opt out of this: https://github.com/electron/electron/issues/42578.

### What changed

- `libs/mngr_forward/imbue/mngr_forward/server.py`: the proxy loader no longer meta-refreshes. It polls its own URL in the background (`fetch`) and reloads only once the backend answers (status != 503). While waiting, the page -- and any focused overlay above it -- is untouched.
- `apps/minds/imbue/minds/desktop_client/templates.py`: the recovery page's `scheduleRefresh` (the "restarting" convergence loop) now polls in the background too, navigating only on a state change -- a healthy 302 (surfaced as an opaque redirect) or any non-"restarting" status. While the status stays "restarting" it just keeps the spinner up and polls again, no reload.
- `apps/minds/imbue/minds/desktop_client/app.py`: `_handle_recovery_page` sets a new `X-Recovery-Status` response header carrying the rendered status, so the background poll can cheaply tell "still restarting" from a state change without parsing HTML.
- As a side benefit, neither loader resets its spinner animation while waiting anymore.

### Motivation

The get-help modal offers a "have an agent help fix the problem" option that spawns an in-workspace `/assist` chat (via `mngr create` inside the workspace container, run through `mngr exec`). Three robustness gaps:

1. **Offered when it couldn't work:** the option was enabled whenever a workspace URL was showing -- including the "Loading workspace" / recovery screens, where the workspace backend is unreachable -- so the spawned chat couldn't be seen or used.
2. **Hung on old templates:** a workspace created from an FCT template predating the `/assist` skill would accept the `mngr create` but hang for the full send timeout on the unknown `/assist` command (an unknown slash command never submits a prompt, so no `UserPromptSubmit` hook fires), leaving a half-created chat behind.
3. **Dead spinner on failure:** on any failure the modal just kept spinning, with no way back.

### What changed

- **Health gate.** `_handle_help_page` (`app.py`) now reads an `assist` query param; the titlebar (`chrome.js`, via `main.js`/`preload.js`) only sets `?assist=1` when the displayed workspace's system-interface is healthy. On a loading/stuck workspace the agent-help option shows disabled ("Available once this workspace is responding") while a bug report stays scoped to that workspace. Threaded through `render_help_page` and `Help.jinja` (`assist_available`).
- **Skill probe before spawning.** New `apps/minds/imbue/minds/desktop_client/assist_chat.py` helpers (`check_assist_support` / `build_assist_support_probe_args`, `AssistSupport` enum): a quick `mngr exec` filesystem check for `.agents/skills/assist/SKILL.md` inside the container, using echoed present/absent sentinels to distinguish "reachable but old template" from "probe never ran (unreachable)". `_handle_help_assist` calls it before spawning and returns **409** (unsupported / old template) or **502** (unreachable) with a human-readable `error` message, instead of spawning a chat that could only hang.
- **Error screen instead of a stuck spinner.** On any failure (skill missing, workspace unreachable, or the create erroring), `Help.jinja` now swaps the modal to an error screen with the reason and a "Back to report" button.
- **Testability.** `create_desktop_client` / state now accept an injectable `mngr_caller` (`state.py`), so `_handle_help_assist` uses `state.mngr_caller or get_default_mngr_caller()`. Covered by `assist_chat_test.py` (new) and additions to `test_desktop_client.py`.

## Incidental

- `libs/mngr_claude/imbue/mngr_claude/plugin_test.py`: gave `test_provision_preserves_existing_remote_project_config` the same 30s pytest-timeout its sibling already carries (it drives the real local `claude --version` subprocess check, which can exceed the default 10s under CI load and was flaking).

## Test plan

- [x] `node --test apps/minds/test/unit/view-layout.test.js` -- passes locally.
- [x] `node --check` on the proxy-loader and recovery-page poll scripts -- syntax OK.
- New/updated automated coverage: `apps/minds/.../assist_chat_test.py`, additions to `apps/minds/.../test_desktop_client.py` (assist route + `X-Recovery-Status` header), `libs/mngr_forward/.../server_test.py`, `apps/minds/.../templates_test.py`. Authoritative full-suite run is via CI.
- [x] Real-app verification of Fix 1: opened the app, killed `mngr forward` to reach the takeover, clicked "Report a bug" -> modal opens, is visible, and its interior is clickable / submittable.
- [ ] Real-app verification of Fix 2 (typing into the report form over a live loading/recovery screen) -- recommended before merge, since Electron focus behavior can't be confirmed by unit tests alone.
- [ ] Real-app verification of Fix 3 (agent-help disabled over a loading workspace; 409/502 error screen on an old-template / unreachable workspace) -- recommended before merge.
- [ ] Full suite green via CI.


libs/mngr_forward/imbue/mngr_forward/server.py
```diff
@@ -450,11 +450,44 @@ def _emit_backend_failure(
...
+              }
+            }, function () {
+              setTimeout(poll, INTERVAL_MS);
+            });
+        }
+        setTimeout(poll, INTERVAL_MS);
+      })();
+    </script>
+"""
+_SERVICE_UNAVAILABLE_HTML = render_loading_page(body_extra=_LOADING_POLL_SCRIPT)
```
gnguralnick, line 490: this is definitely increasing the complexity of this page. but electron auto-steals focus when webcontentviews refresh (in a non-blockable way) so i think it's needed to fix the issue people were having.

apps/minds/imbue/minds/desktop_client/assist_chat.py
```diff
@@ -28,6 +29,75 @@
...
+    workspace missing the skill from a probe that never ran (host unreachable).
+    """
+    check = (
+        f"if [ -f {shlex.quote(_ASSIST_SKILL_PATH)} ]; "
+        f"then echo {_ASSIST_PRESENT_SENTINEL}; else echo {_ASSIST_ABSENT_SENTINEL}; fi"
+    )
+    return ["exec", "--agent", str(workspace_agent_id), check]
+
+
+def check_assist_support(mngr_caller: MngrCaller, workspace_agent_id: AgentId) -> AssistSupport:
```
gnguralnick, line 78: claude code rejects slash-command messages for skills whose files are missing, so without this if you use this feature on an old workspace it'll just spin and load until the message send times out. the other options here are:
- don't use a slash command - have the message be "use the /assist skill. then address: ..."
- update mngr message to treat read the TUI failure text ("Unknown command: /assist") and treat it as a message send failure

happy to do that mngr change now (instead of this probe) or as a follow-up

joshalbrecht: I'm ok with your current approach for this!

libs/mngr_forward/imbue/mngr_forward/server.py
joshalbrecht: I'm ok with this as well

joshalbrecht:
lgtm

feel free to merge once you've fixed the conflict

PR #2314, opened 2026-06-26
minds: live hostname validation in the create-a-mind form

gdenisov:
## What

The advanced **Name** field on the create-a-mind form now validates live as you type, instead of failing partway through creation.

- **Format**: instant, rule-specific messages mirroring mngr's `HostName` rules (no dots, no leading/trailing dash/underscore, only letters/numbers/dashes/underscores).
- **Availability**: a new read-only `GET /api/host-name-available` endpoint reports whether a name is already taken by an *active* workspace on the **selected provider instance** (and account, for Imbue Cloud), reading the discovery snapshot only (no provider/subprocess call).
  - Destroyed-mind names are treated as reusable; matching is case-insensitive.
- **Submit**: clicking Create on a known-invalid/taken name surfaces the inline error and blocks submission (client-side, mirroring the account-picker guard) rather than starting a doomed creation. The existing create-time `HostNameConflictError` remains the last-resort backstop for the rare race.

`provider_instance_name_for_launch` is extracted as the single source of truth that both the create command's address and the availability check share, so "taken" always matches the scope where the create-time conflict actually fires.

Plan: `blueprint/hostname-live-validation/plan-hostname-live-validation.md`.

## Tests

- New: `app_host_name_availability_test.py` (provider/account scoping, destroyed-exclusion, case-insensitivity, auth) and `provider_instance_name_for_launch` tests in `agent_creator_test.py`.
- Local: full `desktop_client` subtree (1070 passed), minds ratchets (56), `ruff`, `ty check`, changelog gate all green.


joshalbrecht:
lgtm I guess

this is slightly (~10%) better UX than I *want* for right now on this form (bc such UX comes with additional complexity, eg, the code in this PR, that will then compound the difficulty of refactoring, and I suspect we will have a bit more refactoring before we're happy with the basics)

merging anyway, but just flagging -- we don't need to make all of these details perfect quite yet -- there's enough other broken things that we really probably should address a lot of those before we get to stuff like this

thanks for fixing it though! I appreciate the instinct to make the user experience good :)

joshalbrecht:
please merge after fixing the conflicts!

PR #2307, opened 2026-06-26
minds workspace API: post-merge work

joshalbrecht:
Continuation of the minds workspace-API work, based on `josh/pre_final_minds_api` (the post-`main`-merge snapshot) so this PR shows only what changed *after* the merge.

## Contents so far

- **docs(minds-workspace-api):** point the blueprint plan + handoff at the shipped latchkey permission design (the `minds-workspaces` scope whose scope/verb/per-target schemas arrive with each grant and merge by name via `computeWorkspaceEffect` — no startup schema-sync, no per-host `anyOf` allowlist). Canonical reference: `apps/minds/docs/latchkey-permissions.md`.
- Handoff rewritten to current status: FCT telegram teardown, per-target permissions, and create backup/tunnel parity are done; the UI repoint is mostly done (remaining old routes enumerated); the SSH remote→local broker + wiring `prune_expired_grant_lines` is the main outstanding item.

Docs only so far; no code changed in this PR.

> Note: the `main`-merge itself (and its conflict resolutions) lives in the base branch `josh/pre_final_minds_api`, intentionally kept out of this diff.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2284, opened 2026-06-25
minds: FCT telegram teardown, /api/v1 create parity, UI repoint

joshalbrecht:
Continues the minds workspace API work (handoff tasks #1, #4, #2). Stacked on `josh/minds-api-capabilities`.

## #1 — FCT telegram teardown (forever-claude-template)
Done in the external worktree `.external_worktrees/forever-claude-template` (branch `mngr/continue-minds-api`, committed separately in that repo). Deletes `libs/telegram_bot/`, the `send-telegram-message` / `read-telegram-history` skills, the `TELEGRAM_*` pass-env vars, the `telegram-bot` dep/workspace-member/source + Dockerfile copy; regenerates `uv.lock`. `send-user-message` falls back to inline as its sole channel. Docs/docstrings scrubbed.

## #4 — create backup + tunnel parity
Extracts the create-orchestration helpers from `app.py` into a new shared `workspace_create.py` (imported by both `app.py` and `api_v1.py`, breaking the import cycle). `POST /api/v1/workspaces` now accepts the `backup_*` fields, provisions restic backups, injects the Cloudflare tunnel token, and resolves/persists the region — matching the desktop create form.

## #2 — browser-JS repoint
Lifecycle (start/stop), destroy (settings button + detail-page status poll / SSE log tail / retry), and backup-export now call `/api/v1/workspaces/...`. Legacy `/api/*` JSON routes stay registered during the transition.

**Deliberately kept on old routes (v1 has no clean equivalent):** the create page (`creating.js`) needs the server-computed `redirect_url` + `status_text`, which the agent-facing v1 ops API omits by design; the backup-status badge needs a live status summary that v1's snapshot list doesn't provide.

## Testing
- `desktop_client` suite (fast markers): 1039 passed; api_v1 21 passed (+2 new); minds ratchets + repo-wide import-layer: green.
- FCT root 12 passed; cloudflare_tunnel 14 passed; system_interface 483 passed.
- **#2 needs Electron verification** (pytest can't exercise browser JS) — pending.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2276, opened 2026-06-25
Minds workspace API: cross-workspace capabilities via latchkey (WIP)

joshalbrecht:
WIP — implementing the versioned Minds workspace API in reviewable, independently-green slices. Full design in `blueprint/minds-workspace-api/plan-minds-workspace-api.md`.

## Goal

Expose minds workspace/backup/version/SSH capabilities through a versioned `/api/v1/workspaces/...` API that in-workspace agents reach through the latchkey gateway's `minds-api-proxy`, gated by a new `minds` detent scope with per-verb × per-target (none/all/selected) permissions. Also refactors the UI onto the same API and removes telegram entirely (minds + FCT).

## Progress

- [x] **Slice 1 (P4 foundation):** restic snapshot listing — `restic_cli.list_snapshots` + `ResticSnapshot` + `parse_restic_snapshots`; pure-parser unit tests + local-repo integration test.
- [ ] Slice 2: `original_minds_version` label at create
- [ ] Slice 3: telegram teardown (minds side)
- [ ] Slice 4: `minds` detent scope + per-target allowlist + UNKNOWN grant-flow fix + startup schema sync
- [ ] Slice 5: `/api/v1/workspaces` resource + UI refactor onto it
- [ ] Slice 6: version-read / backups / SSH routes
- [ ] Slice 7: FCT external worktree (parent.toml pin, update-self merge messages, telegram teardown)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2315, opened 2026-06-26
minds: cross-workspace SSH, /api/schema, and spectree validation of /api/v1

joshalbrecht:
## Overview

This PR builds out the **minds cross-workspace `/api/v1` API** — a versioned, self-describing, latchkey-gated surface that an agent in one workspace can use to act on the desktop client and on *other* workspaces — and converts that surface to **spectree + pydantic** validation as the single source of truth. It also consolidates the old ad-hoc `app.py` routes onto `/api/v1`, and carries a set of supporting mngr/infra fixes uncovered along the way.

Originally stacked on `josh/more-minds-api`; **rebased onto `main`** so it subsumes the whole stack in one reviewable PR (see "Stacked on" below).

### Cross-workspace `/api/v1` workspace API
- **Read**: list/get workspaces, version, backups (restic snapshot listing).
- **Lifecycle**: create / destroy / start / stop / restart, with operation handles polled at type-segmented `/operations/<type>/<id>` (+ `/logs` streamed as SSE).
- **SSH**: `POST /workspaces/<id>/ssh` establishes scoped, TTL'd SSH access into a target; grants are pruned + refreshed (never stacked); a **reverse-tunnel broker** lets a workspace reach a *local* (Docker/Lima) target, not just remote; requester-id/`authorized_keys` injection guards.
- **Settings/sharing/recovery**: color + account association, Cloudflare sharing enable/disable/readiness, host-health probe + restart tiers.
- **Account association** now validates the value against the signed-in accounts and accepts **either the account id or its email** (404 on no match, instead of the previous silent false-success); the workspace detail surfaces `account_id`/`account_email`.
- **`GET /api/v1/accounts`** lists the device's signed-in accounts so an agent can turn a known email into an id — gated by a **must-ask `minds-accounts-read` permission** (modelled on file-sharing: a fixed permission under `latchkey-self`, **not** in the baseline, granted on request).

### Latchkey gating
- Cross-workspace routes are gated by a dedicated **`minds-workspaces` detent scope** with one grantable verb per capability (read/create/destroy/lifecycle/backups-export/ssh/update/recover/sharing), target-scoped per workspace where applicable; the verb catalog is a single shared JSON read by both the gateway extension and Python.
- Fixes the domain-only `latchkey-self` catch-all to stay **last** so narrower same-domain grants take effect.

### spectree + pydantic conversion
- Extracts the `/api/v1` request/response models into `api_models.py` (single source of truth), adds the spectree instance + a 422 `{"errors":[{field,message}]}` hook, then **validates request bodies** and **enforces responses** via pydantic on the core routes, with a drift-guard test asserting the documented model equals the enforced one.

### Self-describing schema
- **`GET /api/schema`** returns an OpenAPI 3.1 document generated from the live routes (so it can't drift), default-allowed so an agent can discover the API.

### Route consolidation
- Collapses the legacy `app.py` routes onto `/api/v1` (single browser create path, host-lifecycle extraction, removal of `/api/create-agent` and other dead routes), and repoints the Electron shell + e2e harness onto the versioned routes.

### Detail/recovery enrichments
- Immutable create-time labels surfaced on the detail: `original_minds_version`, `original_branch`, `git_url`.
- Recovery: adds a **`HEALTHY` dispatch tier** and makes the classifier consult the live in-container HTTP probe, so a responsive workspace isn't misclassified `interface_unresponsive` (and isn't needlessly restarted).

### Cleanup
- Removes all **Telegram** functionality from the desktop client and the `MINDS_WORKSPACE_NAME` dev override.
- Adds a headless-Electron (xvfb) full-lifecycle **e2e harness** and a testing-overview doc.

### Supporting mngr / infra fixes (separate libs)
- **`mngr_forward`**: respawn a dead per-agent events stream — fixes a workspace permanently stuck on "Loading workspace" (503) after a host restart.
- **`mngr` docker/gVisor**: refuse a gVisor host whose rootfs would be ephemeral (`--overlay2` not `none`), with actionable remediation incl. the `--tmpfs /run` note for supervisord images.
- **`mngr_aws`**: bound boto connect/read timeouts + disable the IMDS credential probe by default, so a slow/blackholed AWS endpoint can't freeze discovery.
- **`mngr`**: name the slow/hung provider in parallel discovery; test-helper/flaky-timeout fixes.

## Stacked on / supersedes

This work was developed as a stack; with the base now `main`, this PR contains all of it. Prior associated PRs:
- #2276 — Minds workspace API: cross-workspace capabilities via latchkey (WIP)
- #2284 — minds: FCT telegram teardown, /api/v1 create parity, UI repoint
- #2297 — Workspace permissions
- #2307 — minds workspace API: post-merge work

## Changelog
Per-project changelog entries are included for every touched project (`apps/minds`, `libs/mngr`, `libs/mngr_aws`, `libs/mngr_forward`, `libs/mngr_latchkey`, and `dev`).


joshalbrecht:
lgtm

PR #2319, opened 2026-06-27
minds: per-run CI environment + manual release tier for the snapshot test pipeline

joshalbrecht:
Adds a per-run minds `ci-*` environment to the CI snapshot test pipeline so live tests can run against a real, isolated cloud env that is created before the run and destroyed after, plus a manual release tier for the heavier deploy tests.

Spec: `blueprint/minds-ci-env-pipeline/plan-minds-ci-env-pipeline.md`
Phase 0 (Vault/GitHub) runbook: `blueprint/minds-ci-env-pipeline/phase-0-runbook.md` (deployed out-of-band: `imbue-ai/vault` role `minds_ci_env_gh`, GitHub Environment `minds-ci-env`, `CI_TEST_USER_*` credentials, dedicated `minds-dev` Modal service token).

## Orchestrator (`apps/minds/scripts/test_deployments.py`)

Replaces the previous stubs with the real lifecycle:

- `_deploy_shared_env`: deploys a `ci-*` env, creates a fixed verified CI test user against it, and publishes the env's freshly-minted per-env dynamic secrets (its own SuperTokens app + Neon DSNs) to Vault.
- `_destroy_env`: destroys an env by name, reconstructing `secrets.toml` from Vault (`_reconstruct_env_secrets_file`) so teardown works on a different runner than the one that deployed, then deletes the per-env Vault secrets.
- `_sweep_stale_envs` + new `sweep` command: enumerates leaked `ci-*` Modal envs and destroys ones older than a threshold (timestamp parsed from the env name), as a backstop for failed per-run destroys.
- Raises the pytest global-duration deadline (`PYTEST_MAX_DURATION_SECONDS`) for the deployment/services suites, which run for minutes; softens the FCT-worktree check to a warning (no current test needs it).

## Secret handoff and fixtures

- Per-env dynamic secrets are handed between jobs via an env-name-keyed Vault path (`secrets/minds/ci/runs/<env-name>/shared-<role>`), so the build, test, and destroy jobs (which may land on different runners) all resolve the same values.
- The `shared_env` fixture resolves these secrets from injected env vars when present (the CI path) and otherwise from Vault (local runs); a new `ci_test_user` fixture supplies the fixed CI credentials. The same test body runs in both places.

## Tests and tiers

Two capability marks mapped onto two tiers (see `apps/minds/deployment_tests/README.md`):

- **Integration** (every push, in `test-minds-snapshot`): `minds_snapshot_resume` runs via offload — it needs only the built snapshot image, not a remote ci env.
- **Release** (manual, `workflow_dispatch` with `run_minds_release_tests=true`): everything that needs a remote `ci-*` env. `minds_services` (incl. new `test_ci_env_litellm`: log in as the fixed CI user → mint a LiteLLM key → make a live `claude-haiku-4-5` call) runs in `test-minds-snapshot` against the per-run env from `build-minds-ci-env`; `minds_deployment` (`test_deploy_new_version`, `test_deploy_rollback`, `test_deploy_round_trip`) runs in `test-minds-release`, each minting + destroying its own ephemeral env.

Gating the entire remote-ci-env path behind a single opt-in means **a normal push to any branch never stands up a `ci-*` env** (Modal env + Neon project + SuperTokens app); it only builds the snapshot image and runs the resume tests. The full suite is one dispatch away: `gh workflow run ci.yml -f run_minds_release_tests=true --ref <branch>`.

`test_litellm_via_workspace` and `test_signup_tunnel` are wired into the flow but remain `@pytest.mark.skip`ped with explicit notes: their bodies are still stubs and need debugging (FCT Docker workspace creation, Cloudflare tunnels, the mail.tm signup flow) before they will pass.

## `minds env recover` bug fix

`test_deploy_rollback` surfaced a real recover gap: after an auto-rollback, the rolled-back app's broken containers were never terminated because the Modal app-id lookup matched the app name against `Name`/`name`/`App`, but `modal app list --json` reports the name under `Description`. The lookup silently found nothing, so the failed version's containers kept serving (and `/version` kept reporting the failed `deploy_id`) until they idled out. `_find_modal_app_id` now also checks `Description`; the parsing is extracted into a pure `_select_deployed_app_id` with regression unit tests against the real JSON shape. The rollback test is un-skipped and passing.

## CI workflow (`.github/workflows/ci.yml`)

- New jobs: `build-minds-ci-env` (deploy the per-run env), `cleanup-minds-ci-envs` (parallel leaked-resource sweep), `destroy-minds-ci-env` (per-run teardown after offload). These use the `minds-ci-env` GitHub Environment, the `minds_ci_env_gh` Vault role, and dedicated `minds-dev` Modal service tokens.
- `test-minds-snapshot` runs the snapshot-resume tests on every push and additionally runs the `minds_services` step (role `minds_ci_test_gh`) only on the release opt-in.
- New `workflow_dispatch` trigger with a `run_minds_release_tests` input. It gates the entire remote-ci-env path: `build-minds-ci-env`, `cleanup-minds-ci-envs`, `destroy-minds-ci-env`, the `minds_services` step in `test-minds-snapshot`, and `test-minds-release`. Off by default, so normal pushes create no ci env. `test-minds-snapshot` uses `!cancelled()` + a `build-minds-snapshot` success gate so the (normally skipped) ci-env build doesn't skip the snapshot-resume tests.
- offload pin unified to 0.9.10 across the minds-snapshot jobs.

Verified locally end-to-end (deploy → login → mint key → live LLM call → destroy) and in CI: the integration tier is green on every push, and a `workflow_dispatch` release run passed all three `minds_deployment` tests. No leaked `ci-*` envs.

## Electron e2e: product-level create+chat test, consolidated into the snapshot stage

- New `test_create_apikey_workspace_and_chat_via_electron` (in `apps/minds/test_snapshot_resume.py`, mark `minds_snapshot_resume`): drives the **real Electron desktop app** to create a local Docker workspace via the manual **`api_key`** AI provider (typing a real Anthropic key into the create form), then sends a chat message and asserts the agent in the workspace's `system_interface` answers — end-to-end through the Electron renderer and the desktop client proxy. The Anthropic key is read from `ANTHROPIC_API_KEY` (already fetched from Vault into the snapshot stage) and typed into the form; the Electron child env scrubs that var, so the key reaches the agent only via the form, exercising the real manual-key UX.
- The create driver (`e2e_workspace_runner.py`) gained an `API_KEY` AI-provider path and an `on_workspace_ready` hook so the chat round-trip runs on the same attached Electron session. A new `xvfb_display` conftest fixture supplies a display in the offload sandbox (it starts Xvfb via `-displayfd`; local `xvfb-run` runs are detected and reused).
- **Consolidation:** the dedicated `test-docker-electron` CI job (which cold-installed Node/pnpm/Electron/xvfb on every push) is removed; all Electron e2e now runs in `test-minds-snapshot`, reusing the snapshot image's already-warm Electron/Playwright/Xvfb toolchain. Deleted the standalone `test_desktop_client_e2e.py` and the `minds_electron` marker; `just minds-test-electron` now runs the consolidated test locally.

Validated locally (the create+chat test passes both under `xvfb-run` and with `DISPLAY` unset, i.e. the fixture's Xvfb-start path used in the sandbox) and in CI's `test-minds-snapshot` stage.


joshalbrecht:
lgtm after:
1. updating PR description
2. adding a few release tests
3. disabling the env creation unelss we're running the release tests (I'm worried that creating one env per push is going to start causing problems)

PR #2328, opened 2026-06-29
mngr_latchkey: leash discovery producer to the embedder instead of reaping orphan forwards

gnguralnick:
## Problem

PR #2285 fixed duplicate background discovery producers (multiple `mngr observe --discovery-only` processes writing the one shared `discovery-events.jsonl` with divergent, frozen-at-startup provider config) — the cause of "disabled providers keep erroring" and a running mind intermittently disappearing. It did so by **hunting the process table on every minds startup** and reaping every other `mngr latchkey forward` bound to the same latchkey directory (plus that forward's descendants).

That approach is the parent policing orphans after the fact. The orphans still exist for the entire time the app is closed (and after a crash, until the next launch sweeps them). This PR removes the root cause instead.

## Approach

The `mngr latchkey forward` supervisor bundles two responsibilities with **opposite ideal lifetimes**:

- the **discovery producer** (`mngr observe`) — the thing that, when duplicated/stale, causes the bug, and that nothing-while-closed depends on;
- the **gateway + reverse tunnels** — which *should* outlive the app so agents stay reachable and can still deposit permission requests (persisted to disk) while the app is closed.

So we split them by lifetime:

- **core `mngr`**: `mngr observe` gains `--exit-alongside-pid PID` — it SIGTERMs itself when that PID exits, via a cross-platform `os.kill(pid, 0)` watcher (`start_pid_death_watcher`). This is needed because the producer's *parent* is the surviving supervisor, so the existing `getppid()` parent-death watcher would watch the wrong process; and the laptop may be macOS, so the `/proc`-based grandparent watcher doesn't apply.
- **`mngr_latchkey`**: `mngr latchkey forward` gains `--observe-exit-alongside-pid PID`, forwarded to its `mngr observe` child **only**. The detached forward and its gateway/tunnels are not tied to the PID. The startup reaper (`_reap_duplicate_forwards` + process-table scanning, `--latchkey-directory` cmdline parsing, descendant capture, PID-reuse guards) is deleted.
- **`minds`**: passes its backend PID. A prior instance's producer now dies with the app — clean quit, or an Electron crash that the grandparent watcher turns into a SIGTERM — instead of lingering to race the shared file. The fix also lands *earlier* (on app exit) than the old next-launch sweep.

## Why `restart()`-on-launch is kept (not dropped)

It's natural to ask: if the gateway is meant to survive a restart, why does minds still tear it down and respawn the supervisor on every launch? Because `restart()` is load-bearing for the gateway's `minds-api-proxy` extension:

- The proxy authenticates to the minds backend with a bearer key that is **regenerated every launch and never persisted** (`run.py`), and points at the backend port, which **may change across launches**.
- Both reach the gateway only through the supervisor's `extra_env` at **spawn time** → gateway env → extension. Adopting an old supervisor would leave its gateway holding the previous launch's key/port, so agents' minds-API calls would **401 after every reopen**.
- `restart()` also guarantees current code after an upgrade, and re-leashes the new producer to the new backend PID.

Keeping `restart()` does **not** conflict with this change: the gateway survives the *closed* window (where its value is — permission requests deposited then are persisted to disk and survive the rebuild), and gets a clean, correctly-keyed start on *reopen*.

## Behaviour change / caveat

While the app is closed there is now no producer, so the shared discovery file goes stale. Existing agents keep their tunnels; the gateway keeps serving (and persisting) permission requests. The one regression: a **brand-new agent that first appears while the app is closed** won't get a reverse tunnel until the next launch re-runs discovery. New agents generally originate from the app, so this is an accepted edge.

## Test plan

- [x] `libs/mngr` `parent_process_test.py`: `start_pid_death_watcher` starts its thread, stays alive while the watched PID lives, and SIGTERMs the process once the watched PID dies. (5 passed, 1 env skip)
- [x] `mngr_latchkey` `discovery_stream_test.py`: `_observe_command` appends `--exit-alongside-pid` only when set.
- [x] `mngr_latchkey` `_spawn_test.py`: `spawn_detached_mngr_latchkey_forward` forwards `--observe-exit-alongside-pid` into the forward argv.
- [x] Removed the ~316 lines of reaping tests; `forward_supervisor_test.py` / `cli_test.py` still green.
- [x] Local: `parent_process`+`discovery_stream`+`_spawn`+`forward_supervisor`+`cli`+`observe`+`minds run` tests pass; both `test_ratchets.py` suites pass (no new violations).
- [ ] **Manual cold-start re-test** (unit tests use fakes): launch minds → exactly one `mngr observe`; quit → the `mngr observe` producer is gone but `mngr latchkey forward` + `latchkey gateway` survive; reopen → a single fresh producer; disabled providers no longer error.


gnguralnick:
dropping this - we've decided to allow `mngr observe` processes to stick around while the app is closed, and the whole `mngr forward` tree will be reaped on next startup

PR #2329, opened 2026-06-29
minds-launch-to-msg: raise build timeouts; split slack binary link into todesktop(mac-arm64)

weishi-imbue:
## Why

The scheduled `minds launch-to-first-message` health check has been failing in the **`build`** job on back-to-back fresh builds (runs [28312671748](https://github.com/imbue-ai/mngr/actions/runs/28312671748), [28389420591](https://github.com/imbue-ai/mngr/actions/runs/28389420591)). Both hit the `pnpm dist` step's **45-min timeout** while ToDesktop was *still actively notarizing* — steady progress, no wedge.

Evidence from the run logs:

- A fresh dual-arch ToDesktop cloud build runs **~43 min even with a fast source download**: serial sign + notarize x64 + arm64, then a **~15-min final packaging phase** (`Packaging arm64 (41%)` -> `Build complete` measured at ~15 min on a passing run). So ~43 min is the *floor*, not a rare tail.
- When ToDesktop's `Downloading application source` phase stalls (~16 min, seen in both failing runs), `pnpm dist` needs **~60-62 min**.
- Fresh-build durations have crept right up to the ceiling: 24, 25, 28, 30, 31, 35, 38, **43.4 (barely passed)**, then **45.5 / 45.7 (timed out)**.

Because the workflow **cancels the in-progress ToDesktop build on timeout**, the SHA never completed a reusable build, so every subsequent run rebuilt from scratch and timed out again — a permanent fail loop for that SHA.

## What

- `pnpm dist` step timeout: **45 -> 70 min**
- `build` job timeout: **60 -> 85 min**

The 70/85 split preserves the ~15-min margin the post-failure cancel step needs to tear an orphaned build down, so a genuine signing wedge still fails fast at the step level rather than burning the whole job budget.

Notarization ("up to 30 min per arch") remains the long pole; this only widens the budget to cover the observed worst case with margin.

## Not addressed here

- x64 Mac + Linux targets are built serially alongside arm64 (set in the ToDesktop dashboard, not `todesktop.js`). Dropping x64 would roughly halve the Mac build but is a product decision (Intel Mac is a supported run target per `desktop-app.md`).


weishi-imbue:
Added a second, unrelated tweak to the same workflow file (commit e9ccf5071): the Slack notification's `binary` link (which pointed at the ToDesktop dashboard) is now rendered as **`todesktop(mac-arm64)`** — `todesktop` links to the dashboard build page, `mac-arm64` links to the arm64 `.dmg` download.

Both URLs are derived from the build's `app_zip_url`. Example render:

```
[<…/apps/26032588hqdzk/builds/260629qmkn4071s|todesktop>(<…/builds/260629qmkn4071s/mac/dmg/arm64|mac-arm64>) | <run logs> | <mngr commit> | <fct commit>]
```

joshalbrecht:
lgtm

PR #2331, opened 2026-06-29
discovery: carry agent lifecycle state through DiscoveredAgent

gnguralnick:
## What

`DiscoveredAgent` now always carries the agent's lifecycle `state`, and the discovery event stream always reports a real one.

- **Full-listing snapshots** carry the state the listing already computed (`discovered_agent_from_agent_details` copies `AgentDetails.state`). Previously this was dropped during conversion, so `mngr observe` consumers couldn't tell a stopped agent from a running one without re-listing.
- **Incremental `agent_discovered` events** (emitted by create/start/stop/archive/cleanup and rename) now probe each agent's liveness in `emit_discovery_events_for_host` before emitting, so they carry a real state instead of none. `discover_agents()` itself stays cheap (no probe) for its other callers; only the emit boundary probes.
- The field is **non-optional** now, with an `UNKNOWN` default. References built without a probe read `UNKNOWN` ("present per persisted records, liveness not checked"); offline hosts and offline rename report `STOPPED` (a down host cannot have a running agent process); online rename probes the renamed agent.

## Why

Consumers of the discovery stream (notably the minds system interface, via `mngr observe`) need to reflect whether an agent's process is actually running. With the state optional, the only path that reached them without one was the incremental event, which forced the consumer to guess `RUNNING` for a just-surfaced agent -- so a just-*stopped* agent briefly read as running until the next snapshot. Probing at the emit boundary (a rare, already-connected lifecycle action) removes the guess for one cheap probe per agent, and making the field required means no consumer has to invent a value.

## Notes

- **Wire compatibility:** older discovery-event lines with a missing or explicit-null `state` still parse, reading as `UNKNOWN` (default + a null-coercing `field_validator`).
- Documented that `DiscoveredAgent.certified_data` is the full raw `data.json` on the persisted-read paths (`discover_agents`, offline listings, rename) but a reconstructed subset on the full-listing path (`type`, `work_dir`, `command`, `create_time`, `start_on_boot`, `labels`, `plugin`).

## Tests

`libs/mngr/imbue/mngr/api/discovery_events_test.py`, `hosts/offline_host_test.py`, and `primitives_test.py` pass (147 locally), including new coverage that the emit path probes and emits a real state, that missing/null `state` lines still parse as `UNKNOWN`, and that offline discovery reports `STOPPED`.


joshalbrecht:
I don't love this -- it makes the interface too complicated ("this field can be None if the log is old *or* if the agent was offline" is too complicated for a human or LLM to properly code around)

Could we *always* set the state?  Don't we already know enough information to properly set the state when an agent is offline too?

Then the only reason it would be None is "this is from a historical log entry", and in *that* case, perhaps we could make the field non-Optional, and simply give the agent a state of UKNOWN in that case?

That would be much easier for consumers to work with

gnguralnick:
let me check over where else this instance gets built to see if/when the value would be None and how we can avoid that. having it be non-optional makes sense to have things be clearer

gnguralnick:
Closing in favor of a different approach. Carrying lifecycle `state` on `DiscoveredAgent` meant the continuous `mngr observe --discovery-only` producer had to emit it, but that path is a lightweight metadata stream and does not probe liveness -- so it emitted `UNKNOWN` for every agent, which the system_interface consumer then trusted, breaking both the liveness dot and the activity indicator. Rather than make the discovery poll probe (heavier, wrong layer), the system_interface now runs its own small per-agent lifecycle poll for process-liveness and the discovery stream stays metadata-only. That removes the need for this change.

PR #2332, opened 2026-06-29
Auto-sync FCT vendor/mngr to mngr main (daily)

weishi-imbue:
## Why

`forever-claude-template`'s `vendor/mngr` is only refreshed at release time, so between releases it drifts hundreds of commits behind mngr `main` (today: 337 commits / 3 days). The twice-daily `minds-launch-to-msg` health check builds the binary from mngr main but creates the in-VM agent from **FCT main's committed `vendor/mngr`** — the further they drift, the more the e2e risks a vendor-skew wedge (binary expects an API the stale vendored mngr doesn't have).

## What

Adds `.github/workflows/sync-fct-vendor-mngr.yml`:

- **Twice daily** (15:00 / 03:00 UTC, an hour before each launch-to-msg cron) it re-archives mngr `main` into FCT `vendor/mngr` using the same `git archive` flow as `just sync-vendor-mngr`, and **pushes the refresh straight to FCT `main`**.
- **No-ops** when mngr main hasn't moved since the last sync (no empty commits).
- Runs on free `ubuntu-latest`; the archive is seconds.

So the health check sees a fresh, coherent `(binary, vendor)` pair, and FCT main stops drifting.

## Prerequisite: `FCT_VENDOR_SYNC_TOKEN` secret

The workflow's built-in `GITHUB_TOKEN` is scoped to mngr only, so the cross-repo push needs a secret:

- **Fine-grained PAT**, resource owner `imbue-ai`, repository access **only `forever-claude-template`**, permission **Contents: Read and write** (nothing else).
- Add as a repo secret named `FCT_VENDOR_SYNC_TOKEN` in mngr.

A PAT push (vs `GITHUB_TOKEN`) also lets FCT's own CI fire on each sync commit as a post-push signal. The `check FCT token` step fails fast with instructions if the secret is missing.

## Notes

- Direct-push (not a gated PR) by design: vendor refreshes are generated/not-reviewed, production pins the `minds-v*` tag (not FCT main), and FCT CI still runs post-push. Blast radius of a transient bad vendor is the health check + dev clones, not shipped app users.
- Validated locally: the archive/extract steps reproduce mngr main's tree byte-for-byte (modulo FCT's `**/.minds/` ignore, the one expected delta).


weishi-imbue:
Moving this to the FCT repo instead: since mngr is public, an FCT-side workflow can clone mngr with no credential and push to its own main via the built-in GITHUB_TOKEN -- zero secrets needed. New PR will be on forever-claude-template.

PR #2333, opened 2026-06-30
mngr create --reuse: drop redundant target_host_ref, scope by address host only

weishi-imbue:
## Context

Follow-up to PR #2085 review comment [r3390807364](https://github.com/imbue-ai/mngr/pull/2085#discussion_r3390807364):

> I think `target_host_ref` is now redundant once you pass in `address.host_name`. Everything can be now derived from the address.host_name. (See at #1694)

PR #2085 fixed the `--reuse` ambiguity bug by threading `address.host_name` into `_try_reuse_existing_agent`, but left the pre-existing `target_host_ref` parameter in place, so the function scoped by host twice.

## Why `target_host_ref` is redundant

`target_host_ref` was only ever non-`None` when `target_host` is a `DiscoveredHost`, which happens only on the existing-host path of `_parse_target_host`, where the host is resolved via `_find_existing_host(address.host_name, address.provider_name, ...)`. That resolver returns **exactly one** host matching the address (raising before reuse is ever reached if the address is ambiguous). So whenever `target_host_ref` was set, the `host_name` filter (plus `provider_name`) already narrowed the candidate set to that exact same single host -- the `host_id` filter added no further constraint.

## Change

- Remove the `target_host_ref` parameter from `_try_reuse_existing_agent` and its call site.
- Adopt PR #1694's approach: a single pure `_is_host_in_reuse_scope(discovered_host, provider_name, host_name)` helper holds the provider/host-in-scope predicate; the lookup loop calls it.

No user-visible behavior change. Fresh-host create still skips same-named agents on other hosts; existing-host re-create still reuses the agent on the named host; bare-name `--reuse` with multiple matches still raises the disambiguation error.

## Tests

`libs/mngr/imbue/mngr/cli/create_test.py`:

- Replaced `test_try_reuse_existing_agent_filters_by_host` (which constructed a `target_host_ref`) with `..._filters_by_address_host_id` and `..._filters_by_address_host_name`, covering both branches of the host filter via the address.
- Added `test_is_host_in_reuse_scope_matches_documented_examples`, a direct unit test of the new pure helper across the full scoping matrix.
- Dropped the `target_host_ref=None` argument from the remaining call sites.

Local run: `create_test.py` fast subset -- 131 passed; tmux reuse subset -- 2 passed; `mngr` ratchets -- 56 passed. `ty check` clean on both changed files. `vet` (agentic) -- no issues. Full suite via CI.

### DIFF
--- a/libs/mngr/changelog/mngr-followup-1.md
+++ b/libs/mngr/changelog/mngr-followup-1.md
@@ -1,0 +1,3 @@
+Simplified the internal `--reuse` agent-lookup scoping introduced for `mngr create <agent>@<host>.<provider>`: the lookup is now driven entirely by the host named in the create address. No user-visible behavior change -- a fresh-host create still skips same-named agents on other hosts, an existing-host re-create still reuses the agent on the named host, and a bare-name `--reuse` with multiple matches still raises the disambiguation error.
+
+The redundant `target_host_ref` parameter (which scoped by host id only for already-resolved existing hosts) has been removed; its constraint was already covered by the address's host name plus provider. Host-in-scope matching now lives in a single pure `_is_host_in_reuse_scope` helper.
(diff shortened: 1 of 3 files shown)


joshalbrecht:
lgtm, thanks for the follow up!

PR #2334, opened 2026-06-30
Add async def / await ratchet across the monorepo

joshalbrecht:
## Summary

Adds a new common ratchet that freezes `async def` / `await` usage across the monorepo and prevents new async code from being added. We strongly prefer synchronous code: it is far easier to debug, and our software is intentionally low-scale, so async provides no benefit. This complements the existing `PREVENT_ASYNCIO_IMPORT` ratchet by also freezing the language-level constructs regardless of which async runtime backs them.

## Changes

- **Rule + wrapper** (`libs/imbue_common`): add `PREVENT_ASYNC_AWAIT` (`RegexRatchetRule`, pattern `\basync\s+def\b|\bawait\b`) and the `check_async_await` wrapper.
- **Fan-out**: ran `scripts/sync_common_ratchets.py` to add `test_prevent_async_await` to every `libs/*` and `apps/*` `test_ratchets.py`; baselines pinned per-project via inline-snapshot (scans all files, test + production).
- **`mngr_robinhood` exempt**: its `test_prevent_async_await` is a no-op `pass` (it wraps the fundamentally-async Claude Agent SDK; the meta-ratchet still requires the test to exist).
- **`imbue_common` baseline of 3** is purely self-references — the rule's own `rule_name`/`rule_description` text contains the literal words `async def`/`await` (annotated in the test).
- Per the dead-simple-regex decision, any matches inside comments/strings are absorbed into the per-project baseline rather than specially handled.

## Notes

- Per-project only (`libs/*` + `apps/*`); repo-root `scripts/` is not covered.
- `async with` / `async for` only appear inside an `async def`, which is already matched, so they are covered transitively.
- One changelog entry added per touched project (45 projects + `dev`).
- Design doc: `blueprint/ratchet-async-await/plan-ratchet-async-await.md`.


joshalbrecht:
lgtm

PR #2335, opened 2026-06-30
mngr: per-provider discovery (no more global full snapshot)

joshalbrecht:
## What & why

Makes mngr's provider discovery **per-provider** instead of producing a single global "full discovery snapshot". Previously discovery waited for *all* providers and emitted one merged `FullDiscoverySnapshotEvent`, so a single hung or slow provider froze discovery for everything — particularly bad in the Minds app. Now each provider is discovered, snapshotted, and emitted independently, so a wedged provider can no longer block any other.

## Core changes (in `libs/mngr`)

- **New event + config model.** Added `DiscoveryEventType.DISCOVERY_PROVIDER` / `ProviderDiscoverySnapshotEvent` (one provider's agents/hosts, optional `DiscoveryError`, `unknown_host_ids`/`unknown_agent_ids`, and a `discovery_started_at`/`discovery_finished_at` span). Added per-provider cadence/timeout settings to `[providers.<name>]`: `discovery_poll_interval_seconds` (30), `discovery_warn_seconds` (20), `discovery_error_timeout_seconds` (120), `host_discovery_timeout_seconds` (30), `agent_discovery_timeout_seconds` (30), validated so the sub-timeouts stay below the provider error timeout.

- **Per-provider producer.** `mngr observe --discovery-only` now runs one decoupled poll loop (thread) per provider (`run_per_provider_discovery_stream`). A hung provider is bounded *without killing threads*: it warns after the warn threshold, then emits a per-provider `DiscoveryError` after the error timeout while the abandoned read keeps running; its late result is accepted on a later poll. `mngr list` writes one per-provider snapshot per provider as a side-effect.

- **Per-host/agent sub-timeouts + UNKNOWN.** Providers expose a bounded `discover_hosts_and_agents_within_timeouts`: each host's read runs on an abandonable daemon thread, and a host past its per-host timeout is omitted and marked **UNKNOWN** (its previously-known agents retained as unknown by the aggregator) rather than stalling the whole provider's snapshot. Batch providers (modal / vps / imbue_cloud) override it to delegate to their batch path (provider-level timeout only — documented).

- **Shared span-aware aggregator.** New `DiscoveryStateAggregator` folds per-provider snapshots + incremental events into one consistent view. It fixes a latent clobber bug: a host/agent state-change or destroy event that lands *during* a provider's discovery span is no longer overwritten by that older in-flight snapshot (the snapshot is authoritative only for items with no newer event since its `discovery_started_at`). The span-rule helpers live in a leaf module `discovery_reconciliation.py` so they can be shared without a circular import.

- **Deprecate (do not delete) the legacy global snapshot.** `FullDiscoverySnapshotEvent` / `DISCOVERY_FULL` are no longer produced and all live writers/usages were removed, but the type + parse path are kept and marked deprecated so **historical on-disk discovery logs still parse** (the discovery models use `extra="forbid"` + a discriminated union, so deleting the type would break old logs). `complete_names.py` and the resolution replay were made per-provider-aware.

- **Read-after-write consistency.** A per-provider snapshot's authority window reaches back to its `discovery_started_at`, earlier than its line position in the log. The replay-based resolution (`mngr stop --stop-host`) and the cached-snapshot attach (`mngr forward --observe-via-file`) now replay back to each provider's span start and apply the same span rule as the aggregator, and `mngr stop` falls back to a live discovery for any agent not yet in the stream — so an agent created (or destroyed) during an in-flight discovery span is no longer lost (or resurrected), and stopping a just-created agent works immediately.

## Consumers migrated

All discovery consumers now reconcile through the shared aggregator (never wholesale-replace across providers): `mngr observe`'s `AgentObserver`, `mngr_forward`, `mngr_latchkey`, and the Minds desktop client (per-provider freshness for the recovery-redirect gate; `discovery_health` still keyed on "any event"). The forever-claude-template `system_interface/agent_manager.py` is migrated on a **matching branch** (`mngr/per-provider-discovery`) in that repo; its verification lands when this mngr branch is released and re-vendored.

## Testing

Full `libs/mngr` unit/integration suite passes (4834), plus the consumers (`mngr_forward` + `mngr_latchkey` + Minds desktop client, 1853) and the touched plugin suites (1184). `ty`, `ruff`, and the ratchets are clean. CI (`just test-offload`, incl. acceptance) is the final gate. Per-PR changelog entries are present for every touched project (`mngr`, `minds`, `mngr_forward`, `mngr_latchkey`, `mngr_modal`, `mngr_vps`, `mngr_imbue_cloud`, `dev`).

## Known / deferred (intentional)

- **`complete_names` (TAB completion)** is left eventually-consistent (stdlib-only, hot path; a just-created name self-heals on the next poll) rather than carrying a third copy of the span rule.
- **Producer-side "intervening-event immediate re-poll"** (a convergence-speed optimization) is not implemented; correctness is already guaranteed by the span-aware aggregator, so it would only shorten the snapshot-freshness window by ~one poll cycle.
- **Per-host timeout under a permanently-hung host**: a host that hangs forever spawns one abandoned daemon thread per poll (bounded in practice by SSH-level timeouts); cross-poll per-host de-duplication is a possible follow-up.


joshalbrecht:
(roughly lgtm up to this point)

joshalbrecht:
basically lgtm, though this logic is tricky enough that it's a little hard to confirm without more testing

PR #2336, opened 2026-06-30
Send SIGWINCH repaint on every tmux attach (client-attached hook) (#2322)

### LINKED ISSUE #2322: Send SIGWINCH repaint on every tmux attach (client-attached hook), not just on `mngr connect`
## Summary

Move the post-attach `SIGWINCH` repaint nudge from mngr's `connect` wrapper into a tmux **`client-attached` hook** on the agent's window, so it fires on **every** attach (`tmux attach`, the ttyd agent terminal, a web-shell `tmux attach`, `mngr connect`) instead of only when mngr drives the attach. Then **remove the now-redundant SIGWINCH step from the `mngr connect` path**, since the hook will cover it automatically.

## Background — what we observed

While testing a minds Docker workspace, the agent's tmux pane became badly visually corrupted (input-box border characters bleeding through conversation text, overwritten/dropped characters, overlapping lines). The same corruption also broke message sending: the send path's paste-visibility check reads the pane via `capture-pane` and fuzzy-matches the message text, and a corrupted pane makes that match fail, so the message is pasted but never submitted ("failed to send", text left sitting in the input box).

We confirmed the corruption was a **stale, unpainted frame, not buffer damage**: manually running mngr's existing SIGWINCH script against the session immediately triggered a full clean repaint (132 lines changed, corruption gone, stable).

## Root cause chain

1. mngr creates agent tmux sessions at `200x50` (`_DEFAULT_TMUX_WIDTH/HEIGHT`, `libs/mngr/imbue/mngr/hosts/host.py`), and leaves `window-size` at tmux's default `latest` (it is only pinned to `manual` if a provider/caller sets it, which minds does not).
2. Because `window-size` is `latest`, any client that attaches at a different size resizes the agent's live window. In our case a `186x85` viewer (a `tmux attach` run manually inside a minds web "workdir" shell) resized the window `200x50 -> 186x84`.
3. minds provisions `~/.tmux.conf` with `set -g alternate-screen off` (deliberate — it keeps the agent's output in the main-buffer scrollback so the ttyd web terminal can scroll/copy it via OSC 52). As a result, Claude Code's Ink TUI renders into the shared main buffer and does **not** repaint cleanly across a resize; left idle (e.g. awaiting the model), the garbled frame persists.
4. The repaint nudge that fixes this — `SIGWINCH` to the pane PID and its children (`build_post_attach_sigwinch_script`, `libs/mngr/imbue/mngr/api/connect.py:36`) — is **only sent from mngr's own attach/connect wrapper** (`_build_ssh_activity_wrapper_script` -> `sigwinch_step`, `libs/mngr/imbue/mngr/api/connect.py:113`). So a raw `tmux attach`, the ttyd agent terminal, or a web-shell `tmux attach` never receives it. (The corruption persisted precisely because the attach that resized the window was *not* `mngr connect`, so no nudge was sent.)

Note: tmux does send `SIGWINCH` to the pane's foreground process group on resize, but mngr signals the pane PID **and its children** because the agent (claude) is typically a child of the pane shell and may not be the foreground process group leader — so the automatic signal is not reliable here, which is why the explicit nudge exists.

## Proposed fix

Set a persistent tmux `client-attached` hook on the agent's primary window at session creation that runs the existing SIGWINCH script. mngr already does almost exactly this for the onboarding popup:

- `libs/mngr/imbue/mngr/hosts/host.py:3691` sets a one-shot `client-attached[99]` hook (it self-removes via `set-hook -u`).

Mirror that pattern with a **persistent** hook in a different slot (e.g. `client-attached[98]`, without the self-removal) whose body runs `build_post_attach_sigwinch_script(session_name)` (already written). The session name is baked into the script, so no `#{client_session}` interpolation is needed.

Then **remove `sigwinch_step` from `_build_ssh_activity_wrapper_script`** (`libs/mngr/imbue/mngr/api/connect.py:113,133`): once the hook fires on every attach (including `mngr connect`'s own attach), the wrapper's copy is redundant. This is the "remove the existing SIGWINCH on connect" the hook makes safe.

## Implementation notes / caveats

- **Quoting.** The SIGWINCH script is a multi-stage pipeline with command substitution (`kill -WINCH {} $(pgrep -P {})`). Nesting it inside `set-hook '... run-shell "..."'`, itself inside the create-session `&&` chain, is 3+ levels of quoting. Prefer extracting it into a shipped resource `.sh` (mngr already ships resource scripts under `libs/mngr/imbue/mngr/resources/`) and having the hook call e.g. `run-shell "bash /.../sigwinch_panes.sh <session>"`. This avoids quoting hell and is more maintainable.
- **Fire ordering.** Verify `client-attached` fires *after* the window has resized to the new client (so claude re-queries the new size). The old wrapper used `(sleep 3; ...)`; if there's a race, wrap the hook body in a short backgrounded delay.
- **Window pinned to `manual`.** The wrapper skips the nudge when `window-size` is `manual` (no resize, nothing to redraw). Either keep that guard in the script or accept a harmless no-op repaint.
- **Applies to new sessions.** Setting the hook at creation means only newly-created agents get it; existing running sessions are unaffected until recreated. A global `set-hook -g client-attached` in mngr's tmux config would cover all current+future sessions in one place, but then the script must resolve the attached session via `#{hook_session}` (or just signal all panes) rather than a baked-in name. Either placement is acceptable; per-session-at-creation matches the existing onboarding-hook pattern most closely.

## Scope

This is the targeted "always repaint on attach" fix. It is independent of (and complementary to) two deeper items we also found, which should be tracked separately if we want them:

- minds does not pin `window-size manual` for agent sessions, so viewers attaching at a different size resize the agent's live TUI at all (pinning would avoid the resize entirely).
- The send path's paste-visibility check (`_check_paste_content` / `wait_for_paste_visible` in `libs/mngr/imbue/mngr/agents/tui_utils.py`) is fragile to a corrupted/garbled pane capture; it could be hardened independently.

## Acceptance criteria

- Attaching to an agent session via a plain `tmux attach` (not `mngr connect`) triggers a `SIGWINCH` to the agent process (and its children) and a clean repaint.
- The SIGWINCH step is removed from the `mngr connect` / `_build_ssh_activity_wrapper_script` path, and `mngr connect` still produces a clean repaint (now via the hook).
- A test covers the new hook wiring (and the removal), consistent with how the existing `client-attached[99]` onboarding hook is tested.
- Changelog entry added for `mngr`.

joshalbrecht:
## Summary

Moves the post-attach `SIGWINCH` repaint nudge out of `mngr connect`'s SSH wrapper and into a persistent per-session tmux `client-attached` hook, so the agent repaints cleanly on **every** attach -- a plain `tmux attach`, the ttyd agent terminal, a web-shell attach, or `mngr connect` -- not just when mngr drives the attach.

Fixes the garbled-pane / failed-send symptom that occurred when a client attached at a different terminal size than the agent's session was created at (the TUI was left showing a stale, unpainted frame, which also broke message sending). Closes #2322.

Plan: `specs/sigwinch-attach-hook/spec.md`.

## Changes

- **New** `libs/mngr/imbue/mngr/resources/sigwinch_panes.sh`: sends `SIGWINCH` to every pane's child processes across all windows (`pgrep -P`), skips windows pinned to `window-size manual`, and delays before signaling (overridable via `MNGR_SIGWINCH_DELAY_SECONDS`, used by tests).
- Installed at host level during provisioning (`_ensure_shared_shell_libs`), so it exists for local, docker, ssh, and modal hosts.
- A persistent `client-attached[98]` hook (via `run-shell -b`) is set at session creation in `_build_start_agent_shell_command`, passing the session and primary-window names to the script.
- Removed `build_post_attach_sigwinch_script` and the `sigwinch_step` from `_build_ssh_activity_wrapper_script` (and its now-unused `primary_window_name` parameter).
- Tests: unit string-shape coverage for the hook wiring and for the wrapper no longer sending `SIGWINCH`; `@pytest.mark.tmux` integration tests for delivery and the manual-pin guard.
- Docs note in `conventions.md` + `mngr` changelog entry.

## Behavior note

Only newly created agents get the hook. Agents already running at upgrade time will not get the repaint nudge on attach (the old `mngr connect` nudge was removed) until they are recreated -- an accepted gap.

## Testing

- `connect_test.py`: 45 passed
- `host_test.py` (onboarding + sigwinch): 7 passed
- `test_connect.py -m tmux`: 2 passed
- `test_ratchets.py`: 56 passed
- Full `just test-offload`: see CI.


libs/mngr/imbue/mngr/resources/sigwinch_panes.sh
```diff
@@ -0,0 +1,61 @@
...
+# Sends SIGWINCH to every pane's child processes across all windows in a session
+# so the in-pane process (e.g. claude) re-queries its terminal size (TIOCGWINSZ)
+# and redraws at the newly attached client's dimensions. Without this nudge a
+# client that attaches at a different size than the session was created at can be
+# left with a stale, unpainted frame.
+#
+# We signal the pane PID *and its children* (via pgrep -P) rather than relying on
+# the SIGWINCH tmux itself sends on resize: the agent is typically a child of the
+# pane shell and may not be the pane's foreground process group leader, so tmux's
+# automatic signal does not reach it reliably.
```
qi-imbue, line 13: This explanation of SIGWINCH is wrong in two places:
- tmux doesn't send SIGWINCH: it only stores the new size of the pty, and the kernel sends SIGWINCH
- the kernel sends SIGWINCH to all processes in the foreground process group (https://man7.org/linux/man-pages/man2/TIOCSWINSZ.2const.html; look for SIGWINCH)

Also, this uses `pgrep -P` which only finds direct children - do we want descendants instead?

Moreover, I'd like to know if we have a reliable reproduction of the "missed SIGWINCH" bug this is supposed to fix in the first place. Launching an agent the "normal" way will put it in the foreground process group so the normal SIGWINCH mechanism should be enough...

PR #2339, opened 2026-06-30
minds: custom tooltips (titlebar, dialogs, landing) on a unified overlay surface

gdenisov:
## What

Adds custom hover/focus tooltips across the desktop client, and reworks the overlay layer into a single always-warm `WebContentsView` that hosts them (and the existing dialogs) above both the chrome and the workspace content.

Tooltips now cover:
- **Titlebar buttons** — Main Menu (renamed from "Workspaces"), Home, Back, Forward, Get help, Requests, and the window controls.
- **Dialog close buttons** — the inbox, help, and sign-in dialogs' X buttons show a "Close" tooltip.
- **Landing-page workspace-row buttons** — Start, Stop, Restart, Open in new window, Settings.

They replace the OS `title=` tooltips, which were unreliable in the desktop window and couldn't extend past the titlebar. Icon-only buttons keep an `aria-label` so screen readers still announce them.

Plan + design rationale: `blueprint/overlay-surface-tooltips/plan-overlay-surface-tooltips.md`.

## Why this shape

- **Overlay pointer model = dynamic bounds.** `setIgnoreMouseEvents` is **window-level only** in Electron 40 (verified) — there is no per-`WebContentsView` click-through, so the planned "click-through overlay view" is impossible. The only lever a `View` exposes is `setBounds`, so the overlay view shrinks to just the tooltip's rectangle (everything else stays interactive) and expands to full-window only while a modal is open.
- **Two render backends.** The chrome titlebar and the overlay-hosted dialogs render tooltips on the overlay surface via the `window.minds` bridge. The landing page renders in the workspace **content view**, which deliberately has **no bridge** (it also hosts foreign, untrusted workspace content), so it can't reach the overlay — it draws its tooltips **in-page** instead: a `position: fixed` bubble on `<body>` that escapes the row cards' `overflow-hidden`. Both backends share one `.minds-tooltip` style so the bubbles look identical.
- **Dialogs mount on demand.** Each dialog (workspace menu / inbox / help / sign-in) is created when opened and torn down when closed. An earlier "kept-warm iframes" approach was dropped — a hidden warm page kept doing background work and every open reloaded anyway — and modal switches are double-buffered so they're flash-free.

## Notable details

- Titlebar "Workspaces" button renamed to **"Main Menu"**.
- Tooltip appearance: **centered** text, up to **400px** wide, ~**250ms** hover-intent delay plus keyboard focus; repositions to stay on-screen; hides on leave / click / scroll / resize / blur.
- The bubble width is measured with `Math.ceil(getBoundingClientRect().width)` rather than the integer `offsetWidth` — `offsetWidth` rounds the shrink-to-fit width *down*, which pinned the width a fraction short and wrapped the last word (e.g. "Open in new window").
- The shared trigger script (`tooltip_triggers.js`) is included once via `Base.jinja` and picks the backend by whether the `window.minds` bridge is present.

## Tests

- Route + tooltip-attribute tests (titlebar, dialog close buttons, landing row buttons).
- `desktop_client` subtree green (**1218 passed**), minds ratchets green (**57**), ruff clean, all JS `node --check` clean.
- The Electron shell has no JS test suite; the window/view behavior and tooltip geometry are verified manually.

## Testing steps

Verified in a running app (`just minds-start` — it rebuilds `app.min.css`, which is gitignored):

- [x] Each dialog (workspace menu / inbox / help / sign-in) opens, dismisses (backdrop + Escape), and toggles; the sidebar list and inbox badge populate; window dragging still works.
- [x] Tooltips appear on hover (~250ms) and keyboard focus across the **titlebar**, the **dialog close buttons**, and the **landing row buttons**; they float over the content area, stay centered on one line (e.g. "Open in new window"), reposition near window edges, and hide on leave/click/scroll.
- [x] Hovering a titlebar button does **not** block clicks/hover on the content or the rest of the titlebar (the dynamic-bounds shrink works).

The overlay backend measures the bubble while the overlay view is hidden-but-full-window; confirmed it computes layout and sizes tooltips correctly (no mis-sizing). The in-page backend measures in the live content document, so it isn't subject to this.


joshalbrecht:
(I only reviewed the other PR that seemed stacked on this, lmk if this one was meant to be separate though)

PR #2340, opened 2026-06-30
tmr CI: fix broken pre-trust step by extracting it into a type-checked script

qi-imbue:
## Why

The scheduled TMR workflow (and the other TMR workflows that share `tmr-setup`) has been failing at the `tmr-setup` step since June 24:

```
ImportError: cannot import name 'find_user_claude_config' from
'imbue.mngr_claude.claude_config'
```

`tmr-setup` pre-seeds the global Claude config (directory trust, dismissed effort callout, completed onboarding) so a Claude agent can run unattended on the CI host. It did this via an inline Python heredoc. Commit `9cf80b0d2` ("Rename claude config resolvers for isolated/unisolated symmetry") renamed `find_user_claude_config` → `find_user_config_in_isolated_mode` and updated every call site **except** the one buried in the workflow heredoc -- which neither grep nor the type checker can see. Every TMR run aborted before the orchestrator started.

## What

- Extract the heredoc into a real, type-checked module: **`scripts/pretrust_claude_checkout.py`**. The `tmr-setup` action now just runs the module.
- Fix the stale function name in the process.
- The script is generic: TMR is the current caller, but any CI job that launches a local Claude agent against the checkout can use it. The docstring reflects this.

The key win is that the import now lives in a `.py` file in `ty`'s scope (the repo-root `test_no_type_errors` in `test_meta_ratchets.py`), so a future rename of the `claude_config` API is caught at test time instead of only when a TMR run fails in CI.

This is a `dev`-only change (root `scripts/` + `.github/`).

## Manual testing

- Trigger a TMR run (e.g. the `TMR (scheduled)` workflow via `workflow_dispatch`, or wait for the schedule) and confirm `tmr-setup` now passes through the "Pre-trust checkout for Claude" step and reaches the "Run TMR orchestrator" step.
- Confirm the agent does not hang on / fail at Claude's trust dialog (i.e. no `ClaudeDirectoryNotTrustedError`).
- Optional local sanity check (writes to a throwaway HOME, leaves your real `~/.claude.json` untouched):
  ```
  TMPH=$(mktemp -d); env -u ORIGINAL_CLAUDE_CONFIG_DIR -u CLAUDE_CONFIG_DIR HOME="$TMPH" \
    uv run --project libs/mngr_claude python scripts/pretrust_claude_checkout.py
  cat "$TMPH/.claude.json"   # expect hasTrustDialogAccepted, effortCalloutDismissed, hasCompletedOnboarding
  rm -rf "$TMPH"
  ```

### DIFF
--- a/.github/actions/tmr-setup/action.yml
+++ b/.github/actions/tmr-setup/action.yml
@@ -43,17 +43,4 @@ runs:
       # integrator config then inherits the trust via the plugin's
       # copy_project_config_from path.
       shell: bash
-      run: |
-        uv run --project libs/mngr_claude python - <<'EOF'
-        from pathlib import Path
-        from imbue.mngr_claude.claude_config import (
-            add_claude_trust_for_path,
-            complete_onboarding,
-            dismiss_effort_callout,
-            find_user_claude_config,
-        )
-        config_path = find_user_claude_config()
-        add_claude_trust_for_path(config_path, Path.cwd().resolve())
-        dismiss_effort_callout(config_path)
-        complete_onboarding(config_path)
-        EOF
+      run: uv run --project libs/mngr_claude python libs/mngr_claude/scripts/pretrust_ci_checkout.py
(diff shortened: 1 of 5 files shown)


joshalbrecht:
lgtm

PR #2341, opened 2026-06-30
Revive a ctrl-c'd / crashed / OOM-killed agent when it's next messaged

gnguralnick:
## Problem

When an agent's main process exits but tmux keeps the session alive (a `ctrl-c`, a crash, or an OOM kill of just the agent process), messaging the agent through the chat/API did not deliver the message or bring the agent back. The text was typed via `tmux send-keys` into the leftover shell — run as a shell command and lost — while the send was still reported as **successful**. (Sending to an agent whose session was *fully* gone already worked.)

This is the path the OOM-revival design depends on: a shed agent is meant to come back when it's next messaged.

## Root cause — two layers

1. **Lifecycle misclassification.** A real claude agent's pane keeps mngr's own in-pane helpers running as descendants — the transcript streamers and the background-task script, each a `sleep` loop. After the agent process is killed they linger, so `determine_lifecycle_state` saw a non-shell descendant and returned `REPLACED` rather than `DONE`. (Verified against a real ctrl-c'd staging agent: pane foreground `bash`, claude gone, `claude_background_tasks.sh`/`stream_transcript.sh`/`common_transcript.sh` still running `sleep`s → `REPLACED`.) The `non_shell_descendant → REPLACED` heuristic predates those in-pane helpers, which is why it now misfires.

2. **Revive gate too narrow.** `REPLACED` is treated as *live* everywhere (`is_running()` true, `ensure_agent_started` won't restart it), and the send path only auto-restarted `STOPPED`. So a non-running agent was never revived on message.

## Change

- **`hosts/common.py`** — only treat a non-shell *descendant* as `REPLACED` when the agent type is **unknown** (there the descendant might be the agent under a name we don't recognize). For a known type the agent's process was already checked and is absent, so a non-shell descendant is a background/auxiliary process (e.g. mngr's own helpers) and must not mask that the pane dropped to a shell prompt → `DONE`. A genuine replacement runs in the *foreground*, still detected via `pane_current_command`. Unknown-type behavior is unchanged.
- **`api/message.py`** — restart an agent that is `STOPPED` *or* `DONE` before sending (neither has a live process to receive the message).
- **`api/find.py`** — `ensure_agent_started` tears down a lingering session (`stop_agents`) before relaunch for a `DONE` agent, mirroring `mngr start --restart`, so the restart actually happens rather than no-op'ing on the existing session. The agent comes back with its prior conversation intact (normal stop/start resume).

## Tests

- `common_test.py`: a known-type agent with a non-shell descendant but a shell foreground is now `DONE` (and a non-shell *foreground* is still `REPLACED`); unknown-type cases unchanged. Verified the patched detector returns `DONE` on the real staging agent's process tree.
- `message_test.py`: `test_send_message_to_agents_revives_done_agent_when_start_desired` — kills an agent's process while leaving its tmux session, then asserts a message revives it to a running state. Verified it fails without the restart change.

`common_test.py` + `base_agent_test.py` + `message_test.py` + `find_test.py` pass locally (202).


joshalbrecht:
@gnguralnick I see no change to hosts/common.py?

also the tests are failing

not sure whether you wanted to make that change or update the PR description!

gnguralnick:
you're right, thanks for the catch! had a commit that needed to be pushed :/

joshalbrecht:
lgtm

Issue #2342, opened 2026-06-30
Apply pending workspace renames to offline/unreachable providers


joshalbrecht:
## Context

Follow-up from the workspace-names refactor (see `blueprint/simplify-workspace-names/`).

That work makes a workspace rename a two-step, minds-orchestrated operation: `mngr rename --host <new>` (updates the provider's `host_name`) followed by a `workspace_display_name` label write on the `system-services` agent. v1 only supports rename when the provider's name store is writable **right now**:

- `imbue_cloud` / `mngr_vps` / `mngr_modal`: always (cloud control plane / volume reachable regardless of container state)
- `docker` / `lima`: when the local daemon / `LocalVolume` store is reachable
- `ssh`: unsupported (host names are user-owned — "edit your config")

If the control plane is not reachable, v1 surfaces an error rather than queuing the rename.

## Problem

For providers/states where the name store cannot be written at rename time, the user's rename is simply rejected. We want a rename initiated while the workspace is offline (or its control plane is temporarily unreachable) to be applied automatically once it comes back.

## Proposed approach (to be designed)

A minds-side **pending-rename** record:

- On rename, write a pending-rename record (keyed by the workspace's `system-services` `agent_id` / `host_id`) capturing the target `host_name` slug + `workspace_display_name` **before** attempting the provider rename.
- Attempt the provider rename + label write immediately; on success, clear the record.
- On failure (control plane unreachable), leave the record in place.
- A reconciler applied on next discovery / when the host next comes online replays the pending rename. It is **idempotent**: it no-ops the `host_name` change if the host already matches, and (re)writes the display label. This mirrors mngr's existing resumable-rename semantics (`rename` re-run completes an interrupted rename).
- Collision handling at apply time: if the target slug is now taken on the provider, surface the conflict rather than silently dropping the rename.

## Out of scope here

- `ssh` host rename (stays user-owned).
- The synchronous rename path itself (delivered in the main PR).

## Notes

- `imbue_cloud` (minds' production mode) does **not** need this — its connector is always reachable — so this is a robustness improvement for the edge providers/states.

PR #2343, opened 2026-06-30
Simplify workspace names: renamable workspaces, host_id identity, decoupled display name

joshalbrecht:
Refactors and simplifies how workspaces are named. Plan: `blueprint/simplify-workspace-names/plan-simplify-workspace-names.md`.

Companion PR (forever-claude-template): https://github.com/imbue-ai/forever-claude-template/pull/236 — removes the FCT in-container `mngr list` `has(labels.workspace)` filter and the FCT-side label writes. (Currently reverted on FCT main pending the snapshot-test investigation noted below.)

## The model

One clean concept per datum, each with a single canonical home:

- **`host_id`** — the immutable identity. Never derived from a name.
- **`host_name`** — a normalized, mutable, lowercase slug. Lives on the host. Per-provider unique. This is what gets renamed.
- **display name** — an arbitrary human-readable string (any characters, mixed case). Lives in a `workspace_display_name` label on the workspace's `system-services` agent. This is what the UI shows.

Renaming a workspace updates the `host_name` and the display name together, so the two never drift.

## What's in this PR

**mngr core (`libs/mngr`)**
- `mngr rename --host <agent> <new>` — renames the host of the referenced agent via the provider's `rename_host`, leaving the agent name, tmux session, env file, and git branch untouched. (Was a `[future]` placeholder.)
- `HostName` length cap of 63 (the DNS-label limit), so a host name is always safe to embed in provider-side identifiers. Generous on purpose: it does not retroactively invalidate existing long names (`mngr_mapreduce`, `mngr_vps`). Auto-generated names stay within the cap.

**imbue_cloud (`libs/mngr_imbue_cloud` + `apps/remote_service_connector`)**
- `ImbueCloudProvider.rename_host` via a new connector endpoint `POST /hosts/{host_db_id}/rename` (UPDATE pool_hosts SET host_name, with ownership checks) + `client.rename_host`. The lease `host_db_id` stays the durable identity, so a rename never touches the VPS/container and works online or offline.
- Dropped the bake-time `workspace` label on pre-baked pool hosts.

**lima (`libs/mngr_lima`)**
- `rename_host` implemented: rewrites the logical host name on the local (offline-writable) record store, so it works whether or not the VM is running, uniformly for legacy and current-scheme VMs.
- New VMs derive their limactl instance name from the immutable `host_id` (not `host_name`), so a rename never desyncs the instance name. Existing `<prefix><host_name>` VMs keep working (discovery reads the name from the record, never the instance name).

**minds (`apps/minds`)**
- `POST /api/v1/workspaces/<id>/rename` + an ultra-basic rename field on the workspace settings page. A same-slug rename is a display-only label write (works on every provider, online/offline); a slug change renames the host then writes the label, ordered and idempotently re-runnable. Per-provider slug collisions are rejected (409).
- Create accepts an arbitrary name, derives the slug, and sets the display label.
- Arbitrary-text → lowercase-slug normalizer (minds targets 32 chars; rejects names that reduce to nothing).
- Workspace discovery keys off the `is_primary` label; collision checks use the actual `host_name` (threaded through discovery via a new `host_name_by_host_id` map); display reads `workspace_display_name` (falling back to the host name for legacy workspaces).
- LiteLLM key no longer minted with `host_name` metadata.

## The `workspace` label (nothing reads it; write retained transitionally)

Nothing in minds *reads* the `workspace` label anymore — discovery keys off `is_primary`, collisions off the host name, display off `workspace_display_name`. minds still *writes* `workspace=<host_name>` on the primary agent as a transitional shim, because the deployed forever-claude-template's in-container `mngr list` still filters on `has(labels.workspace)`; dropping the write hides the `system-services` agent from that listing and breaks the snapshot-resume tests. The FCT-side filter removal (companion PR #236) must land and be picked up by the snapshot test before the shim can go.

Note: a full-removal attempt (merge #236, drop the shim) unexpectedly still failed the snapshot test — in-container `mngr list` dropped `system-services` even with the FCT filter gone, which could not be reproduced locally (`mngr list` lists `is_primary` agents normally) or root-caused from CI artifacts (it looks like CI snapshot-image caching serving a stale container). That attempt was reverted; the shim stays until this is investigated in a live workspace. Tracked in #2347.

## Deviations from the original plan (all improvements, flagged)

- **`HostName` cap is 63, not 32.** 32 retroactively broke production host-name generators in `mngr_mapreduce`/`mngr_vps`. The type cap is now 63; minds applies its own smaller 32-char target to derived slugs.
- **lima rename is uniform across legacy + new VMs.** The plan assumed discovery parses instance names (it reads the record store), so rejecting legacy VMs was unnecessary.

## Follow-ups

- #2342 — pending-rename for offline/unreachable providers (the synchronous path is here; queuing for the rare not-immediately-writable case is deferred).
- #2347 — remove the transitional minds-side `workspace` label write once the FCT filter-removal (#236) is re-landed and picked up by the snapshot test.

## Testing

CI green: Unit+Integration, Acceptance, test-offload (incl. coverage), test-docker, test-offload-acceptance, test-minds-snapshot, check-changelog. Per-project changelogs included.


apps/minds/imbue/minds/desktop_client/agent_creator.py
```diff
@@ -679,6 +680,19 @@ def _build_mngr_create_command(
...
+        # Transitional backwards-compat: nothing in minds reads the ``workspace``
+        # label anymore (discovery keys off ``is_primary``, collisions off the
+        # host name, display off ``workspace_display_name``). But the deployed
+        # forever-claude-template's in-container ``mngr list`` still filters on
+        # ``has(labels.workspace)``, so the primary agent must keep the label or
+        # it disappears from that listing. Set it to the host-name slug (the old
+        # convention) for presence. Remove this once the FCT change that drops
+        # the filter has shipped (FALLBACK_BRANCH advanced). See follow-up issue.
         "--label",
         f"workspace={host_name}",
```
joshalbrecht: wait -- I thought we changed this? We should remove this comment, and the "--label workspace=blah" args so that we can see if this actually *does* work without that label!

the whole point was to remove this :-P

joshalbrecht:
lgtm other than the one comment

PR #2344, opened 2026-06-30
minds: find and read past agents' preserved transcripts

pseay-imbue:
## What

Lets a minds agent recover the chat history of *past* agents (including destroyed ones) that a user references -- the "old stuff" agents previously had no way to find.

## Changes

**mngr (`libs/mngr`)**
- New public `imbue.mngr.api.transcript` module: the human/json/jsonl transcript rendering used by `mngr transcript`, extracted so it can be reused programmatically (CLI behavior unchanged).
- New `imbue.mngr.api.preservation` helpers: `list_preserved_agents`, `find_preserved_agent_by_id`, `read_preserved_common_transcript`, and `render_preserved_agent_transcript`, which read the durable preserved copy under `<host_dir>/preserved/{agent_name}--{agent_id}/` that survives full host destruction.

**minds (`apps/minds`)**
- `GET /api/v1/workspaces/preserved` -- lists agents whose transcripts were preserved on destroy, newest-first (name, id, preserved-at).
- `GET /api/v1/workspaces/<agent_id>/transcript` -- returns an agent's transcript (the durable preserved copy for destroyed agents, falling back to the live agent via `mngr transcript`), with `format`/`role`/`head`/`tail` filters.
- Both reuse the existing `minds-workspaces-read` gateway permission -- no new permission entry.

## Permission model (verified)

The `minds-workspaces-read` verb is `path.kind: tree`, whose gateway pattern covers every GET under `/workspaces`, so both new endpoints are authorized with no new gateway entry, and cross-agent reads are allowed. The user must still have granted `minds-workspaces` (it is not in the agent baseline) -- the same grant used to list workspaces.

## Agent-facing skill (separate repo)

A `find-past-transcripts` skill plus a `CLAUDE.md` pointer were prepared in the **forever-claude-template** repo (committed locally on a matching branch, not pushed) so agents know by default that past transcripts are findable. Pushing that needs a separate decision since it ships to all agents via release tags.

## Tests

- mngr: unit tests for the rendering module and the preservation enumeration/read helpers.
- minds: integration tests through the API for both endpoints (preserved listing order, preserved transcript retrieval with filters, live fallback via a fake mngr, and the 400/404/auth cases).


pseay-imbue:
Reverted all the mngr/minds backend + CLI changes (this branch is now empty vs main). The feature ships instead as a single host-local skill in forever-claude-template (imbue-ai/forever-claude-template#232): agents read their own host's /mngr/preserved/ directly, so no Minds API endpoint, gateway permission, or mngr/minds code change is needed.

PR #2346, opened 2026-06-30
Fix VPS-resident latchkey gateway tunnel using the wrong container port

joshalbrecht:
## Problem

On a minds lima-slice host (docker-in-VM), an agent reaches latchkey via two gateways: a **primary** (the desktop gateway, reverse-tunneled from the user's laptop) on `127.0.0.1:1989`, and a **secondary** VPS-resident gateway on `127.0.0.1:1990` that is meant to keep latchkey working when the laptop is offline.

The secondary never worked on slices. `provision_remote_gateway`'s reverse tunnel runs **on the outer host (the VM)** and SSHes to the container at `127.0.0.1:<container_ssh_port>`, but discovery passed it `ssh_info.port` — the container's **externally-routable** port (the bare-metal box's forwarded port, e.g. `22005`). On the VM's own loopback the container's sshd is published at a **different** fixed port (`config.container_ssh_port`, e.g. `2222`), so the tunnel was refused and `:1990` never bound inside the container.

When the user's laptop went offline (so the primary dropped), the agent then had **no** latchkey access at all — both gateways down — which is exactly the production incident observed (Gmail/Slack calls failing with connection-refused on both `1989` and `1990`).

A second defect masked it: the tunnel is launched `nohup ssh … &` inside a wrapper that `exit 0`s immediately, so a dead `ssh -R` was reported as success and the host was cached in `_provisioned_hosts`, never retried.

## Fix (minimal)

- **Right port:** add a provider seam `get_container_loopback_ssh_port` (default `None`); `ImbueCloudProvider` returns `config.container_ssh_port` (the in-VM publish port, which it already treats as authoritative — see `is_slice` detection). Discovery uses it for the reverse tunnel, falling back to `ssh_info.port` when `None` (no behavior change for other providers / plain docker-on-VPS, where the two coincide).
- **De-silence:** after backgrounding the tunnel, the launch confirms the `ssh` process survives a short window (it runs under `ExitOnForwardFailure=yes`), so a refused/failed-bind tunnel becomes a failed provision that retries instead of being cached as success.

## Tests

- New: `ImbueCloudProvider.get_container_loopback_ssh_port` returns the in-VM publish port, not the lease's connect port.
- Updated: the reverse-tunnel script test asserts the liveness check is present.
- All `mngr_latchkey`, `mngr_imbue_cloud`, and `mngr` unit + ratchet tests pass locally; `ty` type-checks clean for all three packages.

## Not included (follow-up)

The VM→container tunnel still has no reconnect/health loop, and re-provisioning is gated by the in-memory `_provisioned_hosts` cache (cleared on supervisor restart). A tunnel that dies mid-session while the laptop is offline won't self-heal until the laptop returns. A health-check mirroring the desktop tunnel's 30s loop would make the secondary dependable through long offline stretches.


joshalbrecht:
I don't necessarily love this, but this *was* a problem, and this probably *does* fix it

Curious if @hynek-urban has any better ideas here

hynek-urban:
LGTM :) Copying my answer from Slack:

> I think the fix is fine :) If I understand it correctly, it will also auto-repair the existing remote workspaces. (The only other solution that comes to mind is not relying on SSH tunnels for this purpose, as [discussed in the past](https://imbue-ai.slack.com/archives/C0ASLEFFD62/p1780953745039019?thread_ts=1780673165.030299&cid=C0ASLEFFD62). I still think it would introduce other downsides and sources of bugs, though. (For example when trying to bind to a proper non-0.0.0.0 address on the host service's side.))

joshalbrecht:
lgtm for now then

this adds a provider interface method that I don't really like, but will let that slide for now

PR #2349, opened 2026-07-01
Separate mngr and minds release test suites by tag

qi-imbue:
Implements the release-test restructuring agreed in Slack (Josh + Qi): make the whole release suite -- mngr and minds -- discoverable by the `@pytest.mark.release` tag, name the mngr jobs explicitly, and run all minds release tests from the minds release job. Also fixes the two pre-existing test failures that folding surfaced.

## Restructuring

- **Rename** the `release-tests.yml` jobs to the *mngr* release suite: `test-docker-release` -> `test-mngr-release-docker`, `test-release` -> `test-mngr-release`.
- **Tag** the minds release-tier tests: every test under `apps/minds/deployment_tests/` (`minds_deployment` / `minds_services`) now also carries `@pytest.mark.release`, so the whole release suite is discoverable by tag rather than path.
- **Exclude minds from the mngr jobs by path** (`--ignore apps/minds`), simplifying the `-m` filters.
- **Fold all minds `@release` tests into `test-minds-release`** (`ci.yml`, manual `run_minds_release_tests` dispatch): after the `minds_deployment` group it installs Chromium and runs the plain minds `@release` tests, selected by tag. Keeps the two release procedures separate (mngr = `v*` tag / dispatch; minds = manual dispatch).
  - Sets a fixed `PLAYWRIGHT_BROWSERS_PATH` (the minds conftest re-points `$HOME` per test, hiding the default browser location).
  - Clears `apps/minds`'s `--timeout=10` addopts (picked up because the target is `apps/minds`) and sets `--timeout=300`.

## Fixes to the two pre-existing failures

- **`test_claude_version_alignment`** (real drift): the release `Dockerfile` pinned Claude Code `2.1.141` but forever-claude-template `main` is `2.1.160`. Bumped the Dockerfile pin and the other pins that track it (`release-tests.yml`, `tmr-setup` action, `snapshot_minds_e2e_state.py`) to `2.1.160`.
- **`test_sse_redirect`** (stale test): the creating page now takes a `CreationId` and polls the v1 operations resource (`/api/v1/workspaces/operations/create/<creation_id>`) for completion, not the old core SSE endpoint. Reworked the test to key the fake creation by a `CreationId`, mount `/api/v1` (pass `paths`), and assert the canonical `/goto/<agent>/` redirect.

## Verified locally

- Tag/marker collection all correct; `--ignore apps/minds` prunes all minds tests from the mngr collection; folded selector collects exactly the 3 plain minds `@release` tests.
- Folded step runs with a 300s per-test timeout.
- All three plain minds release tests now behave correctly: `test_claude_version_alignment` **passes**, `test_sse_redirect` **passes**, `test_aws_workspace_release` skips (no AWS opt-in).
- `-m minds_deployment` (3) / `-m minds_services` (4) selection unchanged (orchestrator + minds CI jobs unaffected).
- Changelog gate passes (touches `dev` + `minds` + `mngr`); `vet` clean.

## Follow-up (not in this PR)

- If any GitHub branch-protection required-status-checks reference the old job names `test-release` / `test-docker-release`, rename them in repo settings. (Confirmed unlikely: this workflow only runs on `workflow_dispatch` + `v*` tags, never on PRs.)


joshalbrecht:
lgtm!

PR #2350, opened 2026-07-01
Make remote host-lock acquisition resilient to transient SSH failures

qi-imbue:
## Why

A scheduled TMR run ([28506362822](https://github.com/imbue-ai/mngr/actions/runs/28506362822)) crashed the whole orchestrator when a single mapper host's SSH connection dropped:

```
paramiko.ssh_exception.SSHException: SSH session not active
ConcurrencyExceptionGroup: SSH session not active (1 sub-exception)
```

The mapreduce/TMR orchestrator launches one agent per task across many remote hosts and already isolates per-agent failures (`except (MngrError, OSError, BaseExceptionGroup)` in `launch_all_mappers`). But `_hold_remote_host_lock` (reached by every `mngr create` on a remote host, via `create` → `host.lock_cooperatively`) called `transport.open_session()` with **no retry and no error wrapping**. When a host's transport died ("Connection reset by peer" → dead transport → `open_session` raises `SSHException("SSH session not active")`), the raw paramiko exception leaked past the orchestrator's `except`, escaped the `ConcurrencyGroupExecutor` `with` block, and was re-raised as a fatal `ConcurrencyExceptionGroup`.

The lock path was the **one** remote operation that neither retried transient SSH errors nor translated them, unlike `_run_shell_command` and the file-transfer paths.

## What

- Move channel acquisition into `_open_remote_lock_channel`, decorated with the existing `@_retry_on_transient_ssh_error` and disconnecting between attempts so a dropped connection is rebuilt before the retry (mirrors `_run_shell_command_with_transient_retry`).
- Wrap post-retry failures via the existing `_translate_ssh_errors` / `_notify_on_connection_error` so a persistently unreachable host surfaces a structured `HostConnectionError` (a `MngrError`) instead of a raw paramiko exception.
- The channel is closed on every acquisition failure path, so no channel leaks.

No mapreduce code change is needed: its per-agent `except (MngrError, ...)` already isolates failures — it just needs the host layer to raise an `MngrError`, which it now does. So one flaky host becomes a single recorded launch failure instead of a fatal crash.

## Testing

- Two new unit tests in `host_test.py`:
  - a transient `open_session` failure disconnects, reconnects, and retries to success;
  - a persistent SSH failure surfaces as `HostConnectionError`.
- Full `host_test.py`: 193 passed. mngr exception-handling ratchets: pass (no new broad/base `except`). `ty check` clean. vet: no issues.

## Manual testing

- A fresh TMR run should complete end-to-end; if a mapper host's SSH drops, that mapper should appear as a single failed launch in the report while the rest of the run proceeds (rather than the whole run aborting with `SSH session not active`).

### DIFF
--- a/libs/mngr/imbue/mngr/hosts/host.py
+++ b/libs/mngr/imbue/mngr/hosts/host.py
@@ -22,6 +22,7 @@ from typing import assert_never
 from uuid import uuid4
 
 from loguru import logger
+from paramiko import Channel
 from paramiko import SSHException
 from paramiko import Transport
 from pydantic import Field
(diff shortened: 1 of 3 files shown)


libs/mngr/imbue/mngr/hosts/host.py
```diff
@@ -858,25 +859,29 @@ def _hold_remote_host_lock(self, lock_file_path: Path, timeout_seconds: float |
...
+                failed="Could not acquire host lock due to connection error",
+                closed="Connection was closed while acquiring host lock",
+            ),
+        ):
+            channel = self._open_remote_lock_channel(lock_file_path, timeout_seconds)
         try:
-            with log_span("acquiring host lock at {} (over SSH)", lock_file_path):
-                channel.exec_command(_build_remote_lock_command(lock_file_path, timeout_seconds))
-                _wait_for_remote_lock_acquired(channel)
-            is_lock_acquired = True
```
qi-imbue, line 873: This has been moved into the new `_open_remote_lock_channel` above and thus subject to the same exception wrapping and retrying.

This makes sense to me (and it slightly simplified things by getting rid of `is_lock_acquired`), but I'm not 100% confident.

joshalbrecht:
lgtm, conditional on us doing the fix on the other side (disconnection) as well (per that other design PR)

PR #2352, opened 2026-07-01
Generalize TMR into named variants (mngr vs minds suites)

qi-imbue:
## Summary

Generalizes TMR so distinct test suites (the mngr release suite and the minds release suite) run as separate, independently reviewable runs. Per the design discussion, variants are expressed as **CLI flags composed in CI/justfile**, not a new in-package registry. Builds on main's path-based split of the two release suites (`libs/...` vs `apps/minds`).

## Changes

**Framework (`mngr_mapreduce`)**
- `MapReduceRecipe.name` is now a per-instance value instead of a `ClassVar`, so one recipe class can serve multiple named variants.

**TMR (`mngr_tmr`)**
- `--name <slug>` — the variant prefix on agent/branch/host names (`tmr-mngr/<run>/*` vs `tmr-minds/<run>/*`), keeping the two suites' branches and PRs separate. Distinct from `--run-name`. Validated as a slug.
- `--mapper-prompt` / `--reducer-prompt` — override the packaged Jinja templates (an override may `{% extends %}` / `{% include %}` the packaged ones by name).
- Reintegrate hint carries `--name` for non-default variants.

**Minds variant**
- `apps/minds/tmr/mapper.j2` — a minds-tailored mapper prompt (drops the mngr-only tutorial-block guidance; adds minds setup-blocker context: Docker daemon, snapshot image, deployed env, secrets → flagged with `FIXME(tmr)` + BLOCKED).
- `justfile`: `tmr-mngr` / `tmr-minds` recipes as the canonical per-suite flag sets. `tmr-minds` targets `apps/minds` with the minds prompt and defaults to the plain `@release` tests; capability suites are documented as opt-in extra args.
- `tmr.yml`: `name` / `mapper_prompt` / `reducer_prompt` workflow inputs plus a documented minds dispatch preset.

## On the capability suites (correcting my earlier note)

An earlier version of this PR said the minds capability tests "can't run on a plain Modal agent host." That was too strong. Each capability suite has an infra prerequisite TMR can satisfy by injecting the right secrets/config via `--env` / `--snapshot`:

- `minds_snapshot_resume` — needs the pre-baked snapshot image + a Docker-capable (`vm_runtime`) host + `ANTHROPIC_API_KEY`. TMR already boots all mappers from a snapshot (`--snapshot <id>`), exactly like the offload snapshot stage; the real constraint is provisioning `vm_runtime` + dockerd hosts (provider config).
- `minds_services` — needs a running shared `ci-*` env; the fixtures already read endpoints/secrets from injected env vars, so `--env MINDS_DEPLOYMENT_TEST_ENVS_JSON=... ` + tier secrets points every mapper at one env you stand up once.
- `minds_deployment` — each test mints/destroys its own ephemeral `ci-*` env; injectable via `--env`, with an eyes-open cost (N parallel mappers = N real cloud deployments) and leak-cleanup (`just minds-test-deployment-cleanup`) tradeoff.

The wired `tmr-minds` default targets the plain minds `@release` tests that need none of this; the capability suites are opt-in.

## Testing

- `just test-quick libs/mngr_tmr` (126) and `libs/mngr_mapreduce` (100) pass; ratchets pass.
- `uv run ty check`: all checks pass.
- vet (agentic): no issues on the first commit.


joshalbrecht:
lgtm!  thanks Qi :)

PR #2353, opened 2026-07-01
mngr: bound per-host discovery reads (no abandoned-thread leak)

joshalbrecht:
## What & why

Follow-up to per-provider discovery (PR #2335), **stacked on `mngr/per-provider-discovery`**. Implements the two changes specced in `blueprint/per-provider-discovery/spec-bounded-per-host-discovery.md`: make the per-host-bounded discovery read time-bounded so a wedged host cannot leak background threads, and de-duplicate in-flight per-host reads across polls.

Per-provider discovery runs each host's agent read on an abandonable daemon thread and marks a host UNKNOWN if it doesn't finish within `host_discovery_timeout_seconds`. But the abandoned thread kept running until the underlying read returned, and the discovery reads were **not** time-bounded, so a host that connects and then stalls mid-read (sshd hung after auth, network drops mid-transfer) could leak one thread per poll, forever.

## Change 1: bound the per-host read with a hard timeout

`host_discovery_timeout_seconds` is now threaded down as a hard per-command timeout on the two reads discovery issues:

- the agent-directory listing (`ls`, via `execute_idempotent_command(timeout_seconds=...)`), and
- each `data.json` read (SFTP, via a new `read_text_file_within_timeout` that sets the SFTP channel socket timeout).

On timeout, pyinfra/paramiko raises, which the existing `hosts/` machinery converts to `HostConnectionError`; the discovery path already catches that and falls back to the host's last-known offline agents. So a stalled host resolves to its offline agents *within the timeout* rather than hanging. The bound is threaded only through the discovery path (a new optional `timeout_seconds` on `discover_agents` / `_list_directory` / a new `read_*_within_timeout` helper); the shared `read_file` / `read_text_file` methods and their interface are untouched, so all other callers are unaffected.

Semantics: the bound is per-command, so a host with N agents can take up to ~N x the timeout; the outer per-host wall-clock wait remains the overall guarantee. Change 1 only guarantees each op self-terminates, which is what stops unbounded thread accumulation. (The residual `recv_exit_status()` after a read has no timeout -- a narrow corner, left as-is.)

## Change 2: cross-poll per-host in-flight de-duplication + warn

Added `HostDiscoveryReadRegistry`, a long-lived per-host `host_id -> Future` map owned by each provider's discovery poller. On each poll, a host whose prior read is still in flight is **not** re-read -- the in-flight read is reused and the host is reported UNKNOWN again -- so at most one abandoned read exists per host at a time. Every such skip is logged at **warning** with the host id and provider: with Change 1 in place this should essentially never fire, so it is a precise "host wedged past its timeout" alarm. When no registry is passed (one-shot discovery), a fresh per-call registry is used, so there is no cross-poll state.

## Non-goals (unchanged, per the spec)

- **Connect-phase timeouts** -- already bounded (~10s pyinfra `CONNECT_TIMEOUT` + ~15s paramiko banner); left untouched.
- **`complete_names.py`** -- intentionally eventually-consistent; not touched.
- **Batch providers** (modal / vps / imbue_cloud) accept the new `registry` param for interface compatibility but are unaffected: they read all hosts in one bounded pass and spawn no per-host reads.

## Testing

- New unit tests: the per-host timeout is threaded into `read_host_agents_for_bounded_discovery` (not left `None`); `discover_agents(timeout_seconds=T)` uses the bounded directory-listing and bounded file-read helpers (asserts it does *not* fall back to the unbounded `read_text_file`); and two consecutive polls sharing one registry start a permanently-gated host's read exactly once, report it UNKNOWN both times, and emit the wedged-host warning.
- Touched-area suites pass locally (`host_test`, `provider_instance_test`, `provider_discovery_stream_test`, `gc_test`, `destroy_test`, `outer_host_test`), plus the ratchets for all four touched projects. `ty` and `ruff` are clean. Full `just test-offload` is the final gate (CI).
- Per-PR changelog entries added for every touched project (`mngr`, `mngr_modal`, `mngr_vps`, `mngr_imbue_cloud`, `dev`).

### DIFF
--- a/blueprint/per-provider-discovery/spec-bounded-per-host-discovery.md
+++ b/blueprint/per-provider-discovery/spec-bounded-per-host-discovery.md
@@ -1,6 +1,8 @@
 # Spec: bound per-host discovery reads (no unbounded abandoned-thread leak)
 
-Status: proposed follow-up to the per-provider-discovery branch (`mngr/per-provider-discovery`, PR #2335). Not yet implemented.
+Status: implemented on branch `mngr/bounded-per-host-discovery` (stacked on
+`mngr/per-provider-discovery`, PR #2335). Both changes below are in place; the
+connect-phase timeouts were deliberately left untouched per the non-goals.
 
 ## Context
 
--- a/blueprint/per-provider-discovery/spec-bounded-per-host-discovery.md
+++ b/blueprint/per-provider-discovery/spec-bounded-per-host-discovery.md
@@ -1,8 +1,22 @@
 # Spec: bound per-host discovery reads (no unbounded abandoned-thread leak)
 
-Status: implemented on branch `mngr/bounded-per-host-discovery` (stacked on
-`mngr/per-provider-discovery`, PR #2335). Both changes below are in place; the
-connect-phase timeouts were deliberately left untouched per the non-goals.
+Status: implemented on branch `mngr/bounded-per-host-discovery` (PR #2353,
+stacked on `mngr/per-provider-discovery` / PR #2335). Both changes below are in
+place; the connect-phase timeouts were deliberately left untouched per the
+non-goals.
+
+Correction (post-implementation): the original draft of this spec assumed the
+per-file `data.json` read (`read_text_file`) also went through
+`execute_idempotent_command` / pyinfra `_timeout`. It does not -- `read_text_file`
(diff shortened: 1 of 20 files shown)


joshalbrecht:
roughly lgtm, does seem pretty useful to bound these

PR #2357, opened 2026-07-01
minds: overlay modals as in-page JS/DOM instead of iframes

gdenisov:
Renders the minds desktop app's overlay modals (workspace menu, inbox, help, sign-in) as in-page DOM instead of iframes.

Stacked on top of #2339 (custom titlebar tooltips on a unified warm overlay surface), which is still open — so until #2339 lands, this PR's diff against `main` also includes that tooltip work. The net-new delta here (~20 commits on top of #2339) is the iframe-to-JS modal migration.

## What changed

Opening a modal no longer mounts a nested iframe. The always-warm overlay host fetches the modal's markup as a bare fragment (`?fragment=1`) and injects it directly into the shared overlay surface, and the modal's per-modal JS runs as a module in that host.

This removes a layer of iframe machinery from the overlay surface:

- the per-frame IPC fan-out and the SSE priming handshake,
- the `nodeIntegrationInSubFrames` subframe bridge,
- the front/back-buffer iframe swap.

The SSE-driven modals (workspace menu and inbox) now read the live workspace list / request state from a single cache the overlay host keeps — primed on load and kept current as events arrive.

Each modal's JavaScript is single-sourced in one module (the `auth.js` pattern) that drives both the injected fragment in the desktop app and the modal's standalone full page (the browser / dev path), so there is no duplicated per-modal script. The dead `sidebar.js` (an Electron-only iframe duplicate) is removed.

Behavior is unchanged — this is a structural simplification of the overlay surface, not a UX change.

## Scope

The delta on top of #2339 touches the minds desktop client only: `electron/main.js` + `preload.js`, the overlay statics (`overlay.js`, the new `overlay_{signin,help,sidebar,inbox}.js` modules, `auth.js`, `tooltip_triggers.js`), the modal templates (`Base.jinja` bare-fragment mode, `Help`/`Inbox`/`Sidebar`/`SigninModal`/`OverlayHost`), and their tests. The design plan lives under `blueprint/overlay-js-modals/`.

Generated with [Claude Code](https://claude.com/claude-code) via Sculptor.


apps/minds/imbue/minds/desktop_client/static/overlay.js
```diff
@@ -0,0 +1,330 @@
...
+    var m = el.getBoundingClientRect();
+    var w = Math.ceil(m.width);
+    var h = Math.ceil(m.height);
+    root.style.width = '';
+    root.style.height = '';
+    var a = cmd.rect || { x: 0, y: 0, width: 0, height: 0 };
+    // Centered under the trigger by default; flip above if it would overflow the
+    // bottom; clamp horizontally to stay on-screen.
+    var tx = a.x + a.width / 2 - w / 2;
+    var ty = a.y + a.height + TOOLTIP_GAP;
```
joshalbrecht: eh, "el" is too short to be a variable name, similarly for "a" and "tx", etc

We want to keep the LLM away from abbreviations and super short variable names, because that kind of code can be more confusing for it

I didn't see *too* much of that elsewhere in this PR, but worth having an LLM do a quick pass for otherwise abbreviated or super short names and correct those too

(my complain is both for individual variable names, and for the "El" in "ensureTooltipeEl" -- LLMs dont really mind typing out longer identifiers, so we might as well have them do so!)

apps/minds/imbue/minds/desktop_client/static/overlay_help.js
```diff
@@ -0,0 +1,224 @@
...
+      var idValue = find('#help-event-id');
+      if (eventId) {
+        idValue.textContent = eventId;
+        idRow.classList.remove('hidden');
+      } else {
+        idRow.classList.add('hidden');
+      }
+    }
+
+    var copyBtn = find('#help-copy-id-btn');
```
joshalbrecht: another example of needless abbreviation -- Button isn't many more tokens than Btn, so, might as well

joshalbrecht:
lgtm other than my stylistic comments above!

PR #2360, opened 2026-07-02
Local pages in chrome: de-mix the content surface (content view = agent-only)

gdenisov:
## Goal

De-mix the desktop client's two content surfaces so the app's **trust boundary matches its view boundary**: every trusted local/native page (Landing, Create, Settings, Accounts, Consent, Welcome, Creating, Destroying, workspace settings, sharing, recovery, auth flow) renders on the **chrome surface**, and the **content surface hosts only untrusted agent content** (`agent-*.localhost` / `/goto/<id>/`). Full design: `specs/local-pages-in-chrome/spec.md`.

Before this change, local pages shared the `contentView` with foreign agent content and were forced through the caged `content-relay-preload.js`; now they render on the trusted chrome surface and call the `window.minds` bridge directly.

## What ships (all phases + a review-hardening round)

**Templating.** Extracted the trusted app shell (fixed titlebar + browser-mode floating sidebar + accent / `chrome.js` wiring) into a reusable `ChromeShell.jinja`. Every local page renders it directly (the titlebar travels with the page); `Chrome.jinja` (the agent-content surface) is a thin wrapper passing `is_agent_content_surface=True`.

**Electron.** A pure `surface-routing.js` classifier + `navigateBundle()` route each navigation: agent URLs → the content view (shown over the `/_chrome` wrapper), local pages → the chrome view (content view hidden + parked, releasing the agent renderer). Accent/title/current-workspace are split across both surfaces' `did-navigate`. Back/forward, DevTools, and zoom follow an explicit `activeSurface` (the visible view).

**Lockdown + cleanup.** The content-relay allowlist is down to what foreign agent content legitimately needs (`open-request-modal`, `open-help`, `reload-crashed-view`); the sign-in / stop-mind / open-in-new-window / accent-preview relays are gone (they're `window.minds` bridge calls from trusted pages now). The content→default cookie watcher was removed (sign-in happens on the trusted default session); only the default→content push for `/goto` remains. Symmetric **defense-in-depth guards**: the content view can never navigate to a trusted backend-origin page, and the chrome view can never navigate to agent content — both cover server `302`s (`will-redirect`), with the chrome guard *rerouting* the agent URL onto the content surface.

**Recovery page** now renders under the app titlebar: the shared loading card was split out of the (byte-for-byte unchanged) `mngr_forward` proxy loader and wrapped in ChromeShell, so a window keeps Home / drag / window controls while a workspace restarts.

**Browser (non-Electron) mode reconciled.** Opening a workspace routes through the agent wrapper `/_chrome?workspace=<agent-id>` (the titlebar + sidebar wrap the workspace iframe) instead of full-navigating to the bare agent origin; the wrapper iframe defaults to `about:blank` (never `/`, which used to double the titlebar) and the requested workspace id is validated server-side. Browser navigation is now uniformly full-page.

## Review-round fixes (a full-diff review after the merges)

- **Crash**: the workspace-crash "Reload" called a function the split deleted (`ReferenceError`); both crash-recovery reloads route through the surface router now.
- **Agent-content leaks**: the recovery healthy-`302`, agent-controlled notification URLs, and the sharing-page link can no longer load agent/foreign content into the privileged chrome view.
- **Correctness/UX**: TLS scheme for workspace links (`https`), a Flask request-context guard, three titlebar CSS regressions (in-page modals under the titlebar, the Settings sticky nav, the sync banner), and backing out of a quit no longer force-reloads a live workspace.
- The e2e workspace runner + full-flow Home step were updated for the split.

## Testing

- Python `desktop_client` subtree: **1468 passed**. `mngr_forward`: **275 passed**. Ratchets: **57 + 57**. Node unit (`surface-routing` / `startup-routing`): **20 passed**. Both e2e DOM contract specs (`landing-stopped-mind-restart`, `recovery-redirect`) verified **green in a headless browser**. Ruff + design-system guards clean.
- New coverage: the local-page titlebar invariant now asserts across all converted pages (incl. recovery); a route test asserts the `/_chrome` iframe targets a workspace only for a valid id; a browser contract test asserts a running mind opens through the wrapper.

## Known follow-ups (minor, browser-mode only)

- The browser agent-wrapper's titlebar shows the neutral accent + generic title (the workspace iframe is cross-origin, so the poll can't read it) — the workspace still renders; only the tint/title is generic.
- Inbox / Help from the titlebar in browser mode full-navigate to their overlay pages (no titlebar there) — pre-existing; Electron uses the modal overlay.

## Needs a GUI pass

The Python/route/JS-DOM layers are covered by the suites above, but the Electron-runtime behavior (surface swap, guards, `activeSurface`, zoom, history, recovery titlebar) has no unit coverage — `main.js` can't be required. It's covered by CI's Electron acceptance test; a manual pass of the running app is still worth doing before merge.


gdenisov:
Superseded by #2476

PR #2366, opened 2026-07-06
[MIND-123] Managed git binary for the minds desktop app (design + implementation)

DanverImbue:
## Summary

Implements `specs/minds-managed-git/concise.md` (the spec is the first commit; implementation follows).

The minds desktop app's bundled git is now fully managed on macOS and Linux: a version-pinned, SHA256-verified [dugite-native](https://github.com/desktop/dugite-native) payload (git 2.53.0) acquired per `apps/minds/scripts/git-manifest.json`, replacing the unpinned Xcode-CLT copy on macOS and the incomplete `cp $(which git)` stub on Linux. Windows stays on already-pinned MinGit.

### Key pieces:
- **Manifest** (`git-manifest.json`): tag, git version, per-target assets + SHA256 for darwin-arm64/x64, linux-x64/arm64, win32-x64 (shipped + CI-gated: darwin-arm64, linux-x64). All five hashes independently recomputed from downloaded bytes.
- **Acquisition** (`download-binaries.js`): manifest-driven download/verify/extract, `.dugite-tag` staleness marker (`ensure-binaries.js` re-downloads stale dev payloads), GPLv2 `COPYING` + provenance `NOTICE` staged into the payload.
- **Symlink-free payload**: the payload's 145 symlinks (142 `libexec/git-core` builtins plus the `git-remote-https`/`ftp`/`ftps` helpers) are replaced at download time with tiny POSIX-sh shims. ToDesktop's app-files glob silently DROPS symlinks, so the raw payload would arrive on the build servers missing every builtin (breaking https clones cloud-side), while dereferencing copiers (electron-builder's extraResources) materialize a full copy per link. Shims are regular files and survive every copier, globber, and signer.
- **Runtime env contract** (`backend.js`/`paths.js`): `GIT_EXEC_PATH`, `GIT_TEMPLATE_DIR`, `GIT_CONFIG_SYSTEM`, Linux-only `GIT_SSL_CAINFO`, in dev and packaged modes; dev PATH now also runs the bundled git so binary and exec-path can't skew. Required because the payload bakes in an empty prefix — without it, https clones fail with "'remote-https' is not a git command" (reproduced, then proven fixed).
- **ToDesktop upload size**: `resources/` no longer uploads twice (app-files glob + extraResources), cutting the app-source upload from ~701MB to ~467MB against the 600MB `uploadSizeLimit`; `build.js` now fails fast at build time with an estimate that mirrors the CLI's pricing.
- **Freshness nag** (`.github/workflows/minds-git-freshness.yml`): weekly check for a dugite-native release carrying a newer upstream *git version* that has cleared the repo's 14-day dependency cooldown (the same minimum-release-age posture as `pnpm-workspace.yaml` and the packaged pyproject). It picks the highest git version among cooldown-cleared releases -- not the newest by date, so an old-line patch released after a newer line can't mask it -- and maintains a single tracking issue by stable title. Same-git-version dugite rebuilds are deliberately not chased; a maintainer may bump before the cooldown for an urgent CVE.
- **Tests**: 7 manifest/downloader drift-guard unit tests plus shim-conversion, upload-estimator, and ToDesktop-config guard tests in `build_test.py`; `test_bundled_git.py` acceptance test proving version, templates, and a hermetic HTTPS clone through `git-remote-https` against a local self-signed TLS server, including through the shimmed payload shape the app actually ships. CI runs it on both shipped targets: linux-x64 via offload (sandboxes have no network, so the mngr Dockerfile pre-fetches the pinned asset into an image cache, `MINDS_DUGITE_NATIVE_CACHE_DIR`, with the manifest in acceptance checkpoint `build_inputs`) and darwin-arm64 via a GitHub-hosted macOS runner (`test-minds-bundled-git-macos`, live download).
- **Docs**: `desktop-app.md` rewritten for the new reality + an "Updating the bundled git" runbook.

Verification: full CI green on the final HEAD -- offload unit+integration and acceptance suites, docker, minds snapshot, and the new `test-minds-bundled-git-macos` job all pass. The acceptance test also passes locally on both mac payload paths (image-cache and live-download); manual end-to-end clone battery against the staged payload under the exact backend env. A pre-existing extras-test flake (the real claude-CLI probe vs the global 10s offload pytest-timeout) is fixed by a shared `stub_claude_on_path` fixture -- a stub `claude` on PATH reporting every plugin installed, keeping the shell-out and JSON parsing exercised end to end; the real probe stays covered by a new acceptance test with a Node-sized timeout.

image

Packaged-app beta verification: `minds-launch-to-msg` dispatched against HEAD (`3f67f7e190`) + `minds-v0.3.8` is **green** -- [run](https://github.com/imbue-ai/mngr/actions/runs/29967174507) (17m18s on mac-runner), [#minds-notifications post](https://imbue-ai.slack.com/archives/C0B9TC4TMC2/p1784765070780799).


.github/workflows/minds-git-freshness.yml
```diff
@@ -0,0 +1,114 @@
+name: minds git freshness
```
weishi-imbue, line 1: just fyi, there's a pattern in mngr repo where mngr/minds will intentionally decline any new deps version if it's too recent. (2 weeks) This is meant to avoid any new security vulnerabilities before they're fully vetted. 
Depending on the nature of dugite updates, checking for security updates seems useful, but we shouldn't force to latest feature updates.

DanverImbue: Thanks for pointing that out, Weishi! Addressed.

DanverImbue:
Note, still needs testing before this is ready t omereg.

.github/workflows/minds-git-freshness.yml
DanverImbue: Good catch, thank you -- this was inconsistent with the cooldown we already document. Updated in `b623e1174d`: the workflow now ignores releases younger than the 14-day cooldown (matching `minimumReleaseAge` / `exclude-newer`) and nags only when a cooldown-cleared release bumps the upstream *git version* (`v<gitversion>-<build>`), so same-git dugite rebuilds aren't chased. A maintainer can still bump before the window for an urgent CVE.

joshalbrecht:
lgtm assuming that @weishi-imbue is onboard!

.github/workflows/minds-git-freshness.yml
```diff
@@ -0,0 +1,147 @@
...
+          releases_json=$(mktemp)
+          gh api "repos/desktop/dugite-native/releases?per_page=100" > "$releases_json"
+
+          candidate=$(mktemp)
+          jq '
+            [ .[]
+              | select(.draft == false and .prerelease == false)
+              | select((.published_at | fromdateiso8601) <= (now - (env.COOLDOWN_DAYS | tonumber) * 86400))
+            ]
+            | sort_by(.published_at) | reverse | (.[0] // null)
```
weishi-imbue: This is picking up the latest date, not necessarily the highest version. 
we probably want to sort by version desc, then pick the highest version that satisfies the cooldown.

libs/mngr/changelog/danver-embed-git.md
```diff
@@ -0,0 +1,3 @@
+The interactive `mngr extras` CLI tests (`test_extras_interactive_mode`, `test_extras_interactive_includes_default_type`) now carry the same 30s per-test timeout as `test_extras_no_args_shows_status`: they probe claude-plugin status (about 7s even locally), and one tripped the global 10s pytest-timeout on a contended offload sandbox.
```
weishi-imbue: this could be stale? didn't see the test change?

weishi-imbue:
> Remaining manual step: dispatch minds-launch-to-msg against this SHA for the packaged-app beta verification (ToDesktop build; not triggered from this branch automatically).

This is still the best way to capture bugs and issues today. Worth doing and link the slack msg in #minds-notifications

apps/minds/docs/desktop-app.md
```diff
@@ -101,11 +101,21 @@ The CLI separates two channels, following the same conventions as mngr:
...
+
+1. Pick the new dugite-native tag from the freshness tracking issue (or, for an urgent CVE, directly -- you may bump before the cooldown window at your discretion; the automated nag waits it out).
+2. Update `apps/minds/scripts/git-manifest.json`: the `dugiteNativeTag`, the `gitVersion`, all five asset names (each embeds a dugite-native commit short-SHA, so record them verbatim), and each target's hash taken from the release's `.sha256` companion asset.
+3. Independently download each tarball and recompute its SHA256, then compare against the values you just recorded (pinning defends against future substitution, not against copying a wrong value in).
+4. Run the bundled-git acceptance test locally on a mac; CI covers linux-x64.
```
weishi-imbue: github provides mac runner too. (owned by github, not our office self hosted ones). 
there shouldn't be any reason that CI can't cover this bundled-git acceptance test for mac as well. 
Can defer to later separate pr.

.github/workflows/minds-git-freshness.yml
DanverImbue: Good catch -- fixed in fc52ec29b1. The workflow still filters to cooldown-cleared releases first, but then sorts by numeric git version (newest-published as tie-break among same-git rebuilds) instead of published date. Verified against the live dugite-native release list, plus a simulated case where an old-line patch (v2.52.1-1) is the most recently published: the old date-sort picked it, the version sort correctly still picks v2.53.0-3.

libs/mngr/changelog/danver-embed-git.md
DanverImbue: Right, it was stale -- the paragraph described the intermediate 30s-timeout approach, which 805732cd78 replaced with the stub_claude_on_path fixture. Rewritten in 75c5a00b15 to describe what actually shipped.

apps/minds/docs/desktop-app.md
DanverImbue: Implemented directly rather than deferred: ci.yml gains a test-minds-bundled-git-macos job on GitHub-hosted macos-latest (3f67f7e190) that runs the acceptance test with a live asset download -- no Vault secrets needed, the dugite-native release is public. Verified locally on an arm64 mac with the exact CI pytest invocation; its first real CI run is on this branch's latest push. The runbook step now reflects that CI covers both shipped targets.

PR #2368, opened 2026-07-06
Avoid ordering fragility

hynek-urban:
This has been done in response to https://github.com/imbue-ai/mngr/pull/2297.

- Do not introduce unnecessary dependencies on ordering in the workspace permissions setup. (I believe it should make this a little less fragile going forward.)
- In the process, introduce a simple data migrations mechanism for `mngr_latchkey`.
  - It doesn't try to be a particularly robust system. Just something that's usable for now.
  - Adds the first migration to reflect the changes from this PR.
  - Ideally: migrations are very rare, simple and fool-proof.


joshalbrecht:
lgtm, and that seems better :)

thanks hynek!

PR #2370, opened 2026-07-06
minds: keep the recovery page checking so a healthy workspace isn't stranded after sleep

gnguralnick:
## Problem

Diagnosed from Sentry event `fc54dc12`: a user (on macOS) lands on the static **"Workspace unresponsive"** recovery page after their computer wakes from sleep, even though the workspace was healthy the whole time. Root cause, from the bug-report logs:

- The recovery page's in-container health probe (30s cap) was launched just before the laptop suspended and spanned the entire sleep; on wake it was declared timed-out and — combined with a pre-sleep `RUNNING` snapshot — misclassified as `HOST_UNRESPONSIVE`.
- That tier rendered a terminal verdict page that **never re-checked**, so even though the workspace answered ~3s later, the user stayed stranded (38 min, until they filed the report).
- The STUCK→recovery redirect was gated on discovery *freshness*, so healthy wakes also sat on a long "Loading workspace…" loader, and a persistently-stalled discovery pipeline could strand the user on that loader with no recourse.

## Fix (three coordinated changes)

1. **Cheap-probe-first, and always** — the recovery page arms an idempotent cheap liveness poll on load and under *every* waiting/terminal state (including the unresponsive/consent and restart-failed states, plus the new "Reconnecting" state). The instant the workspace answers, the page returns the user home (~1s), so a self-recovering workspace is never stranded on a verdict.

2. **A timed-out probe is non-evidence** — a timed-out in-container probe now classifies as a new `INDETERMINATE` dispatch tier ("keep checking"), not `HOST_UNRESPONSIVE`. `_classify_dispatch_tier` lets direct in-container evidence (a live `GET / 200`) short-circuit to `HEALTHY` first, then returns `INDETERMINATE` when the probe timed out or the snapshot isn't trustworthy, then the host-state verdicts. A clean-exit-no-sentinel (ssh dead) stays `HOST_UNRESPONSIVE`.

3. **Freshness gate moved from the redirect to the verdict** — the STUCK-redirect freshness suppression (and its now-redundant per-wake flip-check block) is removed; the recovery page appears promptly. The freshness predicate is relocated to the verdict path (`is_recovery_classification_trustworthy` in `workspace_recovery.py`, computed inside `probe_workspace_health`), so a negative verdict or auto-restart still waits for a trustworthy post-onset snapshot — while a stale snapshot or timed-out probe shows a live "Reconnecting…" state that self-heals instead of an indefinite loader.

All changes are in `apps/minds/imbue/minds/desktop_client/`.

## Tests

- Unit: `recovery_probe_test` (51), `workspace_recovery_test` (15, incl. relocated freshness-predicate tests + new INDETERMINATE/direct-evidence/precedence cases), `templates_test` (187, incl. INDETERMINATE-render + no-dead-end + cheap-probe-first), `test_desktop_client` (117), `api_v1_test` (82), `test_ratchets` (57) — all green locally.
- `ruff` clean, `ty` (type check) clean, changelog gate passes.

## Not covered / verify in review

- **Real-app sleep→wake** isn't unit-testable (it hinges on cold-start discovery timing); the JS wiring and classifier logic are asserted, but the end-to-end lid-close→reopen flow should be confirmed in the actual Electron app.
- The timeout→`INDETERMINATE` end-to-end path is covered at the classifier level rather than by a slow integration test (the in-container probe cap is a hardcoded 30s).


joshalbrecht:
lgtm

PR #2371, opened 2026-07-06
ratchets: bump bash strict-mode snapshot 11 -> 12 for sigwinch_panes.sh

joshalbrecht:
## Why

PR #2296's CI failed on `test_meta_ratchets.py::test_prevent_bash_without_strict_mode` (`assert 12 <= 11`).

The 12th violating script is `libs/mngr/imbue/mngr/resources/sigwinch_panes.sh`, added on `main` with the per-session SIGWINCH client-attached hook (#2322) *after* the ratchet snapshot was last pinned to 11. It deliberately uses `set -uo pipefail` (omitting `-e`, a best-effort tmux repaint sweep), so it counts against the strict-mode ratchet. The ratchet was already over its bound on `main`; #2296 merely inherited it by merging current `main`.

## What

- Bump the snapshot from 11 to 12.
- Refresh the test's docstring, which had gone stale (it referenced minds verify scripts that no longer exist and a local/CI count divergence that no longer holds).

Per instruction, this just bumps the ratchet rather than adding strict mode to the script.


joshalbrecht:
lgtm

PR #2372, opened 2026-07-06
Fix unbounded memory leak in long-running mngr observe processes

gnguralnick:
## Problem

Long-running `mngr observe --discovery-only` processes grow in memory without bound — observed climbing to multiple GB (~1 GB/day) on live minds hosts. `ps` RSS hides it (most of the footprint is compressed/swapped); the real growth shows up in `footprint`.

## Root cause

Each discovery poll spins up worker threads (via `mngr_executor`, running pyinfra) that each create a per-thread gevent Hub. `cleanup_thread_local_resources` destroyed each hub to reclaim its OS pipe descriptor, but `Hub.destroy` alone leaves the hub greenlet parked mid-throw of the `LoopExit` it raises into its parent. That exception's traceback pins the worker's frames — and transitively the poll's entire result graph (`ListResult`/`AgentDetails`/parsed JSON) — in a reference cycle anchored in greenlet C state the garbage collector cannot break. A stranded hub (plus everything it retained) accrued on every poll.

This is a known gevent bug: [gevent issue 1601](https://github.com/gevent/gevent/issues/1601). The maintainer diagnosed the same "cycles that live in C outside of where the GC can get to them" and the partial fix in [PR 1604](https://github.com/gevent/gevent/pull/1604) (v20.5.1) only helps if the hub is joined/switched-to after destroy — which this code did not do. The defect persists through the latest gevent (26.5.0, the vendored version) and current master.

## Fix

Join the hub before destroying it — the pattern gevent itself prescribes for this leak. Running `Hub.join` lets the loop finish and releases the pinned frames. The join is bounded by a timeout so an idle hub (the normal case, once a task's work has finished) cleans up instantly, while the pathological case (a task left a greenlet/watcher running) falls through to destroy rather than hanging the discovery loop.

## Verification

- Reproduced the leak and validated the fix on the affected gevent version (26.5.0): destroy-only grows RSS ~66 MB per 400 polls, stranding one hub each; join-then-destroy holds RSS and live-object counts flat.
- Added a regression test (`thread_cleanup_test.py`) driving repeated executor polls and asserting live gevent Hub objects do not accumulate. Confirmed it **fails** without the fix ("grew by 25 over 25 polls") and **passes** with it.
- ruff format/check, repo-wide type check, mngr utils ratchets, and the full mngr utils suite (461 passed, 1 skipped) all green.

## Note on rollout

The leaking processes run the **vendored** mngr in `~/.minds/.venv`. This fix reaches the minds app only after a release re-vendors mngr; already-running observe processes keep growing until then.


joshalbrecht:
lgtm

PR #2373, opened 2026-07-06
Add `just default-workspace-template-worktree` for developing default-workspace-template from mngr

weishi-imbue:
## What

Adds `just default-workspace-template-worktree [branch] [base]` (backed by `scripts/default_workspace_template_worktree.sh`): stands up an **independent** default-workspace-template checkout at `.external_worktrees/default-workspace-template` nested in the current mngr checkout, defaulting the template branch to the current mngr branch. It's the one-command version of the `CLAUDE.md` `.external_worktrees/` convention for developing default-workspace-template alongside mngr.

## Why

Developing the template from an mngr agent required a hand-rolled `git worktree add` from a hardcoded `~/project/...` path -- a person-specific path that doesn't reach agents (`mngr create` doesn't copy gitignored files) and breaks if the base clone is ever removed.

## How it works

- **Destination is repo-scope**: derived from `git rev-parse --show-toplevel`, so it lands at the right `.external_worktrees/default-workspace-template` whether run from an operator clone or an agent's `~/.mngr/worktrees/<agent>/` checkout -- no path baked in.
- **Independent clone, not a worktree**: each checkout is a full clone with its own `.git`, so it survives deletion of any other clone or cache (a `git worktree` hung off a base repo does not).
- **Zero-config for agents**: clones the template directly from the remote via `GH_TOKEN`; the token is never persisted (`origin` is reset to the plain URL).
- **`DEFAULT_WORKSPACE_TEMPLATE_DIR` is an optional speed hint**: `git clone --reference-if-able "$DEFAULT_WORKSPACE_TEMPLATE_DIR" --dissociate` borrows objects from a local clone, then dissociates so no lasting dependency is created.
- **Rejects rather than silently reuses**: errors if a checkout already exists (with an `rm` hint) or if the branch already exists on the remote (with delete/rename hints), so a reused branch name never inherits stale history.

## Also removes the `~/project/...` hardcode

- `minds-start`, `propagate_changes`, the `minds-dev-workflow` skill, and `CLAUDE.md` now point at `just default-workspace-template-worktree`.
- `bake-slice-dev` resolves its workspace dir from the explicit arg, else `DEFAULT_WORKSPACE_TEMPLATE_DIR` (shell or gitignored `apps/minds/.env`), else the `.external_worktrees/default-workspace-template` checkout -- no personal path baked in.
- `create-new-mind-repo` no longer requires a local template clone: it clones the template from the remote, so the vestigial local-clone check was removed.
- `.mngr/settings.toml` COPYs the gitignored `apps/minds/.env` into each mngr agent worktree, so an agent inherits the operator's `DEFAULT_WORKSPACE_TEMPLATE_DIR` speed hint.

## Rebased onto the repo rename

The template repo was renamed `forever-claude-template` -> `default-workspace-template` (mngr #2394, template #256, both merged to main). This branch was merged with `main` and its additions rewritten to the new names via `scripts/rename_template_repo.py` (`--check` reports no remaining live references): `just fct-worktree` -> `just default-workspace-template-worktree`, `scripts/fct_worktree.sh` -> `scripts/default_workspace_template_worktree.sh`, `FCT_DIR` -> `DEFAULT_WORKSPACE_TEMPLATE_DIR`.

## Testing

- `just --list` parses and shows `default-workspace-template-worktree`; `scripts/default_workspace_template_worktree.sh` passes `bash -n`.
- End-to-end: `just default-workspace-template-worktree <branch>` clones `default-workspace-template` into `.external_worktrees/`, checks out the branch, leaves `origin` tokenless, and produces a fully independent `.git`.

Changelog entries: `dev/` and `apps/minds/`.


weishi-imbue:
Superseded by #2483 (renamed branch to wz/dwt-worktree to drop the last `fct` reference; adds `dwt` as the documented shorthand). Same change, current with main.

PR #2375, opened 2026-07-06
Fix macOS grandparent-death watcher no-op that orphaned the minds backend (MIND-103)

DanverImbue:
## Original bug

[MIND-103](https://linear.app/imbue/issue/MIND-103/macos-minds-backend-mngr-tree-orphaned-on-unclean-electron-exit): on macOS, when the Minds Electron GUI exits *without* running its clean shutdown path (force-quit, crash, `kill -9`, OOM — anything that skips the `before-quit` handler), the `minds run` Python backend and its entire `mngr` subprocess tree (`mngr forward`, `mngr event`, `mngr observe`, `mngr latchkey forward`) are orphaned, reparent to `launchd` (PID 1), and run forever — one leaked tree per session, holding ports and tailing logs.

## Root cause

The safety net for unclean Electron exits is the in-process **grandparent-death watcher** (`start_grandparent_death_watcher`, armed at `apps/minds/imbue/minds/cli/run.py:325`). It watches the *grandparent* (Electron) because the backend's immediate parent is the `uv run` wrapper, which survives Electron's death (it merely reparents). Arming the watcher depends on `_read_grandparent_pid()` in `libs/mngr/imbue/mngr/utils/parent_process.py`, which resolved the grandparent PID **only** by reading Linux's `/proc/<ppid>/status`. macOS has no `/proc`, so `open()` raises `OSError`, the function returns `None`, and the watcher logs *"no resolvable grandparent; skipping"* and never arms. The cross-platform `getppid`-based `start_parent_death_watcher` does not help — the backend's immediate parent (`uv run`) never dies, so `getppid()` never changes.

Confirmed on a macOS dev machine before the fix: `os.getppid()` returned a valid PID, but `_read_grandparent_pid()` returned `None`.

## Fix

Split PID resolution into two helpers and compose them:

- `_read_ppid_via_proc(pid)` — the existing `/proc/<pid>/status` fast path (Linux).
- `_read_ppid_via_ps(pid, concurrency_group)` — new `/proc`-free fallback that runs `ps -o ppid= -p <pid>` (POSIX; works on macOS and Linux).
- `_read_grandparent_pid(concurrency_group)` tries `/proc` first, then falls back to `ps`.

`start_grandparent_death_watcher` threads its `ConcurrencyGroup` down so the `ps` child is spawned via `ConcurrencyGroup.run_process_to_completion` (not raw `subprocess`) and is tracked/reaped like every other child — honoring the repo's `PREVENT_DIRECT_SUBPROCESS` policy. Linux behavior is unchanged: the `/proc` path still wins and `ps` is never spawned there.

**Follow-up (commit `4d2c843569`):** `concurrency_group` was initially `Optional[...] = None`, with the helpers self-provisioning a throwaway group when omitted — a convenience used only by test call sites; production always passed the watcher's group. That optionality made ownership of the spawned `ps` call-path-dependent and left future callers an escape hatch out of their own cancellation tree, so the parameter is now required: ownership is explicit at every call site, the signature enforces at the type level what the `None` fallback used to backstop at runtime, and the tests open their own `ConcurrencyGroup`.

## Reproduction

**Repro steps (from the ticket):**
1. On macOS, launch the Minds desktop app (spawns `uv run ... minds run` → python backend → `mngr` children).
2. Force-quit / hard-kill the Electron GUI so the `before-quit` handler does not run (`kill -9` the Electron process, or a crash/OOM).
3. `ps aux | grep mngr` — the backend and its `mngr` tree are still resident, reparented to PID 1, and never exit.

**Expected vs actual**
- Expected: the backend + `mngr` tree are SIGTERM'd shortly after Electron disappears (as already happens on Linux).
- Actual (before fix): the tree is orphaned under `launchd` and runs indefinitely.

## Proof of work

This bug has no UI surface; the evidence is process-tree/behavioral. (The transcripts below were captured before commit `4d2c843569` made `concurrency_group` a required argument — at the time, the helpers were callable bare.)

**Before (buggy resolution on macOS):**
```
os.getppid()            = 7355         # a real, live parent
_read_grandparent_pid() = None         # -> watcher logs "no resolvable grandparent; skipping"
```

**After (resolution + full reaping on macOS):**
```
_read_ppid_via_proc(ppid) = None       # no /proc on macOS
_read_ppid_via_ps(ppid)   = 14450      # ps fallback resolves it
_read_grandparent_pid()   = 14450      # watcher now arms
```

Live end-to-end proof — a real 3-level tree (grandparent = stand-in Electron → parent = `uv run` wrapper → leaf = backend) where the leaf arms the watcher and the grandparent is then `SIGKILL`ed:
```
Grandparent death watcher started (grandparent PID=23255)
leaf pid=23257 armed; leaf resolved grandparent=23255
SIGKILL-ing the grandparent (unclean exit)
Grandparent process (PID 23255) no longer exists, sending SIGTERM
PASS -- leaf reaped ~3.1s after grandparent died (result=SIGTERM)
```

**Test**
- Path: `libs/mngr/imbue/mngr/utils/parent_process_test.py`
- Kind: unit test (`test_read_ppid_via_ps_returns_parent_pid_of_a_child_process`). The fault is isolated ppid-resolution logic; the `ps` resolver is the exact mechanism that runs on macOS, and testing it directly gives a real red→green in the Linux per-PR CI (an end-to-end "kill Electron, inspect `ps`" test would only fail on macOS, which per-PR CI never runs, and would be slow/flaky).
- Failing-test commit: `870beb5062`
- Fix commit: `70d752b156`

Local verification (all green): the 5 `parent_process_test.py` tests (the new one now passes; `test_read_grandparent_pid_returns_alive_grandparent` now *passes* on macOS instead of skipping), 461 `libs/mngr/imbue/mngr/utils` unit tests, the `PREVENT_DIRECT_SUBPROCESS` ratchet (count unchanged at 15), `ruff` (lint + format), and `ty` (types). After the follow-up commit, the full suite was also run locally on macOS via `just test-offload`: 15060 passed, 0 failed; `vet` (agentic mode, with session history) found no issues. The full offload suite runs here on CI as well.

## Deferred follow-ups

- **Simplify to `psutil`.** `psutil` (already a `libs/mngr` dependency and used in `conftest.py` / `providers/local/instance.py`) offers a cross-platform `psutil.Process(pid).ppid()` that would replace the `ps` shell-out entirely. The `ps` fallback was chosen deliberately here (approved during the fix); `psutil` is a clean future simplification worth considering.
- **Broader `/proc`-first lifecycle code.** As the ticket notes, other lifecycle/process code is Linux-first (`recovery_probe.py`, `host.py`, etc. read `/proc`). This PR fixes only the grandparent-death watcher — the reported symptom — and leaves the broader `/proc`-free audit as separate work.

## Review notes

Self-review (`/code-review`, high effort) found no confirmed correctness bugs. Declined findings:

- `libs/mngr/imbue/mngr/utils/parent_process.py:100` — `except (FileNotFoundError, OSError)` is technically redundant (`FileNotFoundError` subclasses `OSError`). Kept intentionally to match the existing `run_process_to_completion` error-handling idiom at `libs/mngr_notifications/.../notification_verifier.py:89`.
- `psutil` reuse — see Deferred follow-ups above (declined here per the approved `ps` approach).
- A finder flagged the trailing `return None` in `_read_ppid_via_proc` as dead code — refuted: it is reached when the status file opens but contains no `PPid:` line, matching the original function's behavior.
- A finder noted `int(line.split()[1])` could raise on a malformed `/proc` line — this is pre-existing (unchanged from the original) and not realistic on a functioning Linux `/proc`; left as-is to preserve exact prior semantics.

---


libs/mngr/imbue/mngr/utils/parent_process.py
```diff
@@ -55,27 +57,81 @@ def start_parent_death_watcher(concurrency_group: ConcurrencyGroup) -> None:
...
                 if line.startswith("PPid:"):
-                    grandparent_pid = int(line.split()[1])
-                    return grandparent_pid if grandparent_pid > 1 else None
+                    return int(line.split()[1])
     except OSError:
         return None
     return None
 
 
+def _read_ppid_via_ps(pid: int, concurrency_group: ConcurrencyGroup | None = None) -> int | None:
```
DanverImbue: I hate that it made this an Optional out of nowhere. So lazy.

Going to scold it and get this back in line.

joshalbrecht:
lgtm, see slack for further discussion

PR #2376, opened 2026-07-07
Fast Lima create: pre-baked image fixes and release tooling

weishi-imbue:
Completes the pre-baked Lima image path: a local Lima create of the default workspace boots a baked image instead of building the toolchain in-VM (~320s -> ~45s).

**This merges dormant.** No environment sets `lima_image_base_url`, so no image source is built and every create takes the current path. Turning it on is a separate config PR.

## Issues found and fixed

**The baked image was tied to the machine that baked it.** Lima names the guest user after the host account and passes that account's uid into cloud-init. macOS gives its first account uid 501, so on a host whose account had a different name, `useradd <name> --uid 501` failed with "UID 501 is not unique", the users module aborted, and no user existed to accept Lima's SSH key. Creates hung on `Waiting for the essential requirement 1 of 3: "ssh"` until they timed out. The bake now removes all user accounts before capture and fails if any survive, and the build boots the finished image as a host user it has never seen (uid 501) so a regression fails the build rather than shipping.

**`desync` was not reachable in packaged builds.** It was staged only by the ToDesktop `beforeInstall` hook, whose output is folded into `app.asar`, while the app resolves bundled binaries under `process.resourcesPath`. Packaged apps carried an x86_64 copy inside the asar (verified by reading its Mach-O header) and no usable one. `build.js` now stages it, it is resolved by absolute path, and it is signed for notarization alongside `limactl` and `restic`.

**Raw sparse image, no `qemu-img`.** Lima embeds a pure-Go image reader and boots raw directly, so the qcow2 round-trip is gone: the bytes the signed manifest covers are now exactly the bytes Lima boots, an upgrade does one multi-GB pass instead of three, and the app bundles no converter. Sparse raw costs no extra space (4.9 GiB on disk vs 5.1 GiB for qcow2).

**A republished image could never reach a client.** The cache keyed on `(version, arch)` plus "the file exists", so an image corrected under an existing tag was ignored forever. The installed image is now re-checked against the hash the signed manifest names (recorded at install, not re-hashed each launch) and re-fetched when they disagree.

**SHA-pinned creates missed the fast path.** The gate compared refs by name, and CI pins a commit for reproducibility, so the end-to-end run meant to exercise the fast path never reached it. The release tag is resolved to its commit once at startup and either form is accepted.

**The baked disk was larger than the create's disk.** Lima grows a smaller image but fatals on shrink, so a 40GiB image broke every default create. The bake now matches the 20GiB the create template passes.

## Create resilience

The image being *absent* never leaves a user worse off than they are today; only an image that fails to verify stops anything.

- Nothing published, download failed (network, disk, missing tool), download stalled, or the wait ran out -> build in-VM. The prefetch keeps running, so the next create is fast.
- A download that fails to verify against the signed manifest is never booted and fails loudly, and is not retried.
- Waiting is bounded: 10 minutes overall, and a download that gains less than 32MiB in two minutes is treated as dead and abandoned at once (a healthy one moves that much every few seconds).
- The create log reports the download's progress and says why it fell back. `desync` draws a progress bar only on a tty, so a packaged app previously showed nothing while it waited.

## Release and deploy

- `scripts/lima_image/setup_tier.py`: provisions the per-environment bucket, custom domain, and a bucket-scoped publish credential (Cloudflare API only; no Mac needed).
- `scripts/lima_image/publish.py`: chunks, uploads, and signs; the signature is uploaded before the manifest, so a client never sees a manifest it cannot verify.
- `docs/release.md`: publish step plus a gate (§8c) that fails the release if the released tag has no published image. Skipping the publish is otherwise silent -- clients quietly fall back to the slow path and nothing turns red.
- The image must be served from a custom domain: an extract pulls ~66,000 chunks and the managed `r2.dev` origin rate-limits it into `429`s. Publishing must use R2's S3 API; the Cloudflare REST object API caps at 1200 requests per 5 minutes and cannot carry the store.

## Verification

- **End-to-end on a Mac runner:** cold create 167s (including the download), warm create 30s, `total_create_s` 45.1 vs ~320s on the slow path. Zero SSH failures.
- **Bakes at `minds-v0.3.6` and template `main`**, both booting as a foreign user in ~25s.
- **Full bake on a real Linux/KVM host** (Lima's qemu driver, which macOS never exercises). It surfaced two undocumented requirements for the amd64 bake host -- EDK2 firmware (`ovmf`) and the virtio option ROMs (`ipxe-qemu`), without which `limactl start` fails before the bake begins. Both are now checked up front and documented.
- **Cold download measured:** 349s for 66,034 chunks. **1.8 GB over the wire** (compressed chunks), decompressing to a **5.8 GB** sparse image whose hash matches the signed manifest. (The store ships zstd-compressed chunks; the ~5.8 GB is disk-write, not transfer.)
- **Provisioner verified on Debian**, including the negative cases: a surviving `/home` user fails the bake, and a missing `SUDO_USER` refuses to run.
- Full suite green.

## Measurements (reproducible)

All commands use the bundled desync (`node apps/minds/scripts/ensure-binaries.js` stages `apps/minds/resources/desync/desync`) and the dev bucket `$BASE=https://lima-images-dev-weishi.minds-dev.com`.

**Download vs install size** — compressed transfer is 1.8 GB, not the 5.8 GB on disk:
```bash
curl -fsS "$BASE/indexes/minds-v0.3.6/aarch64.caibx" -o img.caibx
desync extract -k -n 32 -c cache -s "$BASE/store/" img.caibx out.raw
du -sh cache    # network (compressed chunks fetched) -> ~1.8G
du -sh out.raw  # install  (decompressed image)        -> ~5.8G
```

**Storage/transfer format = per-chunk zstd** — `file cache/**/*.cacnk` -> "Zstandard compressed data" (magic `28 b5 2f fd`); filename is the content hash.

**Raw vs qcow2 chunk churn is equal** (so raw is chosen for layout-stability guarantees, not a measured delta) — 256 MB image, change 1 MB, both `desync make`d and diffed with `--seed`: raw fetches 20 chunks, uncompressed qcow2 fetches 20. Identical. Only `qcow2 -c` (compressed) churns more, which the publish path never produces.

**qcow2 needs no bundled qemu-img either** — `limactl start` on a qcow2 base with `qemu-img` off `PATH` logs `Converting "…/image" (qcow2) to a raw disk` (in-Go `nativeimgutil`), zero qemu-img calls, and boots. So dropping the bundled qemu-img is a win over the *old* design, not over qcow2.

**Sparse raw ≤ qcow2 on disk** — `du -h out.raw` ~4.9 G real (20 G apparent) vs ~5.1 G for the qcow2 the old flow shipped.

## Next steps

1. Merge this (dormant).
2. Provision staging/production buckets and generate a minisign keypair per tier.
3. Bake and publish per arch (arm64 on Apple Silicon, amd64 on a Linux/KVM host).
4. Config PR setting `lima_image_base_url` + `lima_image_minisign_public_key`, which turns it on.


joshalbrecht:
Hmmm, something seems to have gone wrong here:

> Cold download measured: 349s for 66,034 chunks (~5.5GB transferred), and the assembled image hashes to exactly what the signed manifest names.

I remember seeing something closer 1.4GB, so... not quite sure what that's about?

PR #2378, opened 2026-07-07
Add is_vm_runtime_enabled setting to Modal provider

qi-imbue:
## Summary

Adds an `is_vm_runtime_enabled` setting to the Modal provider (`ModalProviderConfig`), defaulting to `true`, which controls whether Modal sandboxes run on Modal's [VM runtime](https://modal.com/docs/guide/vm-sandboxes) instead of the default gVisor runtime.

When enabled (the default), sandboxes are created with `experimental_options={"vm_runtime": True}`. Set `is_vm_runtime_enabled = false` under `[providers.modal]` to opt back out.

## Changes

- `libs/mngr_modal`: new `is_vm_runtime_enabled` config field; `_build_experimental_options()` helper wired into both sandbox-creation paths (fresh create and snapshot restore); README config example updated.
- `libs/modal_proxy`: `ModalInterface.sandbox_create` now accepts an optional `experimental_options` mapping (mirroring `modal.Sandbox.create`), passed through in `DirectModalInterface`.

## Testing

- New unit tests assert the setting defaults to VM runtime enabled and that disabling it yields no `experimental_options`.
- `just test-quick libs/mngr_modal` (394 passed) and `just test-quick libs/modal_proxy` (102 passed) locally; `uv run ty check` and `ruff` clean.


joshalbrecht:
lgtm, but would it be possible to also change `scripts/snapshot_minds_e2e_state.py` to actually use our mngr code instead now?

Currently it does raw access to Modal Sandboxes, I *think* just because it needed access to this option that previously did not exist?

qi-imbue:
## Investigated: can `scripts/snapshot_minds_e2e_state.py` reuse mngr code now that we have vm_runtime support?

Conclusion: **no — routing it through mngr would add complexity, not remove it**, so I left the script unchanged.

The vm_runtime opt-in isn't why the script talks to Modal directly — it's a single self-contained line (`experimental_options={"vm_runtime": True}`). The script uses raw `modal` for three things, none of which get simpler through mngr:

1. **The image build is bespoke and unsupported by mngr's seam.** The script chains `debian_slim(python_version=…).apt_install(…).run_commands(…).env(…).add_local_dir(staged_repo, "/code/mngr").workdir(…).add_local_dir(fct_worktree, …)`. `modal_proxy`'s `ImageInterface` deliberately exposes only `apt_install` / `dockerfile_commands` / `build` — no `run_commands`, `env`, `workdir`, or `add_local_dir`. And `mngr_modal` builds images **only from a Dockerfile or a base registry image**. Routing through mngr means converting the whole fluent build into a Dockerfile + build context, including re-expressing the two `add_local_dir` bakes (staged repo + FCT worktree) as COPY layers.

2. **`create_host` is a much heavier abstraction than this flow needs.** It starts sshd and waits for SSH readiness, creates/mounts a persistent host Volume, writes a host-record JSON, sets tags, and configures an activity watcher / idle-shutdown. The snapshot script just wants a few `sandbox.exec(...)` calls.

3. **The snapshot output shape mismatches.** The script calls `sandbox.snapshot_filesystem()` and needs the **raw Modal image id** for offload's `--override-image-id`. mngr's snapshot path wraps this in `SnapshotRecord` bookkeeping tied to a host record and doesn't return a bare image id in that form.

The script already reuses mngr domain code wherever it cleanly can (`start-dockerd.sh`, `e2e_workspace_runner`, `materialize_paired_fct_worktree`, `ensure_minds_env_defaults`). What remains in raw `modal` is exactly the infrastructure layer that mngr's host abstraction is a poor fit for.

PR #2380, opened 2026-07-07
Design spec: host lock safety across SSH reconnects

qi-imbue:
Design doc only (no code changes). Follow-up to #2350.

## Why

#2350 fixed the *acquisition* side of the remote cooperative host lock (retry transient SSH failures, wrap them as `HostConnectionError`). This spec addresses the *hold* side: a pre-existing, silent mutual-exclusion violation.

The remote lock is a `flock(2)` held by a remote shell over one SSH channel, so **lock ownership is bound to SSH connection liveness, and nothing re-verifies ownership once the lock is held.** If the connection drops mid-critical-section, the lock silently releases and the per-operation transient-retry machinery silently reconnects and keeps running — now unlocked. Another actor can enter its critical section in the gap. No exception is raised. (This was raised by a reviewer on the #2350 Slack thread.)

## What the spec proposes

- A **monotonic acquisition counter** on the host (incremented under the flock, one per acquire), plus a **re-acquire-and-verify** step on reconnect: if no other actor acquired during the gap (`V == token + 1`), the critical section transparently continues; otherwise raise a new `LockLostError` (a `MngrError`, already isolated per-host by the `mngr_mapreduce` launch loop).
- This is deliberately *less strict* than "never reconnect while locked" — in the TMR environment (each host provisioned by one `create`, so contention is ~nil) almost every blip is uncontended and should recover, not fail the run.
- **Enforcement points** with a completeness argument: the paramiko reconnect chokepoint (`_ensure_connected`, fail-fast); rsync pre-checks (separate connection); lock-channel-death with a live transport; and a single-exit **release backstop** that makes completeness a guarantee rather than a convention.
- **Refinement:** while a lock is held, never `disconnect()` a still-active transport (pyinfra's read timeout currently force-disconnects a possibly-live transport, needlessly dropping the lock) — retry on the same transport and only reconnect-and-re-acquire on a confirmed-dead one.
- Failure-mode table, known assumptions (notably "all lock acquirers must increment the counter"), alternatives considered, testing plan, and implementation sequencing.

## Files

- `specs/host_lock_reconnect_safety.md`
- `dev/changelog/mngr-host-lock-reconnect-safety-spec.md`

Not implemented yet — this PR is for design review. Implementation follows the sequencing section once the approach is agreed.


specs/host_lock_reconnect_safety.md
```diff
@@ -0,0 +1,395 @@
...
+  lock command (and thus the counter). Remote-host operations all go through
+  `_hold_remote_host_lock`, so this holds. The in-host idle-shutdown watcher only
+  *probes* the flock (non-blocking) and does not hold it for mutation, so it does
+  not need to count. **Implementation must confirm** no other in-host actor holds
+  `host_lock` for mutation without incrementing; if one exists, it must be updated
+  to increment, or documented as a gap.
+- **Idle-shutdown during the gap is out of scope.** When the connection drops the
+  flock releases, so the in-host idle watcher could shut the host down mid-create
+  if its idle timeout elapses. This risk is inherent to connection-bound locking
+  and exists regardless of this design; it is not addressed here.
```
joshalbrecht, line 340: idk -- couldn't the idle shutdown code simply acquire the lock in roughly the same (incrementing) way, just via the host itself?

joshalbrecht:
I agree with the PR description, and just had the one comment above

Assuming this doesn't end up being too much complexity, I guess worth doing this for correctness

specs/host_lock_reconnect_safety.md
qi-imbue: I suppose we can, but it's not going to make much difference? The incrementation done by the shutdown script will only be observed if an actor happens to re-connect successfully between the script starts running and the host actually shuts down.

weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#8 from the same branch. This public repository is now a read-only mirror.

PR #2382, opened 2026-07-07
Onboarding fixes: Lima host-key mismatch + macOS desktop-app launch

qi-imbue:
Two independent fixes found while onboarding to minds against a personal dev env.

## 1. Lima workspace creation failed with a host-key mismatch

Creating a local Lima workspace intermittently failed with:

```
Error: Failed to connect to host: SSH host key error (Host key for 127.0.0.1 does not match.)
```

The Lima provisioning script installed the pre-trusted sshd host key **after** a network-dependent `apt-get install`, all under `set -e`. A transient Debian mirror failure made `apt-get` return non-zero, which aborted the script **before** the host-key swap ran -- so the VM kept its default host key while mngr had already pinned the key it expected (`StrictHostKeyChecking=yes`), and the connection was rejected.

**Behavior change:** host-key trust is now established first, before any package fetch, so it no longer depends on the network. The package install is retried to ride out transient apt mirror blips; if it still fails, the error surfaces clearly on an SSH-reachable host instead of as a cryptic host-key mismatch.

## 2. macOS desktop app launched with no Dock icon and no keyboard input

`just minds-start` ran `electron .`, which execs the Electron binary directly. On recent macOS that process is not registered as a foreground GUI app: the window draws but has no Dock tile and can never become the key window, so it takes no keyboard input.

**Behavior change:** on macOS the desktop client now launches through LaunchServices (`open`) via a small supervisor (`apps/minds/scripts/launch_electron.js`), so it comes up as a normal foreground app with a Dock icon and keyboard focus. The activated environment is forwarded explicitly (`open` otherwise drops it). The supervisor forwards SIGTERM/SIGINT to the (LaunchServices-detached) app and stays alive for its lifetime, so `Ctrl-C` / `just minds-stop` still stop it cleanly and the CSS watcher is unaffected. Non-macOS platforms are unchanged (`electron .`).

`just minds-stop` is also fixed and hardened: it depended on `pstree` (absent on stock macOS), which aborted the recipe under `set -e` so it never worked; it now walks the process tree with `ps` and additionally stops the LaunchServices-launched (launchd-reparented) app, force-killing specific survivors after a grace period.

## Manual testing

- **Lima create (happy path):** on macOS, create a Lima workspace from the desktop client (or `mngr create ... --provider lima`) and confirm it reaches a running, SSH-reachable agent.
- **Lima create (transient apt failure):** hard to force deterministically; confirm that if the guest's `apt-get` fails, the failure now shows as a clear provisioning error rather than a `Host key for 127.0.0.1 does not match` mismatch.
- **macOS launch:** `eval "$(uv run minds env activate <env>)"` then `just minds-start`; confirm the app shows a Dock icon, becomes frontmost, and accepts keyboard input.
- **macOS live CSS:** while the app is running, confirm the CSS watcher is still active (edit a style / confirm `tailwindcss --watch` is alive).
- **Stop paths:** `just minds-stop` fully stops the app + Python backend + recipe tree; `Ctrl-C` on a foreground `just minds-start` also stops the app; a second `just minds-start` is not blocked afterward.
- **Non-macOS (Linux):** `just minds-start` still launches via `electron .` and the window behaves normally.


qi-imbue:
Splitting this up. The Lima host-key fix now lives in its own PR against latest main: #2424. The macOS desktop-launch change is being kept as a local-only workaround (other macOS users on the team can't reproduce the underlying macOS 26 issue), so it won't be merged. Closing this combined PR.

PR #2383, opened 2026-07-07
mngr_latchkey: raise gateway --max-body-size for git smart-HTTP pushes

pseay-imbue:
## What

Both `latchkey gateway` spawn sites — the shared desktop gateway (`core.py`) and the VPS-resident gateway (`remote_gateway.py`) — now pass `--max-body-size` raised to **512 MiB** (upstream default: 10 MiB), via a shared `GATEWAY_MAX_BODY_SIZE_BYTES` constant.

## Why

The latchkey gateway natively proxies GitHub's **git smart-HTTP** endpoints (`/gateway/https://github.com/<owner>/<repo>.git/...`, gated by the `github-git` scope with `github-git-read` / `github-git-write` — upstream support since latchkey 2.14.0; the `--max-body-size` flag exists in all versions we pin/bundle: 2.19.1 on the VPS, `^2.19.1` in minds). That lets agents run `git push` / `git fetch` through latchkey with the credential injected server-side — no GitHub token in the container.

But the gateway buffers each request body in memory with a 10 MiB cap, and a push packfile scales with repo history — a minds template full-history push is ~30 MiB today — so gateway-authenticated pushes failed. 512 MiB caps transient per-request memory and costs nothing until a request is actually that large.

The forever-claude-template's inspirations publish flow ([FCT PR #243](https://github.com/imbue-ai/forever-claude-template/pull/243)) now pushes through this path instead of `GH_TOKEN`; this PR also updates the `blueprint/minds-inspirations/` planning docs, which previously asserted a git push was "not an HTTP call latchkey can inject into".

## Testing

- `just test-quick libs/mngr_latchkey` — 395 passed, 0 failed.
- Verified `--max-body-size` is accepted by latchkey 2.14.0 / 2.18.0 / 2.19.1 (unpacked the npm tarballs and checked the CLI option), and that the fake latchkey test binary matches on `argv[1] == "gateway"` so the added flags don't affect it.
- Functional git-over-gateway shape verified separately with `git ls-remote` against a stub server (git emits `GET /gateway/https://github.com/...` with both extra headers intact).

## Caveat

Already-running gateways keep the old 10 MiB cap until they restart: the desktop gateway is respawned with each minds-app session, and a VPS gateway (nohup-detached) keeps running until its recorded process exits — the pidfile guard deliberately never kills a live gateway, so old VPS gateways pick the flag up only on their next launch.


joshalbrecht:
lgtm

PR #2384, opened 2026-07-07
kanpan: attach, peek, and reply from the board; footer, ? help, terminal title

weishi-imbue:
Lets you work with an agent without leaving the kanpan board: enter its session and come back, or peek at its conversation and reply in place. Board grouping/columns and mouse handling are unchanged.

## What's new

- **Attach** (`Enter`): suspends the board's `MainLoop` and runs `mngr connect <agent>` as a child process (connect `execvpe`s, so it cannot run in-process without tearing down kanpan). On detach (`Ctrl-b d`) the board repaints immediately instead of leaving the connect output on screen. The screen clears to a brief `Connecting to <agent>...` line on the way in, `TMUX` is stripped from the child env so attach works when kanpan itself runs inside tmux, and a failed connect shows its exit code as a footer message instead of vanishing with the repaint.
- **Peek** (`Space`): a bordered panel below the board shows the focused agent's recent user/assistant conversation, verbatim (via `mngr transcript --role user --role assistant`), refreshed every ~2s, with the board still visible above. Tool calls and framework-injected turns don't appear. Each `[timestamp] role:` header is trimmed to a dim role cue; the panel tails the newest lines, so a long final message shows its end under a `⋯` marker. The border title (`name · state · provider`) re-renders on every board refresh, so it can't go stale while the panel is open. `Esc` closes; to peek a different agent, close the panel, move the board selection, and press `Space` again. Transcript reads run on their own executor, so a slow read can't freeze the board.
- **Reply**: type into the panel's `›` input and press `Enter` to send via `mngr message` (an empty reply does nothing). The send runs in the background on a single-worker executor, so several replies typed in a row are delivered in order without interleaving. The typed text is echoed into the panel immediately as a `›` line and replaced by the real message once the agent accepts it and it appears in the transcript; a failed send drops its echo and renders a dim `(reply failed: ...)` line instead of silently looking delivered. The input supports readline-style editing (via `urwid-readline`, new dependency): word movement/delete, `Ctrl-A`/`Ctrl-E`, `Ctrl-U`/`Ctrl-K`. The `enter: send · esc: close` hint sits dim in the panel's bottom right.
- **Footer + `?` help**: the footer is one continuous blue belt -- on the left a relative, self-aging refresh stamp (`Refreshed just now · 3.2s` while fresh, aging to `Refreshed 5m ago`; the fetch duration only shows while fresh), on the right the command keys one forgets (`r: refresh  m: mute  d: mark delete  x: execute  q: quit  ?: more keys`, names from the live command map so overrides rename them) with highlighted keys. Everything else (`space` peek, `enter` attach, marks, configured shell commands) is listed by the `?` overlay: a bordered, padded panel anchored above the footer's `?`, keys accented in their own column; `Esc`, `q`, or `?` closes it.
- **Terminal title**: kanpan sets the title (`kanpan`) while running, re-takes it after you detach from an attached session (which sets its own via mngr's tmux `set-titles`), and restores the previous title on exit in terminals with title-stack support.
- **Fix**: the row focus highlight is now one continuous band across muted and stale cells; inverting their dim gray used to punch dark holes in it.

## Notes

- `mngr transcript` previously rendered assistant turns from old-emitter agents as `(no content)`; that fix landed separately in #2392 (MIND-104) and this PR relies on it.
- `mngr message` waits up to ~90s for the agent's submission signal. A busy agent accepts the reply when its current turn ends -- the optimistic echo covers the gap. TUI-local slash commands that open a selector (e.g. `/login`) are pasted and do open the menu, but never fire the submission signal, so the send times out; attach to drive menus.
- Markdown rendering in the peek was implemented (an accent styler, then a full `markdown-it-py` token renderer) and deliberately withdrawn: links cannot be made clickable inside wrapped urwid text -- which was the value that justified transforming -- and a transcript viewer should show exactly what was said rather than restyle it on the guess that it is markdown. Both implementations live in this branch's history if real clickability or a per-harness transcript-format signal ever lands.

## Verification

Manually verified against real local agents via tmux `send-keys`/`capture-pane` (interactive TUI behavior, so not crystallized into pytest per repo convention): attach/detach returns to a repainted board; peek renders the conversation with dim role cues and a live title; replies echo instantly and reconcile once the transcript catches up; queued replies deliver in order; the footer stays on one line and the `?` overlay opens/closes at 90-column width; the terminal title reads `kanpan` (tmux `pane_title`) and the refresh stamp ages across tick boundaries; a focused muted row renders as a single reverse-video run.

Unit tests cover the non-interactive logic (transcript-to-panel rendering, header trimming, pending-reply echo/pruning and failure surfacing, panel lifecycle, legend/overlay binding assembly including user-configured commands, refresh-stamp formatting, config plumbing). 473 kanpan tests pass (tui + ratchets).

https://claude.ai/code/session_01TiyGwn7fTmuP6FStFAz9Ru


qi-imbue (approved):

qi-imbue:
A few minor issues - feel free to defer to followups!

In the peek panel:

截屏2026-07-15 12 50 05

- Can we render the transcript's Markdown instead of showing it raw? This probably depends on the agent type, but at least for Claude we know for sure that it's Markdown. Maybe this can be a property that each agent type implementation can declare.

- The legends for keys should use a consistent style as those of the main board.

On the main board, the legends for keys now consistently takes up two lines for me: it looks a bit ugly and wraps at the wrong place (in my example it separates `enter:` from `attach`). We could use a style refresh here, and we can probably make the key and the description visually more distinct. (You can ask Claude to come up with several different styles and pick one.)

截屏2026-07-15 12 52 27

Finally, we should probably set the terminal title in Kanpan - as you can see my terminal title is stuck on what was set by my attaching to `fix-host-locking`. There's a terminal escape sequence for that and I'm sure Claude knows how to do it.

PR #2392, opened 2026-07-08
mngr transcript: render assistant text from flat fields when parts[] is absent (MIND-104)

weishi-imbue:
Carves the `mngr transcript` reader fix out of #2384 (kanpan agent-view) into its own standalone PR off `main`, so it is tracked and reviewed independently of the kanpan UX. Tracked by [MIND-104](https://linear.app/imbue/issue/MIND-104). #2384 keeps the kanpan work.

## What's here

**`mngr` (`transcript.py`)** — `mngr transcript` (human format) rendered every assistant message as `(no content)` when the event has no ordered `parts[]` list. The current common-transcript emitter always fills `parts[]` (since the 2026-06-15 OTel alignment), but agents on an *older* emitter version emit only the flat `text`/`tool_calls`, so their assistant turns -- including ones with real text -- collapsed to `(no content)` (the text was present, visible via `--format jsonl`). The formatter now falls back to the flat `text`/`tool_calls` when `parts[]` is empty.

This is a **back-compat shim** for old-emitter agents, not a fix for a broken emitter. The gap is keyed on emitter version, not line age, and is not self-healing (`convert()` dedups by `event_id` and only appends). It is scoped for removal once pre-`parts[]` agents age out: [MIND-113](https://linear.app/imbue/issue/MIND-113).

**`specs/common-transcript-standard/spec.md`** — corrected the Compatibility section, which wrongly claimed transcripts are "continuously re-derived from the raw stream" (they are not), and marked the spec implemented.

## Not here

The companion `mngr_claude` change that reclassified slash-command scaffolding (`<command-name>`, `<local-command-stdout>`) as tool results was **dropped**: there is no structured signal for scaffolding (verified — `origin`/`promptSource`/`isMeta`/`isVisibleInTranscriptOnly` are all null on such turns), so it can only be string-matched, and reclassifying is an opinionated, dedup-frozen transform for a purely cosmetic goal. Left as faithful `user_message`s.

## Verification

Unit tests on this branch: `libs/mngr/imbue/mngr/cli/transcript_test.py` — 38 passed (new cases: flat-text renders, tool_calls render, truly-empty still `(no content)`). CI green.


joshalbrecht:
sure, minor but acceptable

lgtm

PR #2395, opened 2026-07-09
Fix stale Docker SSH ports after daemon restart by reconciling with live mapping

pseay-imbue:
## Problem

Docker does not preserve randomly-published SSH host ports across a daemon restart (a reboot can even reshuffle ports between containers), but the docker provider resolved the port once at create time, persisted it in the host record, and trusted it on every later connection. After a reboot, every connection dialed the dead port ("Unable to connect to port <old> on 127.0.0.1"), the host showed STOPPED/UNAUTHENTICATED, discovery fell back to offline agent enumeration, and the minds UI hung on "Loading workspace..." forever. Real-world reproduction: after a host reboot, workspace "care3" had recorded port 52918 vs live 49315; patching the record was the only thing needed to fully restore it.

## Fix

The recorded ssh_port is now treated as a hint. `_create_host_from_container` -- shared by connection setup (`get_host`), discovery (`discover_hosts`, so `mngr list` heals records passively), and `start_host` on a running container -- reconciles it against the container's live port mapping via the new `_reconcile_recorded_ssh_port` (with a reload-free `_read_container_ssh_port` reader split out of the existing port resolver):

- When the live port differs, the host record is rewritten with it and one INFO line (old -> new) is logged. If the live mapping cannot be read (e.g. the container stopped in the window since the caller's status check), reconciliation is best-effort: it warns and keeps the recorded port rather than breaking paths that previously worked.
- The existing known_hosts update now runs with the healed port, replacing a stale same-key-type entry that a *different* container may have left on that port before the reboot, so host-key verification succeeds.
- The cached-Host reuse check compares against the healed port, so a cached connector still holding the old port fails the match and is replaced (and disconnected). All three callers install the reconciled Host in the cache -- `start_host`'s already-running branch previously skipped this, which would have let a stale dead-port Host keep being served from the cache by later `get_host` calls.

The live port is read from the container attributes the callers just reloaded, so the happy path adds no extra Docker API call.

To make the hint-not-truth semantics explicit, `HostRecord.ssh_port` is renamed to `last_discovered_ssh_port` in code. The persisted JSON key stays `ssh_port` (via validation/serialization aliases): host records live on the shared state volume and are read by other mngr clients, including older versions, so the on-disk format is unchanged and old records load as-is. Two host_store tests pin the stable wire format in both directions.

## Tests

- Unit (`instance_test.py`, fake container + LocalVolume-backed store, 5 tests): stale record heals to live port and record is rewritten; matching port causes no rewrite, no log, and reuses the cached Host; stale known_hosts key on the new port is replaced with the recorded key; a cached Host holding the old port is invalidated; an unreadable live mapping falls back to the recorded port without rewriting the record.
- Integration (`test_docker_lifecycle.py`, real daemon): create a host, rewrite its record with a dead port, drop caches, and verify `get_host` reconnects and heals the record (passed locally in ~40s, with the heal visible in the INFO log).
- Full local suite: `just test-quick "libs/mngr -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'"` -- 4673 passed, 0 failed. `ruff` and `ty` clean.
- Also marked `test_check_deps_scope_and_install_are_independent` flaky with a bumped timeout in a separate commit (unrelated intermittent global-timeout trip under xdist load, observed on a full local run; matches the `test_destroy_single_agent` precedent).

## Verification note

Manual Docker Desktop restart verification was intentionally skipped (instructed not to touch the daemon); the integration test simulates the identical post-reboot state (running container, record pointing at a dead port). Known residual: `get_connector` still reads the raw recorded port -- it has zero production call sites and benefits from records healed by discovery, so it was left untouched to avoid extra docker API calls.


joshalbrecht:
lgtm once tests pass

PR #2396, opened 2026-07-09
Fix imbue_cloud husk: persist and re-attach sticky agent labels when a leased host is transiently unreachable

gnguralnick:
## Summary

A leased imbue_cloud minds workspace could vanish from the sidebar (and 404 on restart) during a transient unreachable window — a Mac sleep/wifi blip or a brief box outage. Root cause: when discovery's outer SSH to a leased host fails (or a successful outer-listing returns zero agents), the provider synthesized a single **label-less "husk" agent** named by the lease id. Every consumer that filters on agent labels — most importantly the minds forward's `--agent-include has(agent.labels.is_primary)` and the sidebar/restart `is_primary` guard — silently dropped the workspace.

This is the sibling failure mode to #2381: that PR made a *successful* streaming pass carry labels; this one makes the *fallback* paths carry them too, so the workspace survives the unreachable window instead of collapsing to a husk.

## The fix (contained in `libs/mngr_imbue_cloud`, `providers/instance.py`)

- **Persist last-known identity.** After each successful outer-listing pass that finds real agents, the provider writes `agent_id -> {name, certified_data}` to `hosts/<host_id>/last_known_agents.json` via `atomic_write` (best-effort; an `OSError` is logged, never fails discovery). Persisting to disk — not just in-memory — is what makes the identity survive an app/forward relaunch into a flaky-network window, which is the observed production failure mode.
- **Re-attach in both fallback paths.** When outer SSH is unreachable (CRASHED / UNAUTHENTICATED) or a successful pass lists zero agents (stopped container / empty `data.json`), discovery now re-attaches the **full** cached agent set — each carrying its cached `certified_data` plus a synthetic `"stale": true` marker so consumers can tell cached identity from a live listing — instead of a single lease stub. This also preserves the system-services agent (and any other non-primary agents) that minds tracks per host. The bare lease stub is kept only when no cache exists yet (first-ever discovery), so there is no behavior change until one good pass has been observed.
- **Liveness unchanged.** The host still surfaces as CRASHED / UNAUTHENTICATED with its real `failure_reason`; only the agents' identity (names + labels + `data.json` fields) is restored, marked stale. No consumer treats `certified_data` presence as reachability, so a restart attempted during a genuine outage now fails at the stop/start step with a real, retryable error instead of "Unknown workspace".
- The rich `get_host_and_agent_details` path needs no change: `_build_offline_details_from_lease` already builds `AgentDetails` from the passed refs, so cached identity (including `is_primary`) flows through automatically.
- Destroy/delete cleanup is already covered by the existing `_cleanup_local_host_state` rm-tree of the per-host state dir, so the cache is removed with the rest of the host state (and host_ids are not reused).

## Tests

- Unit tests (`providers/instance_test.py`): a successful pass then an unreachable pass carries the prior labels and the full agent set, each marked stale (and live refs are not stale); persistence across a fresh provider instance (the relaunch case); UNAUTHENTICATED behaves like CRASHED; the empty-agents successful-pass site; and no-cache first-discovery is unchanged (bare lease stub).
- Integration-level round-trip (`providers/test_sticky_agent_labels.py`): real on-disk persist → unreachable fallback → `get_host_and_agent_details`, asserting the cached name and `is_primary` label reach the final `AgentDetails`.

## Changelog

Per-project entry for `mngr_imbue_cloud`.


joshalbrecht:
lgtm!

PR #2399, opened 2026-07-09
Minds discovery logging cleanup

joshalbrecht:
Cleans up discovery logging and provider treatment for the Minds app, per `specs/discovery-log-cleanup.md` (included in this PR).

- **Consumer-side error-log suppression**: new shared `DiscoveryErrorLogSuppressor` in libs/mngr, wired into the three stream consumers (minds `forward_cli`, `mngr_forward` `stream_manager`, `mngr_latchkey` `discovery_stream`). A provider stuck on the same failure (e.g. missing credentials) logs one warning per process with a note that repeats are suppressed; a different error logs immediately; a clean snapshot from the provider logs an info-level recovery line and re-arms suppression. Host/agent-attributed errors always log. This eliminates the thousands of repeated "Discovery error from ..." warnings per session in minds.log.
- **Startup snapshots for skipped providers**: `mngr observe --discovery-only` now emits one `DISCOVERY_PROVIDER` snapshot per provider skipped at startup (error snapshot for unavailable/unauthorized, clean zero-agent snapshot for known-empty), carrying the provider's config, so the providers panel sees their state from the stream. The full `mngr observe` inherits this via its `--discovery-only` subprocess.
- **AWS region blocks always written**: minds' bootstrap no longer gates the four `[providers.aws-<region>]` blocks on plausible AWS credentials -- a credential-less region errors visibly in the panel like vultr/ovh/gcp. The rewrite preserves a panel-toggled `is_enabled`.
- **Observe bounce on settings change**: `_ensure_mngr_settings` reports whether it modified the settings file, and the signin path bounces `mngr observe` when the provider set changed for any reason.

Verified manually: a real `mngr observe --discovery-only` run against a scratch host dir emits exactly one startup error snapshot per unauthorized provider (azure/gcp/aws/ovh/vultr) with quiet stderr across poll cycles, and the suppressor driven with real emitted error events logs once per provider, recovers, and re-arms.

Full offload suite: 15,247 passed, 0 failed.


joshalbrecht:
roughly lgtm

PR #2400, opened 2026-07-09
Parallel pool-host destroy with atomic lease-race claim

joshalbrecht:
## Summary

- `mngr imbue_cloud admin pool destroy` (and the `minds pool destroy` wrapper / `just destroy-pool-hosts` recipe) now accepts multiple pool-host ids and destroys their slice VMs concurrently (`--max-concurrency`, default 8), regardless of which bare-metal box each slice is on — only the carve-time reservation needs the box lock, so parallel deletes are safe even within one box.
- The destroy-vs-lease race is closed with an atomic claim: each row is flipped to `status='removing'` (from an eligible status set) in a committed transaction before any teardown. The connector's `/hosts/lease` only selects `available` rows, so a claimed row can never be handed to a user; a row that got leased first is skipped and reported.
- `--force` is re-scoped to mean "also destroy leased rows" (nothing writes `released` anymore, so the old guard forced `--force` on every real destroy). `--skip-vps-cancel` is replaced by `--drop-row-only` (row-only drop for rows whose box record is gone or machine is dead; the default path already tolerates a merely-absent VM).
- The command keeps going through all ids on failure and emits a per-host outcome report (`{requested, destroyed, skipped, failed, hosts}`); exit is non-zero only when a teardown actually failed. Re-running the same ids converges (`already_gone` counts as success; rows left `removing` by a failed teardown are re-claimed and retried).
- `teardown-slices` (the `minds env destroy` path) claims rows with the same status set, includes rows stranded in `removing` by a crashed release, and reuses the same parallel helper. DB-side slot accounting now counts `removing` rows as occupied.
- Docs: `host-pool-setup.md` Cleanup section documents the multi-id destroy, retry semantics, and the recommended pool-upgrade sequence. The generated `imbue_cloud` CLI reference is regenerated, and the pre-commit docs hook now watches plugin CLI files.

## Verification

- Full offload suite: 15,249 passed, 0 failed (`just test-offload`).
- Live verification against the `dev-josh-1` pool DB with synthetic rows: parallel multi-id destroy, idempotent re-run (`already_gone`), leased-row skip + `--force`, failed-teardown → `removing` → retry convergence, unexpected-status classification, non-UUID id → per-host `failed` outcome, and a simulated in-flight lease (`SELECT ... FOR UPDATE` held during the destroy) correctly resulting in `skipped_leased`.
- Multi-agent code review ran over the diff; confirmed findings (worker exceptions aborting the batch, stale generated docs, claim-miss misclassification, predicate drift between teardown selection and claim) were fixed in the follow-up commit.


joshalbrecht:
roughly lgtm

joshalbrecht:
lgtm

PR #2403, opened 2026-07-09
Speed up most `mngr` calls in Minds.

hynek-urban:
Some time ago, we introduced `MngrCaller` which tries to speed up `mngr` invocations by always keeping a pre-warmed `mngr` process handy. It was only used in a few places so far.

This PR migrates most of the remaining `mngr` invocations in Minds to `MngrCaller`, too. The main exceptions to this are invocations that are meant to be long-running (`mngr observe`, ...) because MngrCaller currently doesn't stream the output while in progress.

Empiricially, the "cloudflare sharing" use case went from 70s to 46s after this change.


joshalbrecht:
lgtm, some questions on slack

PR #2405, opened 2026-07-09
Improve the comprehensibility of workspace management permission requests.

hynek-urban:
As a user, I realized I was confused by the workspace management permission dialogs.

There are two kinds of relevant permissions. 1) For operations that are not targeted towards any specific workspace (e.g. "list all workspaces" or "create new workspaces") and 2) For operations that are targeted towards specific workspaces (e.g. "start / stop workspace X").

To complicate matters further, operations from the second category can also operate on "wildcards" (= any workspace).

The original dialog didn't distinguish between these two. As a user, when a request came that wasn't targeted at any specific workspace (e.g. "read the list of existing workspaces"), I didn't know what checking e.g. "Permanently destroy the selected workspace" would do.

Newly, the permission dialogs clearly distinguish between "targetless" and "targeted" operations. And when a request without a target comes, there's a warning that makes it clear that the workspace-specific permissions will in fact apply across all workspaces.

The screenshots below show how the dialogs look before and after the changes in this PR.

**Before - operations without a target:**

before_targetless

**After - operations without a target:**

after_targetless

**Before - operations with a specific target:**

before_with_target

**After - operations with a specific target:**

after_with_target


joshalbrecht:
lgtm

PR #2407, opened 2026-07-09
Speed up minds remote first-load: seed mngr forward from a last-known service-map cache

gnguralnick:
## TL;DR

Time-to-first-render of a restored **remote** mind window goes from scaling with the number of *other* open workspaces to roughly constant. The blocking path used to be that window's slow per-agent service-URL lookup — spawned as an unthrottled herd across every `is_primary` agent — so the focused window's latency grew with n. Seeding the route from a last-known on-disk cache moves that lookup off the critical path: the window resolves from a single cheap file read + the fixed-cost discovery-snapshot replay (~3s), independent of n.

Caveat on the framing: this is O(1) on the *critical path to first render*, not O(n) work eliminated. The per-agent streams still spawn and run in the background (total work stays O(n)); they're just no longer blocking. A never-before-cached agent still pays its one-time ~10s cold cost.

## Problem

When a minds desktop window is restored onto a **remote** (imbue_cloud) mind, it sits on the `mngr forward` 503 "Loading workspace" loader for ~50-55s even when the host is healthy, versus ~20s for a local docker mind.

Root cause: `mngr forward` starts with an empty routing table. Membership + SSH info arrive fast (~t+2.5s) from the cached discovery-snapshot replay, but the per-agent service URL only arrives once the slow `mngr event <id> services --follow` stream connects. Measured live (staging/prod, read-only): a cold single stream takes ~10s, and under spawn contention across many `is_primary` agents the tail workspace waits ~50s. Until the URL lands, `resolve()` returns `None` and the proxy serves the loader.

Service ports are effectively fixed (measured: `system_interface` pinned to `localhost:8000` across container restart epochs), so the previous run's service URL is almost always still correct.

## Fix

Persist the resolver's per-agent `{service -> url}` map to a small on-disk cache (`$MNGR_HOST_DIR/plugin/forward/service_map.json`) while the plugin runs, and **seed** the resolver from it at startup. A restored window becomes routable as soon as discovery supplies membership + SSH info, instead of waiting on the event stream. The live stream still runs and overwrites the seed via the existing full-replace `update_services` path, so a stale seed self-corrects within one stream connect.

Located entirely in the `mngr_forward` plugin (kept app-agnostic):

- New `service_map_cache.py` (`ServiceMapCache`): load/persist the map as JSON, reusing the sanctioned `atomic_write` / `read_json_dict` helpers. Both `load()` and `persist()` are best-effort; a corrupt/missing/unreadable/non-UTF-8 file degrades to `{}` (today's behavior).
- `resolver.py`: optional `service_map_cache` field + a `_publish_services_snapshot()` helper that emits the existing `resolver_snapshot` envelope **and** persists the cache, wired into the three service-map mutation points (so removals drop cache entries too). New `seed_services()` seeds the map at startup.
- `cli.py`: construct the cache, pass it to the resolver, and `seed_services(cache.load())` in the observe branch.

### Safety

`resolve()` still gates on **this run's** known-agent set, so a cache entry for an agent this run doesn't rediscover is never served. The one accepted risk (documented in the blueprint): if a cached URL is wrong (same agent, running, but a service genuinely moved ports), the proxy dials a dead port and emits the normal `CONNECT_ERROR` envelope until the live stream corrects (~10s) — near-impossible for the fixed-port services in practice, so no grace window is added.

## Design

Blueprint spec: `blueprint/forward-services-cache/plan-forward-services-cache.md` (root-cause grounding, live measurements, and the alternatives considered — discovery-snapshot seed, priority-agent hint — are captured there).

## Testing

- New `service_map_cache_test.py` (roundtrip, missing/corrupt/non-object/non-UTF-8/malformed-entry coercion, best-effort swallow of read/write errors) + 5 new `resolver_test.py` cases including the end-to-end persist->seed->resolve round-trip and the membership-gating safety property.
- `just test-quick libs/mngr_forward` (excluding tmux/modal/docker/acceptance/release): **258 passed, 0 failed**. `uv run ty check` clean; ruff clean. Ratchets: no new violations.
- Full offload suite runs in CI on this PR.

### Not yet verified

Not exercised against a live cloud mind end-to-end (the runtime path writing/reading the file against a running remote container). The logic is fully unit-covered, but the meaningful final check is a real quit/reopen of a remote mind confirming ~3s resolve + a populated `service_map.json`.


joshalbrecht:
lgtm

(this is needed for when the *app* is restarted, which does restart this)

PR #2408, opened 2026-07-09
mngr observe: event-driven local-agent process-death detection

gnguralnick:
Part 1 of a two-repo change. Companion: imbue-ai/default-workspace-template#250 (switches `system_interface` to consume this). The two land/merge separately.

## Summary
The full `mngr observe` observer now detects a **local** agent's main process dying on its own (OOM, crash, or normal exit) within seconds, instead of only at the next 5-minute snapshot:

- Watches each local agent's main (`claude`) process PID via psutil's event-driven `Process.wait()` (`os.pidfd_open` on Linux, kqueue on macOS). On exit it enqueues the agent's host onto the existing activity re-probe/emit path, so the death is emitted as real state. Remote agents keep their existing detection paths (per-host activity streams + periodic snapshot).
- New `mngr observe --stream-events` mode: echoes each agents-stream event (`AGENT_STATE` / `AGENTS_FULL_STATE` / `AGENT_REMOVED`) to stdout as JSONL, in addition to the event files, so a parent process can consume live state. The stdout write is injected as a sink from the cli layer (the api layer stays free of cli output concerns).
- New `AGENT_REMOVED` event on the agents stream, plus prompt re-probe on agent-level discovery deltas — preserving near-instant create/destroy membership latency.
- `AgentDetails.main_pid`: new optional field, populated for running agents on **every** provider — it is a PID in the host's own PID namespace, extracted from probe data the listings already collect (no extra SSH). The local path gets it from the shared lifecycle probe; the remote batch-listing paths in `mngr_modal` / `mngr_vps` / `mngr_imbue_cloud` extract it from their pre-collected tmux/ps data. Filterable/sortable in `mngr list`.
- `HostDetails.is_local`: new required field, populated at every construction site (core + plugin providers; the preserved-agents path in `mngr_usage` uses the same provider-named-`local` convention as the `--local` filter). The observer PID-watches only local agents, gated on `host.is_local` rather than `main_pid` presence — a remote PID must never be watched in-process, since psutil would bind an unrelated same-numbered local process. Also filterable/sortable (`host.is_local`).
- Lifecycle-probe consolidation: one pure function `determine_lifecycle_probe_result` returns a `LifecycleProbeResult` frozen model (state + `main_pid`), replacing the `(state, pid)` tuples with paired `None`s and the state-only wrapper; agents expose `probe_lifecycle()` plus a `get_lifecycle_state()` convenience; `provider_instance` no longer branches on `host.is_local`.
- psutil floor `>=5.9` -> `>=7.2` (guarantees the event-driven `wait()` path); `uv.lock` refreshed; CLI docs regenerated; changelog entries for every touched project (`mngr`, `mngr_modal`, `mngr_vps`, `mngr_imbue_cloud`, `mngr_usage`, `mngr_kanpan`, `dev`).

## Testing
- Local: full `libs/mngr` suite (minus tmux/modal/docker/acceptance/release marks) — 4706 passed; targeted runs of observe, hosts/common, agents/base_agent (incl. tmux integration proving the real probe returns a live, correctly-named PID), interfaces/provider_instance, api/list, interfaces/data_types, cli/field_catalog — 290 + 738 passed; `mngr_vps`/`mngr_imbue_cloud`/`mngr_modal` — 1203 passed (1 pre-existing macOS-only lima sed failure, untouched by this branch); root meta-ratchets (ruff/ty/CLI-docs) — 19 passed. `ty` clean across `libs`.
- Manually verified `mngr observe --stream-events` emits `AGENTS_FULL_STATE` to stdout; a unit test kills a watched subprocess and asserts the host is enqueued for re-probe; a unit test asserts a remote agent carrying a `main_pid` never opens a watcher (and closes a stale one).
- The full offload suite runs here in CI (not runnable locally without `MODAL_TOKEN_ID`).


libs/mngr/imbue/mngr/hosts/common.py
```diff
@@ -358,34 +380,60 @@ def determine_lifecycle_state(
...
+    is_active: bool,
+    expected_process_name: str,
+    ps_output: str,
+    is_agent_type_known: bool = True,
+) -> tuple[AgentLifecycleState, str | None]:
+    """Determine agent lifecycle state and the main process PID from tmux info and ps output.
+
+    Same logic and inputs as :func:`determine_lifecycle_state`, additionally
+    returning the PID (as a string, matching ps output) of the running
+    ``expected_process_name`` process when the agent is RUNNING or WAITING, else
```
gnguralnick: this implementation seems a bit awkward - having to have lots of Nones everywhere and conditionals and such. i wonder if we can consolidate things a bit better

libs/mngr/imbue/mngr/interfaces/data_types.py
```diff
@@ -611,6 +611,14 @@ class AgentDetails(FrozenModel):
     state: AgentLifecycleState = Field(
         description="Agent lifecycle state (STOPPED/RUNNING/WAITING/REPLACED/RUNNING_UNKNOWN_AGENT_TYPE/DONE/UNKNOWN)"
     )
+    main_pid: int | None = Field(
```
gnguralnick: one option - does this actually need to be None for remote hosts? yes we don't intend to watch the process on the remote ones, but i wonder if we can still just set it to make the rest of the logic simpler?

libs/mngr/docs/commands/primary/list.md
```diff
@@ -126,6 +126,7 @@ Fields marked `(cel only)` cannot be used in `--fields`/`--format` template stri
 - `create_time` - Creation timestamp
 - `start_on_boot` - Whether agent starts on host boot
 - `state` - Agent lifecycle state (STOPPED/RUNNING/WAITING/REPLACED/RUNNING_UNKNOWN_AGENT_TYPE/DONE/UNKNOWN)
+- `main_pid` - PID of the agent's main process (e.g. claude) in its host's PID namespace, populated when the agent is running; None for stopped agents. Only watchable in-process (e.g. via psutil) when the host is local (host.is_local).
```
joshalbrecht: not to be annoying, but this field really should be `pid`, not `main_pid` :(

libs/mngr/docs/commands/primary/list.md
```diff
@@ -143,6 +144,7 @@ Fields marked `(cel only)` cannot be used in `--fields`/`--format` template stri
 - `host.id` - Host ID
 - `host.name` - Host name
 - `host.provider_name` - Provider that owns the host
+- `host.is_local` - Whether the host is the local machine (rather than a remote host)
```
joshalbrecht: I'm skeptical of this... is lima local? docker? or just the local provider?

why do we need this at all?

libs/mngr/imbue/mngr/api/observe.py
```diff
@@ -715,10 +851,158 @@ def _process_snapshot_agents(self, agents: Sequence[AgentDetails]) -> None:
...
+            self._reconcile_watcher_for_agent(agent)
+        for agent_id_str in ids_to_drop:
+            self._close_watcher(agent_id_str)
+
+    # === PID Watchers (local agents only) ===
+
+    def _reconcile_watcher_for_agent(self, agent: AgentDetails) -> None:
+        """Open, replace, or close the PID watcher for one agent from its probed details.
+
+        Only local agents are watched: a remote agent's ``main_pid`` is a PID in
```
joshalbrecht: not for this PR, but... why can't remote agents be watched?

esp if the remote agents are running linux (they all are, in practice), you really ought to be able to SSH in and run a command that waits for the remote process to die (ofc sometimes when you get disconnected you'll end up in "this host cannot be reached" rather than "stopped", but that seems ok...)

libs/mngr/imbue/mngr/primitives.py
```diff
@@ -280,6 +280,20 @@ class AgentLifecycleState(UpperCaseStrEnum):
     UNKNOWN = auto()
 
 
+class LifecycleProbeResult(FrozenModel):
+    """Agent lifecycle state plus the agent's main process PID, from a single probe."""
+
+    state: AgentLifecycleState = Field(description="Agent lifecycle state")
+    main_pid: int | None = Field(
```
joshalbrecht: I mean... everything *has* a PID...  I don't love that this is nullable

joshalbrecht:
lgtm after the above, feel free to merge this and the paired PR

PR #2410, opened 2026-07-09
Serve the workspace UI over HTTP/2 (TLS + h2 at the mngr forward proxy)

gnguralnick:
## Summary

- The workspace UI was served over plain HTTP/1.1 from a single origin (`agent-<id>.localhost:<port>`), which Chromium caps at ~6 held-open connections per origin. Once a workspace had enough long-lived streams (one SSE per open chat tab, plus any `/service/` app's own streams), the pool was exhausted and every further request — including the plain GETs that bootstrap a new terminal/app iframe — queued indefinitely, hanging the UI even though the backend was healthy.
- `mngr forward` gains a `--use-http2` flag (off by default). When set, the proxy generates a fresh in-memory self-signed cert at startup (SANs `localhost`, `*.localhost`, `127.0.0.1`), terminates TLS, and negotiates HTTP/2 via ALPN — served by **hypercorn** (replacing uvicorn), with the pre-bound listen socket handed off by fd (`fd://`, via `os.dup` so hypercorn's shutdown-close doesn't double-close). HTTP/2 multiplexes many streams over one connection, so the per-origin ceiling stops binding.
- When TLS is active, the proxy's client-facing URLs become `https`/`wss` (login URL, `/goto` bridge, unauthenticated-subdomain redirect) and its `mngr_forward_session` cookie is marked `Secure`. The proxy→container hop and the WebSocket transport are unchanged; WS upgrades negotiate `http/1.1` per-connection via ALPN.
- Minds always runs its single proxy with `--use-http2`: the Python readiness/recovery probes and `/goto` redirect URLs are `https` (loopback probes use `verify=False`, since they dial `127.0.0.1` with a `Host: agent-<hex>.localhost` header against a self-signed cert), and the Electron shell unconditionally builds `https`/`wss` URLs, marks the cookie `Secure`, and trusts the loopback cert.
- The Electron shell builds `https`/`wss` workspace URLs, marks the cookie `Secure`, and registers a `setCertificateVerifyProc` that trusts the self-signed cert **only** for loopback hosts (`localhost`, `*.localhost`, `127.0.0.1`) in the `persist:workspace-content` partition — every real https origin still gets Chromium's normal verification. The minds bare-origin backend stays plain HTTP (a deliberate, non-blocked mixed transport).
- `cryptography` is declared explicitly and `uvicorn` dropped from `mngr_forward`. CLI docs regenerated; per-project changelog entries added for `mngr_forward`, `minds`, `mngr`, and `dev`. Two ratchets bumped with documented justification (`asyncio import` 1 to 2 for `asyncio.run(hypercorn.serve(...))`; `__init__ in non-exception classes` 1 to 2 for the `InMemoryTLSConfig` subclass of hypercorn's third-party `Config`).

## Implementation notes

- The hypercorn API was confirmed against its installed source, not assumed: `serve(..., shutdown_trigger=None)` installs SIGINT/SIGTERM handlers (matching uvicorn's graceful behavior minds' `terminate()` relies on), `create_sockets` handles `fd://` by wrapping the fd without re-binding, and `ssl_enabled`/`create_ssl_context` are the hooks the `Config` subclass overrides.
- One deviation from the plan: stdlib `ssl.SSLContext.load_cert_chain` requires a filesystem path, so "never written to disk" isn't achievable — the PEM is written to a private `mkstemp` (0600) file, loaded, and unlinked immediately. It is never persisted.

## Verification

- Local: mngr_forward full suite (255 passed), affected minds tests (~140 passed), all ratchets (114 passed), and the CLI-docs meta-ratchet. Re-verified green after merging latest `origin/main`.
- A throwaway spike proved the transport mechanism end-to-end: ALPN negotiates `h2`, a real HTTP/2 request succeeds with `verify=False`, and the fd handoff does not double-close on shutdown. A second check served the **real** forward app over TLS+h2: `GET /` returns 200 HTTP/2 (login page), `/authenticate` returns 307 HTTP/2 with a `Secure` cookie, clean shutdown.
- **Not yet verified (needs real-app / CI e2e):** WebSocket-over-TLS (`wss`) forwarding in a running workspace, and the reported symptom itself — opening more than 6 streaming tabs and then loading a new terminal/app tab without a hang. These require the full Electron app plus a live backend.
- The full offload acceptance suite runs in CI on this PR.


apps/minds/imbue/minds/cli/run.py
```diff
@@ -455,7 +456,8 @@ def run(
             f"{_MNGR_FORWARD_LISTEN_TIMEOUT_SECONDS:.0f}s; the plugin likely failed to start. "
             "Check the logs above for its stderr and retry."
         )
-    logger.info("  mngr forward: http://127.0.0.1:{}", mngr_forward_port)
+    forward_scheme = "https" if MNGR_FORWARD_USE_HTTP2 else "http"
```
gnguralnick: not sure if we need an env var for this or if it should just be a full update

libs/mngr_forward/imbue/mngr_forward/tls.py
```diff
@@ -0,0 +1,113 @@
+"""Ephemeral in-memory TLS for ``mngr forward`` (self-signed cert + hypercorn config).
+
+Used only when ``--use-http2`` is set: the proxy terminates TLS and negotiates
+HTTP/2 (via ALPN) so the workspace origin is not capped by Chromium's
```
gnguralnick: this docstring ("the workspace origin...") doesn't make sense to have here - mngr forward stands on its own

joshalbrecht:
lgtm, thanks for fixing!

this feels likely to break something in the installed version, will flag to weishi on this one

PR #2411, opened 2026-07-09
Spec: mngr message delivery signals

weishi-imbue:
## Summary

`mngr message` drives Claude through a tmux pane, so it has no ACK and must reconstruct "did it land" from side effects. Almost every side effect Claude Code emits is **semantic** -- it records what the input *meant*, not that input *arrived* -- and the two signals mngr races (the `UserPromptSubmit` hook, the `enqueue` transcript event) are both model-bound. Input that never reaches the model satisfies neither.

That is why a typo, `/login`, and `/cost` each hang for the full 90s despite being delivered, and why `/clear` and `/compact` each needed a bespoke hook.

This PR adds a problem-statement spec. **No production code changes.**

## What's in the spec

- **Overview** of why this resists simplification.
- **Seven input categories**: prompt, local command, modal local command, session-restarting command, invalid command, queued-while-busy, empty.
- **Signal inventory**, split by producer (Claude Code vs mngr), with author and firing time for each.
- **Signal-to-case matrix**, measured against a real local Claude agent (Claude Code 2.1.205, macOS).
- **Twelve issues**, each reproduced unless marked otherwise.
- A **design direction** (proposal, not a decision) and a test plan.

## Headline findings

Two silent GNU-coreutils dependencies break `mngr message` on stock macOS, independent of any slash-command behavior:

- **`timeout`** is absent, so `timeout <n> tmux wait-for` returns 127 and the waiter is never registered. When the `UserPromptSubmit` hook fires `tmux wait-for -S`, the signal wakes nobody. A plain prompt that Claude demonstrably answered was reported as `Failed to send message` after 93.6s.
- **`tac`** is absent (`mngr_transcript_lib.sh:105`), so transcript offset reconciliation fails and its stderr is painted into Claude's input box.

Beyond those:

- `mngr message` can **return before Claude's input buffer is clear**, so the next message is concatenated onto residue (`/clear /zzzrace1`, reproduced 1-in-3).
- Modal local commands (`/cost`, `/model`, `/login`) leave **no input row**; every later send fails after a 30s generic timeout, and `_preflight_send_message` does not detect it.
- Busy and idle **disagree about the same input**: a typo sent to a busy agent returns `exit 0` in 2.6s (it enqueued); sent to an idle agent it fails after 93.7s.
- The Slack thread's premise is wrong -- an invalid command **does** write a structured transcript entry (`type=system, level=warning`), already tailed into `events.jsonl`.
- `~/.claude/history.jsonl` is the only artifact written **before** interpretation, and the only one that fires for every input class. Its one deviation is a consecutive-duplicate dedup.

Also adds a top-level `uncertainties.md` recording two comments/docstrings that assert behavior which does not hold.

## Review focus

The design section is deliberately a proposal. The open questions -- what a sound "settled" predicate is, which dedup guard to use, and whether modal commands should fail or warn -- are the parts most worth arguing about.


specs/mngr-message-delivery-signals.md
```diff
@@ -0,0 +1,300 @@
...
+
+`mngr message` should return on **delivered**, must guarantee **settled** before the next send, and cannot in general observe **executed**.
+
+## How mngr message works today
+
+The call chain is `mngr message` -> `send_message_to_agents` -> `_send_message_to_agent` -> `InteractiveTuiAgent.send_message` (`libs/mngr/imbue/mngr/agents/tui_agent.py`).
+
+1. Reject the agent if its lifecycle state is `STOPPED`.
+2. Acquire an exclusive per-agent file lock.
+3. `_preflight_send_message` -- check the `permissions_waiting` marker and match the pane against `_DIALOG_INDICATORS`. Fail fast on a hit.
```
weishi-imbue, line 43: why can't we extend this to capture those modal commands like /cost and /login, so they can fail fast?

PR #2416, opened 2026-07-10
Fix the macOS coreutils portability bugs in transcript streaming and gc

weishi-imbue:
## What

mngr shells out to GNU-coreutils binaries and flags from code that runs on the **agent's host** — which, under the `local` provider, is the user's machine, where stock macOS has none of them. This PR fixes the two files 2475 did not touch. Each was reproduced on macOS 26.4.1 under a stock `PATH`, then re-verified fixed.

| binary | site | macOS effect |
|---|---|---|
| `tac` | `mngr_transcript_lib.sh` | transcript re-emitted already-emitted lines |
| `du -sb` | `api/gc.py` ×2 | `\| cut -f1` masked the error exit → orphan dir reported as **0 bytes** |
| `stat -c %Y` | `api/gc.py` ×2 | `created_at` fell back to `now()` → **age-based GC never ran on macOS** |

Rule applied throughout: **call the tool; don't reimplement it.** `tac` is deleted, `du -sb` becomes POSIX `du -sk`, `stat -c %Y` reuses the existing portable `get_file_mtime` helper.

## Scope change: the `timeout` fix was dropped

This branch originally also replaced `timeout(1)` with a perl `alarm` fallback in the tmux message-submission path. **#2475 (merged) rewrote that path wholesale** — durable-evidence polling, no `tmux wait-for`, no `timeout(1)`, no background jobs — so that fix is superseded, and the merge from main takes 2475's `tui_utils.py`. The perl technique it established had no remaining caller, so rather than leaving dead code it is preserved as a new "Portable shell in host commands" section in the repo-root `style_guide.md` (the perl form and its two pitfalls, plus the general `du -sk` / `stat -c…||stat -f…` / forward-scan rule these fixes all follow).

## `tac` → forward scan + fork elimination

Forward scan keeping the last match is the same `O(n)`, needs no reverse, and fixes a latent off-by-one (old paired `wc -l`, which drops an unterminated final line, with `tac`, which keeps it).

The real speedup was orthogonal: both scanners pulled each field through `value=$(…)`, **forking a subshell per line**; they now match inline with `[[ =~ ]]` (a bash builtin — verified on macOS bash 3.2 and Linux 5.2 — so no external binary at all). `build_id_set`, which must read the whole output, at 50k lines:

| | old | new |
|---|---|---|
| macOS | 31.7s | 0.7s |
| Linux | 13.6s | 0.6s |

~45×/~22× (3 trials each). This is where essentially all the win is, and it needs no early exit. `reconcile_offset` gave up `tac`'s early exit (now a constant ~0.7s@50k), but since it no longer forks, its old worst case — a match near the start, ~12s reverse-fork — is now that same bounded ~0.7s.

## `gc.py` → the host interface

`gc.py` now calls `get_file_mtime` (already portable) and a new `get_directory_size` — **one `du -sk` call, no `is_local` branch** (`execute_idempotent_command` already dispatches local/remote). The `test -d P/ && du -sk P/` invocation was verified on macOS BSD, Linux GNU, and busybox `du` to agree on hardlinks (counted once), nested symlinks (not followed), top-level symlink (resolved via the slash), and non-dirs (0). `du`'s non-zero exit after an unreadable subdir is tolerated (it still prints the total). `get_directory_size` had zero prior tests; added a real-local-host tree (hardlink + nested symlink) and the non-dir case.

## Tests

- `tac` shim that exits 127 on `PATH`, so a reintroduced dep fails on Linux CI too.
- `get_directory_size` real-local-host tree + non-dir case.
- Local run: `just test-quick` on `mngr_transcript_lib_test.py`, the two `get_directory_size` tests, and `gc_test.py` — **115 passed, 0 failed**. Full suite runs in CI.

## Still to land (deferred)

- Ratchet: POSIX-command allowlist for unknown-host shell, plus a denylist for GNU-only flags on POSIX-named commands. The new style-guide section is what it would enforce.
- macOS CI job (local-agent subset). No job runs mngr's suite on macOS today — the root cause of this whole class, and 2475 had to hand-fix `stat` portability in its new file for the same reason.


libs/mngr/imbue/mngr/agents/tui_utils.py
```diff
@@ -301,19 +299,32 @@ def _send_enter_and_wait_for_signal(
...
+    the tmux client itself, so the kill reaps it. A killed ``tmux wait-for`` exits
+    0, so the watchdog records the deadline in ``$tmo`` before killing, and the
+    wait status then separates a signal from an outright tmux failure. Background
+    jobs redirect stdout, which they would otherwise hold open until they exit,
+    stalling the caller for the full deadline.
     """
-    return (
-        f"bash -c '"
-        f'( sleep 0.1 && tmux send-keys -t "$1" Enter ) & '
-        f'timeout {full_timeout} tmux wait-for "$0"'
```
joshalbrecht: idk, maybe we could use this alternative suggested by claude instead?

Perl one-liner (no install, works everywhere)

```sh
perl -e 'alarm shift; exec @ARGV' 10 ./slow-thing
```

This execs the command, so it replaces the perl process — signals and exit codes behave sensibly. It's the closest drop-in.

libs/mngr/imbue/mngr/agents/tui_utils.py
```diff
@@ -343,20 +359,21 @@ def _build_signal_or_marker_command(
...
-        '( sleep 0.1 && tmux send-keys -t "$2" Enter ) & '
+        '( sleep 0.1 && tmux send-keys -t "$2" Enter ) >/dev/null 2>&1 & submit=$!; '
         f'end="$(( $(date +%s) + {int(full_timeout) + 1} ))"; '
+        "hook_pending=1; "
         'while [ "$(date +%s)" -lt "$end" ]; do '
-        'if [ -s "$sig" ]; then exit 0; fi; '
+        'if [ "$hook_pending" -eq 1 ] && ! kill -0 "$waiter" 2>/dev/null; then '
+        'wait "$waiter" 2>/dev/null && [ ! -s "$tmo" ] && exit 0; '
+        "hook_pending=0; fi; "
         f'cur="$({accept_marker_command})"; '
```
joshalbrecht: I'd want a similar change here instead (the above to replace timeout, but only on osx)

these other changes are changes to very delicate code that doesn't seem worth messing with right now

joshalbrecht:
lgtm *except* the timeout changes -- those are too complex for me to verify, and I'm very skeptical that they are still correct

and it seems like we should be able to easily just replace with the other alternative if timeout is unavailable?

PS: did you make issues for the "Filed separately (pre-existing, not macOS-specific)" issues?

weishi-imbue:
[Agent] 
**Reverted the delicate code.** Both builders are back to their original structure; only the deadline changed. The watchdog, the `$tmo` marker, the three tracked pids and the mktemp guard are gone.

And your skepticism was better founded than I realized. The original marker variant wrote its sentinel **on success** (`timeout … && echo 1 > "$sig"`), so *every* way of failing to confirm — deadline, tmux failure, even a failed `mktemp` — landed on the same non-zero exit. My rewrite **inverted** that into a deadline marker, where an *absent* marker meant success. That inversion is what made the unguarded `mktemp` dangerous and the killed waiter's exit status ambiguous. **Both "bugs" I claimed to fix were regressions I had introduced myself** — I checked the original against a real tmux and it was already correct on a tmux that fails outright (rc=1 passes through) and on a failed `mktemp` (falls through to failure). Removed those claims from the description.

**Took the perl suggestion**, with two fixes — the pasted one-liner has real bugs:

| edge | measured | guard |
|---|---|---|
| the builtin `alarm` takes whole seconds, and **`alarm(0)` *cancels* the timer** | `alarm 0.5` + `sleep 3` → **rc=0 after 3s** — no deadline at all, reported as success | `-MTime::HiRes=alarm` (core perl, no install) |
| a failed `exec` **returns to perl**, which runs off the end of the program | `exec` a missing binary → **rc=0** (a missing `tmux` would read as a successful send) | `or exit 127` |

So the final form is:

```sh
perl -MTime::HiRes=alarm -e 'alarm shift; exec @ARGV or exit 127' <secs> <cmd...>
```

Worth recording *why* this works so much better than what I built: **`tmux wait-for` catches SIGTERM and exits 0 from it, but does not catch SIGALRM.** My watchdog killed with `kill` (SIGTERM), which is the *entire* reason it needed a marker file to tell a timeout from a signal. `alarm` sidesteps that completely, and since `exec` means the command *is* the process, there's no supervisor to leak either.

One deviation from your comment: I probe with `command -v timeout` rather than testing for macOS. A host can have either binary without the other — macOS has perl but no `timeout`; a minimal Linux image can have busybox `timeout` but no perl — and the probe is simpler than a `uname` branch. Same behavior on macOS as you asked for.

Verified against a real tmux server on a PATH with no `timeout`, both builders: hook fires → 0, deadline → non-zero, no tmux server → non-zero, zero orphaned `wait-for` clients.

**Issues:** no, I hadn't — filed now as #2477 (wait-for signals latch), #2478 (a client whose server dies mid-wait exits 0), #2479 (marker variant doesn't fail fast). All three are pre-existing and present on Linux.

PR #2415, opened 2026-07-10
Add the mngr_donate plugin (spend spare Claude capacity on a donation skill)

jean-imbue:
## What

Adds the standalone **`imbue-mngr-donate`** plugin — `mngr donate`, which spends spare Claude capacity on a donation skill instead of letting idle quota expire.

> **Stacked on #2414** (the `document-review` skill). Base is `donate-skill`; review that one first. The diff here is just the plugin + wiring.

One tick: read the account-level usage snapshot (from `imbue-mngr-usage`, a hard dependency), and *if* there's spare capacity (5h window under budget **and** the week under its pace line), launch a headless Claude agent that runs the donation skill to completion, then auto-cleans up. `--start`/`--stop` install/remove a **launchd LaunchAgent** (macOS) so idle quota drains on a schedule.

## Why a separate plugin

Measuring usage (`mngr usage`) and donating spare quota (`mngr donate`) are orthogonal. This plugin depends on `imbue-mngr-usage` (installing donate pulls usage in), and it's registered in the plugin catalog as a default-installable `DEPENDENT` gated on `imbue-mngr-usage`. `mngr_usage` itself is unchanged.

## Notes for review

- **Skill split (pinned code, dynamic prompts).** The skill's *code* is pinned/reviewed in `.claude/skills/<skill>/` (imbue owns what executes — the agent runs with `--dangerously-skip-permissions`). The *prompts* are pulled fresh each run from the skill's upstream git repo (`--skill-repo`, default the lab's GitLab). `prepare_skill_dir()` assembles a working copy under `<host_dir>/donate-skills/<skill>/` (pinned code + pulled prompts) and points the agent at it. So a reworded rubric flows through automatically; a client/protocol change is a deliberate, reviewed bump of the pinned code.
- **launchd, not cron.** A cron job runs outside the login session and can't reach the macOS keychain where Claude's subscription token lives (every tick fails `Not logged in`); a LaunchAgent runs in the login session and authenticates. It also catches up after sleep. `--start`/`--stop` are macOS-only and error clearly elsewhere.
- **Glue-heavy, so coverage floor is modest.** The pure decision logic (`evaluate_capacity`, pace line, the argv/plist builders) is fully unit-tested; the launchd/subprocess/git IO is verified end-to-end (`fail_under = 25`, noted in the pyproject).
[...]

PR #2414, opened 2026-07-10
Add the document-review donation skill

jean-imbue:
## What

Adds the **`document-review`** Claude Code skill under `.claude/skills/document-review/` — a self-contained worker for a central review-coordination server (the Sinnott-Armstrong lab's ELSI-checklist project).

Per item, the skill:
1. Leases one assignment for the active prompt (`GET /next`), the server picks the item.
2. Fetches the linked document text from public biomedical APIs (PubMed/MEDLINE/PMC/NIH RePORTER/bioRxiv/medRxiv).
3. Applies the active versioned ELSI prompt to the documents.
4. Submits a structured JSON review back (`POST /reviews`) with the prompt version+hash and worker metadata.

## Notes for review

- **Manual mode only.** The reviewing model is the in-loop Claude session — there is **no `ANTHROPIC_API_KEY` / direct-API path** (the `run`/`auto` subcommands were stripped). A run can therefore only ever spend Claude **subscription** quota, never billable API tokens.
- **Prompts are not committed.** `prompts/` is gitignored; the prompts (the experiment rubrics) are pulled fresh from the lab's upstream git repo at run time by `mngr donate` (see the follow-up PR). This keeps the lab's repo the source of truth for experiments, so a reworded rubric doesn't need an mngr change.
- The skill code (`client.py`, `sources.py`, `prompts.py`) lives under `.claude/skills/` deliberately — it's the reviewed, pinned code that `mngr donate` copies + runs; it's not part of a `libs/*` package.

## Follow-up

A stacked PR adds the **`mngr_donate` plugin** that schedules this skill against spare Claude capacity and pulls its prompts dynamically. This PR is just the skill itself and is reviewable on its own (e.g. by the science collaborators).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2417, opened 2026-07-10
Add the mngr_donate plugin (spend spare Claude capacity on a donation skill)

jean-imbue:
## What

Adds the standalone **`imbue-mngr-donate`** plugin — `mngr donate`, which spends spare Claude capacity on a donation skill instead of letting idle quota expire.

One tick: read the account-level usage snapshot (from `imbue-mngr-usage`, a hard dependency), and *if* there's spare capacity (5h window under budget **and** the week under its pace line), launch a headless Claude agent that runs the donation skill to completion, then auto-cleans up. `--start`/`--stop` install/remove a **launchd LaunchAgent** (macOS) so idle quota drains on a schedule.

> Supersedes the earlier stacked pair (#2414 skill + #2415 plugin). We moved the skill *out* of this repo entirely — see below — so it's now a single PR with `mngr_usage` untouched.

## Skill lives in the lab's repo, checked out by ref

The donation skill (**code + prompts**) is owned by its **upstream git repo** — the single source of truth (for `document-review`, the Sinnott-Armstrong lab's GitLab). `mngr donate`:
1. **checks it out** into a host-dir cache (`<host_dir>/donate-skills/<skill>/`), and
2. points the headless agent at it to run.

`--skill-ref` controls what's checked out:
- a **branch** to *track* — each tick adopts the lab's latest (good for a fast-moving experiment), or
- a **pinned commit** for a reviewed, reproducible version — imbue bumps the ref to adopt updates.

So the lab iterates on the skill without an mngr release, and there's **no vendored copy in this repo to drift**.

⚠️ **Security note:** the agent runs the checked-out skill with `--dangerously-skip-permissions`, so whatever code is at `--skill-ref` executes on the host. The default is therefore **pinned to a reviewed commit** (`c2e9bbe…`), not `main` — so unattended/scheduled runs execute an audited version. Bump the SHA in a reviewed PR to adopt the lab's updates; pass `--skill-ref main` to track the branch instead.

## Why a separate plugin

Measuring usage (`mngr usage`) and donating spare quota (`mngr donate`) are orthogonal. This plugin depends on `imbue-mngr-usage` (installing donate pulls usage in) and is registered in the plugin catalog as a default-installable `DEPENDENT` gated on `imbue-mngr-usage`. `mngr_usage` is unchanged.

## Notes for review

- **launchd, not cron.** A cron job runs outside the login session and can't reach the macOS keychain where Claude's subscription token lives (every tick fails auth); a LaunchAgent runs in the login session and authenticates, and catches up after sleep. `--start`/`--stop` are macOS-only and error clearly elsewhere. It also forces shared Claude config (`-S …isolate_local_config_dir=false`) so the agent uses/refreshes the real token instead of a stale isolated copy.
- **Glue-heavy, modest coverage floor.** The pure decision logic (`evaluate_capacity`, pace line, argv/plist builders) is fully unit-tested; the launchd/subprocess/git IO is verified end-to-end (`fail_under = 25`, noted in the pyproject). A few ratchets are grandfathered for legit patterns (launchd `time.sleep`, stdout-tee, the deliberate `subprocess` shell-outs).

## Verified

`mngr donate --help`/`--dry-run` load from the new plugin; `mngr plugin list` shows both `donate` and `usage`; `prepare_skill_dir` cloned the lab's repo and checked out the skill (code + prompts); plugin-catalog, dependency-pin, changelog, and wheel/readme gates pass locally; `mngr usage` unaffected.


jean-imbue:
### CI note: `test-offload` red only from a pre-existing `minds` ratchet — not from this PR

After the latest push (`bdfe65ea8`), every donate-related failure is fixed:

- ✅ ruff **format** (`donate.py`, `donate_test.py`)
- ✅ `Commands missing generated docs: ['donate']` — `donate` registered in `SECONDARY_COMMANDS`, `docs/commands/secondary/donate.md` generated
- ✅ donate package tests + its own ratchets (70 passed)
- ✅ check-changelog, test-docker, test-offload-acceptance, Acceptance Tests

The **only** remaining red is `test-offload → test_prevent_inline_functions`, which scans **`minds`**, not this plugin:

```
RATCHET TEST FAILURE: inline functions in non-test code have increased!
apps/minds/imbue/minds/desktop_client/app.py:1072   def _on_change() -> None:
apps/minds/imbue/minds/desktop_client/api_v1.py:433
```

This is **pre-existing on `main`** and unrelated to donate:

- This branch touches **zero** `minds` files (`git diff --name-only origin/main...HEAD | grep minds` → none).
- The branch base is exactly `origin/main` HEAD (0 behind).
- The offending inline `_on_change` was introduced in `c15f7bad5` (Gabriel Guralnick, 2026-06-23, *"minds: discovery-pipeline health watchdog"*), long before this branch existed. The ratchet snapshot for `minds` was never updated for it, so `main` is already red on this test.

**Nothing in this PR can turn that check green** — it needs a separate `minds`-owned fix (extract the inline callbacks or bump the `minds` ratchet snapshot). Flagging so this can be merged with that context.

jean-imbue (commented):
Please review! Hoping to get this in so people can install and try at all hands.

libs/mngr/imbue/mngr/plugin_catalog.py
```diff
@@ -376,6 +376,14 @@ class CatalogEntry(FrozenModel):
         is_recommended=True,
         gate=RequiredPackagesGate(packages=("imbue-mngr-pi-coding", "imbue-mngr-usage")),
     ),
+    CatalogEntry(
+        entry_point_name="donate",
+        package_name="imbue-mngr-donate",
+        description="Spend spare Claude capacity on a donation skill (`mngr donate`)",
+        tier=PluginTier.DEPENDENT,
+        is_recommended=True,
```
joshalbrecht, line 384: Given that this is fairly experimental, I'd probably keep is_recommended False by default *for now*

Once we're ready to release it (and after we've done some testing, have some initial science partners, etc), then we can flip it to True

libs/mngr_donate/imbue/mngr_donate/donate.py
```diff
@@ -0,0 +1,788 @@
...
+   "fully used", so a no-data tick never looks spare -- and that case is flagged
+   so the CLI can say "can't tell" rather than "maxed out".
+3. If spare, launch a *headless* Claude agent (:data:`DONATE_AGENT_TYPE`) that
+   runs the donation skill (default ``document-review``) unattended and
+   auto-destroys when done; otherwise do nothing. The agent's stream is tee'd to
+   a durable per-run log under ``<host_dir>/donate-logs/``.
+
+A single tick spends at most one skill run's worth of quota. To drain spare
+capacity over time, ``mngr donate --start`` installs a launchd LaunchAgent
+(macOS) that re-runs the tick on an interval (``--stop`` removes it) -- the
```
joshalbrecht, line 20: this should work on osx and linux (those are the supported platforms for mngr)

(in practice this generally means it works on WSL too, though we dont test it there)

libs/mngr_donate/imbue/mngr_donate/donate.py
```diff
@@ -0,0 +1,788 @@
...
+DEFAULT_INTERVAL_MINUTES = 10
+# The LaunchAgent's label, and thus its plist filename in ~/Library/LaunchAgents.
+LAUNCHD_LABEL = "com.imbue.mngr.donate"
+
+# Spare-capacity thresholds, mirroring spare-capacity.sh exactly:
+#   spare  <=>  five_hour.used% < 80  AND  weekly.used% < pace_line(weekly.elapsed%)
+# where pace_line starts ~30% under the plain used%==elapsed% line early in the
+# 7-day cycle and tapers up to meet it by the cycle's end.
+FIVE_HOUR_USED_CEILING = 80.0
+WEEKLY_EARLY_MARGIN = 0.30
```
joshalbrecht, line 103: we'll definitely want these to be configurable, and we'll have to think about how the initial setup works properly here

libs/mngr_donate/imbue/mngr_donate/donate.py
```diff
@@ -0,0 +1,788 @@
...
+# Bump this SHA in a reviewed PR to adopt the lab's updates; pass `--skill-ref main`
+# to track the branch instead.
+DEFAULT_SKILL_REF = "c2e9bbe799c20c9da3896c2205991164f10555fd"
+
+# Optional macOS keychain entry holding a long-lived OAuth token from
+# `claude setup-token`. Headless agents can't refresh Claude's short-lived (~8h)
+# session token -- the desktop app refreshes only its own copy -- so without this,
+# scheduled ticks start failing with `401 Invalid authentication credentials` as
+# soon as the session token lapses. See README "Authentication".
+OAUTH_KEYCHAIN_SERVICE = "mngr-donate-oauth"
```
joshalbrecht, line 88: this is interesting, but we'll need to talk about this more and think about this -- the auth stuff gets fairly complicated, and we'll need a decent amount of testing here

and we'll need to think about how users are notified if / when it stops working too probably

libs/mngr_donate/imbue/mngr_donate/donate.py
```diff
@@ -0,0 +1,788 @@
...
+DEFAULT_AGENT_NAME = "donate-extra-quota-bio"
+
+# The donation skill (code + prompts) lives in the lab's upstream git repo -- the
+# single source of truth. donate checks it out at DEFAULT_SKILL_REF into a host-dir
+# cache and runs it from there, so the lab can revise the skill without an mngr
+# release. The ref is a branch to *track* (auto-adopt updates) or a pinned commit
+# for a reviewed, reproducible version (imbue bumps the ref to adopt updates).
+# Both overridable via --skill-repo / --skill-ref.
+DEFAULT_SKILL_REPO = (
+    "https://gitlab.com/sinnott-armstrong-lab/elsi-checklist/credits-for-science/document-review-skill.git"
```
joshalbrecht, line 75: I think we want one more level of indirection -- ideally it goes to *us*, and then we supply the list of (currently approved) science projects

we want the projects to be able to update even without people updating this package, for example

libs/mngr_donate/imbue/mngr_donate/donate.py
```diff
@@ -0,0 +1,788 @@
...
+
+
+# The donation agent runs headless (``claude --print``) on purpose: a plain
+# interactive ``claude`` agent blocks on the first tool-permission prompt, which
+# hangs ``mngr create`` and spends none of the quota. A headless agent streams
+# and auto-destroys after a successful pass, so a completed tick leaves nothing
+# behind (a tick that dies mid-way can, which is what the pre-launch cleanup in
+# `donate` handles). ``--dangerously-skip-permissions`` (below) is what lets it
+# actually run the skill's commands unattended.
+DONATE_AGENT_TYPE = "headless_claude"
```
joshalbrecht, line 184: hmmm, I'll need to think a little bit more about this. This has billing implications in the near future potentially

jean-imbue: would this PR be blocked by that? if so what would be the next steps?

libs/mngr_donate/imbue/mngr_donate/donate.py
jean-imbue: yes, 100% agree for the final version, i'm hoping to just get this in as a v0 so we can start unblocking Nasa

jean-imbue: and then we can also push out the blog post, which would then actually get us other scientists and make this necessary :)

libs/mngr_donate/imbue/mngr_donate/donate.py
jean-imbue: makes sense, could we enumerate a list of risks we're worried about so we can concretely address them + have an idea of the scope of work?

would it suffice to warn users about this and "use at your own risk" for now?

libs/mngr_donate/imbue/mngr_donate/donate.py
jean-imbue: proposing to punt this to the next PR 😅

libs/mngr_donate/imbue/mngr_donate/donate.py
jean-imbue: do any imbumans use linux?
also proposing to punt this to the next PR...

boweiliu:
https://github.com/imbue-ai/mngr/pull/2457

weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#9 from the same branch. This public repository is now a read-only mirror.

PR #2420, opened 2026-07-10
Speed up Cloudflare sharing.

hynek-urban:
After reducing the `mngr` call overhead in https://github.com/imbue-ai/mngr/pull/2403, the next big thing in terms of slowness in Cloudflare sharing was `mngr imbue_cloud tunnels list`. Apparently, it was unnecessary - on the server side (`ForwardingCtx.list_tunnels`), it iterated over all existing tunnels and retrieved details for each of them in separate requests to CloudFlare, even though the client only needed the details of a single tunnel.

This PR introduces a new command in the CLI and a new endpoint on the Imbue Cloud backend (remote service connector) to only request info about the one tunnel it needs.

The PR consists of two commits. The second one is there just for backwards compatibility - if we don't need new Minds clients to support old backends, I think the second commit can be safely reverted. (I wasn't sure what our deployment strategy is so I included it.)

Disclaimer: This was mostly driven by Claude itself; I'm not really familiar with the parts of the code this touches. It looks and sounds reasonable but I'm not 100% sure and I didn't properly test it because it involves updating the remote backend.


joshalbrecht:
this actually looks pretty good to me!

I think it'd be a good one for you to test (to learn how to deploy stuff), happy to chat about how to do that

joshalbrecht:
I'm working through this in a separate branch, and almost there.

I've got the time down to ~15 seconds, which is much more tolerable :)

(then another few seconds while cloudflare propagates it, out of our control)

hynek-urban:
Great :) So should I close this one?

PR #2421, opened 2026-07-10
Self-cleanup for MngrCaller's prewarmed processes.

hynek-urban:
Use the parent death watcher in pre-warmed mngr processes. (See https://imbue-ai.slack.com/archives/C0ASLEFFD62/p1783646722781289?thread_ts=1783607429.693149&cid=C0ASLEFFD62 for context.)


joshalbrecht:
lgtm!

yes, this will make me sleep easier :)

PR #2422, opened 2026-07-10
Do not show irrelevant old permission requests.

hynek-urban:
This is a simple UI adjustment of an inconvenience which, I believe, may also have contributed to Matt's experience reported [here](https://imbue-ai.slack.com/archives/C0B1JACQ9HT/p1783377794657299).

Previously:
- When a user opened a permission request dialog and Approved or Denied it,
- If there was another premission request dialog from before (for another agent, potentially old, redundant ...)
- It was immediately shown to the user as soon as the previous request was dealt with.

What this PR does is it auto-closes the permission request dialog window after Approval / Denial, unless the user explicitly opened the whole Request Inbox.


joshalbrecht:
lgtm

PR #2423, opened 2026-07-10
Supervise remote latchkey gateway + tunnel with supervisord

hynek-urban:
For the "remote Latchkey" (the secondary gateway that lives on VPS outer hosts, offering connectivity to agents when the user's laptop is down), do the following:

- Run both `latchkey gateway` and the ssh tunnel using supervisord to be robust to crashes.
- Provide approprate flags to the `ssh` invocation so that stale connections are dropped (and subsequently restarted by supervisord).

In the process, we write secrets to `/run` which is ram-backed. That means that the processes will _not_ respawn successfully after rebooting the whole system but we're OK with that for now. (The benefit is that we don't persist secrets to the disk.)


joshalbrecht:
lgtm

it's a *little* weird that supervisord on remote hosts is managed by mngr latchkey, but w/e for now

PR #2424, opened 2026-07-10
Fix Lima host-key mismatch when apt install fails transiently

qi-imbue:
## Problem

Creating a local Lima workspace intermittently failed with:

```
Error: Failed to connect to host: SSH host key error (Host key for 127.0.0.1 does not match.)
```

The Lima `provision[mode=system]` script installed the pre-trusted sshd host key **after** a network-dependent `apt-get install`, all under `set -eux -o pipefail`. When a Debian mirror transiently failed to serve package indexes, `apt-get` returned non-zero and `set -e` aborted the whole script **before** the host-key swap ran. The VM then kept its own default host key, while mngr had already pinned the key it *intended* to inject and connects with `StrictHostKeyChecking=yes` -- so the strict check rejected the connection.

Verified on a failing VM: the served ed25519 key differed from the injected one, the rsa/ecdsa/dsa keys were never removed, `tmux` was missing, and `cloud-init-output.log` showed `apt-get install` failing right before "scripts-per-boot failed".

## Fix

- Reorder the provision script so SSH host-key trust (key swap + sshd tuning + restart) is established **first**, before any package fetch. That block only uses coreutils and the already-present sshd, so host-key trust no longer depends on the network.
- Wrap the package install in a retry loop to ride out transient apt mirror failures. If it still fails, the error surfaces clearly on an SSH-reachable host instead of as a host-key mismatch.

## Tests

- New `test_provision_script_installs_host_key_before_apt` asserts the key swap precedes the apt install and that apt is retried.
- `just test-quick libs/mngr_lima/imbue/mngr_lima/lima_yaml_test.py`: 18 passed.

## Note

Split out of #2382 (onboarding fixes) so the Lima fix lands on its own; the macOS desktop-launch change from that branch is being kept as a local-only workaround instead.

### DIFF
--- a/libs/mngr_lima/changelog/mngr-lima-hostkey-before-apt.md
+++ b/libs/mngr_lima/changelog/mngr-lima-hostkey-before-apt.md
@@ -1,0 +1,5 @@
+Fixed a Lima workspace-creation failure that surfaced as a confusing `SSH host key error (Host key for 127.0.0.1 does not match.)`.
+
+The VM provisioning script installed the pre-trusted SSH host key *after* a network-dependent `apt-get install`, all under `set -e`. A transient Debian mirror hiccup (`apt-get` failing to fetch package indexes) would abort the whole script before the host key was installed, so the VM kept its default host key while mngr had already pinned the key it expected -- causing a strict host-key-check mismatch on connect.
+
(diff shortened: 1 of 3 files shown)


joshalbrecht:
lgtm

PR #2425, opened 2026-07-10
minds: actionable guidance when creating from a private GitHub repo

pseay-imbue:
image

## What

Creating a workspace from a **private** (or nonexistent) github.com URL previously surfaced only raw git stderr on the Creating page — and before that, could hang on a credential prompt. Now:

- The desktop client's local `git clone` of the workspace source runs with `GIT_TERMINAL_PROMPT=0` — no-credential clones fail fast with git's stable error shape (credential helpers like the macOS keychain still work). Mirrors the earlier FCT bootstrap fix.
- The backend classifies auth-shaped github.com clone failures (`could not read Username`/`terminal prompts disabled`/`Authentication failed`/`Repository not found`/403) as a new machine-readable `error_kind: GIT_AUTH_REQUIRED` on the create-operation status API (additive, optional field).
- The Creating page reveals a static guidance block for that kind, explaining in plain language: the repo looks private (or doesn't exist) and needs GitHub credentials on this computer; the simplest sign-in is `gh auth login` (linked to the official GitHub CLI quickstart); or clone/download the repo locally and enter that folder's path in the form instead of the URL.

Guidance is gated to github.com URLs only (the `gh` advice would be wrong elsewhere); other hosts keep the raw error. The copy is static template content gated on the error class — the backend classifies, it never generates prose.

## Testing

`uv run pytest` over the touched desktop_client modules + minds ratchets: **419 passed, 0 failed** (classification matrix, a local 401-server clone test proving fast auth-shaped failure, API round-trip of `error_kind`, and a page-render test asserting the block exists, is hidden by default, and carries the right copy/link). `ty check` and ruff clean on touched files.

Context: this is the follow-up UX for minds "inspirations" (published snapshot repos are private by default, so creating a mind from one hits this path for anyone without credentials configured) — see [FCT PR #243](https://github.com/imbue-ai/forever-claude-template/pull/243).


apps/minds/imbue/minds/desktop_client/agent_creator.py
```diff
@@ -272,6 +296,62 @@ def _is_local_path(repo_source: str) -> bool:
...
+# ("Repository not found" -- GitHub reports inaccessible and nonexistent repos
+# identically), and access blocked by a policy such as SAML SSO (403).
+_GIT_AUTH_OUTPUT_MARKERS: Final[tuple[str, ...]] = (
+    "could not read Username",
+    "could not read Password",
+    "terminal prompts disabled",
+    "Authentication failed",
+    "Repository not found",
+    "The requested URL returned error: 403",
+)
```
joshalbrecht: this is called making something "stringly-typed" (as opposed to "strongly-typed")

we really don't want things to be "stringly typed" :-P

a better approach is to change the underlying logic to raise particular types of errors (and/or have particular exit codes) that correspond to the type of error that you're trying to identify

this keeps the logic local, and prevent brittle coupling based on specific messages

in this case, do we even need to match against these specific strings? like, isn't any failure of "git clone" pretty much fine to show our error message, plus the original git error message?

pseay-imbue:
Noted! Got rid of the string checking and replaced it with error catching.

apps/minds/imbue/minds/desktop_client/agent_creator.py
```diff
@@ -284,6 +309,39 @@ def _is_local_path(repo_source: str) -> bool:
...
+    private (or does not exist) and this machine's git credentials cannot see
+    it -- and the creating page's guidance covers that case while the raw git
+    error stays visible right above it for anything rarer. Every other
+    failure returns ``None``.
+    """
+    if not isinstance(error, GitCloneError):
+        return None
+    if not _is_github_https_url(repo_source):
+        return None
+    return CreationErrorKind.GIT_AUTH_REQUIRED
```
joshalbrecht: there seems to be a conflation between "git" and "github" here -- probably we should be a bit clearer about that, if that makes sense?

pseay-imbue:
There's now a github (`gh auth login`) and a regular git error message providing guidance for any remote installations.

joshalbrecht:
lgtm, thanks!

PR #2369, opened 2026-07-06
Allow workspace renames, isolate provider discovery, and fix terminals

joshalbrecht:
## Main branches

Merges a few other PRs together and fixes the last issues with them:

### fix bug: manually tmux attaching to an agent inside a workspace sometimes broke message sending
- https://github.com/imbue-ai/mngr/pull/2336
- the fix is to be smarter about when we send the signal that the sized changed

### Make discovery  provider-specific (so that slow / broken providers will not cause issues for hosts that are part of some other provider)
- https://github.com/imbue-ai/mngr/pull/2353

### Enable workspace renaming (and simplify some of the sprawl that happened with the names ending up everywhere -- this one has a large blast radius)
- https://github.com/imbue-ai/mngr/pull/2343
- https://github.com/imbue-ai/forever-claude-template/pull/236

### Make terminals "persistent" and named (right now they constantly lose state, eg, if you exit a mind and come back in, any state in there is gone)
- https://github.com/imbue-ai/mngr/pull/2354
- https://github.com/imbue-ai/forever-claude-template/pull/237

## Other changes here:

- *important* one of the above PRs (the renaming one) also fixes how our tests run so that when changes are made to both FCT and this repo, the end to end acceptance and integration tests still pass
- a few small fixes for the renaming
- fixes merge conflicts
- fixes flaky test

## Corresponding FCT PR for all of these changes:
- https://github.com/imbue-ai/forever-claude-template/pull/242

PR #2427, opened 2026-07-10
lima: shorten VM instance name so `limactl start` doesn't hit UNIX_PATH_MAX

gnguralnick:
## Problem

Creating a Lima VM failed for some setups with:

```
limactl start failed: instance name "minds-staging-host-<32-char id>" too long:
".../.lima/minds-staging-host-<id>/ssh.sock.<suffix>" must be less than UNIX_PATH_MAX=104 characters, but is 104
```

Lima derives an SSH control-socket path from the instance name (`filepath.Join(LIMA_HOME, instance, filenames.LongestSock)`) and rejects the VM when `len(path) >= osutil.UnixPathMax`. The mngr instance name is `<prefix>host-<32-char id>`; the fixed 37-char host-id tail plus a longer env prefix (e.g. `minds-staging-`) and a longer home path pushes the socket path to exactly 104 for a 13-char username — one char over. The overflow is home-path/username- and prefix-length-dependent, so it hits some developers and not others.

This regressed in `832f5ff42` (landed on main via #2369, 2026-07-06), which switched Lima VM naming from the short host *name* to the fixed-length host *id* (to make the instance name stable across renames, since `limactl` has no rename).

## Fix

Make `lima_instance_name_from_host_id` length-aware (`libs/mngr_lima`):

- Compute Lima's socket-path budget and truncate the random hex tail just enough to keep the path under the limit.
- Preserve the full 32-char id whenever it already fits, so short home paths / prefixes are byte-for-byte unchanged (e.g. the `minds-` production prefix).
- Raise a new `LimaInstanceNameTooLongError` early, with an actionable message, when the prefix + `LIMA_HOME` leave no room for even a minimal (8-hex-char) id — instead of letting `limactl` abort with a cryptic fatal error.

Truncation is safe: nothing parses the host id back out of the instance name (discovery reads the persisted instance name from the host record), and the id keeps ≥32 bits of entropy.

The budget constants are grounded in Lima source (`pkg/instance/create.go`, `pkg/limatype/filenames/filenames.go`, `pkg/osutil/osutil_{others,linux}.go`): the socket filename is `filenames.LongestSock` = `ssh.sock.1234567890123456`, and `UnixPathMax` is 104 on macOS / 108 on Linux — the conservative macOS value is used so names are valid on both. `LIMA_HOME` is resolved once at the provider boundary (`create_host`) and threaded into the name function, which is a pure function of its inputs.

## Verification

- Reproduced the exact reported case against the real home path: old name = 51 chars → socket path 104 (rejected); new name = 50 chars → 103 (fits). Production prefix (`minds-`) produces the full untruncated id, unchanged.
- `just test-quick libs/mngr_lima`: 135 passed, 0 failed. The branch adds 6 new unit tests (replacing 1 old one): full-id-fits / truncates-to-fit / deterministic / raises-when-impossible / LIMA_HOME-env / LIMA_HOME-default.
- `uv run ty check` on the changed files: clean.
- Full offload suite deferred to CI.

Changelog: `libs/mngr_lima/changelog/gabriel-khaki-shark.md`.


joshalbrecht:
lgtm after discussion

PR #2428, opened 2026-07-10
Persist Electron logs and recover from renderer death across all window views

gnguralnick:
## What & why

From kanjun's blank-screen-after-sleep incident (Sentry `bc6d78e9`): on wake, a workspace window showed a pinned-white screen with no recovery, and the Electron main process logged nothing durable, so the failure was undiagnosable from uploaded logs. The root cause is that a window's `WebContentsView` renderers can be reaped by the OS over a long sleep, and nothing handled `render-process-gone`. Each window has **three** such renderers, in separate processes, that can die independently: the workspace **content** view, the **chrome** (titlebar) view, and the **overlay** (sidebar/inbox/help) view.

This implements `blueprint/electron-log-and-crash-page/`: durable main-process logging, recovery UI for all three views, and automatic Sentry reporting so these deaths are visible without waiting for a manual report.

## Changes

**Logging & rotation**
- New `electron/logger.js` — tees the main process's `console.log/warn/error` (timestamped) into a new `~/.minds/logs/electron.log`, and records uncaught exceptions (via `uncaughtExceptionMonitor`, chosen so Sentry capture / process-exit behavior is unchanged) and unhandled rejections. Reentrancy-guarded so a rotation-failure `console.warn` can't recurse into the tee. Initialized first in `main.js`.
- New `electron/log-rotation.js` — shared 100 MB / keep-10 size rotation, gzipping each rotated file in the background. Used by `electron.log` and (via `backend.js`) `minds.log`, which was previously an unbounded append stream (the source of ~600 MB log files). Async `fs.WriteStream` (never blocks the main thread); on hitting the threshold it ends/flushes the stream **before** renaming so the background gzip never reads a partial file, replays any lines that arrive mid-rotation, and gzips via `stream.pipeline` so a compression failure can't leak fds.

**Renderer-crash recovery (all three views)**
- **Content view** — `render-process-gone` shows a local `crashed.html` ("Aw, Snap!"-style: crash reason/exit code, a manual **Reload**, and **Report a bug**). No auto-reload, by design (crash-loop safety, matching Chrome/Firefox/VS Code). `content-relay-preload.js` relays the allowlisted `minds:reload-crashed-view`; `isCrashPageUrl` guards `onContentNavigate` so the crash page doesn't clobber the pre-crash URL that Reload re-loads.
- **Chrome (titlebar) view** — because the chrome view is a full-window underlay with only its top ~38 px visible, a dead titlebar has no recovery affordance at all. It now shows a compact local `chrome-crashed.html` strip anchored to that band, with a manual Reload (`reload-chrome` IPC → reloads `/_chrome`, re-primed from the cached chrome state). The workspace content view keeps running. Drag region is scoped to just the 38 px strip: macOS unions `-webkit-app-region` across all views in a window, so a full-window drag surface was swallowing clicks on the content view's Reload button when both crashed at once.
- **Overlay/menu view** — silently reloaded warm (`loadOverlayHost`) on death; it's hidden when idle, so there's no visible error state, just a fresh host for the next sidebar/inbox/help open.
- All handlers ignore `clean-exit` / shutdown / error-takeover teardown and defer navigation via `setImmediate` (electron#19887 — never `loadURL` synchronously in that handler).

**Sentry reporting of renderer deaths**
- `sentry.js` broadens the `@sentry/electron` childProcess integration to capture **`oom`** renderer deaths as events (its default only events `abnormal-exit`/`launch-failed`/`integrity-failure`), and sets `getRendererName` so events are labeled by which view died (`chrome`/`content`/`modal`). `crashed` is deliberately **not** added — a native crash already produces a minidump via `SentryMinidump`, so adding it would double-report. `killed` stays breadcrumb-only (OS/sleep reaping is unactionable and noisy); it remains in `electron.log`. Auto-gated by the existing `report_unexpected_errors` opt-in.

**Upload plumbing**
- `sentry.js` — backend-down reports now attach both `minds.log` and `electron.log` (plus the newest jsonl), so a report filed while the backend is dead shows both why it died and what the shell did.
- `imbue/minds/utils/sentry/core.py` — splits the previously-mislabeled `*.log` attachment group into accurately-named `backend_logs` (`minds.log`) and `electron_logs` (`electron.log`), each with a gzipped rotated group; `core_test.py` updated.

## Screenshots

image

This shows the result of killing both the chrome and content view renderer processes. Each state can show independently.

## Testing

**Automated**
- [x] JS unit tests: `node --test apps/minds/test/unit/*.test.js` → 16 passed (adds `log-rotation.test.js`)
- [x] Python: `just test-quick apps/minds/imbue/minds/utils/sentry/core_test.py`
- [x] `/autofix` and `/verify-conversation` gates clean

**Real-app verification (done on a staging build)**
- [x] Killing the **content**, **chrome**, and **overlay** renderers each shows the right recovery UI; Reload restores the view; both crash pages' Reload buttons are clickable independently when all three die at once (the drag-region fix).
- [x] `oom`/`killed`-path events reach Sentry labeled by view — three distinct issues (`'chrome'`/`'content'`/`'modal'` process exited), carrying `git_sha`/`release`/`environment` tags and structured `{exitCode, reason}` breadcrumbs. (`killed` was temporarily un-exempted only to drive this test, then reverted — it does not ship as a reported reason.)
- [x] A manual `/help` bug report uploads both `minds.log` and `electron.log` to S3 under their accurate `uploaded_files_backend_logs` / `uploaded_files_electron_logs` context keys. (The `*_rotated_logs` gzip groups are correctly absent until a log crosses 100 MB.)


apps/minds/electron/main.js
```diff
@@ -1036,6 +1175,26 @@ function createBundleOverlayView(bundle) {
       } catch { /* noop */ }
     }
   });
+  // When the overlay host's renderer dies (it runs in a separate process and can
+  // be reaped over a long sleep like the other views), there is nothing visible to
+  // recover -- the overlay is hidden whenever no modal is open. So we just reset any
+  // open-modal state and silently reload the warm host, so the next sidebar / inbox /
+  // help open lands on a fresh page instead of a dead one. Defer the reload a tick
+  // (never navigate synchronously in this handler -- electron#19887).
+  modal.webContents.on('render-process-gone', (_e, details) => {
```
gnguralnick, line 1184: i think it's reasonable for the modal webcontentsview to auto-reload since it's generally not even visible to the user but lmk if i should change that

joshalbrecht:
lgtm, but would love a post-hoc review from @hynek-urban as well in case I'm missing something

PR #2429, opened 2026-07-10
Fix the inline-function ratchet double-count

weishi-imbue:
## What

`find_inline_functions` in `libs/imbue_common` double-counts a function nested two or more levels deep. It iterated **every** `FunctionDef` in a file (`get_ast_nodes_of_type` returns nested ones too), then `ast.walk`ed under each. A function with two enclosing functions is a descendant of both, so it was emitted **once per ancestor**. `snapshot(N)` never meant "N inline functions" for any file with 2-deep nesting.

Nested defs are now collected keyed by `(lineno, col_offset)` and counted once. Two distinct `def`s cannot share both coordinates, so the dedup is exact.

## Blast radius: one project

Measured across all 44 projects that pin this ratchet, using each test's real `_DIR`: **exactly one recorded count changes — `apps/minds`, 9 -> 7.** Every other project has zero 2-deep nesting (`imbue_common`'s own count, 3, is unchanged). The two counts removed from minds are the same doubly-nested function counted twice.

## Why this exists

`main` briefly went red on `apps/minds` `test_prevent_inline_functions` after a merge race (PR #2409 tightened the snapshot 9->8 against a base that had 8; PR #2413 legitimately added a 9th violation when main still pinned 9; the merge combined them). Main self-healed by bumping the snapshot back to 9. This PR is the honest follow-up: fix the checker so the recorded count is the true count, rather than leaving an inflated snapshot in place.

## Not in scope (deliberately)

An earlier revision of this PR also hoisted the backup-detail endpoint's one genuine inline closure (`_run_check_into_results`) to a module-level helper, which would have taken minds to 6. Reverted: that closure captures five locals and is used once as a thread target, so the inline form reads better than a five-argument module-level function. It stays, and it is honestly counted in the 7.

## Verification

- `find_inline_functions` recount: minds 7, imbue_common 3 (unchanged).
- Blast-radius sweep over all 44 pinning projects with real `_DIR`: only minds changes.
- `apps/minds` + `imbue_common` inline-function ratchets: pass at 7 and 3.
- Ratchet checker unit tests `core_test.py`: 25 passed.
- `ty check` clean on the changed source.


joshalbrecht:
lgtm, thanks for the fix!

PR #2432, opened 2026-07-11
update azure to 26.0.0; tweaks from bumping other deps

evgunter:
new installs of the azure plugin were broken because it would automatically pull in 26.0.0. the root cause is that dependencies aren't pinned when installing plugins, which is fixed by https://github.com/imbue-ai/mngr/pull/2434

also changes the behavior of plain `mngr config` and `mngr plugin` (no subcommand) to print the default click usage instead of the custom message. this is because click updated so that optional subcommands are displayed as `[COMMAND]` instead of `COMMAND`, and `mngr config`/`mngr plugin` were indicated as having an optional subcommand because of their custom help behavior. making the plain commands use our fancy help again is done in https://github.com/imbue-ai/mngr/pull/2442

---

## What

`azure-mgmt-resource` 26.0.0 made `azure.mgmt.resource` a namespace package and moved `ResourceManagementClient` into `azure.mgmt.resource.resources`. The plugin's old `from azure.mgmt.resource import ResourceManagementClient` raised `ImportError` on import, crashing *every* `mngr` invocation once `imbue-mngr-azure` was installed (compounded by there being no per-plugin import isolation).

## Changes

**Azure fix**
- `libs/mngr_azure/imbue/mngr_azure/client.py`: import `ResourceManagementClient` from `azure.mgmt.resource.resources` (the other resource imports already used the `.resources` subpackage).
- `libs/mngr_azure/pyproject.toml`: bump pin to `azure-mgmt-resource>=26`.
- Root `pyproject.toml`: advance the `exclude-newer` supply-chain cooldown from `2026-06-04` to `2026-06-27` so the workspace can resolve 26.0.0 (released 2026-06-24). This normally advances at release time but was stale because the last release predated the dep. This re-resolves `uv.lock` across the workspace (large lock churn is expected: compute 37->38, storage 24->25, etc.).

**Collateral from the dependency refresh** (advancing `exclude-newer` upgraded `ty` 0.0.39->0.0.54 and library stubs like `click`/`urwid`, surfacing pre-existing issues, none of which are azure-related):
- Regenerated the `config`/`plugin`/`usage` CLI reference docs.
- `apps/minds`: marked `ImbueCloudCliError.exit_code` as an intentional per-instance override (click now types `exit_code` as a `ClassVar`).
- `libs/mngr`: gave a create-test's `create_host` double an explicit typed signature (was `**kwargs: Any`).
- `libs/mngr_kanpan`: hyperlink-canvas `coords`/`translate_coords` return `Mapping`; `content()` matches urwid's `int` params.
- `libs/resource_guards`: dropped a `ty: ignore` the newer ty no longer needs.
- Trimmed the mngr broad-exception-catch ratchet 8->7 (recorded count was stale).

**`config`/`plugin` require a subcommand** (folded in while regenerating their docs):
- Both groups were `invoke_without_command=True` solely to render help when run bare. But click already shows help for a required-subcommand group, and this made them exit 0 and advertise an *optional* `COMMAND` in the docs -- inconsistent with every other group. Dropped the flag so they behave like `mngr snapshot`.

### Before / after: plain `mngr config` and `mngr plugin`

**Before** -- rendered the rich man-page help and exited **0** (as if bare invocation were a valid command):

```
$ mngr config
NAME
       mngr config - Manage mngr configuration
SYNOPSIS
       mngr [config|cfg] <subcommand> [OPTIONS]
DESCRIPTION
       ...
$ echo $?
0
```

**After** -- prints the standard usage/help listing and exits **2** (usage error: a subcommand is required), identical to `mngr snapshot` and every other group. The rich man-page help is still available via `mngr config --help`:

```
$ mngr config
Usage: mngr config [OPTIONS] COMMAND [ARGS]...

Options:
  --scope [user|project|local]  ...
  ...
Commands:
  get
  list
  set
  ...
$ echo $?
2
```

(`mngr usage` is unchanged: it remains runnable bare because bare `usage` does real work -- it prints the usage snapshot.)

## Verification

- All 200 `libs/mngr_azure` unit/integration tests pass locally on 26.0.0; full plugin import smoke-test passes.
- `uv run ty check`: clean (was 14 diagnostics after the refresh).
- `test_no_type_errors` + `test_cli_docs_are_up_to_date` pass; the updated `test_plugin_without_subcommand_shows_help` and the config/plugin CLI test suites pass.
- Manually verified bare `mngr config`/`mngr plugin` now match `mngr snapshot` (Usage/Options/Commands, exit 2) and that `--help` still renders the man-page help.
- Full offload suite left to CI, since the `exclude-newer` bump affects workspace-wide resolution.

## Follow-up watch

`azure-mgmt-network` 31.0.1 (new major, released 2026-07-02) sits just past this cutoff and will be pulled at the next cooldown advance. Its import surface is unchanged (`NetworkManagementClient` still top-level; all model classes we use present), so it likely needs no code change, but the azure suite should be re-run when it lands.


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#12 from the same branch. This public repository is now a read-only mirror.

evgunter:
@weishi-imbue what's happening with mngr? how should i submit future PRs? (or should i just fork it?)

PR #2433, opened 2026-07-11
docker isolation (new) + default agent type (moved) in config wizard

evgunter:
previously new users would get a warning about legacy docker isolation; now the installer prompts them to write that config setting up front

---

## Summary

Consolidates one-time user-scope config setup under `mngr config wizard` instead of splitting it across `mngr extras`.

A fresh install has no `[providers.docker]` block, so the default docker provider is resolved via the bare-backend fallback (`providers/registry.py`), instantiating `DockerProviderConfig` with `isolate_host_volumes` unset -- emitting the one-shot default-flip deprecation warning the first time the user touches docker.

Rather than a separate command, this adds the docker setting as a step in the existing `mngr config wizard` (which already prompts for Claude config-dir isolation), and moves the default-agent-type step there too:

- **Docker host-volume isolation** (new step): the wizard asks whether each docker host sees only its own `host_dir` sub-folder (the recommended, forthcoming default; requires Docker Engine >= 25.0) or keeps the legacy shared state volume, writing `providers.docker.isolate_host_volumes`. Opting in silences the deprecation warning. The recommended "Yes" is the default-highlighted option; users on older Docker can pick "No".

- **Default agent type** (moved): the "pick a default agent type for `mngr create`" step moved out of `mngr extras` into the wizard. Because the wizard runs as its own installer step (after plugins are installed), it now sees freshly-installed plugin agent types instead of only those registered at startup.

Every prompt also offers an explicit **"leave unset for now"** choice (which states the default it falls back to), so a decision can be deferred instead of forcing a value or a cancel.

Each step short-circuits when its setting is already configured, checked across **every config scope** (user, project, and local) so a value set in any of them suppresses the prompt. The two settings whose defaults already mark "unset" (the create default type and docker's `isolate_host_volumes`) are read off the merged config; Claude's `isolate_local_config_dir` has a non-None default, so its explicit-presence check reads the raw config files across all scopes instead.

`mngr extras` loses its `config` subcommand and default-agent-type step; user-scope config setup lives entirely in `mngr config wizard`. `scripts/install.sh` already runs `mngr config wizard`, so no installer change is needed. Tutorial references and the generated config command docs are updated.

The code default (`isolate_host_volumes: bool | None = None`) is intentionally **not** changed here.

## Verification

- Unit tests for all three wizard steps: each step now *returns* the chosen value or `None` (returns-picked-value / returns-none-when-already-set / -no-choices / -declined / -non-interactive), and the orchestrator centralizes writing through the same validated path as `mngr config set`. Added tests for the across-scopes explicit-presence read and the value-render helper. **91 passed** in `config_test.py`, **45** across `extras_test.py` + docker `config_test.py`; ruff, ty, and the `functools.partial` / inline-function ratchets pass.
- Manually verified end-to-end: `mngr config wizard` runs all three steps; the interactive path writes `providers.docker.isolate_host_volumes = true` and `commands.create.type` (validated, deprecation warning gone) and is idempotent on re-run. A project-scope Claude setting correctly suppresses that prompt, and "leave unset" skips without writing. Non-interactive prints the suggested `mngr config set` commands without writing.

## History

Earlier commits on this branch implemented an extras-based approach; this supersedes them.


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#13 from the same branch. This public repository is now a read-only mirror.

PR #2435, opened 2026-07-11
Named dockview layouts: blueprint + Docker cross-env name-collision fix

joshalbrecht:
Two things on this branch:

1. **Blueprint** for named dockview layouts at `blueprint/dockview-named-layouts/plan-dockview-named-layouts.md`. The implementation lives in imbue-ai/default-workspace-template#257 (same branch name there).

2. **Bug fix (libs/mngr)**: the Docker provider's create-time host-name collision check (`_find_container_by_name`) matched containers by the `host-name`/`provider` labels alone. Labels carry no environment discriminator, so two mngr environments that differ only in `MNGR_PREFIX` (e.g. two minds envs) label their containers identically — creating a host that reused a name from *another* env failed with a misleading "container already exists" error naming a container that doesn't exist (found while testing the layouts feature: a staging-env workspace named `docker-1` was blocked by a dev-env container `minds-dev-josh-1-docker-1`). The lookup now also requires the container's actual name to equal `<prefix><host-name>` — the same uniqueness scope Docker itself enforces and the same prefix filter `_list_containers` already applies to discovery. Containers are never renamed after creation (`mngr rename` only updates the host record), so name and label stay in lockstep.

Includes a regression test (`test_find_container_by_name_ignores_other_environments`) and verification against the real colliding containers: with the staging prefix the lookup now returns None while the dev-env prefix still finds its own container.


joshalbrecht:
lgtm

PR #2436, opened 2026-07-11
fix test that was broken if sh points to dash

evgunter:
or could just ensure it doesn't use dash

---

## Problem

`libs/mngr/imbue/mngr/hosts/host_test.py`'s `test_build_remote_lock_command_*` tests validated the generated remote lock command with `subprocess.run(["sh", "-n", "-c", cmd])`, intending a no-exec syntax check.

But **dash ignores `-n` when the script comes via `-c`** (it only honors `-n` for stdin/file input). Since `/bin/sh` -> dash on Debian/Ubuntu, the command actually *ran*, executing its `mkdir -p /mngr/...`. This passed on CI (root, writable `/`) but failed locally on an unprivileged or read-only `/` with `mkdir: cannot create directory '/mngr': Permission denied`.

## Fix

Feed the script via stdin (`subprocess.run(["sh", "-n"], input=cmd, text=True, ...)`), where dash honors `-n` and performs a genuine no-exec syntax check. Verified: dash executes `sh -n -c 'echo X'` (prints X) but does not execute `echo X | sh -n`, while still catching syntax errors (rc 2).

Latent test fragility, not a product bug.


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#15 from the same branch. This public repository is now a read-only mirror.

PR #2437, opened 2026-07-11
Add claude-fable-5 to litellm proxy and bump Claude Code to 2.1.207

weishi-imbue:
Makes Claude Fable 5 available in minds workspaces: the LiteLLM proxy learns to route and price `claude-fable-5`, and the pinned Claude Code version moves to a Fable-capable release.

## Changes

**LiteLLM proxy — add `claude-fable-5`**
- `apps/modal_litellm/app.py` (`LITELLM_CONFIG`) and its mirror `litellm_proxy/config.yaml`: new model entry with inline pricing mirrored from litellm's `model_prices_and_context_window` map — $10 / $50 per 1M input/output tokens, cache write 1.25e-5, cache read 1e-6 (2x Opus tier). Inline pricing keeps cost tracking correct regardless of the deployed litellm version's bundled price map; `config_drift_test.py` keeps the two configs in sync.

**Usage pricing tables**
- `libs/mngr_usage/.../pricing.py`: `anthropic/claude-fable-5` entry so token-only usage sources price Fable instead of reporting it unpriced (`mngr_usage_pricing_drift_test.py` enforces sync with modal_litellm).
- `libs/mngr_robinhood/.../_agent_sdk/pricing.py`: fable family entry so Agent SDK sessions report `total_cost_usd` instead of `None`, with a resolve-test assertion.

**Claude Code pin 2.1.160 → 2.1.207** (latest release; supports Fable) at all five sites: release Dockerfile `ARG`, `scripts/snapshot_minds_e2e_state.py`, `release-tests.yml` (2), `tmr-setup` action.

Changelog entries for all five touched projects.

## Verification

- Full suite: `just test-offload` — 15431 passed, 0 failed.
- Empirical, on a real deployed proxy: deployed the `dev-weishi` env from this branch, minted a LiteLLM virtual key, and ran `claude -p --model claude-fable-5` with `ANTHROPIC_BASE_URL` pointed at the proxy (the exact workspace wire path). Request succeeded end-to-end and the spend log recorded `anthropic/claude-fable-5, $0.735645, 60069 tokens` — consistent with the cache-write rate (60k tokens x 1.25e-5 ≈ $0.75).
- Also confirms the Anthropic org meets Fable's 30-day data-retention requirement (no 400s).

## AskUserQuestion mitigation (review feedback)

Recent Claude Code releases no longer treat AskUserQuestion as a permission prompt, so `auto_allow_permissions` alone no longer prevents an unattended agent from blocking on a question. Added an opt-in `auto_disable_questions` field to `ClaudeAgentConfig` that appends `--disallowed-tools AskUserQuestion` to the agent invocation (verified against claude 2.1.207 that repeated `--disallowed-tools` flags accumulate, so a list already supplied via `cli_args` survives), with unit tests.

Where it gets enabled:
- **default-workspace-template**: NOT yet — its vendored mngr predates the field and `ClaudeAgentConfig` forbids unknown keys, so setting it there today would break workspace agent creation. Its existing raw `cli_args` `--disallowed-tools AskUserQuestion,...` keeps workspace agents safe meanwhile; switch to the typed field after the next vendor/mngr sync.
- **This repo's own .mngr/settings.toml**: deliberately not set — agents here run interactively where questions are useful.

## Merge coordination

Companion PR imbue-ai/default-workspace-template#258 bumps `[agent_types.claude].version` to 2.1.207. The release test `test_claude_code_version_matches_default_workspace_template_pin` requires the two pins to match — merge both together (release tests don't run in CI, but agent provisioning fails on a version mismatch until both land).

## Rollout

Deployed proxies pick up the model list on their next `minds env deploy`; staging/production deploys to be coordinated after merge. Workspaces on the SUBSCRIPTION/API_KEY AI providers only need the version bump.


joshalbrecht:
lgtm except for one required follow up change, see slack

PR #2438, opened 2026-07-12
make plugin remove/disable not crash when a plugin errors at import time

evgunter:
## Problem

A plugin that fails to import (e.g. `imbue-mngr-azure` after `azure-mgmt-resource` 26 renamed its import path) crashed **every** `mngr` command at import time -- including `mngr plugin remove` / `mngr plugin disable`, the commands you'd use to recover. Installing a broken plugin locked you out of the tool needed to uninstall it.

Plugin loading happens at module-import time (`create_plugin_manager` -> `pm.load_setuptools_entrypoints("mngr")`), before Click parses argv, and pluggy offers no per-plugin isolation.

## Approach

Rather than swallow import errors and run in a degraded state (against mngr's philosophy), only the two recovery escape-hatch commands bypass plugin loading:

- `mngr plugin remove` and `mngr plugin disable` now run **without** importing third-party plugin entry points. `remove` works off the uv-tool receipt and `disable` just writes config, so neither needs the loaded plugins.
- These two are detected from `sys.argv` (before Click parses it, since the load happens at import). Detection is hardcoded to `{plugin, plug} x {remove, disable}`; a test keeps that set in sync with the real command tree so it can't silently drift.
- They also parse config leniently (as `mngr plugin add` already does), so a `[providers.<x>]` block belonging to the not-loaded plugin doesn't turn recovery into a new "unknown backend" error. Plugin-disable blocking still runs (it imports nothing) so `load_config`'s strict re-block doesn't trip.

**Every other command is unchanged**: it loads all plugins and fails loudly if one is broken.

## Verification

Manually reproduced end-to-end with a simulated broken entry point: a normal command (`mngr ls`) crashes with the `ImportError`; `mngr plugin disable <name>` succeeds and writes config; afterward the normal command no longer crashes on that plugin. `mngr plugin remove` gets past the import to its normal removal path.

Tests added: recovery-detection truth table, command-tree sync test, `create_plugin_manager` entry-point-skip + still-blocks-disabled, and `plugin disable` lenient-parse / skips-soft-validation.

_Note: this PR touches `libs/mngr` (the fix) and `libs/mngr_lima` (a one-line call-site update after `create_plugin_manager` gained a required argument)._

### DIFF
--- a/dev/changelog/ev-plugin-import-isolation.md
+++ b/dev/changelog/ev-plugin-import-isolation.md
@@ -1,0 +1,1 @@
+Update the `scripts/qi/fd_leak/` FD-leak repro scripts to pass `load_entry_points=True` to `create_plugin_manager`, which now requires that argument. Keeps their behavior unchanged (all providers loaded) and keeps the repo-wide type check green.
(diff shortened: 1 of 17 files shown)


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#16 from the same branch. This public repository is now a read-only mirror.

PR #2440, opened 2026-07-12
make config-editing hints use `mngr config` whenever possible + shared hint helpers

evgunter:
ideally we never tell the users to hand-edit config; this PR removes all but one instance of that (for editing `enabled_backends` there's no good config command to remove just one backend; it seemed worse to suggest a command which contained a list of all backends except the one to disable)

---

## What

Standardizes user-facing configuration remediation hints so error and warning messages always point users at a runnable `mngr config set` / `mngr config unset` command instead of telling them to hand-edit `settings.toml`.

Two goals: eliminate the "edit settings.toml" hints, and route every site through one helper so flag-order and scope drift are structurally impossible.

## How

- New helper `imbue.mngr.remediations` with `format_config_set`, `format_config_unset`, and `format_disable_provider`. One canonical flag order (`--scope` immediately after `set`/`unset`), factored into a shared `_scope_flag` renderer.
- Scope is typed with the canonical `ConfigScope` enum rather than a re-declared string literal. To let the low-level `remediations` module import it without an `errors -> remediations -> data_types -> errors` cycle, `ConfigScope` moved from `config/data_types.py` down to `primitives.py` (which `errors.py` already depends on); all importers now pull it from there.
- `format_config_set` / `format_config_unset` take no default scope (per the style guide) — every caller passes `scope=` explicitly.
- `format_disable_provider` always recommends `--scope local`. Config precedence is user < project < local (local wins), so a `--scope user` suggestion was silently overridden — and thus ineffective — whenever the provider was enabled at the project or local layer. Writing to local always takes effect. This also unifies the previous split where most providers used `--scope user` but modal used `--scope local`.
- Routed all the drifting sites through the helper: `mngr create --template` (no templates), custom-agent-type "no command", provider unavailable/not-authorized errors, the list-error disable hint, notifications terminal config, and the azure/gcp/aws/modal/ovh/claude provider hints.
- Made two hints reflect the user's actual input rather than generic placeholders: the "no command" example now uses the agent type they ran (not always `--type command`), and the "template not found / none configured" hint scaffolds the specific template name they asked for. The OVH hint now names the real credential keys (`application_key` / `application_secret` / `consumer_key`) instead of a vague `<key>`.

## Notes / deliberately left alone

- `ProviderNotAuthorizedError`'s "or remove it from `enabled_backends`" line is kept as prose: `enabled_backends` is an allowlist with no single-command remove operation, so a file reference is the honest guidance there. (Possible follow-up: only show that line when a non-empty allowlist actually exists.)
- Value-set hints (subscription_id, project_id, etc.) pass `scope=None` (no `--scope`), letting `config set` use its default (project) scope; only disable hints force `local`.
- The claude "share your Claude config" hint keeps its existing `--scope user` (a per-machine credential preference), now just rendered in canonical flag order.

## Testing

- New unit tests: `libs/mngr/imbue/mngr/remediations_test.py`.
- Updated `list_test.py` disable-hint assertion (`--scope user` → `--scope local`).
- `ty` type-check and `ruff` clean on all touched files. Targeted message-assertion and config-loader/`ConfigScope` unit tests pass. Full suite runs via CI.

### DIFF
--- a/libs/mngr/changelog/ev-standardize-config-tips.md
+++ b/libs/mngr/changelog/ev-standardize-config-tips.md
@@ -1,0 +1,7 @@
+Standardized user-facing configuration remediation hints. Error and warning messages now always suggest a runnable `mngr config set` / `mngr config unset` command instead of telling you to hand-edit `settings.toml`.
+
+A new shared helper (`imbue.mngr.remediations`) renders these hints in one canonical form, so flag order and scope no longer drift between call sites.
+
+Provider-disable hints now recommend `--scope local` rather than `--scope user`. Because config precedence is user < project < local, a `--scope user` suggestion was silently overridden (and thus ineffective) whenever the provider was enabled at the project or local layer; writing to the local scope always takes effect.
+
+Updated the affected messages in `mngr create --template`, the custom-agent-type "no command" error, and the provider-unavailable / not-authorized errors.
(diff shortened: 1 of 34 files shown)


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#18 from the same branch. This public repository is now a read-only mirror.

PR #2441, opened 2026-07-12
separate hints for disabled vs not installed provider

evgunter:
## Summary

The `Provider 'X' references unknown backend 'X'` config error told users to add `plugin = "<plugin-name>"` to the provider block whenever any plugin was disabled — advice that cannot resolve a backend whose plugin is **not installed at all**. Users who followed it hit the same error again, and the "Currently disabled plugins" list (which didn't include the missing backend) added to the confusion.

### Repro (before)
```
$ mngr config set --scope local providers.azure.plugin azure
ERROR Invalid configuration: Provider 'azure' references unknown backend 'azure'.
If this backend is provided by a disabled plugin, either enable the plugin or add
plugin = "<plugin-name>" to this provider block. Currently disabled plugins: claude_subagent_proxy
```

### After
```
Provider 'azure' references unknown backend 'azure'. This plugin is provided by
'imbue-mngr-azure' (Azure Virtual Machines provider backend plugin for mngr). Install
it (e.g. reinstall the mngr uv tool with '--with imbue-mngr-azure') and ensure the
plugin is enabled. If instead this backend is provided by one of the currently disabled
plugins (claude_subagent_proxy), enable that plugin or add plugin = "<plugin-name>" to
this provider block.
```

## Changes

- `config/loader.py` `_parse_providers`: an unregistered backend now leads with the concrete plugin install hint (via the existing catalog-backed `get_plugin_install_hint`), and only mentions the `plugin =` path as a secondary "if instead" option, and only when other plugins are disabled. The installed-but-disabled branch (backend name maps directly to a disabled plugin) is unchanged.
- Updated `test_parse_providers_unknown_backend_mentions_disabled_plugins` for the new phrasing; added two tests asserting the install hint is primary and that `plugin =` is not suggested when the plugin is uninstalled.
- Changelog entry.

## Testing

`uv run pytest libs/mngr/imbue/mngr/config/loader_test.py -k parse_providers` — 18 passed. Full `loader_test.py` run: 134 passed (40 errors are a local-environment artifact — a fixture runs `git init -b main` and this machine resolves an old Windows git without `-b`; unrelated to this change and green in CI sandboxes). Also manually verified all three message paths end-to-end.

### DIFF
--- a/libs/mngr/changelog/ev-provider-error-hint.md
+++ b/libs/mngr/changelog/ev-provider-error-hint.md
@@ -1,0 +1,5 @@
+Improved the "Provider 'X' references unknown backend 'X'" configuration error when the backend's plugin is not installed at all.
+
+Previously the message unconditionally advised adding `plugin = "<plugin-name>"` to the provider block whenever any plugin was disabled. That advice cannot resolve a backend whose plugin is not installed, so users who followed it hit the same error again, and the "Currently disabled plugins" list (which did not include the missing backend) added to the confusion.
+
+The message now distinguishes the two cases: when the backend maps to an installed-but-disabled plugin, it still suggests enabling it or opting out with `plugin =`; when the backend is not registered at all, it leads with the concrete install hint (e.g. naming `imbue-mngr-azure`) and only mentions the `plugin =` path as a secondary "if instead" option when other plugins happen to be disabled.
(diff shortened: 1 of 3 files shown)


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#19 from the same branch. This public repository is now a read-only mirror.

PR #2442, opened 2026-07-12
improve help: section for subcommands; if subcommand is missing, display nice help page, not click default

evgunter:
Two related improvements to the help output for command groups.

**Note on base branch:** based on `ev/azure-update` (PR #2432), which already made `config`/`plugin` require a subcommand and exit 2 with plain click help. This PR builds the unified git-style rendering on top of that groundwork. It should merge after (or together with) #2432.

## Group `--help` now lists subcommands

The git-style help page (`mngr <group> --help`, `mngr help <group>`) previously rendered NAME / SYNOPSIS / DESCRIPTION / OPTIONS but **no command list**, so `mngr config --help` never showed `list`/`get`/`set`/... This adds a `COMMANDS` section (placed before `OPTIONS` for prominence) listing each subcommand with its one-line description and inline aliases.

## Unified bare-invocation behavior

Bare group invocation is centralized in `DefaultCommandGroup`: with no subcommand it renders the same rich, pageable git-style page to **stdout** and exits **2**, instead of click's plain stderr usage line. `config` and `plugin` are migrated onto `DefaultCommandGroup` so they behave identically to `snapshot`/`git`/top-level `mngr`.

Because the new COMMANDS section lands too, switching bare invocation to the git-style page is a strict improvement (it no longer loses the command list that plain click help showed).

The `config`/`plugin` synopsis now enumerates the common subcommands (`[list|get|set|unset|edit|...]`, `[list|add|remove|enable|disable|...]`) rather than a bare `COMMAND`, matching how `snapshot`/`git` already read.

## Known minor limitation

`mngr config --scope user` (a group-level option but no subcommand) still shows click's terse `Error: Missing command.` rather than the git-style page, because click intercepts before the bare-args path. This is a genuine usage error and an uncommon invocation; left as-is.

## Testing

- Updated `snapshot_test`, `default_command_group_test`, `test_plugin` to assert the git-style page + exit 2; added `help_formatter_test` cases for the COMMANDS section.
- Regenerated `libs/mngr/docs/commands/{config,plugin}.md`; `make_cli_docs.py --check` passes.
- Ran help/config/plugin/snapshot/default_command_group suites + ratchets + `ty` locally (green; 25 unrelated errors on this machine are an old-git `git init -b main` fixture issue that passes in CI).


libs/mngr/imbue/mngr/cli/help_formatter.py
```diff
@@ -357,6 +365,71 @@ def _write_git_style_help(
             output.write(f"           $ {example}\n\n")
 
 
+def _write_commands_section(
```
evgunter, line 368: why do all these `_write_*_section` commands write into a buffer instead of being pure functions that return a string and are assembled later

weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#20 from the same branch. This public repository is now a read-only mirror.

PR #2443, opened 2026-07-12
make tabcomplete get config keys from the schema

evgunter:
## Summary

`mngr config` tab-completion only completed keys already present in your config, so schema-discoverable keys under `commands.*`, `create_templates.*`, `plugins.*`, `pre_command_scripts.*` (and `providers.*`) were invisible until you'd already set something under them — backwards for a discovery feature.

Completion for the dict-container config namespaces is now schema-driven and sourced from enumerable key sources instead of a flattened config instance:

- **`commands.<cmd>.<param>`** — every option of a command, discoverable before anything is configured (e.g. `commands.destroy.<TAB>`); boolean/choice options also complete their value. Group subcommands use their `<group>_<sub>` bucket (e.g. `commands.snapshot_create.*`).
- **`commands.<group>.default_subcommand`** — for each `DefaultCommandGroup`, with that group's subcommand names as value choices.
- **`pre_command_scripts.<cmd>`** — for every `command_name` bucket, including the `<group>_<sub>` subcommands.
- **`plugins.<name>.*`** — from the `PluginConfig` schema for every installed plugin.
- **`providers.<name>.*` / `create_templates.<name>.*`** — all settable fields of a *configured* instance from the schema, not just the ones already written (same win `agent_types.*` already had). A template option also completes to the same values as the corresponding `mngr create --<opt>` (e.g. `create_templates.<name>.type` → agent type names, `.template` → template names, `.provider` → provider names).

### Bug fixed along the way

The old instance-dump completion emitted the **internal** `commands.<name>.defaults.<param>` and `create_templates.<name>.options.<param>` shapes. Those don't round-trip through `mngr config set` — the wrapper (`defaults`/`options`) is transparent to users (see `config/key_resolver.py`), so setting that key stores a parameter literally named `defaults`, which is silently ignored at runtime (with an "Unknown parameter" warning). Completion now emits the correct user-facing `commands.<name>.<param>` / `create_templates.<name>.<param>` keys.

### Why a command_name registry

The valid `commands.*` / `pre_command_scripts.*` bucket names are the `command_name` values passed at each call site, which are **not** derivable from the click tree: group subcommands are namespaced `<group>_<sub>` (`snapshot_create`, `git_pull`), while the `config` / `plugin` groups deliberately share one group-level bucket. So they're maintained as an authoritative registry (`cli/command_names.py::KNOWN_CONFIG_COMMAND_NAMES`) with a CI test (`command_names_test.py`) that statically asserts it stays in exact sync with the `command_name="..."` literals in the source — adding/renaming/removing a command forces an update.

Layering is preserved: the config-layer completion writer receives the registry and the `DefaultCommandGroup` default-subcommand mapping as data from the cli-layer caller (same pattern as `topic_names` / `installed_plugin_packages`).

## Testing

- New/updated unit tests in `config/test_completion.py` (schema discovery for each namespace, `<group>_<sub>` bucketing, `default_subcommand` choices minus aliases, template option value choices, and that no internal `.defaults.`/`.options.` keys leak) and `cli/command_names_test.py` (registry sync). All pass.
- Manually verified generated keys against a configured context, and verified round-trip: the new keys validate through `parse_config` while the old `.defaults.` shape is silently ineffective.
- `ruff`, `ty`, import-layer and getattr/inline-import/model_copy ratchets pass.
- Full local `just test-quick libs/mngr` run: the only failures/errors are pre-existing environment issues unrelated to this change (an older local `git` that predates `git init -b`, and tests that shell out to `sh -n -c` which executes rather than syntax-checks under dash). CI runs the suite on modern tooling.

### DIFF
--- a/libs/mngr/changelog/ev-tab-complete-schema.md
+++ b/libs/mngr/changelog/ev-tab-complete-schema.md
@@ -1,0 +1,11 @@
+Tab-completion of config keys (`mngr config set/get/unset <TAB>` and `-S <TAB>`) is now schema-driven for the dict-container namespaces instead of only completing keys already present in your config:
+
+- `commands.<command>.<param>` completes every option of the command before you have set anything under it (e.g. `commands.destroy.<TAB>`), with boolean/choice options also completing their value. Group subcommands use their `<group>_<sub>` bucket (e.g. `commands.snapshot_create.*`).
+
+- `commands.<group>.default_subcommand` completes for each group that supports a configurable default, offering that group's subcommand names as values.
+
+- `pre_command_scripts.<command>` completes for every command, including the `<group>_<sub>` subcommand buckets.
+
+- `plugins.<name>.*` completes from the plugin config schema for every installed plugin, and a configured `providers.<name>.*` / `create_templates.<name>.*` now offers all settable fields from the schema, not just the ones already written.
+
+Fixed: completion previously emitted the internal `commands.<name>.defaults.<param>` and `create_templates.<name>.options.<param>` shapes, which do not round-trip through `mngr config set` (the parameter would be read literally as `defaults`/`options` and ignored at runtime). Completion now emits the correct user-facing `commands.<name>.<param>` / `create_templates.<name>.<param>` keys.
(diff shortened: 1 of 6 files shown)


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#21 from the same branch. This public repository is now a read-only mirror.

PR #2447, opened 2026-07-13
Stop paramiko/pyinfra stdlib logs from flooding Sentry

hynek-urban:
(Addresses https://generally-intelligent-e3.sentry.io/issues/7602414344/. Description generated by Claude.)

The mngr latchkey forward daemon (and the minds backend) reverse-tunnel via paramiko. When a tunnel target goes offline, paramiko's transport thread logs the connection-reset failure at ERROR via stdlib logging. Sentry's default LoggingIntegration patches logging.Logger.callHandlers at the class level, so it captured each such record as a Sentry event regardless of the propagate=False we set when redirecting paramiko into loguru. Those events were also not rate-limited (the stdlib records carry no exc_info or loguru full_location fingerprint), so a handful of users produced tens of thousands of events.

setup_sentry now accepts ignored_loggers (fnmatch globs) and calls Sentry's ignore_logger for each. mngr exposes SENTRY_IGNORED_STDLIB_LOGGER_PATTERNS (paramiko, pyinfra) and both the latchkey daemon and minds backend pass it. Genuine failures still reach Sentry as typed exceptions logged via loguru.


joshalbrecht:
lgtm

PR #2448, opened 2026-07-13
Auto-restart the shared latchkey gateway when it dies

hynek-urban:
The locally running `mngr latchkey forward` now keeps monitoring the running `latchkey gateway` process and restarts it when it gets killed or when it crashes.

I thought that the Minds app itself should also monitor and restart `mngr latchkey forward` when appropriate but it turns out we already notice it's down by seeing that the snapshots in the discovery streams are too old. This PR just increases the related threshold from 35s to 180s to prevent misfirings of the "discovery is down" detection.

(This should hopefully mitigate the [issue one of our users reported](https://imbue-ai.slack.com/archives/C0B1JACQ9HT/p1783797893490489), at least to some extent.)


joshalbrecht:
lgtm

merging now, but pinging @gnguralnick to take a look too in case he sees a reason not to do this / was planning to do something different / related

PR #2450, opened 2026-07-13
minds: bring-your-own cloud accounts (AWS / GCP / Azure)

minhtrinh-imbue:
[Human-written]

UX flow:

Option in creation menu UI to configure your own AWS/GCP/Azure account
image

Clicking on "Configure," there is UI to view your onboarded cloud accounts and add new ones via key input, which will run `mngr prepare`
image
image

Accounts that have been successfully `mngr prepare`ed will be added to the list of compute providers
image
- Question: should we remove that AWS option (which does workspace setup via ambient environment aws CLI, unlike the mngr prepare path this PR has built.)

Also implemented/tested features: start/stop/restart/deletion of BYO cloud workspaces


apps/minds/imbue/minds/desktop_client/templates/pages/Create.jinja
```diff
@@ -1,4 +1,4 @@
-{#def git_url, branch, host_name, launch_modes, selected_launch_mode, ai_providers, selected_ai_provider, docker_runtimes, selected_docker_runtime, backup_providers, selected_backup_provider, backup_api_key_env, accounts, default_account_id, anthropic_api_key, error_message, region_options_by_launch_mode, region_selected_by_launch_mode, selected_preset="remote", start_advanced=False, color="#0b292b" #}
+{#def git_url, branch, host_name, launch_modes, selected_launch_mode, ai_providers, selected_ai_provider, docker_runtimes, selected_docker_runtime, backup_providers, selected_backup_provider, backup_api_key_env, accounts, default_account_id, anthropic_api_key, error_message, region_options_by_launch_mode, region_selected_by_launch_mode, cloud_accounts=[], instance_types_by_backend={}, default_instance_type_by_backend={}, byo_region_options={}, byo_region_selected={}, selected_preset="remote", start_advanced=False, color="#0b292b" #}
```
joshalbrecht: nit: this is really dumb and small, but can you change all of the "byo" to "byok" (for "bring your own key")

"byo" just on its own doesn't read correctly (and "byok" is a term that coding agents will be more familiar with)

joshalbrecht: (applies to all case-variants and the spelled out versions too, eg, BYO, byo, "bring your own", etc)

apps/minds/imbue/minds/desktop_client/templates/pages/Landing.jinja
```diff
@@ -36,6 +36,14 @@
...
     // authoritative state reaches the target or the action fails.
     var pendingMindActionByAgent = {};
+    // When the user last issued a Start/Stop (ms epoch). A host mid-transition
+    // can transiently drop out of the discovery snapshot (a stopping VM is
+    // briefly neither live-listable nor yet offline-reconstructed); within this
+    // grace window a rows-disappeared diff must NOT trigger the reload below,
+    // or the page reloads into the zero-workspace create form and the row
+    // "vanishes" until discovery catches up.
+    var lastMindActionAt = 0;
+    var MIND_ACTION_GRACE_MS = 180000;
```
joshalbrecht: this is being solved for at the wrong level -- it needs to happen on the backend (and probably should be already anyway?)

the reason is that otherwise if you navigate away and then back, the host will be gone (so this can't really just be fixed on the front end side)

apps/minds/imbue/minds/desktop_client/agent_creator.py
```diff
@@ -1876,9 +1971,37 @@ def _create_agent_background(
...
-                    log_queue.put(f"[minds] Ensuring AWS security group is ready in {region}...")
-                    run_mngr_aws_prepare(region, on_output=emit_log, parent_cg=self.root_concurrency_group)
+                    if cloud_account:
+                        log_queue.put(f"[minds] Preparing cloud account '{cloud_account}' in {region}...")
+                        run_mngr_aws_prepare(
+                            region,
+                            on_output=emit_log,
+                            provider_name=cloud_account,
+                            parent_cg=self.root_concurrency_group,
+                        )
```
joshalbrecht: this is kinda weird actually -- shouldn't we be running the "prepare" step much earlier, when we're adding the account?

that seems like the logical place for it to go -- not at workspace create time!

apps/minds/imbue/minds/desktop_client/api_models.py
```diff
@@ -249,6 +261,44 @@ class BugReportRequest(ApiRequestModel):
...
+    backend: str = Field(description="Cloud backend (e.g. 'aws')")
+    region: str = Field(description="The account's current default region")
+    identifier: str = Field(description="Masked credential hint (e.g. 'AKIA…F5X2'); never the secret")
+
+
+class CloudAccountPrepareResponse(FrozenModel):
+    """Result of registering + preparing a cloud account."""
+
+    account: CloudAccountSummary = Field(description="The registered account")
+
```
joshalbrecht: weird to have a type with only a single attribute

apps/minds/imbue/minds/desktop_client/api_models.py
```diff
@@ -249,6 +261,44 @@ class BugReportRequest(ApiRequestModel):
...
+    gcp_service_account_key_json: str | None = Field(
+        default=None, description="Full JSON contents of a GCP service-account key (required for gcp)"
+    )
+    azure_subscription_id: str | None = Field(default=None, description="Azure subscription id (required for azure)")
+    azure_tenant_id: str | None = Field(default=None, description="Entra tenant id (required for azure)")
+    azure_client_id: str | None = Field(default=None, description="Service-principal client id (required for azure)")
+    azure_client_secret: str | None = Field(
+        default=None, description="Service-principal client secret (required for azure)"
+    )
+
```
joshalbrecht, line 293: this feels smelly -- it's weird to have to splat out the different value names for each provider

it feels like you're looking for something more like a discriminated union here?  (discriminate based on backend)

apps/minds/imbue/minds/desktop_client/app.py
```diff
@@ -1090,7 +1091,6 @@ def _build_region_form_context(
     for launch_mode, provider_key in (
         (LaunchMode.IMBUE_CLOUD, IMBUE_CLOUD_PROVIDER_KEY),
         (LaunchMode.VULTR, VULTR_PROVIDER_KEY),
-        (LaunchMode.AWS, AWS_PROVIDER_KEY),
```
joshalbrecht, line 1169: (not strictly related to this block, it just reminded me): as a follow-up PR, we should convert the vultr and ovh and modal and other cloud providers to work the same way as aws/gcp/azure do now (ie, require that they be added as a provider first)

this will help keep things sane

apps/minds/imbue/minds/desktop_client/workspace_lifecycle.py
```diff
@@ -24,8 +24,14 @@
...
-# resolves before returning the outcome.
-_LIFECYCLE_TIMEOUT_SECONDS: Final[float] = 300.0
+# resolves before returning the outcome. Sized for the slowest legitimate case:
+# a cloud VM's FIRST stop mirrors the entire host_dir to the provider's state
+# store before deallocating (observed ~10 minutes on Azure after a fresh
+# workspace build; later stops sync deltas and take ~1-2 min). A timeout here
+# is reported to the UI as a stop *failure* even though the underlying stop
+# keeps running -- so a too-small cap manufactures false failures for exactly
+# the flows that are working.
+_LIFECYCLE_TIMEOUT_SECONDS: Final[float] = 1200.0
```
joshalbrecht: there's at least 2 things going wrong here:

1. this constant appears duplicated in the codebase
2. there's no way this should be a *default* for all mngr commands

we should be specifically passing in this larger timeout for the ~1 place that needs it

finally, it's quite suspicious that anything would ever take 20 minutes--we're really not dealing with that much data... what exactly was taking so long here?

joshalbrecht:
for all of this--I wonder if this might be a good case for a feature flag? then we could merge this (without users seeing it unless they opted in), and we could iterate a little more quickly without worrying about people starting to use it, backwards compatibility, etc quite as much...

joshalbrecht:
also as we discussed--we'll need a follow-up PR to deal with synchronizing these provider configs to imbue_cloud (and either associating them with an account, or with the install, ie, no account)

I'd like to pair with you on that PR (and it should go *after* you convert the modal and vultr clouds to be the same as aws/gcp/azure probably)

joshalbrecht:
besides the comments above, lgtm!

Nice work overall :)

apps/minds/imbue/minds/desktop_client/app.py
MT-GoCode: got it

apps/minds/imbue/minds/desktop_client/agent_creator.py
MT-GoCode: yes this is redundant, removing

apps/minds/imbue/minds/desktop_client/templates/pages/Create.jinja
MT-GoCode: fixed

apps/minds/imbue/minds/desktop_client/api_models.py
MT-GoCode: fixed

apps/minds/imbue/minds/desktop_client/workspace_lifecycle.py
MT-GoCode: about 20 minutes: a cloud VM's first stop mirrors the entire host_dir to the provider's state store before deallocating. I observed ~12 min on Azure after a fresh build. 1200s is 2× headroom so a working slow-stop isn't reported as a false failure -- this happened in testing.

MT-GoCode:
> for all of this--I wonder if this might be a good case for a feature flag? then we could merge this (without users seeing it unless they opted in), and we could iterate a little more quickly without worrying about people starting to use it, backwards compatibility, etc quite as much...

I've added `FEATURE_FLAG_BRING_YOUR_OWN_CLOUDS`

apps/minds/imbue/minds/desktop_client/api_models.py
MT-GoCode: i made a discriminated union

apps/minds/imbue/minds/desktop_client/workspace_lifecycle.py
joshalbrecht: 1. did you end up fixing this to default only to the command that is actually slow?

2. mirroring the data to the host state store on azure seems... kinda busted?  why do we actually have to do that at all? Don't they have persistent disks? Couldn't we have the state store loaded on some kind of persistent disk instead?  (no need to change right now, but might need to be a follow up before azure is really usable)

minhtrinh-imbue: 1. yes
2. I believe this is pre-existing mngr_vps behavior, which lives in libs/mngr_vps/imbue/mngr_vps/instance_offline.py. This just triggers it -- what would you like done?

weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#22 from the same branch. This public repository is now a read-only mirror.

PR #2451, opened 2026-07-13
minds-eval: eval harness app (mngr_minds_eval)

minhtrinh-imbue:
high-level description of evaluation harness architecture: 

On mngr, there exists a CLI minds-evals that takes input of a mngr branch, default-workspace-template branch, and a list of personas with accompanying first prompts. It will spin up a docker container, clone mngr at that branch's head, run minds from that, then start up one modal-backed container per persona using the default-workspace-template branch specified. 

On default-workspace-template, there is a script that is invoked at workspace boot and takes a test_case_config.json that specifies a persona, first prompt, and how many conversation turns to simulate. If this file does not exist, the script exits immediately (which will be the case for normal workspace boots). Otherwise, this script will monitor the chat with the agent and send in the first prompt when the agent waits for user input. The next time the agent waits for user input, the script sends in second prompt, third prompt, so on so forth until the conversation turns count specified in the json is achieved. Right now, all prompts after the first are just hard-wired to be "OKAY" but could use claude to determine a prompt based on current transcript & persona.

The minds-evals CLI populates test_case_config.json in individual default-workspace-template clones in order to set up each persona.

The minds-evals CLI spins up minds inside of a docker. I made this design choice because it made branch-specific testing much easier.

Every turn of the conversation, the codebase and transcript is snapshotted to AWS S3 for future evaluation agents.

A lightweight evaluation agent is included.


joshalbrecht:
it's actually really nice that everything is contained in its own app :)

the overall quality is a little bit low (because there's a ton of duplication with this functionality vs how minds normally works anyway), but I kinda don't mind that too much for running evals (and because it will serve as a reasonable example for how to do this once we have slightly better primitives as well)

maybe my only push would be to switch from S3 -> R2 (since vault dev tokens already define sufficient authorization for all of our developers to use R2, so, no need to go integrating with AWS more tightly, if that makes sense?)

it should be still accessible via the same API and routes--it's just that the auth should end up being simpler

minhtrinh-imbue:
Superseded by #2499 (which carries this work forward, incl. the S3 -> R2 switch).

PR #2453, opened 2026-07-13
Reframe the titlebar help entry point around bug reporting

gdenisov:
## What

Reframes the vague "Get help" titlebar entry point around **bug reporting**, per design direction (Figma bug asset). The change is icon + copy only — the underlying flow is untouched.

| Surface | Before | After |
|---|---|---|
| Titlebar trigger icon | `help-circle` (question mark) | `bug` glyph |
| Tooltip + `aria-label` | "Get help" | "Ran into a bug?" |
| Modal title (`<h1>` + page title) | "Get help" | "Ran into a bug?" |
| Modal intro | "Seems like you're running into a problem — we can help." | "We can try to fix it, or send it to us." |

The two options inside the modal — **have an agent fix it** and **report the bug to Imbue** — and the default behavior (agent-fix is the default in a live workspace) are unchanged. The DOM id cluster (`help-toggle`, `help-*`) is intentionally left as-is so the `chrome.js` wiring is untouched.

## Demo
Screenshot 2026-07-13 at 6 48 51 PM
<br>
Screenshot 2026-07-13 at 11 49 23 PM

## Changes

- Added a `bug` glyph to `_ICONS_16` in `templates.py` (pulled from the Figma bug node, matching the set's filled style) and listed it in the dev styleguide inventory.
- Removed the now-orphaned `help-circle` glyph (the titlebar was its only consumer) so the icon catalog stays a gap-free source of truth.
- Swapped the trigger icon + tooltip/aria-label in `Chrome.jinja`.
- Retitled the modal and reworked its intro copy in `Help.jinja`; the title uses past tense ("Ran into a bug?") and the intro no longer restates the title.
- Updated the tooltip/aria-label assertions in `test_desktop_client.py`.
- Added the `apps/minds` changelog entry.

## Verification

- `ruff check` + `ruff format`: clean.
- Ratchets: 57 passed.
- Full `desktop_client` subtree: 1388 passed, 0 failed.
- Rendered the `bug` glyph through `Icon16` to confirm shape/weight.


joshalbrecht:
lgtm

PR #2454, opened 2026-07-13
minds: drop the restart veto that stranded offline hosts on the loader

gnguralnick:
## Summary

Diagnosed from Sentry bug report `df103b8f` (staging 0.3.6): a lima workspace whose VM was offline since before the app started stayed on "Loading workspace" forever, and docker workspaces in the same state only cold-booted after an extra STUCK-triggered reprobe round.

Root cause: the restart endpoint skipped any `auto_dispatched` restart while the health tracker read HEALTHY -- but the tracker reports default-HEALTHY for workspaces it has never probed (a host offline since before the process started only produces UNRESOLVED backend failures, which never enroll a probe suspect; a probe success also *pops* the record, so "confirmed recovered" and "never observed" were indistinguishable). The landing page routes stopped minds to the recovery page with `intent=restart` precisely *because* of that default-HEALTHY reading, so the same tracker state both forced the recovery entry and silently killed its dispatch.

Fix: remove the veto and the `auto_dispatched` request marker end-to-end. The self-recovery race the veto guarded against is absorbed by `mngr start` itself: the auto-dispatched host tier skips the stop step (`host_already_stopped`), and `mngr start` only targets STOPPED agents and starts the host idempotently, so a restart dispatched against a live workspace degrades to a no-op. Manual restarts were never covered by the veto and are unchanged.

## Test plan

- New `test_restart_dispatches_for_never_probed_workspace` pins the regression: a host-scope dispatch for a workspace with no tracker record proceeds to a real DONE restart operation instead of being vetoed.
- `pytest apps/minds/imbue/minds/desktop_client/{api_v1,templates,workspace_recovery,system_interface_health}_test.py`: 347 passed.
- Full local `just test-quick apps/minds` (minus docker/tmux/acceptance/release): 1836 passed; 15 failures + 1 error are environmental (missing restic binary, known webdav TMPDIR issue, ConcurrencyGroup teardown flake) and untouched by this diff -- offload CI covers them.


joshalbrecht:
lgtm

PR #2457, opened 2026-07-14
Fix host-listing crash on Apple Silicon M4/M5 (psutil cpu_freq)

minhtrinh-imbue:
## Problem

On newer Apple Silicon Macs (M4/M5), `mngr list` -- and everything built on it (`mngr usage`, `mngr donate`) -- logs `ERROR: Error processing host host-...` for the local host, and `mngr donate` then wrongly reports `No Claude usage data ... skipping` (because the failing host is where local Claude agents' usage lives, so no usage events get collected).

## Root cause

`LocalProviderInstance.get_host_resources` reads CPU frequency via `psutil.cpu_freq()`, guarded only by `hasattr(psutil, "cpu_freq")`. On Apple Silicon, psutil reads frequency through the IOKit `pmgr` entry (not the `hw.cpufrequency` sysctl, which Apple removed). Its `has_cpu_freq()` probe only checks that the `pmgr` entry *exists* -- but on M4/M5 Apple changed the private `voltage-states{1,5}-sram` data format psutil parses, so:

- `hasattr(psutil, "cpu_freq")` -> `True` (the probe passes)
- `psutil.cpu_freq()` -> raises (surfaces as `SystemError: <built-in function cpu_freq> returned a result with an exception set`, wrapping psutil's `RuntimeError: invalid CPU frequency data`)

The unguarded call propagated out of host detail collection, so `list_agents` logged "Error processing host" and never emitted that host's agents. This is an unresolved psutil issue ([#2642](https://github.com/giampaolo/psutil/issues/2642) documents the M4 format change; [#2382](https://github.com/giampaolo/psutil/issues/2382) tracks the arm64 read raising).

## Fix

Read the frequency through a new `read_cpu_freq_ghz(read_cpu_freq)` helper that treats any of the documented psutil failure modes (`NotImplementedError`, `OSError`, `RuntimeError`, `SystemError`) as "no reading" and returns `None`. CPU frequency is an optional informational field (callers already handle `frequency_ghz=None`), so this degrades gracefully -- exactly like the `disk_usage("/")` read directly below it already does. The reader is injected so the failure path is unit-testable without patching psutil.

## Verification

- Reproduced the original crash on an M5 Pro (macOS 26.5, psutil 7.2.2): `psutil.cpu_freq()` raises `SystemError` while `has_cpu_freq()` returns `True`.
- After the fix, `read_cpu_freq_ghz(psutil.cpu_freq)` returns `None` and `mngr list` cleanly lists the local host's agents instead of erroring.
- Added unit tests covering the MHz->GHz conversion, the `None` return, and each raising failure mode.
- `instance_test.py` (52), mngr ratchets (57), `ty check`, and `ruff` all pass locally.


boweiliu:
This probably needs to be pulled into #2417

minhtrinh-imbue:
Superseded: folding this Apple Silicon cpu_freq fix into #2417 (add-mngr-donate) instead of landing it separately against main, per request.

PR #2458, opened 2026-07-14
Fix mngr forward --use-http2 TLS teardown noise and 30s hang

### LINKED ISSUE #2455: mngr forward (--use-http2): abandoned TLS connections log 'Unhandled exception in client_connected_cb / TimeoutError: SSL shutdown timed out' after a 30s hang
## Summary

Since `mngr forward` gained `--use-http2` (TLS termination), an abandoned client connection can leave a per-connection task stuck for 30 seconds and then dump an "Unhandled exception" traceback into the minds app's stderr. It is cosmetic — the connection being torn down is already dead, nothing is dropped — but it is alarming log noise, and it holds a connection task open for the full asyncio SSL-shutdown timeout.

Observed twice in one minds dev session (staging env, local Docker workspaces).

## Traceback

```
mngr forward stderr: Unhandled exception in client_connected_cb
mngr forward stderr: transport: <asyncio.sslproto._SSLProtocolTransport object at 0x...>
mngr forward stderr: Traceback (most recent call last):
mngr forward stderr:   File ".../hypercorn/asyncio/run.py", ...
mngr forward stderr:     await TCPServer(app, loop, config, context, lifespan_state, reader, writer)
mngr forward stderr:   File ".../hypercorn/asyncio/tcp_server.py", ...
mngr forward stderr:     await self._close()
mngr forward stderr:   File ".../hypercorn/asyncio/tcp_server.py", ...
mngr forward stderr:     await self.writer.wait_closed()
mngr forward stderr:   File ".../python3.12/asyncio/streams.py", ...
mngr forward stderr:     await self._protocol._get_close_waiter(self)
mngr forward stderr: TimeoutError: SSL shutdown timed out
```

## Cause

Two facts combine:

1. **asyncio waits 30 s for the TLS close handshake.** `TCPServer._close()` calls `await self.writer.wait_closed()`, which waits for `close_notify` to be exchanged. If the peer walks away without answering, asyncio raises `TimeoutError` after `asyncio.constants.SSL_SHUTDOWN_TIMEOUT` (**30.0 s** — verified in this interpreter's `asyncio/constants.py`).

2. **hypercorn does not catch `TimeoutError` on that path.** `hypercorn/asyncio/tcp_server.py::TCPServer._close` catches `BrokenPipeError, ConnectionAbortedError, ConnectionResetError, RuntimeError, asyncio.CancelledError` — but not `TimeoutError`. So it escapes the per-connection coroutine, which asyncio's stream server invoked as `client_connected_cb`, and asyncio's default handler logs it as unhandled.

This path only exists with `--use-http2`: `mngr forward` builds an `SSLContext` (`InMemoryTLSConfig`, ALPN `h2`/`http/1.1`) only in that mode (`libs/mngr_forward/imbue/mngr_forward/cli.py::_build_hypercorn_config`). In plain-HTTP mode there is no TLS shutdown to wait on, so this cannot fire.

**What made the peer walk away.** Both occurrences sit immediately after a workspace creation, while the container is still booting:

```
Transferring git repository... / Creating agent state... / Starting agent system-services ...
{"event": "created", "agent_id": "agent-f798ab31...", ...}
WARNING: [paramiko] Secsh channel N open FAILED: Connection refused: Connect failed
WARNING: Failed to open SSH channel to 127.0.0.1:8000: ChannelException(2, 'Connect failed')   (xN)
Unhandled exception in client_connected_cb ... TimeoutError: SSL shutdown timed out
```

The UI polls the new workspace before supervisord has started `system_interface` inside the container, so the forward's SSH channels to `127.0.0.1:8000` are refused; it answers 502 and closes those client connections. By then the Electron renderer has navigated/reloaded and never completes the TLS shutdown, so the close waiter times out 30 s later.

## Reproduction (rough)

1. Activate a minds env and `just minds-start` (so `mngr forward --use-http2` is running).
2. Create a local Docker workspace and let the app open it **immediately**, while the container is still booting — the "Connection refused to 127.0.0.1:8000" window is the trigger.
3. Reload / navigate the workspace view during that window so the browser abandons its TLS connections.
4. Wait ~30 s and watch the app's stderr for `Unhandled exception in client_connected_cb ... TimeoutError: SSL shutdown timed out`.

A more deterministic harness: point `mngr forward --use-http2` at a port with nothing listening, open a TLS connection, send a request, then hard-close the client socket without a TLS close_notify (e.g. `SO_LINGER` 0 / kill the client), and wait out the 30 s timeout.

## Potential fixes

- **Shorten the SSL shutdown timeout** so an unresponsive peer is dropped in ~1 s instead of 30. Note hypercorn passes `ssl_handshake_timeout` to `asyncio.start_server` but **not** `ssl_shutdown_timeout` (grep of `hypercorn/asyncio/run.py`: `ssl_shutdown_timeout` never appears), so this cannot be done via `Config` alone today — it would need an upstream change, or serving via our own `asyncio.start_server(..., ssl_shutdown_timeout=...)` instead of `hypercorn_serve`.
- **Swallow the `TimeoutError`** at the boundary. Cleanest local option: an `asyncio` exception handler on the loop that drops this specific "already-closing connection" case, or a thin wrapper around hypercorn's connection callback. Upstream-correct option: add `TimeoutError` to the `except` tuple in `TCPServer._close` (a small hypercorn PR — it is plainly the same "already gone" class as the errors it does catch).
- **Reduce the trigger**: don't let the UI hammer the forward before the container's `system_interface` is listening (the 502 storm is what closes the connections in the first place). Fixes the noise at the source but not the underlying 30 s hang for any other abandoned connection.

Recommendation: bound the shutdown wait (fix 1 or 2) so no connection task can ever sit for 30 s, and treat the 502 storm separately if it is worth it.

## Pointers

- `libs/mngr_forward/imbue/mngr_forward/cli.py` — `_build_hypercorn_config` (the `--use-http2` / `InMemoryTLSConfig` branch), `_serve_forward_app` (`asyncio.run(hypercorn_serve(app, config))`).
- `libs/mngr_forward/imbue/mngr_forward/tls.py` — `build_server_ssl_context`, `ALPN_PROTOCOLS`.
- Upstream: `hypercorn/asyncio/tcp_server.py::TCPServer._close`, `hypercorn/asyncio/run.py` (`asyncio.start_server(..., ssl=..., ssl_handshake_timeout=...)`).

## Severity

Low. Cosmetic log noise plus a 30 s idle connection task per abandoned TLS connection; no functional impact observed (the affected connections were already dead). Worth fixing because the traceback looks like a crash and will send people down the wrong path when debugging real forward problems — it already did in this session.

joshalbrecht:
Fixes #2455.

## Problem

With `--use-http2` (TLS termination), a client that abandons a connection without answering the TLS `close_notify` left the per-connection task stuck for asyncio's full 30s SSL shutdown timeout, then dumped an alarming `Unhandled exception in client_connected_cb / TimeoutError: SSL shutdown timed out` traceback to stderr (surfacing in the minds app's logs).

Two upstream gaps combine: hypercorn's `TCPServer._close` catches the other already-gone connection errors but not `TimeoutError`, and hypercorn's `asyncio.start_server` call does not pass `ssl_shutdown_timeout`, so the 30s stdlib default is not configurable via `Config`.

## Fix

Both adjustments are made where `mngr forward` owns the serve loop (`_serve_forward_app`):

- The loop is created via a small `asyncio.SelectorEventLoop` subclass that rewrites the stdlib's `ssl_shutdown_timeout=None` sentinel to 5s in `_make_ssl_transport`, so an unresponsive peer is force-closed in seconds instead of 30. The override is defensive: it only rewrites the sentinel when the keyword is present, so a future CPython signature change degrades to the stdlib default instead of breaking accepts.

- A loop exception handler drops the benign TLS teardown errors of already-dead connections (the SSL-shutdown `TimeoutError` and `ssl.SSLError` handshake failures -- the latter matching what hypercorn's own runner does when it owns the loop) at debug level, and defers everything else to the default handler.

## Testing

- Unit tests for the exception handler (suppression is narrow: unrelated errors and other `TimeoutError`s still reach the default handler).
- An end-to-end test that reproduces the issue: a TLS client handshakes and goes silent; the connection must be torn down within the bounded wait and the escaping `TimeoutError` suppressed. A negative control (same scenario without the handler) reproduced the exact traceback from the issue.
- `just test-quick libs/mngr_forward`: 274 passed, 1 skipped (pre-existing default-port skip); repo meta ratchets + type check pass.


joshalbrecht:
lgtm

libs/mngr_forward/imbue/mngr_forward/cli.py
```diff
@@ -416,16 +419,96 @@ def _build_hypercorn_config(listen_socket: socket.socket, use_http2: bool) -> Co
...
+
+    Covers ``ssl.SSLError`` (handshake failures -- hypercorn's own runner drops
+    these, and we replace its exception handler by running the loop ourselves)
+    and the ``TimeoutError`` asyncio raises when a peer abandons a connection
+    without answering ``close_notify``. hypercorn's ``TCPServer._close``
+    catches the other already-gone connection errors but not that
+    ``TimeoutError``, so it would otherwise escape ``client_connected_cb`` and
+    be logged as an alarming unhandled-exception traceback.
+    """
+    is_handshake_failure = isinstance(exception, ssl.SSLError)
```
gnguralnick, line 465: does this mean we'll be swallowing all SSL Errors? is that safe to do

joshalbrecht: I think so, given that hypercorn's own runner already drops them?

PR #2460, opened 2026-07-14
donate: forward CLAUDE_CODE_OAUTH_TOKEN to the headless agent via --pass-env

boweiliu:
## What

Stacks on top of #2417 (`add-mngr-donate`). Fixes the donation agent failing with "Not logged in" even when a valid `CLAUDE_CODE_OAUTH_TOKEN` was available.

## Why

The token was placed in the `mngr create` subprocess env by `_donation_agent_env()`, but never forwarded into the agent's sanitized environment — `mngr create` only passes through vars named via `--pass-env` (see `resolve_env_vars` in `libs/mngr/imbue/mngr/cli/env_utils.py`, which iterates only `pass_env_var_names` from `os.environ`). Without that flag, the token was dropped at the process boundary and the headless claude agent never saw it, so it failed "Not logged in" immediately on launch.

In other words: donate placed the token in the *parent* env, but the agent runs in a *sanitized* env, and the bridge between them is `--pass-env`.

## Fix

`build_create_argv` now passes `--pass-env CLAUDE_CODE_OAUTH_TOKEN`, so the long-lived OAuth token (from the keychain stash or the caller's env) reaches the headless claude agent and it authenticates. This is independent of *where* the token comes from — it just guarantees the var, if present in the parent env, reaches the agent.

Safe unconditionally: `resolve_env_vars` skips names not present in `os.environ`, so `--pass-env CLAUDE_CODE_OAUTH_TOKEN` is a no-op when the var isn't set.

## Verification

- Donate unit tests: 16 passed. Ratchet tests: 57 passed.
- E2e in a tmux: pre-fix the agent died with "Not logged in"; post-fix the agent authenticated and ran a full document-review tick to completion (10 papers leased, reviewed, all 10 submitted with HTTP 201, `recorded`). The perpetual run in tmux `donate_perpetual` has completed multiple full ticks with this fix applied.

## Files

- `libs/mngr_donate/imbue/mngr_donate/donate.py` — `build_create_argv` adds `--pass-env CLAUDE_CODE_OAUTH_TOKEN` (+ explanatory comment).
- `libs/mngr_donate/imbue/mngr_donate/donate_test.py` — `test_build_create_argv_launches_a_headless_agent_that_skips_permissions` now asserts the full argv (including `--pass-env CLAUDE_CODE_OAUTH_TOKEN`) instead of only the first 10 elements.
- `libs/mngr_donate/changelog/add-mngr-donate.md` — changelog bullet.


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#24 from the same branch. This public repository is now a read-only mirror.

PR #2461, opened 2026-07-14
donate: source agent from skill checkout so donate works from a non-git cwd

boweiliu:
## Problem

`mngr donate` raised **`Not inside a git repository. Either run from within a git repo, or specify --from to set the source explicitly.`** when run from a non-git directory — the common case for users who install mngr from a wheel/PR rather than a monorepo checkout (e.g. someone who installed from the `mngr/donate-auth-fix` branch via `uv tool install`).

## Root cause

`build_create_argv` built the `mngr create` invocation with **no `--from`**. With no `--from`, `mngr create` falls back to sourcing the agent from the current directory (`_resolve_source_location` -> `find_git_worktree_root(cwd)`), which returns `None` outside a git repo and raises `UserInputError`. The donate README even stated the assumption ("Run it from a trusted git repo, since the created agent is sourced from the current directory") — that assumption is the bug for installed-from-source users.

## Fix

`build_create_argv` now passes `--from <skill_dir>`. The skill checkout (`prepare_skill_dir` clones the skill repo into `~/.mngr/donate-skills/<skill>/`) is **always a git repo**, so the agent sources from it regardless of the caller's cwd. The donation agent only talks to the skill's coordination server (its own message says so), so the skill checkout is the correct source anyway.

`_clear_stale_worktree` now runs `git -C <skill_dir>` (where the agent worktree is now rooted) instead of `git` in the caller's cwd, so stale-worktree cleanup still works when the caller's cwd isn't a git repo.

## Verification

- Unit tests: 73 passed (`libs/mngr_donate`).
- Repro: a script exercising the real `build_create_argv` from a non-git cwd — pre-fix hit "Not inside a git repository", post-fix passes source resolution and reaches the auth stage.
- E2e: ran a full `mngr donate` tick from `/tmp/repro_not_a_git` (a non-git dir) with a real token — exit 0, the agent launched, leased 10 papers, and reviewed them. Confirmed the agent worktree was rooted in the skill cache and that `_clear_stale_worktree` cleans up the leftover `mngr/<agent>` branch against the skill dir.

## Stacking

Stacked on top of #2460 (`mngr/donate-auth-fix`, the `--pass-env` auth fix), which is itself stacked on #2417 (the base `mngr_donate` plugin). Merge order: 2417 -> 2460 -> this PR.


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#25 from the same branch. This public repository is now a read-only mirror.

PR #2463, opened 2026-07-14
Fix the root causes of two flaky CI tests (#2456)

### LINKED ISSUE #2456: Two flaky CI tests: claude-CLI shell-out trips the 10s offload timeout, and Modal snapshot test races sshd boot
## Summary

Two unrelated tests fail intermittently in CI. Both bit on an unrelated PR (#2435) purely by timing, and both have now been *stabilized* (retry / timeout headroom) so they stop blocking merges — but neither underlying cause has been fixed. This issue tracks the real fixes.

Stabilization commit: `888e5230be` on `mngr/add-dockview-profiles` (PR #2435).

---

## 1. `libs/mngr/imbue/mngr/cli/extras_test.py::test_extras_claude_plugin_subcommand`

**Symptom (offload / `test-offload`):**

```
py::test_extras_claude_plugin_subcommand.libs/mngr/imbue/mngr/cli/extras_test.py::test_extras_claude_plugin_subcommand:
Failed: Timeout (>10.0s) from pytest-timeout.
```

Passes locally every time (well under a second of actual work).

**Cause.** The test invokes the real `extras claude-plugin` subcommand, which calls `_claude_native_plugin_status()` → shells out to the **`claude` Node CLI** (`["claude", "plugin", "list", "--json"]`, `libs/mngr/imbue/mngr/cli/extras.py:244`). Node process startup dominates the runtime, and on a contended offload sandbox it crosses the global **10 s** `pytest-timeout`.

**This is a known class in this exact file.** Two siblings already carry documented fixes for the identical cause:
- `test_print_extras_status_runs_without_error` — fixed by *dependency-injecting* a fast status fn ("a Node process whose startup dominated this test's runtime and made it flaky under the 10s offload timeout (observed at 10.05s in CI; ~0.4s locally)").
- `test_extras_no_args_shows_status` — `@pytest.mark.timeout(30)` with the comment "can stall on a contended CI sandbox and trip the global 10s pytest-timeout even though it runs in well under a second locally".

**What was done (stopgap):** the same `@pytest.mark.timeout(30)` headroom, because the shell-out *is* what this test exercises end-to-end, so removing it would gut the test.

**Real fix options:**
- Give the `claude-plugin` command an injection seam (like `_print_extras_status` already has: `claude_native_plugin_status_fn`) so the unit test can drive it without spawning Node, and cover the real shell-out once in an acceptance/release test where a 10 s budget isn't the constraint.
- Or: make the whole `extras` CLI's Claude-status probe lazy / cached so no `extras` subcommand pays Node startup unless it actually needs plugin state.
- Or (weakest): raise the *global* offload timeout — rejected, since it hides genuinely slow tests.

The general smell: **any test that shells out to `claude` is a 10 s-timeout landmine.** A grep for other `["claude", ...]` shell-outs under test would be worthwhile.

---

## 2. `libs/mngr_modal/imbue/mngr_modal/test_snapshot.py::test_snapshot_create_then_list_on_modal`

**Symptom (offload / `test-offload-acceptance`):**

```
AssertionError: Create agent failed: Creating host fluorescent-stallion-of-current in modal ...
  Waiting for sshd to be ready...
  Traceback (most recent call last):
    File ".../paramiko/transport.py", line 2369, in _check_banner
      buf = self.packetizer.readline(timeout)
    File ".../paramiko/packet.py", line 395, in readline
      buf += self._read_timeout(timeout)
  ...  (SSH protocol banner read timed out)
```

**Cause.** The test boots a **real Modal host** and SSHes into it. Paramiko occasionally times out reading the SSH protocol banner while `sshd` is still coming up on the freshly-booted container — i.e. Modal-side boot timing, not a defect in the code under test. The test already has `@pytest.mark.timeout(400)`, so the *test* isn't out of time; the *SSH banner read* inside the create path is what gives up.

**What was done (stopgap):** `@pytest.mark.flaky` so offload retries it.

**Real fix options:**
- Harden the SSH-readiness wait in the create path: retry the banner read / reconnect on `SSHException("Error reading SSH protocol banner")` rather than treating the first failed connect as fatal. This is the correct fix and would help *every* Modal (and VPS) create, not just this test — a user hitting this gets "Create agent failed" for what is a transient boot race.
- Check whether the existing "Waiting for sshd to be ready..." poll actually gates on a *successful SSH handshake* vs. just a TCP connect; a TCP-accept-before-sshd-ready window would explain this exactly.

---

## Why this matters beyond the two tests

Both were stabilized rather than fixed, which is deliberate but has a cost: `@pytest.mark.flaky` means a real regression in the Modal create path can now hide behind a retry, and the 30 s timeout means a genuine 20 s slowdown in the extras CLI would go unnoticed. Worth fixing properly.

Item 2's root fix in particular is **user-facing**: the same banner-read race presumably surfaces as a spurious "Create agent failed" for real `mngr create @.modal` users.

## Pointers

- `libs/mngr/imbue/mngr/cli/extras.py` — `_claude_native_plugin_status` (line ~244, the `claude plugin list --json` shell-out), `_print_extras_status` (the existing injection seam to copy).
- `libs/mngr/imbue/mngr/cli/extras_test.py` — the three affected tests and their comments.
- `libs/mngr_modal/imbue/mngr_modal/test_snapshot.py` — `test_snapshot_create_then_list_on_modal`.
- The SSH-readiness wait used by the Modal create path ("Waiting for sshd to be ready...").

joshalbrecht:
Fixes #2456 by addressing the underlying causes of both flaky tests rather than (only) stabilizing them.

## 1. SSH banner-read race on freshly booted hosts (user-facing)

`test_snapshot_create_then_list_on_modal` flaked because a freshly booted Modal sandbox accepts TCP before sshd answers the SSH handshake; paramiko gives up with `Error reading SSH protocol banner`, and the first failed connect was fatal -- surfacing to real `mngr create @.modal` users as a spurious "Create agent failed".

`OuterHost._ensure_connected` now retries connects that fail with exactly that signature (a new `_is_transient_ssh_connect_error` predicate + tenacity decorator, mirroring the existing `_retry_on_transient_ssh_error` pattern). The first retry is immediate (the failed attempt already spent paramiko's banner timeout waiting), the second waits 3s. Refused/unreachable/auth/host-key failures are deliberately not retried, so genuinely-down hosts still fail fast (e.g. during `mngr list` host discovery). Because `_ensure_connected` is shared, this hardens every remote provider (Modal, VPS, Docker, lima, imbue_cloud), not just the flaky test's path.

Note: the existing `wait_for_sshd` gate was already doing a full SSH handshake (not just a TCP connect), so the issue's "TCP-accept-before-sshd-ready window" hypothesis for the *wait* was ruled out -- the race is between the successful readiness probe and the subsequent real connection.

## 2. Extras CLI tests spawning the real `claude` Node CLI

`test_extras_claude_plugin_subcommand` (and siblings) reached `_claude_native_plugin_status` through the real command path, paying Node startup on a contended offload sandbox and tripping the global 10s pytest-timeout.

A shared `write_stub_claude_cli` helper (`cli/testing.py`) + `stub_claude_on_path` fixture (`cli/conftest.py`) now replaces `claude` on PATH with an instant script, still exercising the real shell-out and JSON-parsing plumbing end to end. The stub reports every plugin installed so the install flow short-circuits before its interactive-terminal check -- making the tests deterministic with or without a TTY, and regardless of the machine's real claude/plugin state. The previously inline stub in `test_extras_claude_plugin_yes_flag` was refactored onto the shared helper. The real `claude` probe stays covered by a new acceptance test (`cli/test_extras.py`) with a Node-startup-sized 60s timeout.

Audited the rest of the codebase for other `claude` shell-outs reachable from unit tests (per the issue's suggestion): the extras path was the only one; the plugin catalog's `claude --version` availability signal is only executed by the interactive wizard, and its tests use `true`/`false`/`sh` commands.

## Also

- Marked the preexisting `test_worker_hubs_do_not_accumulate_across_polls` as `@pytest.mark.flaky` with a 30s per-test timeout: it tripped the suite-wide 10s timeout during this branch's full-suite validation run (passes in ~5s in isolation; its 30 executor spinups can cross 10s on a contended runner).

## Testing

- `PYTEST_MAX_DURATION_SECONDS=560 just test-quick "libs/mngr -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'"`: 4692 passed, 0 failed.
- New unit tests cover the retry (fail-once-then-succeed), the no-retry fast-fail for refused connections, and the transient-classifier predicate.


joshalbrecht:
lgtm after a few follow ups, doing now

PR #2465, opened 2026-07-14
Fix permissions and credentials syncing

hynek-urban:
There were two bugs in the logic that synced credentials to the remote Latchkey instance:

- On Linux, the sync ran in an infinite loop because reading caused the sync to trigger, which involved reading, triggering another round of sync, etc. (In effect, the credentials were synced every few seconds.)
- Fixing the above bug revealed another part: when permissions changed, we would only sync `permissions.json` but we also needed to sync `credentials.json.enc` because changing permissions may have changed the set of relevant credentials for the given agent / host.

This PR fixes both.


joshalbrecht:
lgtm

PR #2469, opened 2026-07-14
Add per-host SSH connect-failure backoff to SSHTunnelManager

hynek-urban:
This is another PR in response to the error flood in Sentry.

The [previous fix](https://github.com/imbue-ai/mngr/pull/2447) focused on suppressing the extensive error logging. This one deals with the underlying issue (why do we actually retry the SSH handshake so many times?), see [Bowei's notes](https://imbue-ai.slack.com/archives/C05RX4T4UHJ/p1783969483686209?thread_ts=1783941786.889869&cid=C05RX4T4UHJ) for context.

This is what happened:
- Bowei has a workspace that exists but its sshd doesn't run (because of OOM).
- The `mngr_latchkey` discovery handler keeps (re)trying to establish the SSH connection (which never succeeds because of the unresponsive sshd on the other end).
- So we're adding an exponential backoff (capped at 5 minutes) to at least reduce the cadence of attempts.

This piece of info from Claude is relevant:

> Because the schedule starts at 2s and only exceeds the 30s discovery cadence after the 5th consecutive failure, a slow-booting agent still gets ~5 real connect attempts in its first ~2.5 minutes before the retry rate decays.


hynek-urban:
Closing - we don't mind the high cadence of retries (on the contrary).

PR #2472, opened 2026-07-14
Attach an anonymous per-install user id to Sentry events

hynek-urban:
Persist a randomly generated user ID in `./<minds-dir>/anonymous_user_id` and attach it to Sentry events. The goal is to distinguish "rare events that happen across a large number of installs" from "frequent errors happening only to a single user".

This is an example Sentry event: https://generally-intelligent-e3.sentry.io/issues/7612167425/?project=4511609241862145


joshalbrecht:
lgtm!  thanks for adding this

PR #2473, opened 2026-07-14
Make held remote host lock safe across SSH reconnects

qi-imbue:
Implements `specs/host_lock_reconnect_safety.md` (design PR #2380), plus the idle-shutdown refinement raised by Josh in that review.

## Problem

A remote cooperative host lock is a `flock(2)` held by a remote shell over one SSH channel, so lock ownership is bound to connection liveness and nothing re-verifies it once held. If the connection drops mid-critical-section the flock silently releases, and the per-operation transient-retry machinery reconnects and keeps running -- now unlocked. Another actor can enter its critical section in the gap. No exception is raised: a silent mutual-exclusion violation.

## What this does

- **Acquisition counter.** A monotonic counter (`host_lock.generation`) is incremented under the flock on every acquire. A holder that reconnects while holding the lock re-acquires and checks the counter: if it advanced by exactly its own re-acquire (`V == token + 1`) the blip was uncontended and it transparently continues; otherwise it raises the new **`LockLostError`** (a `MngrError`, already isolated per-host by callers like the `mngr_mapreduce` launch loop).

- **Enforcement at every boundary:** the paramiko reconnect chokepoint (`_ensure_connected`), an independently-dead lock channel, a pre-check before each rsync (which uses a separate connection), and a single-exit **release backstop** that guarantees no critical section is reported successful while the lock was silently lost.

- **Preserve a live transport while locked:** a channel-level or read-timeout error on a still-active transport no longer forces a `disconnect()` (which would needlessly drop the lock); only a confirmed-dead transport reconnects-and-re-verifies.

- **In-host acquirers also increment (Josh's point):** the idle-shutdown watcher acquires the lock and increments the counter when it takes the host down, and the local-filesystem lock path increments on acquire -- so an in-host acquisition is visible to a reconnecting remote holder rather than a silent, unaddressed hazard.

## Notes / known exemptions

- The **release backstop** raises only when the counter actually advanced (an intervening acquisition), matching the spec's stated design principle -- an uncontended momentary loss is safe and does not fail the run. It reads the counter only when the lock channel is not live, so the happy path stays cheap.
- The **debug-only detached lock holder** (`MNGR_DEBUG_RETAIN_LOCK_FOR_FAILED_HOSTS_DURING_CREATE=1`) does not increment: it runs only after the critical section has already failed, with no live holder verifying against it, and embedding the increment in its nested `sh -c '...'` quoting is fragile. Documented as a safe exemption.

## Tests

New unit tests in `host_test.py`: token parsing, uncontended vs contended reconnect, fast-fail without blocking, independent channel-death re-verify, release backstop (cheap/uncontended/lost), preserve-vs-disconnect transport, and local-counter increment. Local run: host + outer_host + ratchet suites green (300+ tests), `ty` clean. Full suite runs here in CI (offload needs Modal creds not available locally).


joshalbrecht:
lgtm (except I'd like confirmation that, in fact, this is backwards compatible)

PR #2474, opened 2026-07-14
Fix right-edge clipping of minds tooltips and modal overlays under classic scrollbars

gnguralnick:
## Problem

On Macs with classic (always-visible) scrollbars, every minds tooltip was clipped on its right side (e.g. "Home" rendered as "Hom"), and the modal overlays (get help / inbox / workspace menu / sign-in) left an un-dimmed strip at the window's right edge.

## Root cause

The global `html { scrollbar-gutter: stable }` rule in app.css (added so content pages don't shift sideways when a classic scrollbar toggles) also applied to the edge-to-edge chrome/overlay surfaces. With classic scrollbars it reserves a 15px right gutter per document that nothing paints:

- the overlay host's initial containing block is 15px narrower than the overlay view, so tooltip bubbles (sized to the view) clip mid-label;
- hosted modal pages lose 15px in the overlay host plus another 15px inside the modal iframe's own document, so the dim backdrop stops 30px short of the window edge;
- the chrome titlebar laid out 15px narrow, shifting its right-aligned buttons inward.

Machines with default overlay scrollbars reserve no gutter, so most setups never showed it.

## Fix

A scoped `html.no-scrollbar-gutter { scrollbar-gutter: auto }` rule opts the edge-to-edge surfaces out; content pages keep the stable gutter.

To keep the opt-out from being enumerative (a future overlay surface would silently reintroduce the bug), the five overlay-view pages (overlay host, workspace menu, inbox, help, sign-in) wrap in a new `OverlaySurface` component — a Base variant that pins the opt-out and the shared transparent body. The chrome page stays on `Base` with an explicit new `html_class` argument, since it's edge-to-edge but not a transparent overlay.

## Verification

- Drove the dev Electron app over CDP before/after: pre-fix the tooltip painted 35 of 50px (transparent tail) and the modal dim ended 30px short; post-fix the tooltip, modal backdrop, and titlebar all paint to the window edge. Re-verified live after the OverlaySurface refactor (tooltip 50/50px, help backdrop = full viewport, sidebar renders).
- `just test-quick apps/minds`: 1853 passed, 0 failed, 3 skipped (includes regression tests asserting the six pages render the opt-out class and the landing page keeps the gutter).
- Changelog entry: `apps/minds/changelog/gabriel-green-hedgehog.md`.


joshalbrecht:
lgtm after conflicts

PR #2475, opened 2026-07-14
Replace tmux wait-for message confirmation with durable-evidence probes

joshalbrecht:
## Summary

`mngr message` could report success for a message that was never submitted. tmux **latches** a `wait-for -S` signal fired with no waiter, so a stale signal on the `mngr-submit-<session>` channel (left by any prompt submission no sender was waiting on, e.g. a task-notification dequeue) instantly "confirmed" the next send — whose EXIT trap then killed the still-pending backgrounded Enter keystroke. The message stayed in the input box while every layer (mngr, minds' system_interface, the chat UI) reported success. This bit in production: a minds chat message sat un-submitted in the agent's input box for 85+ minutes while the UI showed it as queued.

This PR replaces signal-based confirmation with durable-evidence polling:

- **New submit-and-confirm engine** (`tui_utils`): ONE sequential remote script per send — capture per-probe baselines, send Enter, poll evidence — with no background jobs and no traps, so Enter is provably sent before any confirmation check can run.
- **Claude**: content-verified confirmation — the send succeeds only when the message's own content appears in claude's durable transcript (enqueue or user records), checked in BOTH the native session JSONL and mngr's raw copy. Low-content messages fall back to any-new-record evidence.
- **Codex / Antigravity**: confirmation via their existing `active` marker advancing past a pre-Enter baseline (same timing as the old hook signal, minus the latch hazard).
- **Bounded, pane-gated Enter retries** (~3s/10s/30s): Enter is re-sent only while the pane still shows the pasted text; the text itself is never re-pasted, so retries cannot duplicate a message.
- **Slash commands** (leading `/`, all TUI agents): relaxed policy — same retries and a brief evidence poll, but exit 0 with a warning + structured agent event when nothing is observable.
- **Failure UX**: unconfirmed strict sends fail with per-probe baseline/final tokens, retry history, and the pane capture; `mngr create --message` / resume failures explain the agent is up and how to resend.
- **Compatibility both ways**: probes rely only on artifacts that already exist on agents created by older mngr versions (no reprovisioning), and new agents keep firing the now-unlistened `mngr-submit` hooks so older senders still confirm (removal FIXME'd for a future release).
- The bare `command` agent keeps its existing fire-and-forget behavior; opencode/pi-coding/headless paths are untouched.

Design doc: `blueprint/robust-message-delivery/plan-robust-message-delivery.md`.

## Testing

- New engine unit tests (script ordering invariants, no-background-jobs/no-trap, parsing, policies) plus real-tmux integration tests where evidence appears only when Enter actually executes.
- Plugin probe-builder tests, including running the real claude content probe against fixture transcripts (JSON-escaped newlines, repeated identical messages, pre-baseline decoys).
- New release tests: a message-delivery journey per TUI agent type (idle → busy/queued → rapid sequential → long buffer-pasted message → slash command, each delivered exactly once) and a concurrent-delivery test with two claude agents on one tmux server.


joshalbrecht:
roughly lgtm, but that's an awful lot of complexity...

PR #2476, opened 2026-07-14
Rework minds titlebar, navigation, and onboarding: breadcrumb + icon-tabs, centered popups, committed welcome flow

pseay-imbue:
Reworks the minds desktop app's titlebar, navigation, and onboarding to match the minds-options mockup (https://imbue-ai.github.io/mind-sketches/prototypes/minds-options/), converting the app's page detours into centered popups on the shared overlay surface.

## Videos

[Demo](https://www.loom.com/share/0ec4da2796234c4fa45b9162285ccfb8)

[FSM of Changes](https://www.loom.com/share/c7ad29ec079a452cb6aa00d6e923cf01)

## Titlebar and navigation

- The titlebar's left side is a single "(home icon) Minds" button that grows into a breadcrumb ("Minds / workspace-name") on workspace-scoped screens, with Workspace / Workspace Settings icon-tabs and the workspace switcher anchored to the workspace name. The global back/forward arrows and the centered page title are gone; only the sharing editor page and browser-mode fallbacks keep a contextual back arrow. Top-right: the requests inbox button (with a pending-request badge) and the report-a-bug button.
- The breadcrumb never flashes the raw agent id: names survive accent updates, and an unnamed workspace shows a brief ellipsis until discovery delivers the name.
- The switcher menu aligns its rows under the breadcrumb name, highlights the current workspace even on its settings/sharing screens, shows mind status with the mockup's icons (closed eye = stopped, alert triangle = unknown, nothing = running), and gives OTHER workspaces' rows an "open in new window" arrow (the current row and remote rows carry no action buttons; the per-row settings gear is gone).
- On accent-less screens the titlebar uses a subtle neutral grey so the inactive macOS traffic lights stay visible; traffic lights are also re-asserted on focus/blur so they never vanish.

## Centered popups instead of page detours

- Home screen bottom-left launchers open Minds Settings and Manage Accounts as centered modals; the full-page /settings and /accounts routes remain as browser-mode fallbacks.
- The Minds Settings modal shows the real app-level settings (Connectors, Local files, Workspaces, Error reporting, Master password) in a left-nav + panel layout with a fixed height (85vh).
- The sharing editor opens as a centered popup from Workspace Settings' "Manage sharing" buttons (fixed 70vh card; workspace/account names render as plain text so nothing can navigate the overlay iframe). It behaves exactly as the full page always has, except Update/Disable gray the editor out and refresh it in place instead of reloading (a reload would blank the popup). Opened through a new strictly-validated content-relay channel.
- Pending permission requests live in the inbox overlay (badged titlebar button, master/detail list, auto-open on new requests).
- The overlay host keeps a modal that reloads itself in place (fixes the sharing editor after saves and a pre-existing accounts-modal blanking bug).

## Onboarding as a committed choice

- Cold start on a functionally-empty app lands on the welcome splash; while signed out with no workspaces, the home route returns to the splash until the user signs in or explicitly chooses "Continue without an account" (/welcome/skip).
- The splash's Sign Up / Log In open the centered sign-in modal (each leading with its matching tab via a new mode parameter); the titlebar home button is hidden on the splash. The splash self-advances once an account appears (e.g. OAuth completed externally).
- The sign-in modal honors ?return_to=, intercepts "Forgot password?" so the auth flow can't strand the app inside the overlay, and the landing route distinguishes "no accounts" from "the account listing failed" so a transient subprocess failure can't bounce a just-signed-in user back to the splash.
- Startup restore drops saved workspace windows whose workspace no longer exists (previously restored into the "unresponsive" recovery page), and the cold-start route decision is logged.

## Misc

- The workspace-creation loading screen's rotating tips hold for 8 seconds instead of 3.
- The create form loses its titlebar back arrow (home is the escape).
- Home-screen launcher IPC now recognizes the shell view as an event source.

Merged up to origin/main (0 behind). Tests: 1471 desktop-client pytest + 16 node unit tests passing locally; per-project changelog entry included.


joshalbrecht:
per slack discussion: lgtm, but we're going to need to refactor the whole way that this app code is written--it's a bit more complicated than is really necessary

Issue #2478, opened 2026-07-14
A tmux wait-for client whose server dies mid-wait exits 0, so a dead server reads as a successful send


weishi-imbue:
Found while verifying #2416. Pre-existing, present on Linux, **not** introduced by that PR.

`tmux wait-for` exits **0 when the server is killed under it** — byte-identical to a real `-S` signal. Measured against an isolated socket:

| event | exit |
|---|---|
| normal `wait-for -S` | 0 |
| server `kill-server` mid-wait | **0** |
| no server *at start* | 1 |

The submission path treats "the waiter exited 0" as a confirmed submission. So a tmux server that dies **after** the waiter registers but **before** the deadline is reported as a successful send, though nothing was submitted. Only an outright failure *at start* is distinguishable.

`timeout(1)` passed the same 0 through, so this predates and survives #2416. A fix needs a positive confirmation that does not rely on the waiter's exit status alone (e.g. checking the server is still alive, or confirming via the acceptance marker).

weishi-imbue:
Fixed by #2475 (merged), which replaced signal-based confirmation with durable-evidence polling. Confirmation no longer depends on a `tmux wait-for` client's exit status — the script sends Enter, then polls durable evidence (transcript content / marker mtime) and confirms only on a token that changes against a pre-Enter baseline. A `wait-for` client exiting 0 (which is what a server death produced) is no longer anywhere in the confirmation path.

Reverified against `main` (post-#2475), driving the real `build_confirmation_command` output against an isolated tmux server:

```
#2478 -- a tmux server dying mid-run must NOT read as a successful send
  server killed mid-run, no evidence: exit=1 elapsed=19.6s -> no false success (MNGR_UNCONFIRMED)
```

The whole tmux server was killed ~1s in — after Enter was sent, during polling, with no evidence ever produced. The command correctly reported `MNGR_UNCONFIRMED` (exit 1) rather than a false success. (It runs to the end of the window before giving up, since the probe is a local marker file the dead server does not affect; correctness is fine, and fast-fail on tmux death is a separate latency concern, not a false-confirmation one.) Closing as completed.

PR #2481, opened 2026-07-14
Collapse the two tmux submission builders into one and delete the timeout shim

### LINKED ISSUE #2477: tmux wait-for signals latch, so a late hook signal can make the next submit falsely report success
Found while verifying #2416. Pre-existing, present on Linux, **not** macOS-specific and **not** introduced by that PR.

`send_enter_via_tmux_wait_for_hook` used to document that "signals wake exactly one waiter; a signal with none registered is lost". That is **false**. Measured on a fresh isolated socket:

- one unwaited `tmux wait-for -S <chan>` is **retained**, and satisfies the *next* waiter immediately
- a *second* unwaited `-S` **destroys** the latch
- a TERM-d `wait-for` client leaves a stale registration that swallows the next `-S`

The channel is `mngr-submit-<session>`, which is **stable across every submit in a session**. So a hook signal that arrives after its waiter has already timed out stays latched, and the *next* `mngr message` returns "submitted" before the TUI has accepted anything.

The stale docstring was corrected in #2416; the behavior was not. A fix likely means making the channel unique per submission (e.g. suffixing a nonce) so a late signal cannot satisfy a later waiter.

### LINKED ISSUE #2479: Marker-variant submission does not fail fast: it polls the full deadline when there is no tmux server
Found while verifying #2416. Pre-existing; correctness is fine, latency is not.

`_build_signal_only_command` exits immediately when there is no tmux server. `_build_signal_or_marker_command` (the Claude path, which also watches an acceptance marker) instead keeps polling the marker until the full deadline elapses — roughly **92s in production** — before exiting 1.

The result is correct, just slow: a user sending a message to an agent whose tmux server is gone waits the entire deadline for an error that is knowable at once. Matches the pre-PR structure, which also polled.

Fix: probe the server up front, or let the hook waiter's outright failure short-circuit the poll loop.

weishi-imbue:
Stacked on #2416 (`mngr/timeout-tac`) — review that first; this PR's base is that branch, not `main`.

Collapses the two tmux submission-command builders into one and deletes the `timeout(1)`/perl deadline shim that #2416 introduced. Every claim below was measured against a real tmux server on macOS 26.4.1, using the command the builder actually generates (not a hand-copy of it).

## Why touch this code at all

Two measured defects in the shipped submission path, both consequences of the same shape.

**1. The hook waiter outlives the command on every marker-confirmed submission.** The waiter ran inside a subshell (`( _timeout N tmux wait-for "$1" && echo 1 > "$sig" ) &`) and the EXIT trap killed `$(jobs -p)` — the *wrapper subshell*, not the `tmux` client inside it. The client was reparented and survived until its own `timeout` expired, up to the full deadline (~91s in production). This fires whenever the acceptance marker confirms first, which for Claude is the normal case: the marker is recorded the instant the message is queued, while the hook may only fire once the prompt reaches the model.

| shipped marker builder, marker confirms at 0.5s | leftover `tmux wait-for` clients |
|---|---|
| before | **1** (lingers up to 91s) |
| after | **0** |

**2. An unreachable tmux is not reported until the full deadline (#2479).** The marker builder polled its sentinel file and the marker to the end of the deadline, even though the hook waiter had already failed outright. Reading the waiter's *exit status* in the poll loop instead of a file it writes on success ends the wait at once. Measured with a `tmux` that exits non-zero, at the production 90s budget:

| | reports the failure after |
|---|---|
| shipped marker builder (Claude path) | **91.7s** |
| this PR | **0.0s** |

Fixes #2479. Pinned by a test at the production deadline.

## What the new command looks like

```bash
trap 'kill "$waiter" "$submit" 2>/dev/null' EXIT
base="$(<marker command>)"
tmux wait-for "$1" >/dev/null 2>&1 & waiter=$!          # direct job: the trap's kill must reach tmux itself
( sleep 0.1 && tmux send-keys -t "$2" Enter ) >/dev/null 2>&1 & submit=$!
end="$(( $(date +%s) + <deadline> ))"
while [ "$(date +%s)" -lt "$end" ]; do
    if ! kill -0 "$waiter" 2>/dev/null; then
        wait "$waiter" && exit 0      # exited on its own => the hook fired
        break                         # non-zero => tmux failed outright
    fi
    cur="$(<marker command>)"
    if [[ -n "$cur" && "$cur" > "$base" ]]; then exit 0; fi
    sleep 0.25
done
exit 1
```

Two properties keep this correct. A killed `tmux wait-for` exits **0**, indistinguishable from a real signal — so the waiter is killed *only* from the EXIT trap, after the outcome is already decided, which is what makes the status read inside the loop unambiguous. And the deadline is now a plain loop, so the command needs only bash builtins plus `tmux`, `date` and `sleep`: no `timeout`, no perl, no `mktemp`, no sentinel file.

The signal-only builder is gone: waiting on the hook alone is the same command with a marker probe that prints an empty token (`printf ''`), which can never satisfy `[[ -n "$cur" ]]`.

Behavior deltas: hook-confirmed submissions on the former signal-only path (codex, antigravity) now return up to one poll interval (0.25s) later, against a 15–90s budget. Exit status collapses to 0/1 rather than passing the waiter's status through; `_send_enter_and_wait_for_signal` only reads `result.success`, so nothing downstream can see it.

## Correcting the record on #2477

I set out believing this change fixed #2477's root cause. It does not, and I could not reproduce #2477 through the generated command at all. Two measurements say why:

- **`wait-for -S` wakes *every* client registered on the channel**, not one. So a leftover waiter does not starve a submission that has already registered its own waiter — both wake.
- **A *killed* `tmux wait-for` leaves its registration behind in the tmux server.** The process is gone, but the ghost swallows the next `-S` exactly as a live client would, so nothing latches. Since bounding the wait means killing the waiter, *no* design that bounds the wait avoids this — the old shim and this loop leave the same ghost.

That ghost is why #2477 does not fire in practice: a timed-out submission leaves exactly one ghost, which absorbs exactly the one late `-S` its own hook eventually fires, so nothing is left latched to falsely confirm the next submission. Driving the exact #2477 sequence (submission 1 times out → its hook fires late → submission 2 with no hook) through the real command:

| builder | submission 2 |
|---|---|
| shipped signal-only | correctly timed out |
| shipped marker builder | correctly timed out |
| this PR | correctly timed out |

So #2477 stays open, unchanged by this PR — the latch semantics it documents are real, but the path it describes is not reachable while every bounded waiter leaves a ghost. #2478 (a server that dies mid-wait exits 0 → false success) is likewise unchanged: this command still trusts the waiter's exit status.

## Verification

The generated command against a real tmux server, orphan count taken **before** the session is killed (the tmux server exits with its last session and would take a leaked client with it — an earlier draft of the test hid the leak exactly this way):

| outcome | exit | elapsed | orphaned `wait-for` |
|---|---|---|---|
| hook fires | 0 | 0.5s | 0 |
| deadline expires | 1 | 4.0s | 0 |
| marker advances | 0 | 0.6s | 0 |
| no tmux server | 1 | 0.3s | 0 |
| hook fires, Claude marker probe | 0 | 0.5s | 0 |

Real local Claude agent, end to end: `mngr message` → agent replied `PONG` (2.3s); `/clear` → confirmed on the hook alone, since it never moves the marker (2.7s); a follow-up message on the same reused channel → `PONG2` (2.9s). Zero `tmux wait-for` clients left on the user's tmux server afterwards.

New tests: zero leftover `wait-for` clients after a timed-out submission (verified to fail — "left 1 tmux wait-for client(s)" — if the waiter is put back inside a compound subshell), fail-fast on an unreachable tmux at the production deadline, a marker-advance confirmation against real tmux, and a submission that confirms on its hook after an earlier one on the same channel timed out.


weishi-imbue:
Closing as obsolete. #2475 (merged) replaced the entire `tmux wait-for` message-confirmation mechanism in `tui_utils` with durable-evidence polling — it deleted the exact code this PR set out to refactor: `_build_signal_only_command`, `_build_signal_or_marker_command`, `send_enter_via_tmux_wait_for_hook`, `_TIMEOUT_FUNCTION`, and `_PERL_TIMEOUT` are all gone from `main`. There are no longer two builders to collapse or a timeout shim to delete.

The two issues this PR incidentally addressed are both resolved on `main` by #2475's rewrite:
- The orphaned `tmux wait-for` client leak is gone because `main` uses no background waiter at all (single linear script, no traps).
- #2479 (marker path not failing fast on an unreachable tmux) is moot for the same reason.

Verified on `main`: #2477 and #2478 no longer reproduce (see those issues for the evidence). Nothing here is worth rebasing onto the new mechanism.

PR #2482, opened 2026-07-14
mngr transcript: drop the flat-field fallback, restoring the parts[]-only reader (MIND-113)

weishi-imbue:
> **DO NOT MERGE YET.** This is the scheduled cleanup of the back-compat shim added in #2392. It is safe to merge only once no old-emitter agent is still writing parts-less transcript lines. See the gate below. Target: ~4-6 weeks out (early-to-mid August 2026).

Follow-up to #2392 / [MIND-104](https://linear.app/imbue/issue/MIND-104). Tracked by [MIND-113](https://linear.app/imbue/issue/MIND-113).

## What this removes

`_format_event_human` renders an assistant turn from the ordered `parts[]`, with a fallback to the flat `text`/`tool_calls` when `parts[]` is empty. That fallback is a **back-compat shim**, not a permanent design: the current common-transcript emitter has always filled `parts[]` since the OTel alignment (`90ef7a979`, 2026-06-15). Only agents running an *older* emitter emit parts-less lines — and because an agent's emitter version is frozen at create/re-vendor time and `convert()` dedups by `event_id` (it never re-derives), those agents keep emitting parts-less lines for their whole life and their existing lines never heal.

So the shim only ever served agents that are aging out. This PR deletes it and returns the reader to the spec's single code path (`specs/common-transcript-standard/spec.md`, Compatibility).

The `without_parts` test is **inverted, not deleted** — it now asserts a parts-less record renders `(no content)`, pinning the single-path design so a fallback cannot silently creep back.

## Merge gate (run this first)

Merging while old-emitter agents are still live would make their assistant turns render `(no content)`. Before merging, confirm no parts-less assistant lines have been written recently:

```bash
python3 - <<'PY'
import glob, json, os
recent_cut = "2026-08-01T00:00:00Z"   # bump to ~1 week before merge day
bad = 0
for f in glob.glob(os.path.expanduser("~/.mngr/**/claude/common_transcript/events.jsonl"), recursive=True):
    for line in open(f, encoding="utf-8", errors="replace"):
        line = line.strip()
        if not line:
            continue
        try:
            e = json.loads(line)
        except Exception:
            continue
        if e.get("type") != "assistant_message":
            continue
        if e.get("text") and not (e.get("parts") or []) and e.get("timestamp", "") >= recent_cut:
            bad += 1
print(f"parts-less assistant lines written since {recent_cut}: {bad}   (must be 0 to merge)")
PY
```

**Status as of 2026-07-14: NOT met** — 127 parts-less lines written in the preceding week; 51 of 88 local agents are still on a pre-`parts[]` emitter. Recheck in a few weeks.

Note the check is only as good as the hosts you run it on; also confirm no deployed/vendored minds agent is pinned to a pre-`90ef7a979` `mngr_claude`.

## Verification

`libs/mngr/imbue/mngr/cli/transcript_test.py` — 38 passed.


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#28 from the same branch. This public repository is now a read-only mirror.

PR #2485, opened 2026-07-15
Recovery verdict policy: raw-state classifier, host-only restarts, unconditional start dispatch

gnguralnick:
## Summary

**Stacked on #2566** (the mngr-layer `HostState.UNREACHABLE` + degraded-observation vocabulary). Review #2566 first; this PR's diff is against that base and is the minds recovery-policy work.

Recovery should auto-act only when the action is safe without trusting any inferred state, and report failures even when a surface renders them. The core design change (arrived at after live-testing earlier revisions of this branch):

- **The recovery restart is dispatched unconditionally on recovery-page entry**, as a start-only restart (`start_only` in the v1 restart API). No probe, verdict, or host-state knowledge gates it: `mngr start` checks ground truth at commit time, no-ops on a live host, targets only STOPPED agents, and cold-boots a stopped one. A stopped workspace therefore revives on first entry — at app startup, after a laptop reboot, after an out-of-app stop — with no discovery-freshness crawl. The manual "Restart workspace" button keeps the full stop+start bounce, which never fires without a click.
- **The classifier's tiers are display-only** and uniformly freshness-gated: trusted `STOPPED`/`CRASHED` render as offline, `STOPPING` keeps checking, `FAILED` renders the consent page, `UNREACHABLE` (defined in #2566) renders the terminal "Can't connect" page with a canned access-rejected reason. A misclassification now costs wrong page copy, not a wrong action. This relies on #2566's corrected mints, where "could not determine state" surfaces as `UNKNOWN` rather than `CRASHED`.
- **Deleted with the knowledge requirement**: earlier revisions tier-gated the auto-dispatch on a classified `host_offline`, which demanded the app *know* a host was stopped — spawning a freshness-gate exemption for the offline pair and a persisted offline-state map with launch-time override seeding. Live quit/relaunch traces showed that mirror erased by stale in-flight and replayed discovery snapshots (the pipeline structurally produces them), stranding launches on a multi-minute "unresponsive" crawl. The unconditional dispatch removes the need for the knowledge, and the whole subsystem with it.

The rest of the branch:

- **Drop the surgical restart tier**: `INTERFACE_UNRESPONSIVE` and the services-scope in-place restart are gone; the v1 restart API accepts only `scope: "host"` (400 otherwise). Exec-reachable-but-not-answering renders the consent-gated "Workspace unresponsive" page — unless supervisord reports `STARTING`/`BACKOFF` (self-heal in progress), which keeps checking. The dead "Restart system interface" sidebar item is removed.
- **Stops close their windows**: stopping a workspace via the agent-facing `/api/v1/workspaces/<id>/stop` broadcasts a one-shot `workspace_stopped` chrome SSE event from the shared stop path; the Electron shell closes any window open to that workspace (skipped mid-restart), and browser mode navigates home. This is also the safety invariant behind the unconditional dispatch: a window that reaches recovery over a stopped host implies an out-of-app stop, whose revival is intended.
- **No recovery page is a dead-end**: the unresponsive and backend-unreachable verdicts keep re-probing on the slow cadence; every wait state arms the cheap liveness poll; a probe request dropped over a macOS sleep re-probes instead of stranding; a failed top-level content load routes to the recovery page; a window quit on the recovery page persists (and restores) as its workspace, not the port-stale recovery URL.
- **Evidence handling under a broken pipeline**: a completed exec failure is direct evidence (resolves the page even when discovery is stalled and the freshness gate can never open); a probe timeout stays inconclusive; provider errors replayed from before the backend started are dropped at ingest; the persisted-topology reader tolerates unknown fields so a file from another app version cannot void the system-services fallback.
- **Background execs can't sabotage recovery**: diagnostic/cleanup `mngr exec` calls pass `--no-start` (exec auto-starts by default; the backup verification check was observed cold-booting a stopped container mid-diagnosis). The workspace version read's exec argv is also repaired to actually reach the workspace.
- **Every `RESTART_FAILED` branch reports at error level** (reaches Sentry): stop step, start step, readiness timeout and services-agent-not-found (both previously unlogged), and the worker-spawn failure.
- **Diagnostics**: new "Is the system-services agent running?" probe row (scans `/proc/*/environ` for the `MNGR_AGENT_ID` marker), classifier-input logging alongside each verdict, and stop-path target/outcome logging.

Design doc: `blueprint/recovery-verdict-policy/plan-recovery-verdict-policy.md` (includes the history of the tier-gated design and why it was replaced).

## Testing

- CI fully green on the head commit (`5714b2a8f9`): Unit + Integration, Acceptance, Minds Snapshot Resume, test-offload, test-offload-acceptance, test-docker, test-minds-snapshot, check-changelog.
- Locally via `just test-quick`: 490 passed across the five recovery-touched unit-test files (templates, recovery_probe, workspace_recovery, api_v1, backend_resolver), 198 passed for ratchets + `test_desktop_client.py`, 0 failed.
- New/updated unit tests: unconditional entry dispatch (and dispatch-free `applyHealth`), uniform freshness gating incl. stale-offline-is-inconclusive, restart API `start_only` + scope rejection, exactly-one-`logger.error` capture on all five restart-failure branches, `workspace_stopped` broadcast semantics, cross-version topology-file tolerance.
- Manual staging pass pending: one quit ("Shut down all") → relaunch → reopen cycle with a stopped workspace, to observe the immediate "Restarting your workspace" cold boot in the real app.


apps/minds/imbue/minds/desktop_client/recovery_probe.py
```diff
@@ -418,6 +483,51 @@ def _build_can_run_commands_probe(in_container: _InContainerProbe, mngr_exec_com
...
+    Every process an agent launches inherits ``MNGR_AGENT_ID=<id>`` (mngr's own
+    stop path finds agent processes by this exact marker), so a match means the
+    system-services agent is still running. ``grep -l`` prints the matching
+    ``/proc/<pid>/environ`` paths; ``-z`` treats the NUL-separated environ
+    entries as lines so ``^`` anchors each variable name; ``-a`` forces text
+    matching on the binary-looking file. The leading ``unset`` matters when this
+    is re-run through ``mngr exec``, which sources the agent's env file into the
+    shell: without it the grep process itself carries the marker and self-matches.
+    """
+    agent_token = str(services_agent_id) if services_agent_id is not None else "<system-services-agent>"
```
gnguralnick, line 490: this seems like a weird fallback. why not just make this a required parameter?

apps/minds/imbue/minds/desktop_client/recovery_probe.py
```diff
@@ -418,6 +483,51 @@ def _build_can_run_commands_probe(in_container: _InContainerProbe, mngr_exec_com
...
+    elif in_container.agent_processes_error is not None:
+        output, answer = f"error: {in_container.agent_processes_error}", ProbeAnswer.UNKNOWN
+    elif in_container.agent_processes is None:
+        # The scan was skipped: the probe script received no agent id (or an old
+        # script version without the scan answered).
+        output, answer = "(agent-process scan did not run)", ProbeAnswer.UNKNOWN
+    elif in_container.agent_processes.strip():
+        output, answer = in_container.agent_processes, ProbeAnswer.YES
+    else:
+        agent_token = str(services_agent_id) if services_agent_id is not None else "<system-services-agent>"
```
gnguralnick: same as above

apps/minds/imbue/minds/desktop_client/recovery_probe.py
```diff
@@ -601,61 +714,100 @@ def _classify_dispatch_tier(
...
+            return DispatchTier.BACKEND_UNREACHABLE
+        if upper_state == _FAILED_STATE:
+            return DispatchTier.HOST_UNRESPONSIVE
+        if upper_state in _OBSERVED_RUNNING_STATES:
+            return DispatchTier.HOST_UNRESPONSIVE
+        # A trusted UNKNOWN / transitional (e.g. STOPPING) / absent state carries
+        # no verdict of its own; fall through to the completed-exec evidence.
+    if probe_exec_attempted:
+        # The exec completed without reaching the in-container script (a timeout
+        # returned above): direct fresh evidence the container is unreachable,
```
gnguralnick: nit: this comment is slightly confusing (the phrase "a timeout returned above" )

weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#30 from the same branch. This public repository is now a read-only mirror.

gnguralnick:
Continuing on imbue-ai/mngr-internal#105 — development has moved; this mirror is read-only, so PRs cannot merge here.

PR #2487, opened 2026-07-15
minds: skip Docker state-container ops when the daemon is unreachable

weishi-imbue:
## Bug

Minds `docker start`s a Docker "state container" at launch and `docker stop`s it at quit. If the daemon is **wedged** (socket accepts connections but never replies -- Docker Desktop starting up, or a stuck VM), both block the full **60s** command timeout -> 60s slow launch, 60s slow quit.

The old daemon-down fast path only caught a *fast* "Cannot connect" (cleanly-down daemon). A wedge returns nothing, so the 60s elapsed.

## Fix

Gate `start` / `stop` / `remove` on a reachability probe: `docker version --format '{{.Server.Version}}'`, **5s** timeout. Unreachable / wedged / absent -> fast no-op. Healthy daemon answers in <1s -> unchanged. Host-side only; no in-VM agent API change.

## Verified

- **Real wedged Docker Desktop**, real state-container name: patched `start_state_container` / `stop_state_container` returned in **5.0s** each (was ~60s).
- **Original bug, from logs**: launch had a 67s gap ending in `docker start ... exceeded its 60s timeout`; quit had `before-quit` -> `stop-state-container HTTP 500` 67s later.
- **Unit tests** (+5): probe returns True (healthy) / False (unreachable) / False via the 5s timeout (wedged, `sleep`-based fake); `start` / `stop` never invoke `docker start` / `stop` when unreachable. Existing "paused daemon raises" tests updated -- a paused daemon is *reachable* (answers `version`), so the probe lets it through to the real failure.
- **Controlled wedge repro (`kill -STOP`)**: froze all `com.docker.backend` pids on a healthy Docker Desktop -> raw `docker version` hung the full 8s; against that induced wedge the patched `_is_docker_daemon_reachable` returned `False` in 5.1s (probe SIGTERM'd at the 5s bound) and `start_state_container` returned in **5.1s** (was ~60s), logging "Docker daemon unavailable; skipping state-container start"; `kill -CONT` restored the daemon to healthy (0.0s). So the wedge is deterministically reproducible and the fix bounds it to one 5s probe.
- **End-to-end** ([run 29393502938](https://github.com/imbue-ai/mngr/actions/runs/29393502938), build `260715x48qyj0e3`): full ToDesktop build + `launch-to-msg` **green** -- create-agent + first-message round-trip through the built binary, i.e. no regression on a healthy daemon. (CI cannot reproduce a *wedged* daemon, so this covers the healthy path, not the wedge.)

## Dropped

An earlier draft also moved the launch start to a background thread (0s critical path). Dropped -- it re-introduced a discovery race where docker agents could be briefly missing after launch with no obvious cause. The probe gate alone fixes both paths and keeps the original "container up before discovery" ordering.

### DIFF
--- a/apps/minds/changelog/mngr-minds-docker-daemon-probe.md
+++ b/apps/minds/changelog/mngr-minds-docker-daemon-probe.md
@@ -1,0 +1,3 @@
+Fixed a ~60s hang on app launch and on quit when the local Docker daemon is unresponsive. Minds starts and stops an mngr Docker "state container" around its lifecycle; if the Docker daemon was wedged (its socket accepts connections but never replies -- e.g. Docker Desktop still starting up, resuming from sleep, or stuck), each `docker start`/`docker stop` blocked until its 60-second command timeout, delaying the backend coming up (launch) and the app closing (quit).
+
+Launch and quit now probe daemon reachability first (`docker version`, bounded to 5 seconds) and skip the state-container operation when the daemon is unreachable, wedged, or absent. A healthy daemon answers in well under a second, so normal launches are unaffected.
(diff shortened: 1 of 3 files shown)


joshalbrecht:
lgtm!

thanks for the minimal fix :)

PR #2490, opened 2026-07-15
Reduce rotated log size to 10MB for minds and mngr latchkey forward

hynek-urban:
- minds Electron logs (electron.log, minds.log) now rotate at 10MB instead of 100MB, matching the Python backend jsonl sink (already 10MB)
- mngr latchkey forward daemon caps its dedicated events.jsonl at 10MB
- add optional max_log_size_mb override to setup_command_context / parse_output_options so a command can pin its own rotated-log cap


joshalbrecht:
lgtm

PR #2492, opened 2026-07-15
Never tombstone records before their provider's first snapshot; resurrect premature tombstones

joshalbrecht:
## Summary

Fixes a workspace-sync race observed live: right after app start, the absent-host tombstone pass judged a stopped lima workspace "definitively absent" — because a provider that hasn't produced its first listing yet is indistinguishable, via the error map alone, from a healthy provider with no such host — and pushed a spurious `destroyed` record to the connector, silently de-associating the workspace.

- **(a) Per-provider snapshot gate**: `_tombstone_definitively_absent` now also requires the record's *specific* provider to have produced at least one discovery snapshot before absence counts as evidence.
- **(b) Resurrection safety net**: a `DESTROYED` row hosted by this install whose agent is live among the **ACTIVE** discovery ids (not the known ids, which retain genuinely-destroyed hosts for the provider's grace window) is re-activated and re-pushed. Guarded by the destroying marker so an in-flight destroy is never undone. This self-heals rows tombstoned by the old race (e.g. the docker-2 record that triggered this fix).

## Testing

- `just test-quick apps/minds` — 1918 passed, 0 failed, including new tests: tombstone skipped before the provider's first snapshot (then fires after), resurrection of a live-again tombstone (re-pushed as active), no resurrection while a destroy marker exists, no resurrection of other-device tombstones.
- Ratchets, ruff, and `ty` all clean.


joshalbrecht:
lgtm

PR #2493, opened 2026-07-15
Latchkey test e2e

hynek-urban:
Add an end-to-end test for Latchkey gateway connectability:
- Create a (pseudo-)remote workspace.
- Check that the local Latchkey gateway is reachable from the remote workspace.
- Check that the secondary (remote) Latchkey gateway is reachable from the remote workspace, too.
- Check that updating permissions locally results in them getting updated remotely, too.

The related remote VPS provisioning logic now also newly checks the node version.

This test can be invoked like this:

```
latchkey curl -L -X POST -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28" https://api.github.com/repos/imbue-ai/mng/actions/workflows/ci.yml/dispatches -d '{"ref":"hynek/latchkey-test-e2e","inputs":{"run_minds_release_tests":"true"}}'
```


joshalbrecht:
tentatively lgtm, but will want to refactor this in a follow up with @danverbraganza

PR #2495, opened 2026-07-15
minds: pass MINDS_RESTIC_BINARY in dev mode; surface backup status read failures

gnguralnick:
## Bug

Running minds from source (dev mode), the workspace settings page showed configured workspaces as having no backups (Sentry bug report `56d18f3c30ad4d5a84a64d61062e373d`; the same workspaces showed their backups fine under the packaged 0.3.6 app).

The dev-mode env block in `electron/backend.js` never passed `MINDS_RESTIC_BINARY` to the spawned Python backend -- only the packaged-mode block did. `restic_cli` then fell back to bare `restic` on PATH, which fails on any machine without a system restic install, and every snapshot listing raised. Commit `7b490eabef` made `paths.getResticPath()` return the bundled binary in both modes (and its message claims backend.js plumbs it through), but the dev-mode spawn env was never updated. The route degrades the failure into an empty snapshot list + `snapshots_error`, and the UI showed the empty list with only a vague "Backup status unknown." line.

## Fix

- `electron/backend.js`: dev-mode env block now passes `MINDS_RESTIC_BINARY: paths.getResticPath()`, mirroring packaged mode. The bundled binary is guaranteed present in dev by the existing `prestart` ensure-binaries hook.
- `static/workspace_backups.js`: a `snapshots_error` now renders the actual error in the problems list ("Could not read backup status from this machine: ...") and a clearer status line, including when the workspace is offline or verification is disabled.
- `static/backup_health.js`: the sidebar/chrome backup warning badge also fires on `snapshots_error`, not only on `check_state PROBLEMS`. No badge noise for unconfigured workspaces: `snapshots_error` is only set when a canonical restic env exists.

## Verified

- **Real data, real code path**: with `MINDS_RESTIC_BINARY` pointed at the bundled binary (exactly what the fixed spawn env does), `list_workspace_snapshots` against the reporting machine's canonical env listed 9 snapshots for the affected host (read-only restic call). Without the var, the same call reproduces the reported failure.
- JS unit suite (`node --test test/unit/*.test.js`): 16 passed, 0 failed. `node --check` on all three edited JS files.
- Changelog gate passes (`python -m scripts.check_changelog_entries`).
- Not yet exercised in a live Electron app: the two error-surfacing UI branches (static browser JS has no test harness; needs a run-from-source app restart to observe).


gnguralnick:
closed in favor of #2522 for now, will recreate the landing-page stuff as a separate PR

PR #2496, opened 2026-07-15
Fix stale discovery-history replay at minds startup; attach latchkey + discovery logs to bug reports

gnguralnick:
## Summary

Root-causes and fixes the "old workspaces appear at startup, disappear one by one, then the remaining one is renamed" reports from minds 0.3.6 (Sentry events `adc0cf44...`, `694959cc...`, `e6735eb6...`, confirmed against the reporter's actual `events.jsonl`).

**The bug**: `find_discovery_snapshot_replay_offset` anchors the observe-stream replay window at the minimum across every provider's latest `DISCOVERY_PROVIDER` snapshot *and* the latest legacy `DISCOVERY_FULL` line. Nothing has written `DISCOVERY_FULL` since the per-provider migration, so an install that upgraded across it keeps its last pre-migration line forever -- permanently pinning the replay window days in the past. Every app start then re-folds the entire tail of the events file through the UI: ghost workspaces with pre-rename names, their destroy events one by one, the rename, and historical provider-error snapshots flashing as if live. On the reporter's real file this replayed 17,461 events (5 days, ~85% of 36MB) at every launch.

**The fix**: the legacy `DISCOVERY_FULL` timestamp now participates only when the file holds no per-provider snapshot at all (a pure pre-migration log, where it is still the correct anchor). On the reporter's file the replay window moves from July 9 to the current session and the replay shrinks to the last ~15KB.

**Also**: minds bug reports now attach the detached `mngr latchkey forward` daemon's logs (`events.jsonl`, `latchkey_forward.log`) and the shared discovery `events.jsonl` -- the two artifacts this investigation needed and had to obtain manually. `LogAttachmentGroup` gained an optional `base_dir` so a group can sweep a directory outside the process log folder; consent gating is unchanged.

**Note**: an earlier commit on this branch carried a stopgap for the snapshot-resume CI failure (asserting `browser` in place of the removed `web` service). main fixed the same failure by asserting only the always-on services (`system_interface`, `terminal`), deliberately excluding `browser` as flaky; merging main resolved the conflict in favor of main's version, so this PR's net diff no longer touches `test_snapshot_resume.py`.

## Testing

- New regression test: a stale `DISCOVERY_FULL` line plus fresh provider snapshots anchors the replay at the provider snapshots (plus existing legacy-only fallback tests still pass).
- New tests for `base_dir` sweeping (external dir, missing dir, no log folder) and the minds external attachment groups.
- Local quick suites for `libs/imbue_common`, `libs/mngr`, and `apps/minds` passed in full before origin/main was merged in (0 failures); the merged result is validated by CI on this PR. `just test-offload` unavailable in this workspace (no MODAL_TOKEN_ID).


joshalbrecht:
lgtm

PR #2498, opened 2026-07-15
Auto-verify paid users; stop trusting stale email-verification token claims

joshalbrecht:
## Summary

Paid users were getting blocked by the SuperTokens email-verification gate in `remote_service_connector`. This fixes that with three changes:

1. **Auto-verify paid users at signup** (`auth_signup`): a paid email/password signup is marked verified up front (before the session is minted, so the very first token already carries the verified claim) and skips the verification email. A paid-list lookup failure falls back to the normal verify-by-email flow rather than failing signup.

2. **Verify existing accounts when added to the paid list** (`add_paid_email`): after activating the paid-list row, best-effort mark any pre-existing account for that email verified, so a user who signed up *before* being made paid isn't left locked out. Purely best-effort — a SuperTokens hiccup never fails the paid-list write, and an email with no account yet is a no-op.

3. **Stop trusting the stale token claim** (`_authenticate_supertokens`): resolve email-verification from a live core lookup (the same `get_user` the guard already ran to get the email) instead of the token's cached, possibly-stale, `EmailVerificationClaim`. Verification now takes effect on the user's very next request instead of only after their token refreshes.

OAuth (Google/GitHub) sign-ups are intentionally untouched: those emails arrive already verified from the provider.

## Tests

- Rewrote the `_authenticate_supertokens` guard tests for the new live-lookup behavior, including a test proving a stale `email_verified=false` token still authenticates once the core reports the email verified.
- New signup tests: paid email is auto-verified (no email sent); non-paid control still requires verification.
- New paid-add tests: verifies a pre-existing unverified account; no-op when no account exists; still succeeds when SuperTokens is unreachable.
- Added a `create_email_verification_token` fake to `FakeSuperTokensBackend`.

All 294 `remote_service_connector` unit/integration tests pass locally; `ty check` is clean.


joshalbrecht:
lgtm

PR #2499, opened 2026-07-15
minds-evals: eval harness with per-batch Modal envs and Modal-hosted desktop boxes

minhtrinh-imbue:
Ground-up rebuild of the eval harness. Every batch gets its own Modal env (unique name = batch identity, atomically claimed). Boxes are full Minds computers running as Modal sandboxes -- the real Minds Electron app on a virtual desktop, streamed through Modal's encrypted tunnel as one https URL. launch creates the batch's workspaces inside that computer (in-sandbox API discovery) and leaves it watchable; visit-batch finds the same computer by tag; stop terminates it. No local Docker anywhere. Live-verified end-to-end: desktop renders in-browser, 3/3 parallel creates, batches self-complete to S3.


apps/mngr_minds_eval/SETUP.md
```diff
@@ -0,0 +1,65 @@
...
+
+(No region -- R2 ignores it; the CLI defaults it to `auto`.)
+
+## 3. Anthropic key
+
+Needed by `launch` (the workspaces run with `ai_provider=api_key`).
+
+```
+export ANTHROPIC_API_KEY=sk-ant-...
+```
```
joshalbrecht, line 63: I'm getting rid of this / got rid of this in a recent PR

(the auth will be done fully inside of the workspace, not controlled by the API at all)

joshalbrecht: we'll want the env var to be set when launching for testing like this, but it might require a small number of changes to fiddle and make it possible to include the ANTHROPIC_API_KEY in the mngr host env again (just for testing)

joshalbrecht:
roughly lgtm, but we'll need to adapt to the new auth method when that merges soon

weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#32 from the same branch. This public repository is now a read-only mirror.

PR #2508, opened 2026-07-16
Fix the handling of stopped workspaces in mngr_latchkey

hynek-urban:
Stopped workspaces are still discovered. The old logic didn't account for that - it assumed that each discovered agent / workspace was running.

Now we do not try to maintain reverse SSH tunnels to agents who we know aren't running.

This should fix these errors in our production Sentry:
- https://generally-intelligent-e3.sentry.io/issues/7602487397
- https://generally-intelligent-e3.sentry.io/issues/7602521517


libs/mngr_latchkey/imbue/mngr_latchkey/discovery.py
```diff
@@ -263,6 +292,22 @@ def _run_remote_setup(
...
+        """Drop a stopped agent's reverse tunnel and mark its host for re-provisioning.
+
+        The per-agent reverse tunnel points at the container's sshd, which is
+        down while the host is stopped; removing it stops the tunnel manager's
+        health-check loop from indefinitely re-dialing a dead endpoint.
+        Forgetting the host's ``_provisioned_hosts`` marker means a later restart
+        re-runs the idempotent VPS-resident gateway provisioning, since a stopped
+        container may be recreated before it comes back.
+        """
+        removed_tunnel_count = self.tunnel_manager.remove_reverse_tunnels_for_agent(str(agent_id))
```
joshalbrecht, line 305: there's a very minor race condition inside of remove_reverse_tunnels_for_agents because it locks, releases, then calls _drop_tunnel_keys (which then again locks and releases)

idk if it's even worth fixing, but it does seem to be there :-/

joshalbrecht: I mostly bring it up to highlight this *class* of error, which I don't think imbue_code_guardian is properly checking for, and which we have no other real checks for, but which is likely important when dealing with locked data like this

not sure if it's worth doing anything right now...

joshalbrecht:
lgtm, see slack comment

PR #2509, opened 2026-07-16
Pin known hosts during discovery 

hynek-urban:
Previously, the imbue cloud provider would emit discovered agents including their SSH info without actually pinning it in known_hosts. That was mostly fine because the pin would usually happen elsewhere but in some code paths it caused issues. This should hopefully fix https://generally-intelligent-e3.sentry.io/issues/7602464601 and https://generally-intelligent-e3.sentry.io/issues/7605824508/.

(It adds a `logger.warning()` call but I think it's fine because it's consistent with the existing style of the module.)


libs/mngr_imbue_cloud/imbue/mngr_imbue_cloud/providers/instance.py
```diff
@@ -800,6 +809,34 @@ def _ensure_outer_host_key_known(self, lease: LeasedHostInfo) -> None:
...
+                host_id, lease.vps_address, lease.container_ssh_port, lease.container_host_public_key
+            )
+        except OSError as exc:
+            logger.warning(
+                "imbue_cloud[{}] could not update known_hosts for container endpoint of host {} (vps {}): {}",
+                self.name,
+                host_id,
+                lease.vps_address,
+                exc,
+            )
```
joshalbrecht, line 838: this is another one of those `logger.warning` instances that feel sketchy to me

As a human programmer (or even as an LLM) I would be fairly surprised if I called a function named `_ensure_container_host_key_known` and it did not, in fact, ensure that the container host key was known :-P

idk why such errors should *not* be allowed to propagate?  If we cannot set the known host keys, something else has gone deeply wrong, and at least it seems like we need to be learning about how this can fail (in sentry) so that we can then add some UI for the user to go handle such a state?

joshalbrecht:
lgtm (except the above comment, which, per slack discussion, we can resolve along with the rest of the logger.warnings later)

PR #2510, opened 2026-07-16
Fix WorkspaceSyncScheduler vs MngrManager shutdown ordering

hynek-urban:
When exiting the Minds app, the workspace sync could still be in flight while we already tore down the MngrCaller, resulting in errors like this one: https://generally-intelligent-e3.sentry.io/issues/7613995811/

This PR makes sure that we first stop the WorskspaceSyncScheduler and only then do we stop MngrCaller.


joshalbrecht:
lgtm, thanks for the fix!

PR #2512, opened 2026-07-16
Add setup script for WSL2, fix local-directory branch checkout in minds create

joshalbrecht:
## Summary

Two artifacts from getting the full minds stack running under WSL2 on Windows (AWS metal instance):

- **Bug fix**: creating a workspace from a *plain local directory* template (a non-worktree clone) with an explicit `branch` ran `git checkout -B <branch> FETCH_HEAD` in the user's own template checkout without any prior fetch. On a fresh clone (no FETCH_HEAD) the create fails with `'FETCH_HEAD' is not a commit`; with a stale FETCH_HEAD from an unrelated earlier fetch it would silently reset the user's branch tip. Plain local directories now use a new `checkout_existing_branch` (no-op when already on the branch, plain `git checkout <branch>` otherwise, never moves the tip); scratch clones keep the FETCH_HEAD-based `checkout_branch`.
- **Docs**: `apps/minds/docs/wsl.md` -- a verified recipe for running minds under WSL2 on Windows Server, covering the WSL-specific gotchas found along the way (distro auto-termination keepalive, headless latchkey install, `runsc --overlay2=none`, the grandparent-death watcher vs daemonization, CSS build, IPv4-only localhost relay).

## Test plan

- 4 new regression tests for `checkout_existing_branch` in `agent_creator_test.py` (fresh-clone no-op, stale-FETCH_HEAD non-reset, remote-branch DWIM checkout, missing-branch error).
- Full suite via `just test-offload`.
- Manually verified end-to-end on a Windows Server 2022 EC2 metal instance: workspace created from a plain local-directory template via the API, container running under gVisor in WSL2, System Interface serving.


joshalbrecht:
lgtm

PR #2513, opened 2026-07-16
CASA Tier 2 A-list security fixes (minds, connector, CI)

pseay-imbue:
## Summary

Implements the self-contained ("A-list") findings from the CASA Tier 2 (OWASP ASVS v4.0) audit of the minds app. The full audit, with per-requirement verdicts and file:line evidence, is in [`apps/minds/docs/casa-compliance-audit.md`](apps/minds/docs/casa-compliance-audit.md).

Each fix was implemented by an Opus subagent and then reviewed by an independent adversarial verifier. One confirmed regression they caught (the new CSP would have blocked Sentry egress) is fixed. A second conflation — clearing the device-level session on Imbue-account sign-out — was caught in review and reverted (see below).

## Fixes in this PR

| CASA req | What changed | Before → After |
|---|---|---|
| **6.1.1** | `.github/dependabot.yml` (weekly uv/npm/actions updates), a scheduled `.github/workflows/dependency-audit.yml` running `pip-audit --skip-editable` + `pnpm audit --prod`, and a `gitleaks` secrets-scanning pre-commit hook | ❌ NOT MET → ✅ MET |
| **1.1.2 / 1.3.1** | 15-minute TTL on the local one-time login code: `created_at` stamped at mint, expired-but-`VALID` codes rejected and marked `EXPIRED`, missing timestamp treated as expired | 🟡 PARTIAL → ✅ MET |
| **5.1.7** | Flask `after_request` hook sets `X-Content-Type-Options: nosniff` and a first-party CSP, **host-scoped** so proxied `*.localhost` agent content is excluded (agents own their CSP). `connect-src` allows `'self'`, loopback websockets, and the Sentry ingest origin | 🟡 PARTIAL → ✅ MET |
| **3.2.2 (`state`)** | CLI generates a `secrets.token_urlsafe(32)` state; connector reflects it into the authorize URL (`_authorize_url_with_state`); CLI verifies the echoed state in constant time **before** exchanging the code, aborting on missing/mismatched/non-ASCII | ❌ NOT MET → ✅ MET |
| **3.2.1 (PKCE)** | Confirmed the SDK only mints a verifier under `force_pkce`; enabled it on Google and threaded the verifier end-to-end (authorize response → CLI in-memory → callback → `RedirectUriInfo`). GitHub left off (OAuth Apps don't support PKCE) | 🟡 PARTIAL → ✅ MET |
| **5.1.9** | `shlex.quote` the destroy command's `--include` host-id filter (defense-in-depth) | ✅ hardened |

**Verdict corrected without a code change** — 2.2.1 (logout invalidates tokens) and 6.6.1 (browser storage cleared on logout) are ✅ MET as-is: Imbue-account sign-out already revokes the SuperTokens access/refresh tokens server-side, and no user credential is stored in the browser. See the revert note below.

**Reclassified to 🟦 PROBABLY (owned by the external SuperTokens connector, not verifiable from this repo):** 1.1.1 (sign-in brute force), 1.1.3 (password hashing), 2.2.2 (session invalidation on password change).

## Revert: don't clear the device session on account sign-out (2.2.1 / 6.6.1)

An initial revision cleared the `minds_session` cookie + Electron storage on account sign-out. That conflated two separate things:
- **`minds_session`** is the app/device unlock — minted from the one-time code `minds run` emits (payload is the constant `"authenticated"`), independent of any Imbue account, and required for **all** local app use (settings, creating/deleting local workspaces).
- **The Imbue account** (SuperTokens session) is only needed for cloud features.

Because the Settings "Sign out" button POSTs `/auth/api/signout` then navigates to `/`, clearing `minds_session` would have bounced the user out of the entire app after they merely disconnected a cloud account. Reverted. Account sign-out still revokes the real user tokens via `signout_user_via_plugin`, which is what 2.2.1 requires.

## Verification of reviewer assumptions

- **gitleaks not annoying / no false positives:** ran `gitleaks` (v8.30.1) across the whole working tree — **Passed, zero findings**, so no custom allowlist config is needed for the current codebase. (A full git-*history* scan was not run.)
- **Sentry still receives frontend errors under the new CSP:** the CSP's `connect-src` includes the Sentry ingest origin (derived from the DSN table, covered by a unit test asserting the ingest origin is present and `ws:`/`wss:` are loopback-scoped). This has **not** yet been exercised end-to-end in a running Electron build with a live DSN (e.g. a deliberate divide-by-zero in the request inbox) — recommend that manual check before merge, since it can only be observed against a real Sentry project.

## Testing

All tests touching these changes pass (171 ratchets, 298 connector, 450 `mngr_imbue_cloud`, 1400+ `minds/desktop_client`, plus targeted OAuth/auth/CSP suites); `ty check` adds no new errors. Two pre-existing **environment-only** failures remain in files this PR does not touch and will pass on Linux CI: restic/backup tests (`ResticNotInstalledError`, no local `restic` binary) and one lima_slice test (macOS BSD `sed` vs GNU `sed`).

## What's left (deliberately out of scope) and why

**Needs a product/design decision, not a self-contained code fix:**
- **3.3.1 (admin MFA)** — `MINDS_PAID_ADMIN_KEY` gates a privileged op with one static secret; adding MFA is a feature (likely SuperTokens), not a patch.
- **2.2.3 (30-day non-revocable session cookie)** — shortening TTL / adding revocation changes the re-login UX and needs a revocation store.
- **2.4.1 (step-up re-auth)**, **2.3.4 (replay nonce)**, **3.1.5 (synchronizer-token CSRF)** — each adds a UX flow or makes the stateless cookie stateful; the existing `SameSite=Lax` + `HttpOnly` posture may already satisfy an assessor.
- **3.2.2 (`redirect_uri` half)** — optional loopback-only allow-list on the connector as defense-in-depth beyond the IdP's registered-redirect-URI check.

**External service (verify, don't code here):** confirm the SuperTokens connector enforces sign-in lockout (1.1.1) and session revocation on password change (2.2.2).

**Documentation-only:** record 2.3.1 (no `Secure` cookie — loopback HTTP by design) and 4.1.2 (self-signed loopback TLS) as formal accepted exceptions in the CASA questionnaire.


hynek-urban:
I have a doubt related to the new `clear_session_cookie()` call as well as `clearSessionStorageOnSignout()`. I'm not sure but I suspect it may conflate two only semi-related things:

1. The fact that we're trying to enable the frontend app to talk to the backend at all (= we basically authenticate the app, not the user).
2. Super token / Imbue accounts.

What happens after you "sign out"? Is it still possible to continue using the application (change settings, create new local workspaces, delete old ones)?

---

Other than that, with my limited undestanding, LGTM under the following assumptions:

- We have tested that committing with the new `gitleaks` hook is not annoying. We ran gitleaks against the whole codebase and in case any false positives turned up, we prepared custom configuration.
- We have tested that intentionally breaking frontend code (e.g. division by zero in the request inbox) still sends errors to Sentry.

pseay-imbue:
Yup, you're right about the conflation. That's been updated.

Also, when logging out, it now only revokes the SuperToken session, so all the expected local actions work.

Gitleaks isn't annoying and I tested that sentry error reporting still works :)

hynek-urban:
@pseay-imbue Awesome, thanks :) If you think it all works then LGTM :)

pseay-imbue:
Code guardian just caught one issue that the CSP wouldn't allow our iframes on the web version, but that's fixed now. So I think it should be good.

weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#34 from the same branch. This public repository is now a read-only mirror.

pseay-imbue:
Continuing on imbue-ai/mngr-internal#97 — development has moved; this mirror is read-only, so PRs cannot merge here.

PR #2516, opened 2026-07-16
Pin agent tmux windows to a stable size so a stray client can't collapse them

gnguralnick:
## Problem

Agent tmux sessions were created with tmux's default `window-size latest` policy. All agents on a host share one tmux server (no per-agent socket), so a brand-new detached session is born at the geometry of whatever client is already attached to the server — even a degenerate, unsized **2x1** ttyd/web-shell client — regardless of the `new-session -x 200 -y 50` sizing. A 2x1 pane breaks Claude Code's Ink TUI, so the initial `mngr message` (which waits for the TUI to be ready) times out and the message is **silently dropped** (slash commands use the RELAXED confirmation policy).

This is the root cause behind "the agent was created but never got its first message" reports — the desktop `/assist` "have an agent help" flow, launch-task workers, and any message-driven agent. It is general, not flow-specific. Its frequency rose with the persistent-terminals feature, which leaves persistent 2x1 ttyd clients on the shared server.

### Verified on a real host (tmux 3.3a)
- With a 2x1 client attached, `tmux new-session -d -x 200 -y 50` is **born 2x1** (the `-x/-y` guard does not protect against this).
- `set-option window-size manual` + `resize-window` → stable **200x50**.
- `#{client_width}/#{client_height}` expand at hook-fire time to the **attaching** client's geometry.

## Fix

Default sizing policy: pin the primary window to `window-size manual` and `resize-window` it to the intended dimensions (200x50) — deterministic creation, immune to stray clients. The persistent `client-attached[98]` hook now runs in **fit** mode: it re-fits the pane to a real attaching client's size, **floored at 80x24** so a degenerate client can never collapse it below what Claude Code needs — so interactive attach still fits your terminal.

An explicit `--tmux-window-size` is honored verbatim (**nudge** mode): client-following modes (`latest`/`largest`/`smallest`) behave as before, and an explicit `manual` stays truly fixed (the re-fit hook leaves it untouched).

Note: this changes mngr's **default** tmux behavior for all users. Only newly-created agents get the new pinning.

## Changes
- `hosts/host.py` (`_build_start_agent_shell_command`): pin manual + resize by default; install the hook in fit vs nudge mode.
- `resources/sigwinch_panes.sh`: add fit/nudge modes; fit resizes to the clamped client size then repaints; nudge is the prior behavior.
- Unit + tmux integration tests; `libs/mngr` changelog entry.

## Tests
- ruff format + check, ty: clean
- 270 passed across the affected unit test files
- 4 tmux integration tests pass on real tmux (incl. new fit: 140x40 → fits exactly, 2x1 → floored to 80x24; explicit-manual untouched)

## Follow-up (separate)
The ttyd/persistent-terminals side should not attach a 2x1 client in the first place (report a sane minimum geometry for a hidden/unsized xterm). This fix makes agents robust regardless, but fixing the source too is worthwhile.


joshalbrecht:
lgtm!  thanks for the extra testing

PR #2517, opened 2026-07-16
minds: robust window-state persistence (Stream A of 3)

gdenisov:
## Problem

After restarting the minds app -- especially via the ToDesktop auto-update **Install and Restart** prompt -- users sometimes land on the "create a new workspace" screen instead of their restored workspace windows. Intermittent.

Root cause (window-state persistence half): the Electron shell restores previously-open windows from `~/.minds/window-state.json`, which was written **only** at quit/close. `saveSessionState` iterates the live window set and writes `{ windows }` unconditionally. A non-graceful quit (the auto-update restart, a crash, a force-quit) tears the windows down concurrently with that save, so the function iterates already-destroyed windows and overwrites a good file with `{ "windows": [] }`. Next launch computes `restorableCount === 0` and `decideStartupRoute` returns `create`.

Confirmed from a user's `~/.minds/logs/electron.log`: the update-restart quit takes an abnormal path (fires `before-quit isShuttingDown=true` twice) and made no attempt to reload the saved `/goto/agent-.../` window, whereas a normal restart did.

## Fix (Stream A)

- **Continuous, debounced persistence (A1).** Persist `window-state.json` continuously instead of only at quit, triggered on window move, window resize, and content navigation. A trailing-throttle scheduler coalesces bursts (a window drag fires an event per frame) into at most one write per second, so disk churn stays negligible. The saved set now reflects the live windows however the app exits.

- **Empty-clobber guard (A2).** `saveSessionState` now refuses to overwrite a non-empty on-disk file with an empty snapshot -- a teardown-race artifact. An empty snapshot is written only when the on-disk file is already empty, so a genuine empty still persists while a race can't zero out a good file. The normal last-window-close path saves through the quit sequence while the window is still alive (a non-empty snapshot), so it is unaffected.

- **Observability.** Added one `[startup]` log line at the `decideStartupRoute` call site logging `authenticated`, `hasAccounts`, `workspaceCount`, `restorableCount`, and the chosen route -- the headline diagnostic for "why did I land on create".

The debounce scheduler and the guard decision live in a pure, Electron-free module (`electron/session-persistence.js`), mirroring how `startup-routing.js` is kept testable under plain node. New unit tests cover the burst coalescing and the guard.

## Tests

`node --test test/unit/*.test.js` from `apps/minds`: **23 passed, 0 failed** (7 new).

## Scope

Only the Electron shell (`apps/minds/electron/`) and its JS unit tests are touched. No Python or Jinja changes.

**Stream A of 3.** Stream B (PR #2518) fixes the Python backend landing-render half; Stream C (PR #2533) prunes the resolver's persisted last-good topology on observed `DESTROYED` so destroyed workspaces stop counting as restorable.

### DIFF
--- a/apps/minds/changelog/gleb-minds-window-state-robust-persist.md
+++ b/apps/minds/changelog/gleb-minds-window-state-robust-persist.md
@@ -1,0 +1,7 @@
+Fixed a bug where restarting the app -- especially via the "Install and Restart" auto-update prompt -- could drop you on the "create a new workspace" screen instead of restoring your open workspace windows.
+
+The desktop shell now persists its window layout continuously (debounced) as you move, resize, and navigate windows, instead of only at quit. So however the app exits (auto-update restart, crash, or force-quit), the saved layout reflects your live windows and is restored on the next launch.
+
+It also no longer lets a non-graceful quit -- which tears windows down while the save is running -- overwrite a good saved layout with an empty one, which was the direct cause of landing on the create screen.
+
+The startup log now records why the app chose its landing screen (authenticated, account/workspace counts, restorable-window count, and the chosen route), making a bad restore diagnosable from `~/.minds/logs`.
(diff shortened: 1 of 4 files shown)


joshalbrecht:
lgtm, thanks for the fix!

PR #2518, opened 2026-07-16
Fix minds landing dropping users with workspaces onto the create form (Stream B of 3)

gdenisov:
## Summary

Fixes the intermittent bug where restarting the minds app (especially via a ToDesktop auto-update) sometimes lands a user with existing workspaces on the terminal "create a new workspace" screen instead of their workspace list.

**This is Stream B of 3.** Stream A (PR #2517) is the Electron `window-state.json` persistence fix; this PR fixes the landing-render half in the Python backend + Jinja templates; Stream C (PR #2533) prunes the resolver's last-good topology on observed `DESTROYED` (see the risk note below).

### Root cause

On a cold start, `has_completed_initial_discovery()` flips `True` on the first discovery event from the first provider. If that event arrives before the provider hosting the user's workspace (e.g. imbue_cloud) has surfaced it, and there are no remote tiles yet, `/` (`_handle_landing_page`) saw an empty live workspace list and fell through to `render_create_form` -- a terminal page with no SSE subscription and no auto-refresh -- so the user was stuck there even though they have workspaces.

### Background: what "Fix D" refers to

[Fix D (commit `dafb06ec3`, "restore windows against last-good topology, not just the live snapshot")](https://github.com/imbue-ai/mngr/commit/dafb06ec3e64ef26c69e1013d6694a4a5858f407) is a prior fix on the *window-restore* path. Providers enumerate at very different speeds on cold start (imbue_cloud returns its host in ~1s; local docker takes ~15-20s to enumerate stopped hosts), so the first discovery snapshot the Electron shell trusts can be provider-incomplete -- and window restore was dropping every not-yet-re-discovered workspace window onto the create form.

Fix D added `backend_resolver.list_restorable_workspace_ids()` -- the live `is_primary` workspace agents **unioned with the persisted last-good `system-services` topology** (a per-host record of the last complete enumeration, kept on disk across restarts). The Electron shell's window-restore filter keeps a saved window as long as its workspace is in that restorable set, so a workspace merely *absent* from a partial cold-start snapshot is no longer mistaken for a destroyed one (the live discovery flow still navigates genuinely-destroyed workspaces away after restore). This PR reuses that same primitive for the landing-render decision below.

### The fix

**B1 (backend, `_handle_landing_page`).** When the live workspace list and remote tiles are both empty, the handler now consults `backend_resolver.list_restorable_workspace_ids()` (the live primary agents unioned with the persisted last-good `system-services` topology -- the same primitive Fix D added for the Electron window-restore filter). When that set is non-empty -- meaning we *know* workspaces exist even though a partial cold-start snapshot hasn't re-surfaced them -- it renders the auto-refreshing "Discovering agents..." Landing page (which self-heals into the workspace list via its reload timer + chrome SSE) instead of the create form. The create form is now shown only when discovery has completed **and** the restorable set is genuinely empty (a true first-run user). No new resolver primitive was added.

**B2 (template defense-in-depth, `Create.jinja`).** When the create form is rendered as the landing fallback at `/` (new `is_landing_fallback` flag), it subscribes to `/_chrome/events` and navigates to `/` the moment a `workspaces` event shows a workspace now exists (live, remote, or restorable), mirroring `Landing.jinja`'s existing SSE self-heal. It is guarded so an in-flight create submission (which redirects to `/creating`) is never interrupted. The explicit `/create` page leaves the flag `False`, so a deliberate "create another workspace" flow is never bounced away by the user's existing workspaces.

**Observability.** Added one concise debug log at the fallback decision point: `active=<N> remote=<M> discovery_complete=<bool> restorable=<K> -> <create-form|discovering>`.

### Tests

- `test_landing_page_shows_discovering_when_only_restorable_workspaces_remain`: live/active empty + no remote tiles + discovery complete + restorable non-empty -> Discovering page, not the create form.
- `test_landing_page_shows_create_form_when_restorable_set_is_empty`: discovery complete + restorable genuinely empty -> create form (unchanged behavior).
- `test_render_create_form_landing_fallback_wires_self_heal_sse` / `test_render_create_form_explicit_page_omits_self_heal_sse`: assert the SSE/reload wiring is present only in the landing-fallback form.

Local runs (all green): the full `apps/minds/imbue/minds/desktop_client` subtree (1462 passed), the minds `test_ratchets.py` (57 passed), `test_no_type_errors`, and `ruff format` + `ruff check`.

### Risk note

`list_restorable_workspace_ids()` includes the persisted last-good topology, which (before Stream C) is never pruned in-process. If a user destroyed all their workspaces while the app was closed, the stale last-good entry could keep `/` on the "Discovering..." page instead of the create form. This is a pre-existing property of `list_restorable_workspace_ids` (Fix D already relies on it for window restore) and degrades to a self-refreshing spinner rather than a hard failure. **Stream C** (PR #2533) fixes this at the source by pruning the last-good topology when a host is observed `DESTROYED` and excluding `DESTROYED`-host agents from the restorable set.

### DIFF
--- a/apps/minds/changelog/gleb-minds-landing-discovering-not-create.md
+++ b/apps/minds/changelog/gleb-minds-landing-discovering-not-create.md
@@ -1,0 +1,7 @@
+Fix an intermittent bug where restarting the minds app (especially via a ToDesktop auto-update) could drop a user with existing workspaces onto the terminal "create a new workspace" screen instead of their workspace list.
+
+On a cold start, discovery marks itself "complete" on the first event from the first provider. If that arrives before the provider hosting the user's workspace has surfaced it, the landing handler (`/`) previously saw an empty live workspace list and fell through to the create form, a terminal page with no auto-refresh -- stranding the user.
+
+The landing fallback now consults `list_restorable_workspace_ids()` (the live primary agents unioned with the persisted last-good topology). When that set is non-empty -- we know workspaces exist even though a partial cold-start snapshot hasn't re-surfaced them -- it renders the auto-refreshing "Discovering agents..." page (which self-heals into the workspace list) instead of the create form. The create form is now only shown when discovery has completed and nothing anywhere (live, remote, or last-good) says a workspace exists, i.e. a genuine first-run user.
+
+As defense-in-depth, the create form, when it is the landing fallback at `/`, now also subscribes to the chrome SSE and navigates to `/` the moment a workspace appears, so a user shown the form on a cold-start race is taken to their workspace list. The explicit `/create` page keeps its previous behavior (no self-heal SSE) so a deliberate "create another workspace" flow is never bounced away by existing workspaces.
--- a/apps/minds/imbue/minds/desktop_client/app.py
+++ b/apps/minds/imbue/minds/desktop_client/app.py
@@ -1012,11 +1012,32 @@ def _handle_landing_page() -> Response:
         )
         return make_html_response(content=html)
 
-    # No agents discovered yet. If discovery is still in progress, show a
-    # "Discovering agents..." page with auto-refresh. Once discovery has
-    # completed with no agents found, show the create form so the user can
-    # create their first agent instead of polling forever.
-    if not backend_resolver.has_completed_initial_discovery():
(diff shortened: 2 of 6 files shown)


joshalbrecht:
lgtm

PR #2519, opened 2026-07-16
Report and pin the mac runner in minds-launch-to-msg

weishi-imbue:
## What

`minds-launch-to-msg` now reports **which self-hosted mac runner** ran each job, and can pin a run to one machine.

- `build` and `launch_to_msg` export a `runner_name` output (from `${{ runner.name }}`); the Slack status message gains a `[ran on build: <name> | verify: <name>]` line — so with more than one mac runner in the pool you can tell which physical machine a run (or failure) landed on.
- New `runner_label` `workflow_dispatch` input pins build + verify to a single runner by label (e.g. `mac-runner-2`); empty keeps the shared `minds-runner` pool. Use it to isolate a run to one machine and test whether a failure is machine-specific. The target runner must carry the given label.

## Why

We now have a second office mac runner (`mac-runner-2`). Reporting the runner name makes failures attributable to a physical box, and the pinning input lets us reproduce a suspected machine-specific failure on demand.


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#35 from the same branch. This public repository is now a read-only mirror.

PR #2521, opened 2026-07-17
minds: make the OAuth login flow feel responsive

gdenisov:
## Summary

Polishes the Google OAuth sign-in flow on the Minds login page end-to-end so the browser round-trip feels responsive instead of frozen, and removes the non-functional GitHub button. All changes are in the `minds` desktop client; the browser-side "Authorization successful" callback page is intentionally untouched (per the plan).

Plan: `blueprint/login-auth-flow-polish/plan-login-auth-flow-polish.md`

## Changes

- **Provider-button feedback:** on click, that provider's brand logo is *replaced* by a same-size (18px) spinner in one shared slot (no layout shift), and both provider buttons fade for the duration (`OauthButton.jinja`, `auth.js`).
- **Staged "blue box" messaging:** replaces the single frozen line with a sequence anchored to real signals — "Opening your browser..." on click, "Waiting for you to finish signing in with Google..." once the flow starts, and "Finishing up..." while the account is wired up (`auth.js`).
- **Bring the app to the front on success:** a `bring-app-to-front` IPC channel activates the whole app over the frontmost browser (macOS `app.focus({steal:true})`) and raises the window, only if it wasn't already focused. Wired for both the standalone `/auth` content view (via the content-relay `minds:bring-app-to-front` postMessage) and the sign-in modal overlay (via the `window.minds` bridge). The renderer sends the request, then defers the post-auth navigation ~150ms so the message is delivered before the page navigates away — fixing an intermittent no-raise race (`preload.js`, `content-relay-preload.js`, `main.js`, `auth.js`).
- **Earlier + faster focus:** a new server-side `finishing` flow state is recorded as soon as the sign-in is written to disk (before mirroring the session / registering the provider / bouncing the latchkey-forward supervisor), so the app comes forward and shows "Finishing up..." while mngr finishes, then navigates on `done`. The status poll interval dropped from 2000ms to 500ms (`supertokens_routes.py`, `auth.js`).
- **In-page errors:** OAuth failures/timeouts/lost-flows now render in the in-page error box and reset the buttons, replacing the browser `alert()` popups (`auth.js`).
- **Removed the "Continue with GitHub" button:** GitHub OAuth is only registered on the connector when `GITHUB_CLIENT_ID`/`SECRET` are set, which the current deployment does not do, so the button was non-functional. Google is now the only third-party option; the generic OAuth machinery (component, connector provider block, `mngr` CLI choice) is left intact so it can return once credentials are configured (`AuthForm.jinja`).
- **Bug fix (spinner-at-rest):** the click spinner was visible on every OAuth button at rest — `hidden` and `Spinner`'s `inline-block` collided in the compiled CSS. Moved `hidden` onto a plain wrapper span (the `PermissionsForm` pattern) (`OauthButton.jinja`).

login_button_before_after
login_feedback_before_after
login_feedback

## Testing

- `just test-quick apps/minds/imbue/minds/desktop_client` — 1464 passed, 0 failed.
- `apps/minds/imbue/minds/test_ratchets.py` — 57 passed; `ruff check`/`format` clean.
- New tests: the click-spinner wrapper markup; the server flow is `finishing` during mirroring and resolves to `done`.
- Verified in a headless browser: the logo→spinner swap (no layout shift), the staged messages including "Finishing up...", the `finishing → done` UI transition, and that the bring-to-front request is delivered ~150ms before navigation.
- **Not scriptable headlessly:** the macOS cross-app window activation itself (needs a real-app check); the live Electron OAuth round-trip is covered by the electron-flow/acceptance suites in CI.

## Screenshots

Before/after media is generated in the workspace (`login_button_before_after.png`, `login_feedback.gif`, `login_feedback_before_after.png`) — drag them into this description to embed.


joshalbrecht:
lgtm!

these seem like good improvements.  There's probably plenty more to fix about this form / the signin process in general, but at least this makes it better than where we are today

PR #2522, opened 2026-07-17
Pass MINDS_RESTIC_BINARY to dev-mode backends; check desync in ensure-binaries

joshalbrecht:
## Summary

Follow-up to #2512, from auditing which bundled binaries a from-source install actually receives.

- **Dev-mode restic wiring**: `backend.js` only passed `MINDS_RESTIC_BINARY` in packaged mode. Dev-mode backends (`pnpm start` / `just minds-start`) fell back to `restic` on PATH, so backup provisioning failed on any machine without a system restic — even though the `prestart` hook downloads the pinned restic into `apps/minds/resources/restic/` precisely so nobody has to install it (macOS devs were masked by brew's restic). Dev mode now sets the variable, matching packaged mode and the documented intent in `paths.js`.
- **ensure-binaries completeness**: the `REQUIRED` list in `scripts/ensure-binaries.js` omitted desync, so a partially populated `resources/` (e.g. a checkout predating the desync bundling) skipped the download and left desync missing.

## Test plan

- `node --test test/unit/*.test.js` (all pass) and `node --check` on both files.
- Manually verified on the WSL test box: relaunched the dev-mode app and confirmed the backend process env contains `MINDS_RESTIC_BINARY=<repo>/apps/minds/resources/restic/restic`, and that binary executes (`restic 0.18.1 ... linux/amd64`). Before the fix the env var was absent and `restic` was not resolvable on the backend's PATH.
- No Python changes; the env assembly is inline in `backend.js`'s spawn path, which has no existing unit-test seam — left as-is rather than refactoring for testability in a two-line fix.


joshalbrecht:
lgtm

PR #2526, opened 2026-07-17
Fix stale workspace docs: a workspace is a host, not an agent

qi-imbue:
## Why

While answering a question about how minds concepts map onto mngr primitives, the glossary turned out to state the opposite of what the code does.

`apps/minds/docs/workspace/glossary.md` defined a workspace as "a persistent mngr **agent** created from a template repository via `mngr create`... Each workspace is labeled with `workspace=<name>` for discovery." Both halves are stale:

- A workspace is a mngr **host** (every launch path passes `--new-host`), holding one primary agent plus N chat/worktree/worker agents. The one-agent model was replaced by the services/chat split.
- The `workspace` label was removed entirely when workspace names were split into a `workspace_display_name` label on the primary agent plus a normalized host slug. Discovery now keys off `is_primary`.

## What changed

- **glossary.md**: workspace redefined as a host. Added entries for the four agent kinds that live in one: primary (`system-services`, `is_primary=true`), chat (`--transfer none`, shares the primary's work_dir, one per chat tab, uncapped), worktree (`--template worktree`, own git worktree, `user_created=true`), and worker (agent-created via `launch-task`, `agent_created=true`, not tied to a tab).
- **glossary.md**: the launch mode entry listed a nonexistent `CLOUD` mode (the real enum member is `VULTR`) and omitted `AWS` and `MODAL`.
- **design.md**, **user_story.md**: the shown `mngr create` command still passed the removed `--label workspace=<name>` and lacked `--new-host`, so following it would not reproduce what minds actually runs.
- **api_v1.py**: a comment referenced the removed label (`is_primary`+`workspace`). The code below it already didn't use it.

Docs and one comment only -- no behavior change.

## Left for a follow-up

`_build_worktree_create_command`'s docstring in the default-workspace-template repo (`apps/system_interface/.../agent_manager.py`) says "The **worker** belongs to its workspace by sharing the host" while building the **worktree** agent command. That's a different repo, so it needs its own PR.

### DIFF
--- a/apps/minds/changelog/mngr-fix-workspace-glossary.md
+++ b/apps/minds/changelog/mngr-fix-workspace-glossary.md
@@ -1,0 +1,7 @@
+Fixed: documentation drift around what a workspace is. The glossary described a workspace as a single mngr *agent* labeled `workspace=<name>`, but a workspace is a mngr *host* holding several agents, and the `workspace` label was removed some time ago (discovery keys off `is_primary`).
+
+The glossary now defines a workspace as a host, and adds entries for the four agent kinds that live in one: the primary `system-services` agent, chat agents, worktree agents, and worker agents.
+
+Fixed: the launch mode glossary entry listed a nonexistent `CLOUD` mode (the real one is `VULTR`) and omitted `AWS` and `MODAL`.
+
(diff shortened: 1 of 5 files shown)


joshalbrecht:
lgtm

PR #2528, opened 2026-07-17
minds: harden snapshot-resume service-registration test (poll + shed-aware browser)

DanverImbue:
## What

Hardens `test_resumed_workspace_registered_expected_services` in `apps/minds/test_snapshot_resume.py` with two changes.

**1. Re-land poll-until-registered.** The test read `runtime/applications.toml` exactly once after resume, but the app-watcher re-registers services asynchronously (each writes its port via `scripts/forward_port.py` as it comes back up), so a read that landed before a service finished registering failed intermittently. Re-lands the poll-until-registered mechanism (reverted in `6196864805` only because its expected set still included the template-removed `web` service): poll the file shell-side inside `docker exec` until the expected services appear or a 120s deadline passes, then assert. Core set is now `system_interface` + `terminal`.

**2. Restore shed-aware `browser` coverage.** `browser` autostarts and registers via `forward_port.py` *before* launching its memory-heavy browser-service/Chromium, and nothing ever removes an `applications.toml` entry, so the only way `browser` is absent after a resume is that its supervisord program never finished registering. The test now requires `browser` **unless** there is positive evidence it was shed under memory pressure -- supervisord recorded its program dying on an OOM signal (`terminated by SIGKILL`/`SIGTERM`, or `exit status 137`/`143` when the killed child is `forward_port.py` and the bash wrapper propagates it). A bare "never re-registered", with no shed signal, is a real regression and fails. This distinguishes the shed case (which `86a1f15ac8` dropped `browser` to avoid) from a genuine registration failure.

The shed signal is scoped to the **current (post-resume) supervisord instance** via an `awk` pass that resets on each `supervisord started with pid` marker -- `supervisord.log` is append-only and survives the snapshot, so a browser signal-kill logged by the snapshot's own build or `docker stop` shutdown must not be mistaken for a post-resume shed. A missing marker safely yields "not shed" (an absent browser becomes a hard failure), so it can never mask a regression.

## Validation

This is a `minds_snapshot_resume` test that needs the Modal snapshot infrastructure and does not run in normal CI. Validated locally:

- Built a snapshot image (`scripts/snapshot_minds_e2e_state.py`) and ran the test via `just test-offload-minds-snapshot` -- 1 passed.
- Resumed the snapshot directly and confirmed ground truth: `browser`, `system_interface`, `terminal` all re-register in `applications.toml` (freshly `spawned` post-resume); the `supervisord started with pid` scoping marker is present for both the build and post-resume instances; the shed diagnostic correctly reports `BROWSER_SIGNAL_KILLED=no` on a healthy resume.
- Offline scenario tests of the `awk` decision: pre-snapshot kill (scoped out), post-resume shed (SIGKILL / exit 137), code-exit regression, graceful stop, and missing-marker all classify correctly.

## Changelog

`apps/minds/changelog/minds-poll-resume-registered-services.md`.


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#36 from the same branch. This public repository is now a read-only mirror.

DanverImbue:
Continuing on imbue-ai/mngr-internal#138 -- development has moved; this mirror is read-only, so PRs cannot merge here.

PR #2514, opened 2026-07-16
Add proof-of-concept behavioral spec for desktop-client sign-in/session/landing

DanverImbue:
## Summary

First experiment toward behavioral testing for minds: a behavioral specification kept as an artifact distinct from the tests that validate it.

- Adds `apps/minds/specs/authentication.md` covering the desktop client's sign-in (one-time login codes), session lifetime/integrity, landing-page routing, post-sign-in destination, and the one-sign-in-opens-every-workspace bridge.
- Scenarios are Gherkin flow paths with stable tags; cross-cutting **invariants** (INV-1..8) state properties that hold across all scenarios.
- A traceability appendix maps every scenario tag and invariant to the existing tests that verify it (many-to-many), explicitly marking partial coverage and gaps (e.g. no test covers session expiry, and no test asserts that fetching the login URL leaves the code spendable).
- Records two documentation divergences found while writing the spec: `desktop_client/README.md` still describes the old `Domain=localhost` cookie and the single-agent landing redirect.

Documentation only; no runtime behavior changes.

Stacked on #2366 (`danver/embed-git`) because this branch was cut from that work; the spec content is independent of it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2530, opened 2026-07-17
[MIND-124] Define the minds behavioral-spec language: skill, specs CLI, and authentication corpus

DanverImbue:
## Summary

Defines the behavioral-spec language for minds and delivers it end to end: the definitional skill, the `minds specs` CLI, and the first spec corpus.

- **Skill** — `.claude/skills/minds-behavioral-specs/SKILL.md` is the definitional reference for the language: `.feature` files under `apps/minds/specs/`, validity decided by `gherkin-official` (classic syntax, English keywords), kebab-case folders/basenames/tags, first-tag identity with folder-derived coordinates (`@fresh-code` in `authentication/signin.feature` is `authentication.fresh-code` — file basenames never qualify), invariants as `Rule:` blocks scoped by file placement (reserved `invariants.feature` = folder-and-below; reserved `overview.md` for prose), and the `witnesses` test back-link convention. Reviewed by a second agent through the de-complect and crispy-comments lenses before landing.
- **CLI** — `uv run minds specs validate | list | query`: a TDD-built scan engine (`imbue.minds.core.behavioral_specs`) plus click group; `validate` enforces the language rules with located errors, `list`/`query` emit JSONL (one record per unit: coordinate, kind, name, file, line, tags, steps, parent). Adds the `gherkin-official` dependency. Documented in `apps/minds/docs/behavioral-specs.md`.
- **Corpus** — `apps/minds/specs/authentication/`: 32 units (24 scenarios/outlines across `signin`/`session`/`landing`/`post-login`/`workspace-bridge`, plus 8 subtree-scoped invariant Rules), re-expressing the essence of the interim `authentication.md` (created and replaced within this PR), grounded against the actual desktop_client and mngr_forward behavior. A permanent unit test asserts the live corpus always validates.
- **Back-link support** — the shared `witnesses(coordinate, partial=...)` pytest marker is registered monorepo-wide and documented in `testing-overview.md`. Annotating the authentication corpus's tests is deliberately deferred to a follow-up PR.

Plan: `blueprint/minds-behavioral-specs/plan-minds-behavioral-specs.md`.

Supersedes #2514.


joshalbrecht:
re: this:
> 32 units (24 scenarios/outlines across signin/session/landing/post-login/workspace-bridge, plus 8 subtree-scoped invariant Rules),

we actually might be changing a decent number of things around the sign-in experience, so I wouldn't go too crazy on testing it quite yet (ex: we might even want some sign in to be driven in the user's browser, etc)

(we were just talking about it at standup)

libs/mngr_specs/README.md
```diff
@@ -0,0 +1,38 @@
+# mngr-specs
```
joshalbrecht, line 1: I think it'd be better to have a different name -- "mngr specs" is a bit *too* generic for what this is doing / is responsible for

this is something more like an internal tool for us right now -- I'm not even sure it if should be a mngr plugin?  (vs an entry in scripts or something, or at the very least, something more like mngr_imbue_bdd_specs)

libs/mngr_specs/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+- `validate` - parse every spec file and enforce the corpus language rules.
+- `list` - emit the corpus as JSONL, one record per authored unit, with
+  structural filters (`--area`, `--tag`, `--unit`, `--name`, `--step`).
+- `matrix` - join the corpus against the `witnesses` markers in a test tree,
+  reporting per-unit coverage.
+
+## The corpus model
+
+The tool is corpus-generic: a corpus is any `<project>/specs/` directory, and
```
joshalbrecht, line 17: putting stuff in /specs/ conflicts with blueprint, IIRC?

we certainly have a bunch of other stuff in folders called "specs" already, so we should be careful not to conflict with that...

apps/minds/imbue/minds/desktop_client/test_desktop_client.py
```diff
@@ -113,19 +113,16 @@ def _authenticate_client(
...
-    The server-side logic the test is exercising is independent of the
-    Set-Cookie emission path; the bare presence/signature of the cookie is
+    The production path (GET /authenticate?one_time_code=...) sets a host-only
+    ``minds_session`` cookie on the bare origin (workspace subdomains get their
+    own session via the forward server's /goto/ auth bridge, not this cookie).
+    Setting the cookie directly on the jar skips the /authenticate round-trip
+    and its one-time-code bookkeeping in tests that only care about being
+    signed in. The server-side logic the test is exercising is independent of
+    the Set-Cookie emission path; the bare presence/signature of the cookie is
     what ``_is_authenticated`` checks.
```
joshalbrecht, line 123: I'm confused about this comment change (and the below one)

I'm not quite sure why there were changed?

apps/minds/specs/authentication/invariants.feature
```diff
@@ -0,0 +1,69 @@
+Feature: Authentication invariants
```
joshalbrecht, line 1: another reason we might want to be careful about making too many tests of the sign in flow -- there is currently some messy language around "sign in" and "authentication"

*some* of that logic is what you expect (signing in to your imbue account, for example). Other code is about one-time codes that are used for XSS and CSRF reasons, but weirdly shares the same language (and often confuses coding agents)

apps/minds/specs/authentication/invariants.feature
```diff
@@ -0,0 +1,69 @@
...
+  describes.
+
+  @single-use-codes
+  Rule: A one-time code grants at most one session, ever
+    Over its whole lifetime, each one-time code is spent at most once, and
+    only by the authentication step that establishes a session. Every later
+    presentation of the same code is refused; no sequence or interleaving of
+    requests can spend a code twice or sign in twice from one code.
+    Rationale: the login URL is written in plain text to a terminal and an
+    event stream; single use bounds the damage of that exposure.
```
joshalbrecht, line 13: I feel like these could be written more clearly

one minor pet peeve of mine is hard-coded wrapped text line this because it makes merge conflicts MUCH harder to read than my alternative, preferred format ("put every sentence on its own line, don't wrap")

Here I might want to go even further--these descriptions should probably be short bullet points, not complete sentences anyway (arguing even more strongly against line-wrapped text)

apps/minds/specs/authentication/overview.md
```diff
@@ -0,0 +1,45 @@
...
+Component: the minds desktop client -- the bare-origin web UI served by
+`minds run` (`apps/minds/imbue/minds/desktop_client/`) -- plus the
+workspace-origin bridge served by the forward server (`libs/mngr_forward/`).
+
+The features in this folder cover sign-in with a one-time login code, the
+session that sign-in establishes, landing-page routing, the post-sign-in
+destination, and the bridge that opens every workspace from one sign-in.
+The Rules in `invariants.feature` bind all of them.
+
+## Glossary
```
joshalbrecht, line 12: gabe has a separate glossary effort, see here:  https://www.dropbox.com/scl/fi/808pv5up523xbwkp8x6wt/Minds-Glossary.paper?rlkey=8sm4nn0bvuukdlu9c44vf9wki&dl=0

we almost certainly want to:
A) unify these efforts
B) unify the terms / settle on terms we're happy with
C) broadcast those terms / do the refactors to make the code nice and consistent in its use of the terms in the glossary(s)

apps/minds/specs/authentication/overview.md
```diff
@@ -0,0 +1,45 @@
...
+- **goto bridge**: the forward server's `/goto/<agent-id>/` route, which
+  converts a valid bare-origin session into a workspace-origin session without
+  user interaction.
+
+## Out of scope
+
+- Imbue-cloud account sign-in (email/password/OAuth, the `/auth/*` pages) --
+  a separate account system layered on top of the local session. Only its
+  funnel point `/post-login` is specified here, because it decides landing.
+- The workspace-creation flow beyond "the create form is shown".
```
joshalbrecht, line 39: oh that's weird, that's almost certainly what I would have expected "authentication" to cover!  (I think it's just a terminology mismatch)

apps/minds/specs/authentication/signin.feature
```diff
@@ -0,0 +1,54 @@
+Feature: Sign-in with a one-time login code
+  At startup the desktop client mints a fresh one-time code and prints the
+  login URL to its terminal. Opening that URL in a browser is the only way
+  to establish a session in a browser that has none.
+
```
joshalbrecht, line 5: I have the same multi-line wrapping comment here

libs/mngr_specs/README.md
DanverImbue: Seems like calling this `bdd-specs` might be what we need to deconflict with this?

apps/minds/specs/authentication/signin.feature
DanverImbue: Sounds good, I can make that a requirement of the skill, and that should impact all the features at once.

apps/minds/imbue/minds/desktop_client/test_desktop_client.py
DanverImbue: IIRC the comments conflicted with the actual tests as they were identified by Claude.


```
Timeline
1. Comment born accurate — 4674ada0a4 (Apr 22, "Add <agent-id>.localhost subdomain routing") At this point production /authenticate really did set Domain=localhost on the session cookie so it would be valid on subdomains. The httpx TestClient cookie jar genuinely was stricter and silently dropped such cookies, so the test helper set the cookie directly. Comment matched reality.

2. Production behavior changed, comment left behind — 1f47f0295a (Apr 22, "Add /goto/<agent-id>/ auth bridge") This is the key commit. Josh discovered that Chromium and curl treat localhost as a public suffix, so Domain=localhost cookies never cross to subdomains — the whole subdomain flow fell over in a landing-page redirect loop. The fix removed Domain=localhost from /authenticate entirely: the bare-origin cookie became host-only, and each subdomain now gets its own cookie via the /goto/ auth bridge. But the test comment (and the docstring above it) still claimed production used Domain=localhost. Drift begins here.

3. Stack migration reworded but didn't re-verify — cd87a73394 (Jun 20, FastAPI→Flask) The helper moved from client.cookies.set(...) to client.set_cookie(...), and the comment was reworded from "httpx's cookie jar silently drops..." to "the test client cookie jar is strict about..." — preserving the stale premise under a new test client, with no evidence the Flask client jar even has the claimed behavior.
```

libs/mngr_specs/README.md
joshalbrecht: yup, that feels much better!

joshalbrecht:
ok, I finished reviewing and didn't have any other comments

lgtm once we address the above!

weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#37 from the same branch. This public repository is now a read-only mirror.

PR #2532, opened 2026-07-18
Fix mngr list --sort to order numeric CEL fields numerically

weishi-imbue:
## Problem

`idle` is a documented sortable CEL field, but `mngr ls --sort 'idle asc'` returned a non-monotonic order (e.g. `111044, 14040.2, 2.42` interleaved). The sort comparator in `_CelSortKeyExtractor` (`libs/mngr/imbue/mngr/cli/list.py`) built every key as `str(value)`, so all numeric CEL fields sorted lexicographically instead of numerically.

## Fix

`_CelSortKeyExtractor` now derives each key via a new `_typed_cel_sort_key`, which returns a `(type_rank, comparable)` pair. The leading rank groups values by type — so values of different types are never compared to each other (which would raise) — while numbers sort numerically and strings lexicographically. Missing values still sort to the end in both directions.

Because it is one shared comparator, the fix covers every numeric CEL field at once (`idle`, `idle_seconds`, `age`, `runtime`, `pid`, …), not just `idle`. String fields are unchanged. `--sort` and `--include` already share one CEL context (`build_agent_cel_context`); only the sort-key derivation was wrong, so no production code beyond the comparator changed.

## Tests

- **Unit** (`_sort_agents_by_cel`): numeric ascending, numeric descending, and missing-value-sorts-last.
- **End-to-end**: drives the real `list_command` via `CliRunner` (`mngr ls --sort 'idle_seconds asc' --format json`) and asserts numeric ordering in the emitted JSON, reusing the existing `FakeApiListAgents` seam (no production change, no new `monkeypatch`).

Tests use `2 / 10 / 100` (numerically `2 < 10 < 100`, but lexicographically `"10" < "100" < "2"`) and were verified to **fail against the pre-fix `str()` comparator** and pass with the fix.

Changelog: `libs/mngr/changelog/mngr-mngr-ls-sort-idle.md`.


joshalbrecht:
lgtm! thanks for the fix

PR #2533, opened 2026-07-18
minds: prune last-good topology on observed DESTROYED (Stream C of 3)

gdenisov:
## Stream C of 3 — resolver prune of last-good topology on observed DESTROYED

This is **Stream C** of a multi-PR minds bug fix. It is confined to the desktop-client backend resolver (`backend_resolver.py` + its test). It does **not** touch the Electron shell (Stream A, PR #2517: `apps/minds/electron/`) or the Python landing handler / templates (Stream B, PR #2518).

### Background: what [Fix D (`dafb06ec3`)](https://github.com/imbue-ai/mngr/commit/dafb06ec3e64ef26c69e1013d6694a4a5858f407) is
Fix D introduced the persisted "last-good agent topology" and `list_restorable_workspace_ids()`. The resolver remembers the last *complete* per-host agent enumeration on disk (`last_good_agent_topology.json`) so a workspace that exists but hasn't been re-discovered yet on a slow cold start still counts as "restorable". The Electron shell's window-restore filter (`filterRestorableUrls`) and — since Stream B — the landing handler both consult this set so they don't drop live windows / dump the user on the create form against a partial cold-start snapshot.

### The problem C fixes
`list_restorable_workspace_ids()` could stay non-empty for workspaces that are actually gone:

1. **Stale last-good:** `_merge_last_good_topology_locked` only ever *overwrote* a host on a complete enumeration and *never removed* one. A destroyed host lingered in last-good forever, so its `system-services` agent stayed in the restorable union indefinitely.
2. **DESTROYED in the live set:** the live half of the union filtered `is_primary` with **no host-state check**, so a workspace whose host is observed `DESTROYED` (it lingers in discovery for the provider's destroyed-host persistence window) still counted as restorable during that window.

Consequence: after a user destroyed *every* workspace while the app was closed, `list_restorable_workspace_ids()` stayed non-empty — keeping `/` on the self-refreshing "Discovering…" spinner (via Stream B) instead of the create form, and keeping the shell restoring dead workspace windows.

### The fixes (both in `backend_resolver.py`)

- **C1 — prune last-good on observed DESTROYED.** `update_agents` now removes from `_last_good_agents_by_host` any host whose fresh snapshot reports `HostState.DESTROYED`, and persists the pruned topology through the existing write path. The merge's add-loop also skips DESTROYED hosts, so a lingering destroyed host is not thrashed back in tick after tick (no write churn during the persistence window). **Pruning happens ONLY on an explicit `DESTROYED` observation, never on mere absence** — declining to drop on absence is the entire point of last-good, and is preserved.
- **C2 — exclude DESTROYED-host agents from the live half of `list_restorable_workspace_ids`.** The live `is_primary` set is now filtered by `host_state_by_host_id.get(...) is not HostState.DESTROYED`, mirroring `list_active_workspace_ids`.

### C2 safety investigation (did C2, not the C1-only fallback)
Fix D deliberately kept DESTROYED in `restorable` and relied on the shell's "navigate away after restore" flow. I confirmed dropping DESTROYED *earlier* is safe and loses no needed behavior:

- The shell's navigate-away flow (`main.js` ~1866-1907) only navigates a window away when the workspace id was seen in `destroying_agent_ids` **this session** (`everSeenDestroying`, an in-memory set reset every launch). In the exact bug scenario (destroy all workspaces while the app is closed), that set is empty on relaunch and the on-disk `destroying/` markers are already cleaned up — so the navigate-away flow never fires, and keeping DESTROYED in `restorable` would *strand* a dead window (restored, never navigated away). C2 makes that window simply not restored.
- `list_restorable_workspace_ids()` is consulted **only** at cold-start restore filtering. The within-session destroy-navigation flow uses the active `workspaces` list (already DESTROYED-filtered via `list_active_workspace_ids`) + `destroying_agent_ids`, which C2 does not touch. So C2 changes no within-session behavior; the only user-visible effect is that a destroyed workspace's window is not re-opened at launch (the user simply doesn't reopen a dead window).

### Tests (`backend_resolver_test.py`)
Added:
- last-good prunes a host once observed DESTROYED and the prune is persisted to disk (verified via `_read_last_good_agent_topology` and a fresh reloaded resolver);
- a host merely **absent** from a later snapshot is NOT pruned while a sibling observed DESTROYED in the same tick is (regression guard for Fix D's core behavior);
- `list_restorable_workspace_ids()` excludes a DESTROYED-host workspace in the live snapshot, and empties entirely after the sole workspace is destroyed (the spinner-vs-create-form bug), staying empty once the lingering agent drops.

Existing Fix D tests (union, cold-start fallback, partial-loss retention) remain and still pass.

### Test commands run (0 failures)
- `just test-quick "apps/minds/imbue/minds/desktop_client/backend_resolver_test.py"` — **82 passed**
- `just test-quick "apps/minds/imbue/minds/test_ratchets.py"` — **57 passed** (no ratchet changes needed)
- `just test-quick "apps/minds/imbue/minds/desktop_client -m 'not tmux and not modal and not docker and not docker_sdk and not acceptance and not release'"` — **1466 passed**
- `uv run ruff format` + `uv run ruff check` on both touched files — clean


joshalbrecht:
lgtm

PR #2534, opened 2026-07-18
minds docs: correct stale, misleading claims in overview.md

weishi-imbue:
## The defects

Three claims in `apps/minds/docs/overview.md` no longer matched the code, in a
way that actively misleads readers (human and agent) about what is exposed and
how a workspace is reached.

- **D1 -- phantom `global` flag.** The doc documented a `global = true` key in
  `runtime/applications.toml`. `scripts/forward_port.py` only ever writes a
  service's `name` and `url`; there is no `global` key, so nothing toggles
  public exposure by editing that file. Removed from prose and the example TOML.

- **D2 -- "tunnel is automatic" read.** The Cloudflare section read as if
  creating an agent wires up public access. It does not. A workspace's tunnel is
  created with an empty ingress (`wrap_ingress([])` -> `http_status:404`,
  `apps/remote_service_connector/.../app.py:1017,1275`), so it routes no
  service. Exposing a service is an explicit, opt-in share that **requires at
  least one email** and installs a **Cloudflare Access policy**
  (`desktop_client/sharing_handler.py:148` rejects an empty allowlist). A shared
  service is always gated behind Cloudflare Access -- never anonymously public.
  The section now states this loudly.

- **D3 -- wrong subdomain port/owner.** The doc said browsers reach a workspace
  at `<agent-id>.localhost:8420`. 8420 is `DEFAULT_DESKTOP_CLIENT_PORT`
  (`config/data_types.py:22`) -- the desktop client's own bare-origin UI. Agent
  subdomains are served over https by the separate `mngr forward` child process,
  on a port the desktop client chooses that falls back to 8421 only when unset
  (`port = get_state().mngr_forward_port or 8421`, `desktop_client/app.py:228`).
  The doc now describes the mechanism and treats 8421 as a default, not a fixed
  address. The same 8420 claim in `docs/workspace/getting_started.md:52-53` is
  fixed too.

## Scope note

The three enumerated claims were symptoms of broader drift: the doc had also
gone stale on the desktop-client stack (Flask, not FastAPI), on where subdomain
forwarding lives (the `mngr forward` plugin), on the auth model, and on the
launch-mode list. This branch therefore rewrites `overview.md` into an accurate
top-level overview rather than patching three lines and leaving adjacent stale
claims in place. A minimal three-defect diff is the alternative if the fuller
sweep is not wanted.

## Verification

Every load-bearing claim was checked against the current source with `file:line`
citations, and the finished doc was adversarially re-checked against the code.
Docs-only change: no code paths touched, no test references these files, and
there is no markdown link checker; all "where to go next" links resolve.

## Out of scope (left for separate changes, to keep this on-ticket)

The same drift appears elsewhere: `imbue/minds/desktop_client/README.md` still
says "FastAPI" and documents in-process subdomain forwarding;
`docs/workspace/glossary.md`, `docs/design.md`, and `docs/user_story.md` carry
the old launch-mode list and the deleted `MngrStreamManager`;
`docs/environments.md` says "three tiers" where the code has four; and the
README does not link the overview.

### DIFF
--- a/apps/minds/README.md
+++ b/apps/minds/README.md
@@ -46,6 +46,7 @@ Then visit the login URL printed in the terminal to create your first agent.
 
 ## Learn more
 
+- [Architecture overview](./docs/overview.md)
 - [Architecture and design](./docs/design.md)
 - [Desktop client internals](./imbue/minds/desktop_client/README.md)
 - [Glossary of key concepts](./docs/workspace/glossary.md)
--- a/apps/minds/README.md
+++ b/apps/minds/README.md
@@ -46,7 +46,6 @@ Then visit the login URL printed in the terminal to create your first agent.
 
 ## Learn more
 
-- [Architecture overview](./docs/overview.md)
 - [Architecture and design](./docs/design.md)
 - [Desktop client internals](./imbue/minds/desktop_client/README.md)
 - [Glossary of key concepts](./docs/workspace/glossary.md)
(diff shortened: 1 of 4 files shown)


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#39 from the same branch. This public repository is now a read-only mirror.

PR #2535, opened 2026-07-18
Fix Intel (x86_64) minds build: stage per-arch native helpers via afterPack

weishi-imbue:
## Problem
The Intel (x86_64) minds desktop build does not launch on Intel Macs — it bundles **arm64** native helpers (`uv`, `lima`, `restic`, `desync`), so the x64 Electron shell dies the moment `env-setup` spawns `uv` (`exec format error` / "Bad CPU type").

## Root cause
The helpers were downloaded for the **build host's** arch, not the **target** arch. `scripts/build.js` and the `beforeInstall` hook both key off `process.arch`, and ToDesktop's `beforeInstall` hook receives **no `arch` parameter** — so an arm64 builder's binaries end up in the arm64, x64, **and** universal builds alike. (`git` is exempt: Apple's `xcrun` git is already a universal Mach-O.)

## Fix (canonical)
Move native-helper staging into an arch-aware **`afterPack`** hook (`scripts/after-pack.js`), which *does* receive the build's `arch` (`x64=1, arm64=3, universal=4`). Each per-arch build stages its own arch's helpers; the **universal** build is lipo-merged from the x64 + arm64 sub-builds by `@electron/universal`. The downloaders are reworked to be **arch-parameterized** (single arch, or `'universal'` → `lipo`).

`beforeInstall` + the Stage-A host-arch downloads are kept as a safety baseline (`afterPack` runs later and overrides); they can be removed once `afterPack` is proven.

## Verification
- **Binary level (local, Rosetta):** reproduced the bug (`arch -x86_64 ./uv-arm64` → "Bad CPU type") and proved the fix (x86_64 `uv`/`restic`/`limactl` run under Rosetta; the `'universal'` fallback runs both slices).
- **App level (local, Rosetta):** launched minds as x86 (x64 Electron 40.10.1 + the x64 helpers) — the **full Home/Workspaces UI came up with no exec-format error**. macOS itself flagged the process as an Intel app (Rosetta deprecation banner), confirming it was genuinely x86.

## Not yet verified — why this is a draft
The ToDesktop **packaging** wiring needs one real cloud build to confirm: that ToDesktop invokes `todesktop:afterPack` per-arch, that `@electron/universal` merges the per-arch `Contents/Resources` cleanly, and that `additionalBinariesToSign` still resolves.

## Follow-ups
- Confirm end-to-end with one `pnpm dist`.
- Move the `build` job off the self-hosted `minds-runner` onto a GitHub `macos-latest` runner + add a `macos-13-large` Intel launch smoke leg (`minds-launch-to-msg.yml`).
- Remove the now-redundant `beforeInstall` / Stage-A host-arch downloads once `afterPack` is proven.

### DIFF
--- a/apps/minds/changelog/mngr-tart.md
+++ b/apps/minds/changelog/mngr-tart.md
@@ -1,0 +1,15 @@
+Fixed the Intel (x86_64) desktop build, which previously bundled arm64 native
+helpers (`uv`, `lima`, `restic`, `desync`) into every build and so failed to
+launch on Intel Macs.
+
+Root cause: the bundled helpers were downloaded for the *build host's* arch,
+never the *target* arch. `build.js` and the `beforeInstall` hook both keyed off
+`process.arch`, and `beforeInstall` receives no target-arch parameter, so the
+arm64 build machine's binaries ended up in the arm64, x64, and universal
+builds alike.
+
+The native-helper download now honors an explicit target arch and runs from a
+new `afterPack` hook (which does receive the build's `arch`): each per-arch
+build stages its own arch's helpers, and the universal build is lipo-merged
+from the x64 and arm64 builds by `@electron/universal`. The downloaders also
+support `arch: 'universal'` (lipo of both slices) as a fallback.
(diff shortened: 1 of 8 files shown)


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#40 from the same branch. This public repository is now a read-only mirror.

Issue #2538, opened 2026-07-18
minds device id depends on a file only mngr's local provider creates, and is read only once at startup


joshalbrecht:
## Problem

On a fresh minds install, every workspace-record reconcile pass logs:

```
WARNING: Skipping absent-host tombstoning: this install has no device id
```

and absent-host tombstoning is skipped for the entire session.

## Mechanism

- The device id is read from `<MNGR_HOST_DIR>/host_id` (`read_device_id` in `apps/minds/imbue/minds/desktop_client/workspace_record_store.py`).
- Nothing in minds creates that file. It is created only as a side effect of mngr's **local provider** running discovery: `LocalProviderInstance.discover_hosts` touches its `host_id` property, which calls `get_or_create_local_host_id` (`libs/mngr/imbue/mngr/providers/local/instance.py`). minds does not otherwise use the local provider for anything -- the device identity freeloads on a provider that happens to be enabled by default.
- `WorkspaceRecordStore.device_id` is a `frozen=True` field populated once at backend construction (`apps/minds/imbue/minds/cli/run.py`, `device_id=read_device_id(mngr_host_dir)`). On a from-scratch install, construction runs before any mngr process has ever executed, so the file does not exist yet and `device_id` is `""` for the whole first session -- even after discovery creates the file seconds later. Only an app restart picks it up.

## Impact

- The warning repeats on every reconcile pass of the first session (and forever, if the local provider is ever disabled in the minds profile).
- Absent-host tombstoning is deferred for the whole session (the guard itself is correct: with an empty id, hosted rows cannot be attributed to this install, and two id-less installs would match each other's rows and destroy them).
- Any locally-hosted record created during that first session gets `hosting_device_id=""` (`build_record_from_resolver` uses `self.device_id` for non-cloud rows), i.e. record provenance is silently wrong for those rows.
- The device-identity mechanism structurally depends on the local provider staying enabled, which minds has no other reason to require.

Issue #2537, opened 2026-07-18
Backups: snapshot listing regularly exceeds its 12s status budget, degrading the backups view


joshalbrecht:
## Problem

The per-workspace backups route (`GET /api/v1/workspaces/<id>/backups`) queries restic against the workspace's R2 repository from the laptop, with a hard 12s cap per invocation (`_STATUS_RESTIC_TIMEOUT_SECONDS` in `apps/minds/imbue/minds/desktop_client/backup_status.py`). In practice this budget is regularly exceeded at app startup, producing log noise and a degraded UI:

```
Terminating subprocess (pid 311682) with SIGTERM because it exceeded its 12s timeout
WARNING: Backup snapshot listing failed for agent-...: restic snapshots failed (exit 130):
```

(exit 130 is restic's "interrupted" code from the SIGTERM; stderr is empty because the process was killed.)

When the timeout fires, the route degrades to an empty snapshot list plus an error string, so the user sees no backup information for that workspace on that request.

## Why it is slow

- Opening a restic repo over the network is many sequential HTTPS round trips to R2: fetch `config`, list and fetch key files, run the deliberately slow scrypt KDF on the password (roughly 0.5-1s of CPU by design), list the `snapshots/` prefix, then GET each new snapshot object individually.
- The retention policy (24 hourly + 30 daily + 12 weekly + 24 monthly, `libs/host_backup` in default-workspace-template) means a mature repo holds up to ~90 snapshot objects.
- The route pays the full repo-open cost twice per workspace: once for `list_snapshots` and again for the `is_backing_up` lock probe (each is a separate restic invocation with its own 12s budget).
- restic's local metadata cache (`~/.cache/restic`) is active and helps on warm repos, but cold caches, R2 latency spikes, and busy startup network traffic (sync pulls, forward streams) push the total past 12s.

## Impact

- Startup log warnings (two per affected workspace per route hit).
- Backups panel intermittently shows no snapshots for configured workspaces.

Rough direction discussed: let the user manually query for the current backup state, and serve the last-known state along with the time it is from, rather than blocking the route on live restic invocations.

PR #2539, opened 2026-07-18
Fix noisy startup logs: legacy-association limbo, subprocess reap wording, connector concurrency

joshalbrecht:
Fixes three sources of scary-but-wrong (or scary-and-slow) behavior seen in the minds production startup log:

## Legacy association limbo (apps/minds)

`_is_definitively_absent_from_discovery` treated *any* provider error as "this discovery pass proves nothing", but providers without credentials (aws/azure/gcp/ovh/vultr on a typical install) emit `ProviderNotAuthorizedError` on *every* poll, so a legacy association pointing at a long-gone workspace could never be dropped and warned on every reconcile pass forever. Unauthorized providers no longer block the drop; genuinely failed polls still do.

Note: a legacy association records only a bare agent id (no provider), so scoping strictly to "the provider it is associated with" is not implementable; filtering out not-authorized errors is the closest available equivalent, and the retired legacy file is renamed (`.pre-sync`), not deleted, so nothing is unrecoverable.

## Subprocess shutdown log wording (libs/concurrency_group)

`_shutdown_popen` logged "Terminating subprocess (pid X) with SIGTERM because a shutdown was requested" even when the child had already exited on its own -- the routine case for single-use workers like the minds warm `mngr` processes, which made every imbue_cloud call log a phantom kill. Already-exited processes are now reaped quietly at debug level with their real exit code; a genuinely live process logs "Stopping subprocess ... because the parent requested cleanup"; the timeout path keeps its alarming (and accurate) wording.

## Connector request concurrency (apps/remote_service_connector)

The connector's Modal function had no `@modal.concurrent`, so Modal delivered one request per container at a time: one slow request (lease SSH provisioning, cold sync pull) forced every other caller to queue or wait out a cold container boot, even with the production/staging warm pool. Now allows 8 concurrent inputs per container. Audited for thread-safety: per-request psycopg2 connections closed in `finally` (all 20 call sites), `FOR UPDATE SKIP LOCKED` lease selection, thread-safe shared Cloudflare `httpx.Client`, lock-guarded paid-status cache, no request-time `os.environ` mutation.

Related issues filed separately (problem statements only): #2537 (slow restic snapshot listing), #2538 (device id depends on local provider's host_id file).


joshalbrecht:
lgtm

PR #2544, opened 2026-07-20
minds: repaint views on wake to fix blank-white-after-sleep

gnguralnick:
## Problem

Diagnosed from Sentry event `fc1dc9eb` (minds-production, 0.3.8) plus the reporter's own local logs: a window comes back from a Mac sleep (or screen lock) showing a **blank white content area**, recoverable only by moving the window or navigating Home-and-back.

Over sleep a `WebContentsView`'s renderer **survives but stops painting** — Electron leaves its GPU/compositor surface detached. Because the renderer never actually *dies*, the existing `render-process-gone` crash-page recovery never fires (confirmed: the all-console `electron.log` had no `[content-crash]` line, and the app never restarted). The pinned-white background just persists, with no recovery and no log trace.

## Fix

Register `powerMonitor` `resume` + `unlock-screen` listeners (in `onReady`, after app-ready) that force a full, **non-destructive** repaint of every open window's views on wake via `webContents.invalidate()`:

- `invalidate()` (per the vendored Electron 40.10.1 `electron.d.ts`) *"Schedules a full repaint of the window this web contents is in."* No reload, so scrollback / websockets / terminal state are preserved. Idempotent, so the macOS resume+unlock double-fire is harmless.
- Invalidate is applied **per view** (chrome/content/modal, guarded on `isDestroyed`) on purpose: the failure mode is the *content* view going blank while the sibling *chrome* view stays fine, so we don't rely on one view's window-repaint reaching a sibling view's detached surface.
- Each wake logs a `[wake-repaint]` line (teed to `electron.log`) so this previously-invisible failure is diagnosable going forward.

## Verification

- `node --check electron/main.js` — clean.
- `node --test test/unit/*.test.js` — 16/16 pass (unchanged; the new wake wiring lives inline in `main.js` like sibling view-iteration code, which is not unit-tested in this repo).
- `invalidate()` and both `powerMonitor` events confirmed present in the installed Electron 40.10.1 type defs.
- Changelog gate passes against `origin/main` (touches only `minds`).

### Outstanding — needs real-hardware verification

The **behavioral efficacy on a genuinely detached surface is not yet confirmed.** A real Mac sleep/wake can't be triggered deterministically in-session, and `invalidate()` is the documented remedy for the stale-surface case but a truly lost GPU context is the kind of thing only a real cold sleep/wake proves. To verify: with a minds window open, let the Mac sleep (or lock) → wake → confirm the content area repaints instead of staying white, and check `electron.log` for a `[wake-repaint]` line. If it's still blank after that, the escalation is a `reload()` fallback (heavier — reconnects terminals).

## Note

- A preexisting flaky test surfaced during review — `test/unit/log-rotation.test.js` intermittently fails with a gzip `Z_BUF_ERROR`. Unrelated to this change; flagging per repo guidance.


joshalbrecht:
lgtm assuming you tested this and it works!

gnguralnick:
i only ever had the issue that this fixes a single time. i have run from this branch and it's been working fine so far but it's hard to produce the exact scenario here deterministically. in my testing (waking from sleep, which you saw this morning) - the logs confirm that the code here is firing, and i haven't noticed any regression

PR #2546, opened 2026-07-20
fix: suppress local workspace duplicated as remote landing tile (MINDS-PRODUCTION-4)

pseay-imbue:
Cherry-pick of the open-seer fixer's commit from the mirror PR
imbue-ai/agentic-mngr#6 (single squashed commit 582bd961, applied here as
414c6ed49; the only addition is a changelog-entry rename to match this
branch).

## What broke
User-filed bug report (Sentry MINDS-PRODUCTION-4, 68 events / 10 users): the
minds landing page showed two rows for one workspace ("gabespace") and no row
for another ("pr-reviewer").

## Diagnosis
The landing list unions two independently-sourced lists: local rows from
backend_resolver.list_active_workspace_ids() and remote tiles derived from
other devices' synced state. A workspace on a host inside its destroyed-grace
window stays in the remote-derived set while also present locally, so it
renders as both a local row and a remote tile; the row layout then displaces
another workspace's entry.

## The fix
Suppress a remote tile whenever the same workspace id is already present as a
local row (with a regression test in test_desktop_client.py). See
imbue-ai/agentic-mngr#6 for the fixer's full diagnosis, verification notes,
and regression-risk assessment.

Sentry: https://generally-intelligent-e3.sentry.io/issues/7575818434/


pseay-imbue:
Closing per Preston: keeping the fix only as the mirror PR imbue-ai/agentic-mngr#6 for now.

PR #2547, opened 2026-07-20
Give each manual bug report a unique Sentry fingerprint

pseay-imbue:
## Summary

Manually-submitted bug reports (submit_manual_bug_report in
libs/imbue_common/imbue/imbue_common/sentry/core.py) shared Sentry's default
message-based grouping, so every report ever filed collapsed into one
catch-all issue — MINDS-PRODUCTION-4 currently holds 68 unrelated complaints
(crashes, hangs, UI bugs) under one issue id. That makes per-report triage,
assignment, and resolution impossible, and it breaks automated triage
(open-seer) that assumes one issue ≈ one root cause: a fixer resolved the
issue's "recommended" event while other events in the same bucket described
entirely different bugs.

Each report now carries a unique fingerprint ("manual-bug-report" + a fresh
UUID), so every submission becomes its own Sentry issue with its own title,
assignee, and lifecycle.

Existing reports in the catch-all issue are unaffected (fingerprints apply at
event ingestion); MINDS-PRODUCTION-4 remains as the historical bucket and
should not be resolved wholesale.

## Test plan

- [x] Extended test_submit_manual_bug_report_sends_tagged_event_even_when_reporting_disabled to assert the fingerprint shape (29 passed in core_test.py)
- [ ] Full suite via CI

### DIFF
--- a/libs/imbue_common/changelog/preston-fingerprint-bug-reports.md
+++ b/libs/imbue_common/changelog/preston-fingerprint-bug-reports.md
@@ -1,0 +1,1 @@
+Manually-submitted bug reports now carry a unique Sentry fingerprint ("manual-bug-report" + a fresh UUID), so each report becomes its own Sentry issue instead of all reports grouping into one catch-all issue. Per-report triage, assignment, and resolution now work; automated triage (open-seer) can treat each report as a distinct issue.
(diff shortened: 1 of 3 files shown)


pseay-imbue:
Closing per Preston: manual bug reports are level:info, so open-seer's sweep query (level:[error,fatal]) never touches them anyway — the catch-all grouping doesn't affect automated triage. Per-report fingerprinting remains an option later if per-report human triage in the Sentry UI becomes desirable.

PR #2548, opened 2026-07-20
minds: show a recoverable error page when a workspace fails to load

gdenisov:
## Problem

When a workspace's content view fails to load at the **network layer** -- the host unreachable/asleep, a TLS failure, or a **dead/stale `mngr_forward` port** left over from a prior app run -- Electron fires `did-fail-load` but nothing handled it, so the view was left painting its pinned-white background: a **silent blank page** with no way forward.

This is the general "show something when the workspace is broken" gap. The existing recovery page only covers backend-observed *health* failures; it does not cover the content view failing to reach the forward origin. (Prod-relevant, not dev-only: the forward gateway picks an ephemeral port via bind-to-0 every launch, so a persisted absolute workspace URL can go stale after any restart; and a genuinely-down remote host produces the same blank.)

## Change

Add a `did-fail-load` handler on the content view that shows a **local error page** (`content-load-error.html`) instead of the blank, mirroring the existing renderer-crash page (`crashed.html`):

- **Retry** re-opens the workspace against the **current** `mngr_forward` origin (via `workspaceUrlForAgent`) rather than reloading the failed URL, so a stale ephemeral port self-heals.
- **Report a bug** opens the existing help modal.

The show/skip decision is extracted into the pure, unit-tested `shouldShowContentLoadError` helper (mirroring the split-out `decideStartupRoute`), so the benign cases are filtered out and only a genuine failure surfaces the page:

- `ERR_ABORTED` (-3) from superseded loads (workspace switches, `will-navigate` preventDefault, in-flight `loadURL` replaced)
- embedded-iframe (subframe) failures
- our own local file:// pages (no recursion)
- shutdown / the full-app error takeover

HTTP errors (4xx/5xx) never reach here -- the `mngr_forward` "Loading workspace" proxy answers 503 as a *successful* load, handled by `did-navigate`.

## Tests

- New `content-error-policy.test.js` (8 cases) covers the filter: real network failures show the page; ERR_ABORTED / subframe / local-page / shutdown / error-state do not.
- All 24 `apps/minds` JS unit tests pass.

## Notes

- This is the first of two PRs. The follow-up (`gleb/minds-recovery-return-to-stale-port`) removes the root cause of the specific stale-port case by canonicalizing recovery-page URLs on persist; the two compose (this PR is the safety net for *any* unreachable-workspace case).
- No packaging change: the whole `electron/` dir ships in the asar, so the new HTML/JS files are bundled automatically.


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#44 from the same branch. This public repository is now a read-only mirror.

PR #2549, opened 2026-07-20
minds: canonicalize recovery-page URLs on persist so a stale port can't blank the page

gdenisov:
## Problem

Quitting the app while a workspace window was on the **recovery page**, then relaunching, lands you on a **blank page**.

The recovery URL is `/agents/<agent>/recovery?return_to=<absolute mngr_forward url>`. Its `return_to` embeds the run's **ephemeral forward port** (the latchkey/mngr_forward gateway picks its port via bind-to-0 on every launch), so it is dead after any restart. `saveSessionState` persisted that recovery URL **verbatim** into `window-state.json`:

```json
"url": "/agents/agent-85b7…/recovery?return_to=https%3A%2F%2Flocalhost%3A49564%2Fgoto%2Fagent-85b7…%2F"
```

On the next launch the restored window reopens the recovery page, which 302s to the dead `return_to` (`49564`) → `ERR_CONNECTION_REFUSED` → blank. This affects packaged **prod** builds too, not just dev — the forward port is ephemeral everywhere, and recovery pages happen in prod.

Root cause: `toPersistedContentUrl` canonicalizes a live *workspace* URL to the port-independent `/goto/<agent>/`, but a *recovery* URL slips through unchanged because `parseWorkspaceId` doesn't recognize the `/agents/.../recovery` path — so the ephemeral port gets frozen in.

## Change

Canonicalize recovery URLs the same way live workspace URLs already are:

- **Persist:** `toPersistedContentUrl` now maps `/agents/<agent>/recovery` → `/goto/<agent>/` (via a new `parseRecoveryAgentId`), so no ephemeral port is ever frozen into the saved state. On restore the workspace loads directly against the **live** forward origin; if it's still unhealthy, the normal health flow re-issues a recovery page with a fresh `return_to`.
- **Restore / filter (legacy heal):** `toRestoredContentUrl` and `filterRestorableUrls` also recognize a legacy recovery URL, so a `window-state.json` already saved in the broken form self-heals on its first restore (rebuilt to the live workspace) rather than reopening to a dead link.

The pure persist-side helpers (`parseWorkspaceId`, `parseRecoveryAgentId`, `toRelativeBackendUrl`, `toPersistedContentUrl`) are extracted into `electron/workspace-urls.js` so they're unit-testable outside Electron (mirroring the split-out `decideStartupRoute`).

## Tests

- New `workspace-urls.test.js` (8 cases), anchored on the exact real stale-port recovery URL: asserts it canonicalizes to `/goto/<agent>/` and that the persisted result carries **neither** the dead port nor the `return_to`. Also covers workspace/goto/backend-relative round-trips.
- All 24 `apps/minds` JS unit tests pass.

## Notes

- Second of two PRs. The first (`gleb/minds-content-load-error-page`, #2548) is the general safety net: it shows a recoverable error page for *any* unreachable-workspace load. This PR removes the root cause of the specific stale-port case. They compose.


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#45 from the same branch. This public repository is now a read-only mirror.

PR #2552, opened 2026-07-20
Decline panel-opening slash commands in mngr message

weishi-imbue:
## Summary

TUI-local slash commands like `/cost`, bare `/model`, and `/login` open an interactive panel that a headless send can never drive. After the durable-evidence rewrite (#2475) they returned exit 0 under the relaxed policy — and left the agent wedged on the panel, so the *next* message was pasted into the panel's own field and failed with a confusing diagnostic.

Two layers, in order:

**1. Upfront decline (primary).** `InteractiveTuiAgent` gains a `_validate_outgoing_message` hook (default: accept everything); `ClaudeAgent` declines the panel-openers before anything is sent:

```
ERROR: '/cost' opens an interactive panel that cannot be driven over 'mngr message'.
Connect to the agent with 'mngr connect blockprobe' to run it.
```

The list is **measured, not guessed** (Claude Code 2.1.214, per-command input-row/mode-bar probing): `/login /cost /usage /status /mcp /plugin /add-dir /config /help /theme /permissions /hooks /resume /memory` panel regardless of arguments; `/model` and `/effort` panel **only when bare**, so `/model opus` and `/effort high` still pass. `/login` is declined for the unauthenticated case, where it opens an OAuth flow a headless agent can never complete. Unknown commands and user-defined commands are unaffected.

**2. Panel detection (defense in depth).** A send to an agent *already* showing a panel — e.g. one a human opened over `mngr connect` — is declined fast via a new `InteractivePanelIndicator` in the existing preflight dialog check, matching the panels' Escape-to-dismiss footer (`Esc to cancel` / `Esc to clear`). That footer is absent from the ready and mid-response states, so ready and busy agents are unaffected.

## Measured live (macOS, real agent)

| case | before | after |
|---|---|---|
| `mngr message /cost` | exit 0, agent wedged on panel | **declines 1.9s, agent untouched** |
| bare `/model` | exit 0, agent wedged | **declines 1.8s** |
| `/model opus` | works | works (2.4s) |
| normal prompt | works | works (2.6s) |
| followup to a human-opened panel | pasted into panel, ~17s fail | declines ~2s via the indicator |

## Maintenance note

The always-panel list is coupled to Claude Code's command set and may need updating as upstream changes; the indicator layer catches anything the list misses (at the cost of hitting the *next* send rather than the command itself).


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#47 from the same branch. This public repository is now a read-only mirror.

PR #2553, opened 2026-07-20
Don't bake the operator's git identity into pool hosts

gnguralnick:
## Problem

Git commits made in an imbue cloud workspace are attributed to the operator who baked the pool host (e.g. `Josh Albrecht <josh@imbue.com>`), not the user.

Root cause: mngr's cross-host `mngr create` (GIT_MIRROR) copies the *source host's* `git config user.name/email` into the target checkout (`hosts/host.py:1756-1757` -> `1802-1805`). That's the desired behavior for a local worktree agent (you want your own identity), but a pool-host bake is a cross-host create *from the operator's machine*, so it stamps the baker's identity into the baked services checkout at `/mngr/code`. Every adopting user's agent then mirrors/worktrees from `/mngr/code` and inherits it as its commit author.

## Change

`finalize_baked_pool_host` now resets `/mngr/code`'s repo-local git identity to the neutral `minds-bootstrap <bootstrap@minds.local>` (the same value the workspace-template bootstrap uses as its only-if-unset fallback) before the host ships:

- Runs a `git-identity-reset` in-container step **before** the sentinel wait, so it applies even on hosts where the bootstrap never created a chat agent.
- **Best-effort** (logs a warning, non-fatal), matching the sibling `sshd-harden` step: the workspace template's Bash-command rewrite hook is the authoritative per-agent attribution, so this only cleans up leftover *non-agent* commits on the shared checkout, and a transient `git config` failure shouldn't tank an expensive bake.
- Scoped to the bake, not `_transfer_git_repo`: local `mngr` worktree agents take the GIT_WORKTREE path and are unaffected — they still inherit your identity as before.

This complements the per-agent attribution added on the workspace-template side (a PreToolUse hook that sets `GIT_AUTHOR_*`/`GIT_COMMITTER_*` to the agent's live mngr name + `<agent_id>@<host_id>`).

## Scope / caveat

Only newly baked pool hosts get the neutral identity; already-provisioned hosts keep the baked value in `/mngr/code` until re-baked (the per-agent hook still overrides their agent commits).

## Testing

- `uv run pytest .../bake/pool_bake_test.py` -> 17 passed (3 new: neutral-value + ordering, best-effort-on-failure, still-runs-when-no-chat-agent).
- `ruff check` + `ruff format --check` + `ty check` clean on the changed files.
- Not exercised against a real bake (needs imbue cloud infra); unit tests cover the command construction and ordering.


joshalbrecht:
lgtm

PR #2554, opened 2026-07-21
Fast-reject unknown slash commands via Claude's own rejection record

weishi-imbue:
## Summary

A typo like `/zzz` sent through `mngr message` polled the full 15s relaxed window and then reported **plain success**, even though Claude had rejected the command within a second. Claude writes a structured warning record the instant it rejects an unrecognized command:

```json
{"type":"system","level":"warning","content":"Unknown command: /zzz"}
```

This PR watches for that record and turns the typo case into a **~2s return with an accurate message**: `rejected (unknown command); nothing was executed`.

## Design

- **Engine (`libs/mngr`, agent-neutral):** `SubmissionEvidenceProbe` gains `is_rejection`; when the confirming probe is a rejection probe, the outcome carries `is_rejection`, and `send_message` logs the accurate warning and records a `send_rejected_by_agent` delivery event. Exit stays 0 — delivery semantics: the input demonstrably reached the agent; the agent's verdict on it is *reported*, not hidden.
- **Claude plugin (`libs/mngr_claude`):** two rejection probes under the relaxed policy (raw + native transcript), scanning only bytes appended past the pre-Enter baseline. The jq filter requires the structured fields (`type=system`, `level=warning`) **and** the `Unknown command` content prefix, so other system warnings can never register as a rejection.

## What about user-defined commands?

Unaffected, by construction. The probe is keyed on **Claude's own runtime verdict**, not any list of known commands: built-in, plugin, and user-defined commands never write this record (they produce normal accepted records, which confirm via the existing probes — measured ~2.5s). Only input Claude itself failed to resolve fires it — including a typo *of* a user-defined command, which is exactly the case that should reject.

## Measured live (macOS, real agent)

| case | before | after |
|---|---|---|
| `/zzzfastreject` (typo) | 17.1s, reported success | **2.7s, accurate rejection warning + agent event** |
| `/effort high` (valid local) | — | 2.4s, unaffected |
| normal prompt | — | 2.6s, unaffected |

A pre-baseline warning from an *earlier* typo cannot re-fire the probe (byte-offset baseline; pinned by test), and repeated identical typos are safe for the same reason.

## Tests

- Engine: rejection-flag mapping when a rejection probe confirms; no flag for acceptance probes.
- Plugin: fixture-transcript test — pre-baseline warning does not fire; accepted records and non-warning system records do not fire; this submission's warning fires. Relaxed probe-set test updated.
- 311 tests pass locally across `tui_utils`, `tui_agent`, and the Claude plugin.


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#48 from the same branch. This public repository is now a read-only mirror.

PR #2564, opened 2026-07-21
Workspace switching shows a 2-second loading bar instead of a blank screen

pseay-imbue:
[video](https://www.loom.com/share/31dc04595f16491e8a7ced7706ff3071)

Switching workspaces previously showed a blank white card while the target workspace loaded. The titlebar wrapper's loading card now carries a centered loading bar that fills over two seconds (ease-out, holding near-full for slower loads), tinted with the workspace accent when it is known server-side.

The bar rides the loading card itself, so no show/hide coordination is needed: the workspace content covers it the instant it paints, instant flips back to a still-resident workspace never show it, and the animation restarts on every switch because the wrapper re-renders per navigation (full load or in-place swap).

Includes a template test pinning the bar and its 2s duration, and a changelog entry.


gdenisov:
lgtm in terms of the loading bar — nice touch using the title bar color! Of course go from this bar to "Loading workspace" is a bit annoying we should think about if / how we can make this experience feel more continuous on the template side.

Deferring to @joshalbrecht for swap logic updates

weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#51 from the same branch. This public repository is now a read-only mirror.

pseay-imbue:
Continuing on imbue-ai/mngr-internal#100 — development has moved; this mirror is read-only, so PRs cannot merge here.

PR #2566, opened 2026-07-21
Add HostState.UNREACHABLE and stop degraded observations reading as CRASHED

gnguralnick:
## Summary

Base of a two-PR stack (the minds recovery-verdict-policy work in #2485 stacks on this). This PR is the mngr-layer host-classification vocabulary change alone, so it can be reviewed by the provider/mngr owners independently.

- **Split the overloaded `HostState.UNAUTHENTICATED` into two states that need opposite handling:**
  - `UNAUTHENTICATED` now means our access credential was rejected at the host's access boundary (e.g. imbue_cloud's outer SSH refusing this machine's key): observation of the workspace is impossible and a restart routes through the same rejected key, so it is **terminal** rather than restart-worthy. imbue_cloud's outer-SSH auth-rejection fallback mints this.
  - **New `HostState.UNREACHABLE`** means the host was observed up but its inner sshd is not answering (a running container whose inner sshd died, or an inner-SSH connection error), which **a host restart can revive**. This is the condition the generic, docker, and imbue_cloud providers previously reported as `UNAUTHENTICATED`.
  - Both follow `UNKNOWN` semantics in the CLI: listed by default, never GC'd, non-terminal for `mngr wait`.
- **Degraded observations no longer mint `CRASHED`**: imbue_cloud's empty/unrecognized container state and lima's `"Unknown"` (or any unrecognized) status now surface `UNKNOWN`, so "could not determine state" is treated as non-evidence — consumers that auto-restart off `CRASHED` (e.g. the minds recovery page) do not do so off it, and the host is not made GC-eligible. Lima's `"Broken"` (positive breakage) stays `CRASHED`.

Resolves the provider-vocabulary deferral flagged in PR #2247. The minds consumers that render these states (the terminal "Can't connect" page for `UNAUTHENTICATED`; the consent-gated host restart for `UNREACHABLE`) live in the stacked PR #2485.

> **Note:** the `UNAUTHENTICATED`/`UNREACHABLE` names were swapped mid-branch (`Swap UNAUTHENTICATED/UNREACHABLE so the names track the condition`) so each name tracks its condition. The description above reflects the final, merged code; an earlier version of this PR body had the two inverted.

## A known tension: RUNNING vs UNREACHABLE is observer-effort-relative

`UNREACHABLE` ("up, but we couldn't get inside") is not cleanly separable from `RUNNING` across providers, because providers observe liveness through different channels:

- **Out-of-band discovery reads `RUNNING` without ever touching the inner sshd.** docker's periodic `discover_hosts` derives `RUNNING` from the docker daemon's container status (explicitly avoiding an SSH round-trip); imbue_cloud reads container state + certified data via an outer-host `docker exec`. Neither connects to the workspace's inner sshd, so a container whose inner sshd has died still reads `RUNNING`.
- **`UNREACHABLE` is only minted by a path that *attempts* the inside and fails** — the online `get_host_and_agent_details` fallback (`get_connection_error_fallback_state`, docker), or imbue_cloud's "container running but certified-data read empty".

So the **same physical condition (container up, inner sshd dead) can read `RUNNING` via an out-of-band path and `UNREACHABLE` via an inside-attempting one** — which value a host gets depends on the observing path's effort, not only on the host. This is deliberate, not a bug: both mean "observed up", and consumers (minds' recovery classifier, GC, `mngr list`) treat them alike for liveness; `UNREACHABLE` merely adds the diagnostic "and we know we can't get in". This PR documents that contract on both enum members — `RUNNING` previously had no docstring, which invited reading it as "we confirmed the inside".

Consequence worth flagging for reviewers/consumers: **the minds discovery path never observes `UNREACHABLE` for a dead inner sshd** — it reads `RUNNING` out-of-band, and its own in-container exec probe catches the dead sshd independently — so minds-visible `UNREACHABLE` is realistically an imbue_cloud "running but uninspectable" signal, not a docker one. A consumer that branches on `host_state` without such an exec-probe backstop would treat an out-of-band `RUNNING` host with a dead inner sshd as healthy.

## Testing

- `mngr_imbue_cloud` `listing_test.py` (15 passed), the imbue_cloud auth-rejection→`UNAUTHENTICATED` reattach test, and the lima `"Unknown"`→`UNKNOWN` / `"Broken"`→`CRASHED` discovery test all pass locally via `just test-quick`.
- `ty check` clean on the changed source files.
- Full CI to run on this PR.


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#52 from the same branch. This public repository is now a read-only mirror.

libs/mngr/imbue/mngr/primitives.py
```diff
@@ -246,14 +246,36 @@ class HostState(UpperCaseStrEnum):
...
     UNAUTHENTICATED = auto()
+    # The host was observed up but its inner sshd is not answering, so we cannot enumerate inside it
+    # (e.g. a running container whose inner sshd died, or an inner-SSH connection error). A host
+    # restart can revive it. A strict refinement of RUNNING (see above): only a provider that
+    # *attempts* to reach the inside can distinguish this from RUNNING, so an out-of-band observer
+    # reports RUNNING for the very same host -- which of the two a host reads is a property of the
+    # observing path's effort, not only of the host. Distinct from UNAUTHENTICATED (our credential
+    # was rejected at the access boundary, where a restart routes through the same rejected key) and
+    # UNKNOWN (transient).
+    UNREACHABLE = auto()
```
joshalbrecht, line 278: unfortunately this doesn't make sense -- this is a little bit too weird of a state

we cannot, generically, treat "unreachable" as a subset of running

consider, for example, the ssh provider: if it cannot access the machine... how can we say that the state is "running, but unreachable"?

gnguralnick:
Continuing on imbue-ai/mngr-internal#104 — development has moved; this mirror is read-only, so PRs cannot merge here.

PR #2550, opened 2026-07-20
Per-user resource quotas and plans for imbue_cloud (blueprint)

joshalbrecht:
Blueprint plan for replacing the paid-account gate with per-user resource quotas and account plans (explorer/ally), enforced in the remote_service_connector.

Plan: `blueprint/cloud-quotas/plan-cloud-quotas.md`

Planning only -- no behavior changes yet. Implementation follows on this branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2567, opened 2026-07-21
R2 storage-quota fixes: correct sweep measurement, cleanup grants, and the creation-time storage gate

joshalbrecht:
Follow-ups to #2550, fixing the issues found in the R2 storage-quota implementation (design in `blueprint/cloud-quota-fixes/plan-cloud-quota-fixes.md`).

## Connector (remote_service_connector)

- The sweep's GraphQL usage query groups by `bucketName` only (one row per bucket, 3h lookback) and **raises on a full page** instead of warning into unread cron logs. The old shape already filled its 5000-row budget at 144 production buckets (measured live) and would have started silently counting buckets as zero usage as the account grew.
- Downgrades are **confirmed against the real-time REST usage endpoint** before any key flips; the analytics window peak alone can only screen or restore, so an account that just cleaned up is never re-broken on stale data.
- New **cleanup grants**: an over-quota account's read-only keys are temporarily restored to readwrite so client-side restic `forget`/`prune` (which needs full write — prune repacks) can run. Grants settle at `POST /account/storage-recheck` or at their 60-min expiry via the sweep; only grants that free nothing count against a rolling budget (5/24h). Migration `015_r2_cleanup_grants.sql`.
- `POST /account/storage-recheck` works standalone (immediate restore once under quota, no waiting for the cron).
- `POST /buckets` refuses when live storage usage is over quota, and fresh mints (create + roll-key) come out read-only for enforced-over-quota owners.
- `POST /admin/sweep/r2 [?email=]` (operator-key auth) runs one sweep pass on demand.
- Per-owner advisory locks serialize all enforcement flips; the sweep skips owners with an active grant.

## Plugin (mngr_imbue_cloud)

- `mngr imbue_cloud account cleanup-grant`, `account recheck-storage`, `admin sweep r2`; typed error for the exhausted grant budget; `R2KeyInfo.enforced_access` (the model previously rejected the field the connector already returns).

## minds

- "Free up backup space" on the Accounts page when over quota: trims the oldest half of each reachable repo's snapshots (never the latest) under a grant, prunes, rechecks, repeats; unreachable buckets reported by name. Detached thread with live progress + OS notification.
- Backup provisioning treats the structured quota 403 as terminal (no more retry-burn).
- `restic restore` retries once with `--no-lock` on a repository-lock write failure, so restores work against downgraded keys.
- Plans-guard test discovers tiers from disk; new live deployment test covering the full downgrade → grant → settle cycle.

## dev

- `.reviewer/settings.json` CI poll timeout raised to 900s (the 600s default was below the ~620s median CI run and reported timeouts as "CI failed").


joshalbrecht:
lgtm after testing

joshalbrecht:
merged via https://github.com/imbue-ai/mngr/pull/2591

PR #2568, opened 2026-07-21
Add onboarding walkthrough to the workspace-creating page

pseay-imbue:
While a workspace is being created, the loading page now teaches how the system works instead of showing rotating tips.

- The progress bar moves to a slim strip at the top (stage caption + collapsible logs stay with it); the rest of the screen is a three-step walkthrough with Previous/Next.
- Step 1 (minds icon): the laptop as an access terminal for an AI operating system, with an interactive demo tab-space and a theme-color picker that restyles the demo live (demo only; not persisted).
- Step 2 (latchkey icon): credentials are encrypted and stored locally, never visible to agents, plus a scrolling carousel of connectable services sourced from the bundled latchkey services catalog (vendored simple-icons brand SVGs, CC0; monogram fallback for services without one).
- Step 3: the full picture -- computer + latchkey connected over an encrypted tunnel to the workspace machine, with copy adapting to local vs cloud workspaces. The server graphic fills in lockstep with the top progress bar.
- No more auto-redirect: when creation is done and the user has seen all steps, a Begin button appears and enters the workspace with a zoom animation.

Notes:
- No latchkey icon asset exists in the repo, so the latchkey tile is a key glyph drawn in the minds-icon style -- swap in the real asset when available.
- Manually verified in a browser (light + dark, all steps, color demo, ready/Begin flow).


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#53 from the same branch. This public repository is now a read-only mirror.

pseay-imbue:
Continuing on imbue-ai/mngr-internal#95 — development has moved; this mirror is read-only, so PRs cannot merge here.

PR #2572, opened 2026-07-22
mngr_ttyd: refit the terminal on websocket reconnect in the vendored client

gnguralnick:
## Summary

Fixes the minds desktop client's intermittent terminal-tab reflow/size bugs at two layers of the stack. The companion system_interface change is in default-workspace-template (imbue-ai/default-workspace-template#310).

**1. `mngr_ttyd`: refit on websocket reconnect (commit d235bcf38d).** The ttyd web client disposes all of its listeners -- including its only refit trigger, `window resize -> fitAddon.fit()` -- on every websocket close, and re-registers them only when the reconnect completes. Any resize landing in the disconnect window is lost; the reconnect handshake then sizes the new PTY to the stale `terminal.cols/rows` and nothing re-measures afterwards. Terminal websockets in the desktop client traverse the latchkey forward and drop intermittently (observed close code 1006), which is why the bug only reproduces there and not against a direct tunnel. A second source patch (`scripts/ttyd_reconnect_refit.patch`) makes the client re-measure on reconnect before the size handshake; `resources/ttyd_index.html.gz` is rebuilt with it.

**2. `mngr`: converge the agent-window fit hook on the live client size (commit c61b7b5e71).** The agent tmux window is pinned `window-size manual` (the anti-birth-collapse policy) and refit by the client-attached/client-resized hooks running `sigwinch_panes.sh fit` with hook-fire-captured geometry. A resize burst fires many overlapping backgrounded instances, and whichever `resize-window` landed last could pin the window at a stale intermediate size -- seen as a dot-filled unused region / partial-width claude TUI that nothing corrects (reproduced live: window 151 cols vs client 158 after a tab reveal). Fit mode now reads the session's current client size at act time and re-fits again after its settle delay, converging regardless of instance ordering.

## Root-cause evidence

- Instrumented the real Electron client over CDP: on every reflow miss, resize events reached the iframe with correct geometry but the fit addon was never invoked (listeners disposed); a later manual `term.fit()` always healed it.
- Deterministic repros for both layers: (a) force a socket close (`tmux detach-client`) and drag the sash during the reconnect window -> cols stayed stale indefinitely without the fix; (b) reveal a hidden terminal tab / rapid sash drags -> tmux window pinned at a stale intermediate size without the fix.

## Test plan

- `just test-quick libs/mngr_ttyd`: 78 passed, 0 failed.
- `just test-quick libs/mngr/imbue/mngr/api/test_connect.py`: 5 passed, 0 failed (includes a new regression test that attaches a real client through a pre-sized pty and invokes the script with stale arguments).
- Verified in the real desktop client against a staging workspace: forced-reconnect repro passes on both agent and persistent terminals (server-side PTY size confirmed via tmux); 280 tab-switch+drag iterations with 0 client-side misses; 22 reveal+drag-burst iterations with 0 tmux window/client mismatches (pre-fix this diverged within 1 iteration).


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#55 from the same branch. This public repository is now a read-only mirror.

gnguralnick:
Continuing on imbue-ai/mngr-internal#115 — development has moved; this mirror is read-only, so PRs cannot merge here.

PR #2576, opened 2026-07-22
minds: default error reporting on during alpha, collapse to a single consent flag

gnguralnick:
## What

During the alpha, minds now reports unexpected errors (with logs) to Imbue by default, and this can no longer be turned off in the app.

- **One flag.** Collapse `include_error_logs` out of existence everywhere. A single `report_unexpected_errors` setting now gates both whether automatic reports are sent and whether their log/traceback attachments are uploaded — so a report always carries full diagnostics. This spans the shared Sentry lib (`imbue_common` `setup_sentry` / `submit_manual_bug_report`), the `mngr latchkey forward` daemon consent contract (`mngr_latchkey` `ForwardSentryConsent`), the minds backend, and the Electron main process.
- **Default on.** `report_unexpected_errors` defaults to `True` (backend `MindsConfig` + the Electron `sentry.js` mirror). The flag is retained so reporting can be made opt-out-able again after the alpha.
- **Informational screens, no opt-out.** The first-launch consent screen and the Settings "Error reporting" section are now informational notices (single acknowledge button / static text). The `/_chrome/error-reporting` endpoint is removed. The `/help` bug-report form drops its logs checkbox, and the backend-down error takeover drops its "Include recent logs" checkbox.
- **App diagnostics always included.** Bug reports now always include app diagnostics (app version, signed-in accounts, the list of workspaces, host/system info — never workspace contents); the `/help` form drops that checkbox too. Per-workspace details and remote access stay opt-in.
- **Honest, privacy-clear copy.** All three surfaces note that reports include diagnostic details that may identify the user (e.g. their signed-in account email), and that **"Imbue will never look into your workspaces without your consent."** Verified against a real Sentry event: manual reports carry `signed_in_account_emails`, an email-bearing `provider_name`, and the OS username in every path — so an anonymity claim would have been false.
- **Migration.** A one-time startup migration (guarded by a persisted `alpha_error_reporting_migrated` marker so it runs at most once per install) flips installs that previously opted out back on and re-shows the notice, so reporting is uniformly on during the alpha. The marker keeps it forward-compatible with reintroducing opt-out after the alpha: a later explicit opt-out must not be flipped back on at the next startup.

## Screenshots

First-launch notice (informational, no opt-out):

First-launch error-reporting notice

Settings → Error reporting (informational):

Settings error-reporting section

Bug-report form (logs + app diagnostics always attached; remote access opt-in):

Bug-report form

## Test plan

- Updated unit tests across `imbue_common`, `mngr_latchkey`, and minds (config, sentry utils, templates, desktop client).
- Ran locally (all green): imbue_common sentry (29), mngr_latchkey sentry (12), minds config/utils/templates + desktop client (342 in the latest round).
- `ruff`, `ruff format --check`, `ty check` clean; Electron JS passes `node --check`.
- Screenshots captured via Playwright against the JinjaX-rendered pages with the compiled Tailwind sheet.
- Acceptance suite runs in CI.


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#56 from the same branch. This public repository is now a read-only mirror.

gnguralnick:
Continuing on imbue-ai/mngr-internal#106 — development has moved; this mirror is read-only, so PRs cannot merge here.

PR #2580, opened 2026-07-22
MIND-142: cut test-offload CI wall clock via coverage/uv overhead removal

DanverImbue:
## Summary

MIND-142. The `test-offload` CI step degraded from p50 ~6.5min to ~8.0-8.8min. Investigation (`/tmp/ci-analysis/OFFLOAD_INVESTIGATION.md`, run 29880875933) showed the time goes to: serialized local discovery (126s), per-batch fixed overhead (~70-80s x 117 batch executions, batches are ~72% overhead), checkpoint misses (~7% of runs, +2min), and coverage tracing tax. **Offload's speculative retry / hedging behavior is explicitly untouched** (stakeholder constraint).

Per review, this PR ships only the fixes that are repo policy. The fixes that offload can absorb as features were rolled back here and filed upstream:

- **OFFLOAD-3** (HIGH): skip coverage tracing during pytest discovery
- **OFFLOAD-4** (HIGH): first-class venv activation for the pytest framework command
- **OFFLOAD-5**: single-pass discovery partitioned per group
- **OFFLOAD-6**: discovery cache keyed on tree SHA
- **OFFLOAD-7**: overlap sandbox creation with discovery
- **OFFLOAD-8**: batch-count / target-duration knob decoupled from max_parallel
- **MIND-143**: follow-up to adopt them here (offload version bump, config adoption, re-measurement) once they land

## Empirical Results

Final scope, measured on CI run 29967754088 (15,870 tests, checkpoint hit -- same cache condition as baseline) against the investigation baseline (run 29880875933, 15,821 tests):

| Phase | Baseline | This PR | Delta |
|---|---|---|---|
| Local discovery (2 group passes, coverage-traced) | 126s | 68.5s | **-57s** (sysmon) |
| Fan-out | 250s | 182s | **-68s** |
| First test result after fan-out start | +128s | +87s | **-41s** |
| `Run just test-offload` step | 6m41s | 4m30s | **-2m11s** |
| `test-offload` job total | ~8.0-8.8min | 5m44s | **~-2.3 to -3.1min** |

A second sample (same code, checkpoint miss from a note-push race) showed the same systematic deltas -- discovery 72.0s, median batch 77s -> 58s (**-25%**) -- but a +39s fan-out: makespan there was set by one 288s straggler batch (baseline max 241s), and LPT straggler variance (~+/-40s) currently exceeds the per-batch win in any single run.

### 1. Per-batch: suppress unused coverage reports (est. -10 to -25s across 117 executions)

Every batch generated `--cov-report=term-missing` + `--cov-report=xml` + a full term-missing-into-`.test_output/coverage_*.txt` (via `--coverage-to-file`) -- each walking every measured file, none ever consumed (CI combines the raw `.coverage` data files on the runner and enforces per-package + combined gates there). Root addopts no longer carry those flags, and `offload-modal.toml` run_args pass `--cov-report=` (pytest-cov 7 clears all reports only when the empty value is the sole `--cov-report`, which is why they had to leave addopts). The `.coverage` data file is still written (parallel=true + pytest-cov's finish-time combine) -- verified in a real Modal smoke run. `just test-unit`/`test-integration` pass the report flags explicitly, so local behavior is unchanged. Per-package pyprojects keep their addopts (single-package batches measure one package only, so report cost there is negligible).

### 2. `core = "sysmon"` for coverage (est. -7 to -10s wall + less jitter)

PEP 669 `sys.monitoring` is ~3x lower tracing overhead than the default C tracer on Python 3.12 (coverage 7.14 here). Measured on a 463-test suite: internal time 41-42s (sysmon) vs 45-46s (ctrace) vs 39-41s (no-cov). sysmon's documented incompatibilities don't apply: no branch coverage, no dynamic contexts, `concurrency = ["multiprocessing", "thread"]` is supported, no coverage plugins. The first CI run's coverage gates all passed with sysmon measurement (per-package + combined 80% gate green).

### 3. Checkpoint inputs: drop per-package pyprojects (est. -4 to -8s average)

`libs/*/pyproject.toml` / `apps/*/pyproject.toml` churn forced a full checkpoint rebuild (+~2min when missed) because sandboxed `uv run` would otherwise attempt a network-less editable rebuild (9449099be2). That failure mode predates `post_patch_cmd` (64646ba72b), which reruns `uv sync --all-packages` after every thin diff with network available -- verified locally that `uv sync` rebuilds editables whose pyproject drifted (dist-info regenerated, no network needed). Dependency changes still invalidate via `uv.lock` (any per-package dep change necessarily changes `uv.lock`). `offload checkpoint-status` with the new inputs still resolves to a cached image (thin diff).

### 4. Regression guard

`test_meta_ratchets.py::test_offload_configs_suppress_per_batch_coverage_reports` asserts: no coverage-report flags in root addopts, and `--cov-report=` in offload-modal run_args.

## Testing

- A/B batch timings on a 463-test suite (sysmon vs ctrace vs no-cov, reports on/off): numbers above.
- Coverage gate path: `.coverage` from a sysmon run combines and reports correctly; `[tool.coverage.paths]` mapping verified against a file produced in a real Modal sandbox.
- `just test-quick test_meta_ratchets.py`: 20 passed, 0 failed.
- Full `just test-offload`/`test-offload-acceptance`: CI on this PR.


joshalbrecht:
thanks for all of these fixes!

lgtm

PR #2582, opened 2026-07-22
minds: point undeployed-env and missing-psql failures at their fix

gnguralnick:
## What

Two `minds` startup/deploy failures now explain how to fix themselves instead of surfacing a cryptic error. Both were hit in practice while bringing up a fresh dev env.

### App/`minds run` startup: missing client config
The `--config-file` option no longer uses `click.Path(exists=True, ...)`, which produced a raw `Invalid value for '--config-file': File ... does not exist` with no guidance. Validation moves into a small extracted helper `_resolve_client_config_path`, which raises a `ClickException` that:
- distinguishes "no env activated" (`None`) from "config file set but missing",
- for the missing case, explains that a dev env's `client.toml` is only written by a successful deploy, and gives the exact `minds env activate --deploy <name>` + `minds env deploy` commands.

This is the common first-run wedge: `minds env activate --create <name>` mkdirs the env root and exports `MINDS_CLIENT_CONFIG_PATH`, but `client.toml` doesn't exist until a deploy runs, so the app (which inherits that path) failed opaquely.

### `minds env deploy`: `psql` not on PATH (macOS)
The "psql binary not on PATH" guidance — shared by `migrations.py` and `neon_db.py` via a single new `PSQL_INSTALL_GUIDANCE` constant (previously duplicated) — now covers macOS correctly. Homebrew's `libpq` is **keg-only**, so `brew install libpq` succeeds but leaves `psql` off PATH. The old message implied the install was sufficient; it now says to add it explicitly:
```
export PATH="$(brew --prefix libpq)/bin:$PATH"
```

## Tests
- New unit tests for `_resolve_client_config_path` (existing file returns path; `None` → activation guidance; set-but-missing → deploy guidance): 3 passed.
- `ruff check`, `ruff format`, and `ty check` clean on the changed files.
- psql-guidance strings are not separately tested (per repo convention, no string-constant tests).

### DIFF
--- a/apps/minds/changelog/gabriel-minds-config-and-psql-guidance.md
+++ b/apps/minds/changelog/gabriel-minds-config-and-psql-guidance.md
@@ -1,0 +1,5 @@
+Two startup / deploy failures now explain how to fix themselves instead of surfacing a cryptic error.
+
+When `minds run` (including the desktop app, which passes the activated env's config path) is pointed at a client config file that does not exist, it no longer dies with a raw `Invalid value for '--config-file': File ... does not exist`. It now reports that the file is missing and, for a dev env, that its `client.toml` is only written by a successful deploy -- with the exact `minds env activate --deploy <name>` + `minds env deploy` commands to run. This is the common first-run wedge: activating a dev env with `--create` but never deploying it.
+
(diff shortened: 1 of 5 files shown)


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#61 from the same branch. This public repository is now a read-only mirror.

gnguralnick:
Continuing on imbue-ai/mngr-internal#114 — development has moved; this mirror is read-only, so PRs cannot merge here.

PR #2589, opened 2026-07-23
Stream kanpan board refreshes for instant, incremental updates

qi-imbue:
## What

Makes `mngr kanpan` refreshes feel fast. Instead of blocking the whole board on the slowest data source (the GitHub PR/CI fetch), kanpan now lists the agents, paints the board immediately, and fills in each column the moment its data source returns.

## How

- `stream_board_snapshot` / `stream_local_snapshot` (fetcher.py) run in a background thread and `post` a progressively more complete `FetchUpdate` onto a queue as each data source finishes. The urwid main loop drains the queue each poll tick and re-renders (`_launch_refresh` / `_poll_refresh_completion` / `_apply_stream_updates` / `_finalize_refresh` in tui.py).
- The initial paint is **seeded from the last-known cached values** (rendered stale by their age), so agents stay in their previous PR sections and only update in place when fresh data lands — the board no longer collapses to a flat list on every refresh.
- Once every source has reported, a fresh value fully governs its column (a stale seeded value it does not reproduce is dropped), so the final board is byte-for-byte what the old all-at-once fetch produced. There's an acceptance test asserting stream-final == batch.
- This subsumes the old local-refresh `_carry_forward_fields`: a local-only refresh simply runs no remote source, so PR/CI keys are never reclaimed and their seeded values persist. That function and its tests are removed.
- Batch and streaming source runs now share one `_compute_source_safely` helper — a single point where a misbehaving data source is caught.

## Testing

- `just test-quick libs/mngr_kanpan` (unit + integration, non-tmux): 478 passed.
- `test_fetcher_acceptance.py` (real agents via local provider): 17 passed, including 6 new streaming tests (base-before-source, seed-then-replace, seed-drop, stream==batch parity, local carry-forward).
- New tui unit tests for the drain/apply/finalize path.
- Manual tmux smoke test of the live TUI: board paints immediately, stays populated in its sections mid-refresh (footer shows `Refreshing`), settles to `Refreshed just now`.
- `vet` (agentic): no issues.

### DIFF
--- a/libs/mngr_kanpan/changelog/mngr-fast-kanpan.md
+++ b/libs/mngr_kanpan/changelog/mngr-fast-kanpan.md
@@ -1,0 +1,5 @@
+Kanpan refreshes now stream in: instead of waiting for the slowest data source (the GitHub PR/CI fetch) before the board updates, kanpan lists the agents, paints the board immediately, and then fills in each column the moment its data source returns.
+
+To keep the board from visibly reshuffling on every refresh, the initial paint is seeded from the last known values (rendered greyed-out as stale), so each agent stays in its previous section until fresh data lands and updates it in place. The first refresh of a session, which has no cached values yet, simply shows the agent list and fills in as sources return.
+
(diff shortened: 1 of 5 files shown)


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#63 from the same branch. This public repository is now a read-only mirror.

PR #2590, opened 2026-07-23
Use an impersonating dispatch curl in minds' latchkey gateway

qi-imbue:
## What

Lets the minds latchkey gateway make Chrome-impersonating requests **on demand**. A caller whose request carries the marker header `X-Imbue-Impersonate:` gets Chrome TLS impersonation (to clear Cloudflare-style fingerprinting); every other request — and all of latchkey's own credential checks — pass through to the system curl unchanged.

Rather than repointing `LATCHKEY_CURL` at the impersonator directly (which would break callers that need a real curl), minds bundles the **dispatch curl** from the datalib release (imbue-ai/datalib#97, released as **v0.22.0**) and runs *that* as `LATCHKEY_CURL`. The dispatch curl routes on the marker header to either the Chrome-impersonating curl (a sibling in the same dir) or the system curl.

## Changes

**Local gateway**
- `scripts/download-binaries.js` fetches the datalib `curl-<triple>.tar.gz` (macOS arm64, Linux x86_64; SHA256-verified, pinned to `v0.22.0`) into `resources/curl/`. On platforms datalib doesn't build it for (macOS x86_64, Windows) it's a loud no-op and latchkey uses the system curl.
- `scripts/ensure-binaries.js` lists the dispatch curl in `REQUIRED` (gated on the supported platforms) so dev `pnpm start` fetches it too.
- `electron/paths.js` + `electron/backend.js` expose the two bundled binaries (`latchkey-curl-dispatch`, `latchkey-curl-impersonate`) and set `LATCHKEY_CURL` to the dispatch curl on the minds backend env — guarded on **both** binaries existing (the dispatch curl finds the impersonator as a sibling, so no second env var is needed). This propagates via `os.environ` inheritance to the minds process's own latchkey calls, the detached `mngr latchkey forward` supervisor, and the gateway subprocess it spawns.

**Outer-host (VPS) gateway** (`libs/mngr_latchkey`)
- `remote_gateway.py` best-effort installs the dispatch curl + impersonator from the datalib `v0.22.0` release during provisioning (non-fatal — a fetch failure never breaks provisioning), and the gateway run script exports `LATCHKEY_CURL` pointing at the dispatch curl only when both binaries are present, falling back to the system curl otherwise.

In every skip/fallback case latchkey uses the system curl exactly as before, so existing callers and credential checks are unaffected.

## Signature design

The marker is a value-less, namespaced header (`-H "X-Imbue-Impersonate:"`). On a real curl this is a header *removal* of a header curl never sends — a genuine no-op on the wire — so the signature is invisible even if it reaches a non-dispatch curl. Only the dispatch curl gives it meaning.

## Testing

- The pin is verified end-to-end: `downloadLatchkeyCurl` fetches the real `v0.22.0` tarball, the pinned SHA256 matches, and both binaries extract into `resources/curl/`.
- `libs/mngr_latchkey` remote_gateway tests pass (34), including coverage for the VPS install block and the guarded run-script export; the generated VPS scripts pass `sh -n` and `dash -n`.
- `ty` clean on the changed Python; JS syntax-checked.

Depends on imbue-ai/datalib#97 (merged; released as v0.22.0).


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#64 from the same branch. This public repository is now a read-only mirror.

qi-imbue:
Development moved to the internal repo; retargeted to imbue-ai/mngr-internal#89 (the six commits squashed into one). Closing here.

PR #2571, opened 2026-07-22
Move workspace Claude auth to the in-workspace modal (settings.json env)

joshalbrecht:
Implements `blueprint/workspace-claude-auth-settings-env/plan-workspace-claude-auth-settings-env.md` (also in this PR).

**minds side (this PR):**
- AI-provider selection and the API-key field are removed from the create form and `POST /api/v1/workspaces` (a stale `ai_provider` from old clients is silently ignored); no create-time LiteLLM minting or ANTHROPIC env injection remains. Workspaces boot unauthenticated and sign in through their own Claude modal.
- New mint-only `/settings/ai-keys?workspace=<host_id>` page + `/settings/ai-keys/mint` endpoint: resolves the owning account from the workspace-record store, mints a workspace-aliased LiteLLM key (100 USD / 1d rolling budget), and copies an env-var credential blob for pasting into the workspace modal.
- Electron e2e reworked to the new first-boot flow (create bare, modal auto-appears, API-key sign-in, chat); `test_litellm_via_workspace` implemented for real (connector mint, workspace credential submit, chat through litellm, spend assertion).

**Paired branch:** `default-workspace-template` `mngr/better-chat-auth` (pushed) carries the workspace side: settings-env credential writer (strict env-lines endpoint), `claude setup-token` subscription flow with poll-completion, Sign in with Imbue paste path, restart covering `claude`+`worker` agents with state snapshot and please-continue messaging, the host-env migration script, litellm auth-error patterns, and pinned-version (2.1.207) regression tests for the load-bearing Claude Code behaviors.

**Testing:** full offload suite 15,832 passed / 0 failed; all PR checks green (incl. Minds Snapshot Resume, which exercises the new modal sign-in end-to-end). 19 autofix commits applied.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2592, opened 2026-07-23
minds: fix stuck "Discovering agents..." from a vanished-workspace last-good ghost

gnguralnick:
## Problem

The minds landing page could get permanently stuck on "Discovering agents..." (never falling through to the create form) when the user has no live agents. The landing handler shows the create form only when `discovery_complete AND not restorable_agent_ids`, and `list_restorable_workspace_ids()` counts any `system-services` record in the persisted `last_good_agent_topology.json`.

A workspace that vanishes **without ever being observed `DESTROYED`** — e.g. an imbue_cloud pre-baked pool host returned to the pool — left its last-good record in place forever (the old `_merge_last_good_topology_locked` pruned last-good *only* on a positive `DESTROYED` observation), so `restorable` stayed non-empty and pinned the spinner.

## Fix

Generalize the last-good prune in `backend_resolver.py`: a remembered host absent from the live snapshot is now also dropped once **its owning provider** produces a clean, non-errored discovery snapshot that lacks it — proof the workspace is genuinely gone rather than the provider being slow. This mirrors the existing `WorkspaceRecordStore._tombstone_definitively_absent` gate, scoped per-host via a new optional `provider_name` on `_AgentRecord` so an unrelated provider's outage cannot force the removal. Absence with no authoritative provider signal (provider errored, or hasn't reported yet) is still retained — the transient-cold-start-loss protection is preserved.

Records persisted before the `provider_name` field carry no owning provider; they clear via a legacy any-clean-provider gate and are re-recorded (now attributed) the moment discovery re-enumerates them. This is a self-correcting, one-session migration; the only tradeoff is that a pre-upgrade record for a workspace whose provider is down for a whole session may be dropped from window-restore for that session.

## Verification

- New TDD tests in `backend_resolver_test.py`: prune on owning-provider clean snapshot; retain on owning-provider error; prune a legacy provider-less record once discovery reports.
- Manually verified against a copy of a real stuck `last_good_agent_topology.json`: the ghost is pruned and the on-disk file self-heals, so the landing page resolves to the create form.
- Full `desktop_client` suite: 1525 passed. backend_resolver + ratchets: 142 passed. ruff/format/ty clean.

## Test plan

- [ ] CI green (including the offloaded acceptance suite, not run locally)


gnguralnick:
Closing in favor of #2571, which already contains the same last-good-topology fix (prune remembered hosts on a clean per-provider snapshot, with records carrying their owning provider) plus a complementary 'always offer Create on the discovering landing' safety net.

For reference, this branch also had three unit tests in backend_resolver_test.py covering: prune on owning-provider clean snapshot (+ on-disk persistence), retain on owning-provider error (transient-loss protection), and the legacy provider-less record case reproducing the real stuck last_good_agent_topology.json. Happy to port any of those over to #2571 if useful. #2571's wipe-legacy-on-load migration is cleaner than the any-clean-provider legacy gate used here.

gnguralnick:
Reopening PR for continued work.

PR #2593, opened 2026-07-23
Contain the document scrollbar below the titlebar on local pages

pseay-imbue:
Scrolling local pages (Settings, Home) scroll the document, whose native scrollbar spans the full window edge -- riding over the fixed titlebar -- and, with the persistent shell, could linger visually after swapping to a non-scrolling page.

CSS-only containment: the html scrollbar (Chromium-specific, which every minds surface is) gets a custom style whose track carries a 38px top margin, so the thumb can never travel over the bar; the html background follows --titlebar-bg so the strip's top notch paints in the bar's own color and the titlebar reads edge-to-edge. When nothing overflows -- including right after swapping away from Settings -- no thumb paints and the track matches the page surface, so the scrollbar visibly disappears with the page that owned it. Scoped to the html element only; inner scrollables keep native scrollbars.


apps/minds/imbue/minds/desktop_client/templates/ChromeShell.jinja
```diff
@@ -53,32 +53,39 @@
...
-  /* Viewport lock for the agent content surface. Keyed off the html class
+  /* Viewport lock for both shell modes. Keyed off the html class
      (rather than emitted conditionally) because the swap engine replaces the
      html/body classes on every in-place hub swap but never touches the head:
      a conditional rule would leak the previous page's mode across swaps
      (e.g. leaving a workspace wrapper's overflow:hidden on the Home page,
      making it unscrollable). */
-  html.agent-surface, html.agent-surface body { height: 100%; overflow: hidden; }
+  html.agent-surface, html.agent-surface body,
+  html.local-surface, html.local-surface body { height: 100%; overflow: hidden; }
```
gdenisov, line 88: feels like we should be able to just do `html, body {
height: 100%; overflow: hidden;
}`

gdenisov: but also feels like these should really belong in app.css anyway — can't think of a reason why we wouldn't want this for every screen in the app by default

apps/minds/imbue/minds/desktop_client/static/app.css
```diff
@@ -332,6 +332,33 @@ html {
...
+#local-page-scroll::-webkit-scrollbar-thumb {
+  background: var(--c-border-default);
+  border-radius: 5px;
+  border: 2px solid transparent;
+  background-clip: padding-box;
+}
+#local-page-scroll::-webkit-scrollbar-thumb:hover {
+  background: var(--c-text-tertiary);
+  background-clip: padding-box;
+}
```
gdenisov, line 361: Do these work the same on different OSes?

gdenisov (commented):
LGTM with a minor nit.

weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#65 from the same branch. This public repository is now a read-only mirror.

pseay-imbue:
Continuing on imbue-ai/mngr-internal#98 — development has moved; this mirror is read-only, so PRs cannot merge here.

PR #2596, opened 2026-07-24
Fix misleading "admin" naming in the connector and its clients

joshalbrecht:
## Summary

The remote-service connector used "admin" for three unrelated things, and "username" for a value that was never a username. This PR is a naming cleanup with two deliberate operational pieces (an env-var migration shim and a DB column-rename migration):

- **User-session auth**: `AdminAuth` was just a normal signed-in user's SuperTokens session, not an administrator. Renamed `AdminAuth` -> `UserAuth`, `AgentAuth` -> `TunnelTokenAuth` (naming the credential it validates), `require_admin` -> `require_user_auth`, `resolve_entitlements_for_admin` -> `resolve_entitlements_for_user`, plus all locals, docstrings, error messages, tests, and the README. "Admin" now means exactly one thing: the fixed operator key.
- **Operator key**: `MINDS_PAID_ADMIN_KEY` was stale -- the key guards the account admin API and sweeps, not just the paid lists. Canonical name is now `MINDS_ADMIN_KEY` (env var + Vault field); the deprecated spelling is still accepted everywhere it was read (connector, `mngr imbue_cloud` CLI, `minds paid`) with a warning on fallback, and the new name wins when both are set. `minds paid` injects both spellings into its subprocess so older installed mngr versions keep working. `require_paid_admin_key` -> `require_admin_key`. The fallback contract is unit-tested at all three read sites.
- **`username` -> `user_id_prefix`**: the value is the 16-hex prefix of the SuperTokens user id, used to namespace tunnels/leases/buckets. Unified the three coexisting names (`username`, `username_prefix`, `user_id_prefix`) on `user_id_prefix` throughout, and **migration 016** renames the `account_entitlements.username_prefix` / `r2_cleanup_grants.username_prefix` columns (and the entitlements prefix index) to match, applied automatically on the next `minds env deploy`.
- **Release test fixed**: `test_full_lifecycle` still used HTTP Basic auth (`USER_CREDENTIALS`), which the connector stopped accepting when auth moved to Bearer-only, and predated the quota work. It now stubs the SuperTokens auth path + entitlements store like the unit tests do while keeping every Cloudflare call real. Verified green against the staging Cloudflare account.

## Migration notes

- Vault entries (`secrets/minds/<tier>/supertokens`) can adopt `MINDS_ADMIN_KEY` at leisure; until then the connector and CLIs warn and fall back to `MINDS_PAID_ADMIN_KEY`.
- Migration 016 applies before the new connector code rolls out during a deploy; connectors still running pre-rename code would fail quota/grant queries during that brief window (noted in the migration header). Wire formats and HTTP routes are unchanged.

## Testing

- Full `just test-offload` suite: **16143 passed, 0 failed** (run repeatedly through the branch)
- `just test apps/remote_service_connector/.../test_remote_service_connector.py::test_full_lifecycle` against the real staging Cloudflare account: **1 passed**
- New tests: legacy env-var fallback + precedence (connector and both CLIs), migration-rename assertions
- /autofix ran to convergence (validate-diff + iterated verify-and-fix agents); all surfaced issues fixed, none outstanding


weishi-imbue:
Development has moved to the private repository; this PR was recreated there as imbue-ai/mngr-internal#67 from the same branch. This public repository is now a read-only mirror.

PR #2597, opened 2026-07-28
Service-per-origin forwarding redesign (mngr_forward + minds)

amirmohsen-am:
## Service-per-origin forwarding redesign (mngr + minds side)

Implements `blueprint/forwarding-redesign/plan-forwarding-redesign.md`: replaces the `/service/<name>/` path-prefix muxing with service-per-origin routing. Every registered service owns a real browser origin — locally `http(s)://<name>.agent-<hex>.localhost:8421/` (any deeper subdomain routes to the same service), shared `https://<name>--<host>--<user>.<domain>/`. Pairs with the default-workspace-template PR of the same branch name (big-bang cutover; no backward compatibility).

### mngr_forward
- Host parsing accepts optional service labels; the last label before `agent-<hex>` selects the service, deeper labels are the service's own sub-origin space (multi-origin apps, e.g. VS Code webviews).
- `resolve(agent_id, service_name)`: bare origin → the configured shell service; labels → the agent's registered service map.
- `/goto/` bridge carries the service label chain and sets the session cookie with `Domain=agent-<hex>.localhost` — one login hop covers the shell and every service origin at any depth. Unregistered-but-plausible services get the auto-retrying loading page.
- TLS/HTTP-2 mode retained: per-SNI certificate minting covers nested service origins without restart (static `*.localhost` wildcard can't).

### The `system_interface` → `system-interface` rename
Service names are now hostname labels (`<name>.agent-<hex>.localhost` locally, `<name>--<host>--<user>.<domain>` on shares), and underscores are not valid in DNS labels — so the shell's registered service name had to become DNS-safe. `system-interface` was chosen over `shell` for minimal churn. The rename covers the *name* only: the supervisord program, the `forward_port.py` registration, the oom-priority band key, minds' default `--service`, and the recovery probe's `supervisorctl status` target (template PR carries the container side). The python package (`imbue.system_interface`) and the `system-interface` CLI binary are unchanged. `forward_port.py` now rejects non-DNS-safe names outright, so no other service can reintroduce the problem.

### minds
- Forwards the renamed `system-interface` shell service by default; recovery probe follows the supervisord rename.
- Workspace sharing fan-out (interim single-tier model): sharing `system-interface` shares every registered service with the same email grants via the existing per-service connector API; workspace unshare removes them all. The shared shell derives sibling hostnames by swapping the first `--` token of `location.host`.

### Validation (see `blueprint/forwarding-redesign/validation-results.md`)
- Full unit/integration suites green (all remaining failures verified pre-existing/environmental).
- Real-browser e2e on a Docker workspace: shell at the bare origin, terminal as a cross-origin iframe with working WebSocket over per-SNI TLS; openvscode installed by the workspace agent via chat (standard supervisord + forward_port workflow), serving at its own origin and at arbitrary sub-origin depth.
- Cloudflare share e2e in a fresh external browser: Access email-OTP login, shared shell with panels on derived sibling hostnames (silent Access auth inside iframes confirmed), full VS Code workbench through the share hostname.
- Phase-0 browser matrix: Chromium/Firefox pass all nested `*.localhost` cookie checks; WebKit treats each `x.localhost` as its own site (fallbacks documented in `blueprint/forwarding-redesign/decisions-phase0.md`).

### Future TODO (deferred by design)
- **remote_service_connector: two-tier Access model** — one Access Group per shared workspace as the master email list; per-service app policies become `group ∪ service emails`; service-registration reconcile (DNS + ingress + Access app for services registered while shared — today a new service needs a re-save of the workspace share). Replaces the interim fan-out in `sharing_handler.py`.
- **Connector: Access app cookie settings** — `allow_iframe: true`, and a per-service `SameSite=None` opt-in (NEVER global: it re-opens CSRF and authenticated cross-site WebSockets — unacceptable for the terminal). The flag should be declared at service registration by the installing agent and thread forward_port → sharing → connector; opted-in services should validate `Origin` on WS endpoints.
- **Connector: Option B — Cloudflare Advanced Certificate Manager (ACM) for share hostnames.** Shipped as Option C ("ship A, keep B addable"): apps currently use their public multi-origin defaults on shares (vscode webviews → `vscode-cdn.net`). The free universal cert only covers one label (`*.<domain>`), which is why share hostnames are flat. ACM (~$10/mo/zone) lets the connector issue deeper wildcards (e.g. `*.openvscode--<host>--<user>.<domain>`), which buys: service sub-origins on shares → apps self-host their sandbox/webview origins (`--webview-external-endpoint`) → everything becomes same-site/first-party → no third-party cookies, no `SameSite=None` opt-in needed, works in Safari/Firefox too, and share behavior matches the arbitrary-depth sub-origins already validated locally.
- **openvscode webview endpoint** — run with `--webview-external-endpoint={{uuid}}.openvscode.<workspace-host>` locally (deep sub-origin routing already validated) so webviews stop using `vscode-cdn.net`; document in the build-app guidance. Until then, on shares a small subset of webview-context asset loads break in Safari/Firefox (Access cookie vs third-party cookie policies).
- **minds Electron: window-open policy for webview origins** — `*.vscode-cdn.net` webview iframes currently pop out to the system browser; allowlist them inline (or land the self-hosted webview endpoint above and skip this).
- **Safari-local fallback** — a real wildcard loopback domain (verified: `lvh.me` style) or per-panel auth bootstrap, if local Safari support is wanted.
- **Docs** — the plan's Phase-4 `specs/workspace-server-forwarding` / `specs/mngr-forward-plugin` paths turned out not to exist in either repo; template-side docs/skills were updated in the paired PR. Remaining: sweep any mngr-side forwarding docs (e.g. `apps/minds/docs/`) for stale `/service/` references.
- **Migration** — pre-cutover shares are not migrated; document "re-share after upgrade".
- **e2e harness** — the Electron e2e helpers assume a fresh environment (landing-page picker, health-cache routing, restored-layout tab focus); teach them to handle machines with existing workspace records.

### Out-of-scope TODOs (unchanged by this PR; noted for the record)
- **Per-tab Share button — needs a spec.** Still points at workspace settings, which works fine with the new model. With per-service grants now first-class, decide: narrow-grant editor for that service's email list, removal in favor of workspace settings only, and how it presents the two-tier model once the connector master list exists.
- **Revocation semantics — pre-existing Access behavior.** Policy edits refuse new logins immediately but don't kill active sessions (full unshare does, since the Access app is deleted). Decide session duration vs revoke-active-sessions when the Access-Group work lands, where one master-list edit will affect every service.
- **`agent-<hex>` vs workspace-id subdomain coordinate.** Kept `agent-<hex>`. If a rename to a user-meaningful workspace id is ever wanted, a compatibility-breaking moment like this cutover is the cheap time to do it — flagged for a team decision.


private.just
amirmohsen-am, line 1: Leaving here as there are some changes to `private.just` but should be a much smaller change with the commited `private.just` in the internal repo

libs/mngr_forward/imbue/mngr_forward/tls.py
```diff
@@ -90,6 +114,79 @@ def build_server_ssl_context(cert_pem: bytes, key_pem: bytes) -> ssl.SSLContext:
...
 
+def _is_covered_by_static_sans(server_name: str) -> bool:
+    """True when the static cert (``localhost`` / ``*.localhost``) matches."""
+    if server_name == "localhost":
+        return True
+    return server_name.endswith(".localhost") and server_name.count(".") == 1
+
+
+class _SNICertMinter(MutableModel):
+    """SNI callback that mints (and caches) per-hostname certs on demand.
```
amirmohsen-am, line 125: I don't think this is needed since in Electron we trust all `*.localhost` but keeps the certifications valid if we use other clients

amirmohsen-am:
@joshalbrecht document explaining the issue and the solution with openvscode webviews: [blueprint/forwarding-redesign/openvscode-multi-origin.md](https://github.com/amirmohsen-am/mngr/blob/amir/forwarding-redesign/blueprint/forwarding-redesign/openvscode-multi-origin.md)

